PR #103, opened 2014-01-02
Add support for geopackagebinary format loads/dumps

bradh:
This is described in the OGC GeoPackage spec (12-128r9). Its like
WKB, but with some extra noise (including the SRID) on the front.

### DIFF
--- a/docs/manual.txt
+++ b/docs/manual.txt
@@ -2082,6 +2082,21 @@ All of Shapely's geometry types are supported by these functions.
   >>> loads(wkt).wkt
   'POINT (0.0000000000000000 0.0000000000000000)'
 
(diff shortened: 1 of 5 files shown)


DavidMStraub:
@bradh, looking for a working Python implementation of a Geopackage binary parser, this PR is the only thing I found. Did you contribute this code to some other package eventually?

bradh:
No. Noone wantes wanted it.

achapkowski:
Encourage geomet to re-address it.  I was trying to get it in there too. :)

bradh:
I respect the maintainers call. We might like the feature, but they have have to carry the overhead.

PR #137, opened 2014-07-03
Fix typo in 'shapely.ops' docu

tsauerwein:
'epsg:4326' and not 'espg:4326'


pgiraud:
This has been fixed more than 1 year ago but still not deployed?
http://toblerity.org/shapely/manual.html#shapely.ops.transform

PR #219, opened 2014-12-29
Require GEOS to be preinstalled.

### LINKED ISSUE #218: Need better documentation of GEOS dependencies
Document more clearly that 

- Shapely 1.2.x requires GEOS 3.1+
- Shapely 1.3+ requires GEOS 3.3+

So, if your system only provides GEOS 3.2.x, you need to do `pip install shapely<1.3`.

Should Shapely's setup.py sniff out the GEOS version (using ctypes) and fail the installation if the requirement isn't met?

sgillies:
New custom install command fails if we can't locate a GEOS shared library (using same method as in shapely/geos.py).

Closes #218.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,15 +2,19 @@
 
 from __future__ import print_function
 
+import ctypes
+import ctypes.util
 try:
     # If possible, use setuptools
     from setuptools import setup
     from setuptools.extension import Extension
     from setuptools.command.build_ext import build_ext as distutils_build_ext
+    from setuptools.command.install import install
 except ImportError:
     from distutils.core import setup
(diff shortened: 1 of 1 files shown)


sgillies:
@mwtoews you okay with these changes? will you merge if so?

mwtoews:
@sgillies I can't seem to figure out how to merge this commit to master (is it cherry-pick?), but it would be great if you could do that.

sgillies:
If we're careful, we'll always be able to merge maint to master. See https://github.com/Toblerity/Shapely/commit/f75df63f5099eb4a27009286e2e049740083973b.

PR #221, opened 2015-01-05
Making OS X binary wheels that include a GEOS dylib

sgillies:
4 lines added to shapely/geos.py will let us load a shared library included in a [binary wheel for OS X](https://github.com/MacPython/wiki/wiki/Spinning-wheels). This means a completely self-contained Shapely wheel that "just works" on a OS X system Python or python.org Python.

My plan is to use [delocate](https://github.com/matthew-brett/delocate) to copy dependencies and relink the extension modules as the SciPy project does. Currently, I'm using the Kyngchaos Framework libs because they are dual architecture (like python.org's Python). It brings the wheel size up to 1.2MB, which isn't too bad at all.

@mwtoews can you review this for me and merge? I'd like to get a binary wheel for 1.5.2 out for some user testing soon.

### DIFF
--- a/build-scripts/macosx-10.6-intel.sh
+++ b/build-scripts/macosx-10.6-intel.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+# Dependent on the Kyngchaos Frameworks:
+# http://www.kyngchaos.com/software/frameworks
+
+export GEOS_CONFIG="/Library/Frameworks/GEOS.framework/Versions/3/unix/bin/geos-config"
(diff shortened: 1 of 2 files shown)


sgillies:
Pre-release wheel uploaded to: https://www.dropbox.com/s/656rk4boinh5t50/Shapely-1.5.2-cp27-none-macosx_10_6_intel.whl?dl=0

sgillies:
And here's the result of installing that wheel on my family's older iMac (with a new python.org Python 2.7.9):

shapely-wheel

Success!

PR #222, opened 2015-01-06
Fix polygonize to use _other_owned

aronbierbaum:


sgillies:
Oh, so polygonize has been leaking? @aronbierbaum can you do me a favor and re-aim this PR at the maint branch? I'd like to get the fix into 1.5.3.

aronbierbaum:
@sgillies I can't say that I have done a lot of testing, but I found the use of `_owned` and thought it would be an issue. Closing and rebasing on maint (https://github.com/Toblerity/Shapely/pull/223)

Issue #234, opened 2015-02-02
Import time check for libgeos precludes building with RPATH


kynan:
The import time check for `libgeos` requires the shared library to either be installed in a standard library path or have `LD_LIBRARY_PATH` set, which is a *very bad idea*.

I prefer to build Python extension modules by baking in the RPATH to any shared libraries the extension module links against. Unfortunately this (imho completely unnecessary) import time check precludes this and requires `LD_LIBRARY_PATH` to be always set.

Can this please be removed?

sgillies:
Closed via #238

PR #236, opened 2015-02-02
Move system-dependant logic to load dynamic library into minimal module, and use geos-config to get development headers

mwtoews:
The main thing done here is to move the system-dependant loading of the GEOS dynamic library into a separate module `shapely.libgeos`, which can be used by setup.py for checking versions before install, and by `shapely.geos` for run-time use the GEOS of dynamic library.

A TODO item is to figure out a different way to specify the dynamic library (#234). Someone else might need to take the lead on this, but at least it will be in one place, rather than repeated in `setup.py` and `shapely/geos.py`. CC @kynan for ideas.

Secondly, `geos-config`, where available, is used to obtain libraries and headers for developing Cython extensions. This is useful where GEOS is installed on a non-standard prefix. An alternative location for this utility can be proved for setup.py using a `GEOS_CONFIG` environment variable. Note that this is only needed for setup.py, so I don't think it's suitable for setting dynamic library prefix. Also, `geos-config` is a shell script, and is not available for MS Windows.

Other edits use the logging module for setup.py, and fixing the `requires` metadata. Extra debugging info can be shown e.g. `python -W all setup.py ...`. Also add TODO that `free` is not used anywhere.

Please test and comment, as these files are rather sensitive components.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,6 +27,10 @@ from distutils.errors import CCompilerError, DistutilsExecError, \
     DistutilsPlatformError
 from distutils.sysconfig import get_config_var
 
+# Get geos_version from GEOS dynamic library
+# TODO: find a way to load a specific dynamic library
+from shapely.libgeos import geos_version_string, geos_version
+
 logging.basicConfig()
 log = logging.getLogger(__file__)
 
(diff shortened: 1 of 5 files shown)


sgillies:
@mwtoews Fantastic, I really appreciate this. Could you retarget this at master? It's a little much for maint. I may cherry pick to address #234 or I might just unwind the breaking changes I made in order to release 1.5.6.

mwtoews:
Here's something I can't figure out how to do: is it possible to change the target branch from `Toblerity:maint` to `Toblerity:master`? Or do I need to make a new PR?

mwtoews:
[It looks like](http://stackoverflow.com/a/24159161/327026) this is a limitation of GitHub, so see PR #239 for re-targeting on `Toblerity:master`

PR #237, opened 2015-02-03
Modify svg() and _repr_svg_() outputs

### LINKED ISSUE #233: SVG output for LineString is invalid
Found here: [Shapely LineString SVG Error](http://stackoverflow.com/q/28164557/327026)

```python
x = LineString([(0,0),(50,50)])
print(x.svg())
```
So with [validator.w3.org check](http://validator.w3.org/check):
```xml
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.0//EN" "http://www.w3.org/TR/2001/REC-SVG-20010904/DTD/svg10.dtd">
<svg height="50" width="50">
<polyline
            fill="none"
            stroke="#66cc99"
            stroke-width=2.0
            points="0.0,0.0 50.0,50.0"
            opacity=".8"
            />
</svg>
```
indicates the element is invalid. There are two fixes.

Element literals can be quoted:
```xml
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.0//EN" "http://www.w3.org/TR/2001/REC-SVG-20010904/DTD/svg10.dtd">
<svg height="50" width="50">
<polyline fill="none" stroke="#66cc99" stroke-width="2.0" opacity="0.8" points="0.0,0.0 50.0,50.0" />
</svg>
```

Or, they can be moved to the style element, [like this](http://www.w3schools.com/svg/svg_polyline.asp):
```xml
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.0//EN" "http://www.w3.org/TR/2001/REC-SVG-20010904/DTD/svg10.dtd">
<svg height="50" width="50">
<polyline style="fill:none;stroke:#66cc99;stroke-width:2.0;opacity:0.8" points="0.0,0.0 50.0,50.0" />
</svg>
```

I haven't checked the other geometry types, but it would be worthwhile checking they are all valid.

mwtoews:
To `.svg()`
* 1 element per geometry
* multi-part geometries are grouped in `<g>...</g>` elements
* empty geometries are `<g />`

To `._repr_svg_()`
* add the basic attributes to the the svg element (xmlns et al.) to show output in, e.g. Firefox; see http://www-archive.mozilla.org/projects/svg/faq.html#source
* empty geometries are `<svg ... />`

Also, the second parameter is either `color`, `fill_color`, or `stroke_color`, depending on the geometry type. When this is `None`, it is determined if the geometry is valid, as before. (I'm open to just calling this `color`).

Unit testing is added to ensure consistency of output, and only XML is validated (I don't know of an SVG validator).

Fixes #233.

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -360,7 +360,8 @@ class BaseGeometry(object):
     def _repr_svg_(self):
         """SVG representation for iPython notebook"""
         if self.is_empty:
-            return '<svg />'
+            return '<svg xmlns="http://www.w3.org/2000/svg" ' \
+                'xmlns:xlink="http://www.w3.org/1999/xlink" />'
         else:
             # Pick an arbitrary size for the SVG canvas
             xmin, ymin, xmax, ymax = self.buffer(1).bounds
@@ -374,7 +375,9 @@ class BaseGeometry(object):
             buffered_box = "{0} {1} {2} {3}".\
                 format(xmin, ymin, xmax - xmin, ymax - ymin)
             return (
-                '<svg preserveAspectRatio="xMinYMin meet" viewBox="{0}" '
+                '<svg xmlns="http://www.w3.org/2000/svg" '
+                'xmlns:xlink="http://www.w3.org/1999/xlink" '
+                'preserveAspectRatio="xMinYMin meet" viewBox="{0}" '
                 'width="{1}" height="{2}" '
                 'transform="translate(0, {1}),scale(1, -1)">{3}</svg>'
                 ).format(buffered_box, x_size, y_size, self.svg(scale_factor))
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -359,28 +359,41 @@ class BaseGeometry(object):
 
     def _repr_svg_(self):
         """SVG representation for iPython notebook"""
+        svg_top = '<svg xmlns="http://www.w3.org/2000/svg" ' \
+            'xmlns:xlink="http://www.w3.org/1999/xlink" '
         if self.is_empty:
-            return '<svg xmlns="http://www.w3.org/2000/svg" ' \
-                'xmlns:xlink="http://www.w3.org/1999/xlink" />'
+            return svg_top + '/>'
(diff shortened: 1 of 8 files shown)


mwtoews:
Testing one of the `_repr_svg_()` outputs with http://validator.w3.org/check shows an error "there is no attribute "transform"". It's true, there is no "transform" attribute for the svg element. Firefox and Google Chrome attempt to show it without error, but the images are mirror replicas, since they interpret the parameters differently (probably using the translate part, but not the scale part). We might need to move the transform attribute to a nested `<g transform>...</g>` element to make it valid SVG.

sgillies:
More work to do then? Let me know when it's ready and I'll merge. Things are looking fine in IPython notebooks, which is the main use case.

mwtoews:
I think it's ready to review now. SVG output is valid and opens consistently with various viewers, so should be also good with IPython. I also changed the expansion behaviour. Whereas previously it was arbitrarily buffered by 1 unit, it now expands it by 4% of the widest dimension. This should make very large and very small geometries show with about the same amount on each edge. However, I'm getting mixed results with a `small` case. Do you see anything for `small` before/after this commit?
```python

# no problems
large = LineString([(1000, 5000), (1000001, 5000002)])
with open('large.svg', 'w') as fp:
    fp.write(large._repr_svg_())

# does this look OK?
small = LineString([(1000, 5000), (1000.001, 5000.002)])
with open('small.svg', 'w') as fp:
    fp.write(small._repr_svg_())
```

sgillies:
@mwtoews No, the small LineString doesn't show.

geom_svg

mwtoews:
Did it show before? I'm not sure this is fixable, since it looks like different SVG renderers behave differently with these extremes.

Inkscape is OK with it. Firefox does not show it. Google Chrome will show a line rendered about 3px wide if `stroke-width="6.83e-05"`, but nothing if `stroke-width="6.82e-05"`.

sgillies:
No, it didn't show before, either.

But wait, it does show in both pages if I zoom (Cmd+) a few times, which makes sense.

I call it good.

PR #238, opened 2015-02-03
Revert "Require GEOS to be preinstalled."

sgillies:
My previous version checking is messing up others' builds. See #234. I propose to revert e12d1d2 here in maint (the 1.5.x release branch) and try to fix in master using PR #239.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,19 +2,15 @@
 
 from __future__ import print_function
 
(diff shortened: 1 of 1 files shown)


sgillies:
Green light. Unless you see anything wrong, can you merge @mwtoews?

Issue #232, opened 2015-01-20
Use geos-config for setup.py?


mwtoews:
Is the dynamic library checking in setup.py necessary? Could we instead use `geos-config` similar to rasterio? This could also help populate, `library`, `include_dirs` and `extra_link_args` from `--cflags` and `--clibs` options.

Also, by using `geos-config`, a non-standard installation location for GEOS can be used to build Cython extensions.

mwtoews:
This is what I have in mind: https://github.com/mwtoews/shapely/commit/d65417ab0b7c65617b185bc22b732736b3e15fa7

sgillies:
:+1:

kynan:
It's even *much worse* than that: this check is unnecessarily repeated at *import time*, which precludes using RPATH while building the extension module, see #234.

sgillies:
@mwtoews might you have a PR soon?

mwtoews:
PR at #236

PR #239, opened 2015-02-03
Move system-dependant logic to load dynamic library into minimal module, and use geos-config to get development headers

mwtoews:
The main thing done here is to move the system-dependant loading of the GEOS dynamic library into a separate module `shapely.libgeos`, which can be used by setup.py for checking versions before install, and by `shapely.geos` for run-time use the GEOS of dynamic library.

A TODO item is to figure out a different way to specify the dynamic library (#234). Someone else might need to take the lead on this, but at least it will be in one place, rather than repeated in `setup.py` and `shapely/geos.py`. CC @kynan for ideas.

Secondly, `geos-config`, where available, is used to obtain libraries and headers for developing Cython extensions. This is useful where GEOS is installed on a non-standard prefix. An alternative location for this utility can be proved for setup.py using a `GEOS_CONFIG` environment variable. Note that this is only needed for setup.py, so I don't think it's suitable for setting dynamic library prefix. Also, `geos-config` is a shell script, and is not available for MS Windows.

Other edits use the logging module for setup.py, and fixing the `requires` metadata. Extra debugging info can be shown e.g. `python -W all setup.py ...`. Also add TODO that `free` is not used anywhere.

Please test and comment, as these files are rather sensitive components.

This PR supersedes #236 due to GitHub limitations.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,6 +27,10 @@ from distutils.errors import CCompilerError, DistutilsExecError, \
     DistutilsPlatformError
 from distutils.sysconfig import get_config_var
 
+# Get geos_version from GEOS dynamic library
+# TODO: find a way to load a specific dynamic library
+from shapely.libgeos import geos_version_string, geos_version
+
 logging.basicConfig()
 log = logging.getLogger(__file__)
 
@@ -44,6 +48,17 @@ with open('shapely/__init__.py', 'r') as fp:
 if not version:
     raise ValueError("Could not determine Shapely's version")
 shapely_version = tuple(int(x) for x in version.split('.'))
+
+# Fail installation if the GEOS shared library does not meet the minimum
+# version. We ship it with Shapely for Windows, so no need to check on
+# that platform.
+log.debug('GEOS shared library: %s %s', geos_version_string, geos_version)
+if (set(sys.argv).intersection(['install', 'build', 'build_ext']) and
+        shapely_version >= (1, 3) and geos_version < (3, 3)):
+    log.critical(
+        "Shapely >= 1.3 requires GEOS >= 3.3. "
+        "Install GEOS 3.3+ and reinstall Shapely.")
+    sys.exit(1)
 
 # Handle UTF-8 encoding of certain text files.
 open_kwds = {}
@@ -121,7 +136,14 @@ if sys.platform == 'win32':
         include_package_data=True,
     )
 
-# Get configuartion information for GEOS library using command line tool
+# Build cython extensions, which require development parameters
+include_dirs = [get_config_var('INCLUDEDIR')]
+library_dirs = []
+libraries = []
+extra_link_args = []
+
+# Use geos-config utility to get development parameters
+# (Note: not available for Windows, since it is a shell script)
 geos_config = os.environ.get('GEOS_CONFIG', 'geos-config')
 log.debug('geos_config: %s', geos_config)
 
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -1,47 +1,68 @@
...
-import os
-import platform
-import re
-import shutil
-import subprocess
-import sys
 
+# Get geos_version from GEOS dynamic library
+# TODO: find a way to load a specific dynamic library
+from shapely.libgeos import geos_version_string, geos_version
```
kynan: Unfortunately this import brings back the problem: `shapely.libgeos` still does the ctypes check for the shared libraries at module level. What's the point of using `geos-config` if this check is still made?

setup.py
```diff
@@ -183,6 +136,70 @@ def load_dll(libname, fallbacks=None):
...
+        sys.exit(1)
+except OSError as ex:
+    log.error(ex)
+    log.error('Cannot find geos-config to get headers and check version.')
+    log.error('If available, specify a path to geos-config with a '
+              'GEOS_CONFIG environment variable')
+    geos_config = None
+
+
+if geos_config:
```
kynan: Unless the user has `GEOS_CONFIG` set to empty this will always succeed (since you default to `geos-config`), is this intentional?

setup.py
```diff
@@ -183,6 +136,70 @@ def load_dll(libname, fallbacks=None):
...
+        if item.startswith("-I"):
+            include_dirs.extend(item[2:].split(":"))
+    for item in get_geos_config('--clibs').split():
+        if item.startswith("-L"):
+            library_dirs.extend(item[2:].split(":"))
+        elif item.startswith("-l"):
+            libraries.append(item[2:])
+        else:
+            # e.g. -framework GEOS
+            extra_link_args.append(item)
```
kynan: Yes, I think this is the way to go. So if this succeeds, what's the point of *also* determining the GEOS version via ctypes by importing from `shapely.libgeos` above?

kynan:
@mwtoews @sgillies Sorry to be a bit late to the party (it was the middle of the night in Europe...) and thanks for the quick fix to #234. It seems though this is only part of the way to resolving #232 i.e. during *setup* you still need to have `LD_LIBRARY_PATH` set for ctypes to be able to load the shared library.

setup.py
mwtoews: Also, a few lines up `geos_config = None` if the utility does not run / is not found.

setup.py
mwtoews: Shapely depends on the dynamic library. The development library is optional. The dynamic library version needs to match the development library version.

shapely/libgeos.py
```diff
@@ -0,0 +1,116 @@
+"""
+Proxies for the GEOS shared library, versions, and configure options
+TODO: find a flexible way to load a dynamic library from a specified path
```
mwtoews: Should this be set from a `GEOS_LIBRARY_PATH` environment variable, as shown in #240?

mwtoews: Also, I was considering moving `get_geos_config` ([setup.py#L151](https://github.com/Toblerity/Shapely/blob/2e58796977672acfd487607981b4fc2f62b23811/setup.py#L151)) to this module, but was hesitant since it is only a development tool. I'm not sure how `geos-config` can be used to sniff out the dynamic library location. (Somewhere under `geos-config --prefix`?). Is it possible to get `setup.py` to write a configuration file on as to where the dynamic library should always be loaded?

Issue #136, opened 2014-07-02
GEOSversion function not found


nathanhilbert:
I think I'm getting some geos.dll library confusion, but I can't pinpoint the problem.  With standard install I get a GEOSversion function not found.  I've edited the code in two ways that has made it work:
1. in geos.py, on line 87 under the windows platform, I changed geos.dll to geos_c.dll which resolves the error
2. I also ran a find_library("gdal.dll") then pass the full path into CDLL, which makes it work (and resolves to the installed egg).  I suspect in 1, it is resolving to my GDAL 1.11 install, which appears to be an older version of geos.
Any help would be appreciated, or if others have had this issue, I can submit a pull request.
Thanks.

nathanhilbert:
I hate to reopen this issue, but I hit the same problem months later.  @sgillies I like your first solution to try geos.dll then try geos_c.dll.  In addition, I would like to check if the GEOS_LIBRARY_PATH environment variable is set.  

My issues are not related to conda or SciTools, but they sound similar.  

PR coming soon.

nathanhilbert:
Closed with PR #239

PR #240, opened 2015-02-03
adding checks for the geos lib

nathanhilbert:
In [GeoNode](https://github.com/GeoNode/geonode), GDAL-1.11.0-py2.7-win32.egg is used, which has a geos_c.dll precompiled.  There was a confusion between multiple geos loads where django.contrib.gis would try to use geos_c.dll, but Shaply would try to use geos.dll.  This checks all possibilities or throws an import error.  This is reference to #136


nathanhilbert:
I didn't 100% follow the GEOS_CONFIG discussion and commits, but installing the wheel with 1.5.6 works without this PR.  Thanks for your work.

sgillies:
Great!

PR #241, opened 2015-02-10
Load dynamic library using two alternative methods

mwtoews:
The first method is borrowed from #240 / GeoDjango, which is to use an
environment variable GEOS_LIBRARY_PATH to point directly to a dynamic library.
If this seems like a hack, it can be removed. However, it could be useful
for folks that don't have GEOS_CONFIG, such as Windows, and want to use a
different DLL file.

The second method is to use geos-config from the PATH, or can be explicity set
using a GEOS_CONFIG flag. It attempts a few paths to libraries from the output
from `geos-config --clibs`. If used in combination with GEOS_LIBRARY_PATH,
the two library versions are checked by setup.py.

After these attempts, the old code is still in place (with very minor changes).

An unused `free` object is still a TODO.

This is a continuation of #239.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -44,12 +44,13 @@ system library path, and install Shapely from the Python package index.
 
     $ pip install shapely
 
-If you've installed GEOS to a non-standard location, you can use the
-geos-config program to find compiler and linker options.
+If you've installed GEOS to a non-standard location, the geos-config program
+will be used to get compiler and linker options. If it is not on the PATH,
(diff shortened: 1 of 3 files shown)


sgillies:
:+1:

ahmadia:
Thanks, this enables the HashDist project to provide multiple versions of Shapely and GEOS builds without any extra hacking on our side :)

PR #244, opened 2015-02-17
Only check with geos-config in setup.

sgillies:
Do not import shapely.libgeos.


setup.py
```diff
@@ -41,6 +78,32 @@
...
+class GEOSConfig(object):
+    """Interface to config options from the `geos-config` utility
+    """
+
+    def __init__(self, cmd):
+        self.cmd = cmd
+
+    def get(self, option):
+        try:
+            output = subprocess.check_output([self.cmd, option])
```
mwtoews: [`check_output`](https://docs.python.org/2/library/subprocess.html#subprocess.check_output) was annoyingly added in Python 2.7

setup.py
```diff
@@ -41,6 +78,32 @@
...
+                'Could not find geos-config %r: %s' % (self.cmd, ex))
+        if sys.version_info[0] >= 3:
+            result = output.decode('ascii').strip()
+        else:
+            result = output.strip()
+        log.debug('%s %s: %r', self.cmd, option, result)
+        return result
+
+    def version(self):
+        return tuple(map(int, self.get('--version').strip().split('.')))
```
mwtoews: Unfortunately, this does not work with development versions of GEOS, which emits, e.g. `3.4.0dev`: "ValueError: invalid literal for int() with base 10: '0dev'"

sgillies: Yes... I'll do this more like you were doing in `get_geos_config()`. Basically, the goal is to avoid importing anything from shapely in setup and just use geos-config or command line options and environment vars.

sgillies:
What do you think of the comments at the top of setup.py, @mwtoews? Do they help? Am I overlooking any setup use cases?

setup.py
```diff
@@ -55,13 +123,25 @@
...
+geos_config = GEOSConfig(os.environ.get('GEOS_CONFIG', 'geos-config'))
+
+if not os.environ.get('NO_GEOS_CHECK'):
+    try:
+        geos_version = geos_config.version()
+    except Exception as exc:
+        log.critical("Failed to determine system's GEOS version: %s", exc)
+        sys.exit(1)
+
+    log.debug("GEOS shared library: %", geos_version)
```
mwtoews: should have `%s` for the formatter

mwtoews:
I think the changes look good. They seem to find non-standard installs.

Do we really want the script to fail if `geos-config` is not found? This GEOS utility is only installed if either installed from source, or if the user has the -dev packages installed (e.g., [libgeos-dev](https://packages.debian.org/wheezy/amd64/libgeos-dev/filelist)). Windows, for example, does not have the utility, since it is a shell script. Shouldn't it just warn that the version cannot be verified, and skip building speedups?

sgillies:
geos-config only with -dev packages... I'll have to think about this a little more then.

sgillies:
Went with the warning if geos-config can't be found.

If geos-config is found, and its version doesn't meet the Shapely requirement, install will fail ... unless this check is overridden by NO_GEOS_CHECK=1.

PR #246, opened 2015-02-21
Add ImplementationError

### LINKED ISSUE #216: Raise NotImplementedError for missing methods
Currently we look optimistically for methods in [LGEOSBase.methods](https://github.com/Toblerity/Shapely/blob/master/shapely/geos.py#L721) and a KeyError gets raised if our GEOS shared library doesn't support the method. I think we can do better: a NotImplementedError would be less baffling, especially if it had a message with details about required GEOS versions.

sgillies:
Add an ImplementationError, akin to but slightly different from NotImplementedError. It also derives from KeyError and AttributeError for backwards compatibility with code that catches those errors for this particular case.    

This one would close #216.


sgillies:
@snorfalorpagus merge for me if you agree?

Issue #243, opened 2015-02-12
UnicodeDecodeError with version 1.5.6 wheel on MacOS X 10.10


viaregio:
<p>The code</p>
<pre><code>import shapely.geos
shapely.geos._lgeos</code></pre>
<p>results in</p>
<pre><code>&lt;CDLL &#39;/usr/local/lib/python3.4/site-packages/shapely/.dylibs/libgeos_c.1.dylib&#39;, handle 7fbdf49b8720 at 10c0d2400&gt;</code></pre>
<p>and the following code (A example from the Cartopy homepage)</p>
<pre><code>import cartopy
import matplotlib.pyplot as plt

def main():

    ax = plt.axes(projection=cartopy.crs.PlateCarree())

    ax.add_feature(cartopy.feature.LAND)
    ax.add_feature(cartopy.feature.OCEAN)
    ax.add_feature(cartopy.feature.COASTLINE)
    ax.add_feature(cartopy.feature.BORDERS, linestyle=&#39;:&#39;)
    ax.add_feature(cartopy.feature.LAKES, alpha=0.5)
    ax.add_feature(cartopy.feature.RIVERS)

    ax.set_extent([-20, 60, -40, 40])

    plt.show()

if __name__ == &#39;__main__&#39;:
    main()</code></pre>
<p>results in</p>
<pre><code>---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)

/Users/uli/Entwicklung/IPythonNotebooks/_ctypes/callbacks.c in &#39;calling callback function&#39;()

/usr/local/lib/python3.4/site-packages/shapely/geos.py in error_handler(fmt, *args)
    189     if sys.version_info[0] &gt;= 3:
    190         fmt = fmt.decode(&#39;ascii&#39;)
--&gt; 191         args = [arg.decode(&#39;ascii&#39;) for arg in args]
    192     LOG.error(fmt, *args)
    193 

/usr/local/lib/python3.4/site-packages/shapely/geos.py in &lt;listcomp&gt;(.0)
    189     if sys.version_info[0] &gt;= 3:
    190         fmt = fmt.decode(&#39;ascii&#39;)
--&gt; 191         args = [arg.decode(&#39;ascii&#39;) for arg in args]
    192     LOG.error(fmt, *args)
    193 

UnicodeDecodeError: &#39;ascii&#39; codec can&#39;t decode byte 0xbb in position 1: ordinal not in range(128)</code></pre>
[...]

PR #247, opened 2015-02-23
Add tests error handler and a fix for #243

sgillies:
At least towards a fix of #243. I have no test case that can duplicate the original bug, but this gets the error (and never yet used notice) handler in better shape.


sgillies:
@snorfalorpagus @mwtoews what do you think? I can't reproduce #243 but this PR does fix bugs in the error and notice handlers.

sgillies:
Oh crap, wrong target branch.

PR #249, opened 2015-02-24
Tested error handler

sgillies:
Tests for the error and notice handlers, which was buggy, and fixes for the bugs.

There's still some major fragility here. See my comments above the handler functions in shapely/geos.py.

This PR replaces #247

### DIFF
--- a/shapely/ctypes_declarations.py
+++ b/shapely/ctypes_declarations.py
@@ -14,7 +14,7 @@ class allocated_c_char_p(c_char_p):
     '''char pointer return type'''
     pass
 
-EXCEPTION_HANDLER_FUNCTYPE = CFUNCTYPE(None, c_char_p, c_char_p)
+EXCEPTION_HANDLER_FUNCTYPE = CFUNCTYPE(None, c_char_p, c_void_p)
 
 
 def prototype(lgeos, geos_version):
--- a/shapely/geos.py
+++ b/shapely/geos.py
@@ -186,16 +186,16 @@ class PredicateError(Exception):
 
 
 def error_handler(fmt, *args):
-    if sys.version_info[0] >= 3:
-        fmt = fmt.decode('ascii')
-        args = [arg.decode('ascii') for arg in args]
+    fmt = fmt.decode('ascii')
+    conversions = re.findall(r'%.', fmt)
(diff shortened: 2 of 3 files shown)


sgillies:
@snorfalorpagus @mwtoews what do you think? I can't reproduce #243 but this PR does fix bugs in the error and notice handlers.

After this is merged, I'll follow up by merging maint into master.

mwtoews:
I get a TypeError while testing the validity of an invalid polygon:
```python
from shapely.wkt import loads
g = loads('MULTIPOLYGON (((10 20, 10 120, 60 70, 30 70, 30 40, 60 40, 60 70, 90 20, 10 20)))')
g.is_validTraceback (most recent call last):
  File "_ctypes/callbacks.c", line 295, in 'calling callback function'
  File "shapely/geos.py", line 216, in notice_handler
    for spec, arg in zip(conversions, args):
TypeError: zip argument #2 must support iteration
False
```
The error logger message was normally `WARNING:shapely.geos:Ring Self-intersection at or near point 60 70`

snorfalorpagus:
It seems there isn't a cross-platform friendly way to handle `va_list` from ctypes.

I probably wont be able to test this until Wednesday.

sgillies:
@snorfalorpagus no, there isn't.

Adding another c_void_p argument to the callback (https://github.com/Toblerity/Shapely/blob/b8ecce20a4e4a431cefea316ed120b7bee761693/shapely/ctypes_declarations.py#L17) that would be zipped up with the list of conversion specs (and thereby ignored if not needed) might help future proof these handlers. But instead, I think I'll just commit to watching GEOS for the number of args passed to `handle->ERROR_MESSAGE()` and updating as needed.

mwtoews:
Fantastic :+1:

sgillies:
@QuLogic I can't figure out how to get `GEOSRelateBoundaryNodeRule_r()` to cough up a message with `%d`, so will punt on that for now. The main goal, keeping the callbacks from dragging down our programs unnecessarily, is met.

QuLogic:
Sounds reasonable. FYI, I tried out the cartopy test suite, and it no longer triggers the `TypeError` like before. I don't know if the `UnicodeDecodeError` originally reported is fixed, though.

snorfalorpagus:
This looks good to me.

PR #252, opened 2015-03-06
Enable Cython speedups by default (if available)

snorfalorpagus:
This PR addresses the issue(s) raised in https://github.com/Toblerity/Shapely/issues/251

* Fixed a minor bug in `shapely.speedups.disable`
* Enable Cython speedups by default, if available

For instance:
```
>>> import shapely.speedups
>>> shapely.speedups.enabled
True
```

Tests should be called --with-speedups or --without-speedups - this has been amended in `.travis.yml`.

The manual has been updated to reflect the changes.

I've assumed this would be included in 1.6, rather than 1.5.x, although I'm not sure now. Certainly the first commit should be cherry picked into the 1.5 branch. I feel this is worth some good testing, as if it is broken it would break everything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ python:
 
 env:
   - "TRAVIS_SPEEDUP_OPTS=--with-speedups"
-  - "TRAVIS_SPEEDUP_OPTS="
+  - "TRAVIS_SPEEDUP_OPTS=--without-speedups"
 
 before_install:
   - sudo add-apt-repository -y ppa:ubuntugis/ppa
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -2338,13 +2338,12 @@ involves some overhead that might slow down your code.
 .. versionadded:: 1.2.10
 
 The :mod:`shapely.speedups` module contains performance enhancements written in
-C. They are automaticaly installed when Python has access to a compiler and
-GEOS development headers during installation.
+C. They are automatically installed when Python has access to a compiler and
+GEOS development headers during installation. 
 
 You can check if the speedups are installed with the :attr:`available`
-attribute. The constructor speedups are disabled by default. To enable the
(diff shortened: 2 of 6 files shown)


snorfalorpagus:
Hmm. This is passing tests, but dies if `shapely.geometry` or `shapely.coords` is the first thing that is imported. Need to find a better place to put the initial `import shapely.speedups`.

```
Python 3.4.2 (default, Nov 23 2014, 01:27:17) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.54)] on darwin
>>> import shapely.geometry
  File "/Users/snorf/Desktop/Shapely/shapely/geometry/__init__.py", line 4, in <module>
  File "/Users/snorf/Desktop/Shapely/shapely/geometry/base.py", line 9, in <module>
    from shapely.coords import CoordinateSequence
  File "/Users/snorf/Desktop/Shapely/shapely/coords.py", line 8, in <module>
    from shapely.geos import lgeos
  File "/Users/snorf/Desktop/Shapely/shapely/geos.py", line 663, in <module>
    import shapely.speedups
  File "/Users/snorf/Desktop/Shapely/shapely/speedups/__init__.py", line 3, in <module>
    from shapely.geometry import linestring, polygon
  File "/Users/snorf/Desktop/Shapely/shapely/geometry/linestring.py", line 11, in <module>
    from shapely.coords import required
ImportError: cannot import name 'required'
```

snorfalorpagus:
I think this is ready to go, although it would benefit from some testing by other users in some real-world projects. The unit tests aren't a very good test of whether this kind of thing is working as expected.

The version number in the documentation will need to be tweaked, as it has missed inclusion in 1.6.

sgillies:
Agreed about the first commit.

How about this for a plan: have at least one pre-1.6 release and a campaign to get it well tested.

PR #257, opened 2015-03-22
Allow creation of  multipoint from empty list

jinkunw:
`MultiPoint([])` should work like `MultiPolygon([])`.


sgillies:
Thanks!

PR #267, opened 2015-04-20
Fix int check in GeometrySequence.__getitem__

### LINKED ISSUE #266: Indexing multipolygon with numpy's int64
This is python 3.3, shapely 1.5.7, numpy 1.9.2.

```python
import numpy as np
import shapely.geometry as shpg

p1 = [(0, 0), (0, 2), (2, 2), (2, 0), (0, 0)]
p2 = [(-1, -0.1), (-0.5, -0.5), (-1, -1), (-1.5, -0.5), (-1, -0.1)][::-1]
mp = shpg.MultiPolygon([shpg.Polygon(p1), shpg.Polygon(p2)])

mp[np.int64(0)]
```
Throws TypeError: key must be an index or slice

This is relevant because np.argmax() for example now returns np.int64 types

jwass:
Use numbers.Integral to allow for more robust integer index check.

Fixes #266

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -6,9 +6,11 @@ from shapely import speedups
 try:
     import numpy
     numpy_version = numpy.version.version
+    integer_types = [int, numpy.int16, numpy.int32, numpy.int64]
 except ImportError:
     numpy = False
     numpy_version = 'not available'
+    integer_types = None
 
 # Show some diagnostic information; handy for Travis CI
 print('Python version: ' + sys.version.replace('\n', ' '))
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -3,14 +3,14 @@ from shapely.libgeos import geos_version_string
 from shapely.geos import lgeos, WKTWriter
 from shapely import speedups
 
+test_int_types = [int]
 try:
     import numpy
     numpy_version = numpy.version.version
-    integer_types = [int, numpy.int16, numpy.int32, numpy.int64]
+    test_int_types.extend([int, numpy.int16, numpy.int32, numpy.int64])
 except ImportError:
     numpy = False
     numpy_version = 'not available'
-    integer_types = None
 
 # Show some diagnostic information; handy for Travis CI
 print('Python version: ' + sys.version.replace('\n', ' '))
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -5,7 +5,6 @@ import sys
 from warnings import warn
 from binascii import a2b_hex
 from ctypes import pointer, c_size_t, c_char_p, c_void_p
-import numbers
 
 from shapely.coords import CoordinateSequence
 from shapely.ftools import wraps
@@ -15,6 +14,16 @@ from shapely.impl import DefaultImplementation, delegated
 
 if sys.version_info[0] < 3:
     range = xrange
+    integer_types = (int, long)
+else:
+    integer_types = (int,)
+
+try:
+    import numpy as np
+    integer_types = integer_types + (np.integer,)
+except ImportError:
+    pass
+
 
 GEOMETRY_TYPES = [
     'Point',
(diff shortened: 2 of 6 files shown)


mwtoews:
While this is a good start, the `numbers.Integral` test does not seem portable/reliable. E.g., on 64-bit Windows:
```python
isinstance(np.int16(0), numbers.Integral) # False
isinstance(np.int32(0), numbers.Integral)  # True
isinstance(np.int64(0), numbers.Integral)  # False
```
whereas this numpy-specific test appears more reliable, with the exception of the last one:
```python
isinstance(np.short(0), np.integer)  # True
isinstance(np.int16(0), np.integer)  # True
isinstance(np.int32(0), np.integer)  # True
isinstance(np.int64(0), np.integer)  # True
isinstance(np.long(0), np.integer)  # False
```
Also, this fix should fix slices using Python 2.x `long` types, e.g. `0L`. For compatibility with Py3k it will require something like `long = int` to be defined at the top. Or you could borrow [`six`'s `integer_types`](784c6a213c4527ea18f86a800f51bf16bc1df5bc) idea.

mwtoews:
Or, better idea, at the top of base.py, build a `integer_type` tuple:
```python
if sys.version_info[0] < 3:
    range = xrange
    integer_types = (int, long)
else:
    integer_types = (int,)

try:
    import numpy as np
    integer_types = integer_types + (np.integer, np.long)
except ImportError:
    pass
```
and use it
```python
isinstance(0, integer_types)  # True
isinstance(0L, integer_types)  # True
isinstance(np.short(0), integer_types)  # True
isinstance(np.int32(0), integer_types)  # True
isinstance(np.int64(0), integer_types)  # True
isinstance(np.long(0), integer_types)  # True
```

jwass:
Thanks @mwtoews. Just curious what version of numpy you're using on Windows.

I was basing that fix from: https://github.com/numpy/numpy/pull/4547 which looks like it registers all integer dtypes with `numbers.Integral` even on Windows (I don't have it so I can't test myself). I now see that's for numpy 1.9 which might be too strict.

Another alternative is to just wrap `index = int(key)` in a try/except, which seems a little more robust rather than trying to explicitly define everything that might look like an `int`.

Thoughts?

mwtoews:
My Windows NumPy is 1.8.2, from Christoph's [pythonlibs](http://www.lfd.uci.edu/~gohlke/pythonlibs/).

The simple `int(key)` test is good, except that you need to somehow catch types that also can be cast to int, but should not be used for indices, like float. E.g. `[''][0.0]` raises "TypeError: list indices must be integers, not float"

jwass:
@mwtoews Good call, thanks. I basically took your suggestion to explicitly add `np.integer` to the list of integer types. Also updated all the tests to try to index on each of them as well.

tests/__init__.py
```diff
@@ -6,9 +6,11 @@
 try:
     import numpy
     numpy_version = numpy.version.version
+    integer_types = [int, numpy.int16, numpy.int32, numpy.int64]
 except ImportError:
     numpy = False
     numpy_version = 'not available'
+    integer_types = None
```
mwtoews: shouldn't this be `integer_types = [int]`?

jwass: Originally the idea was that `integer_types` was only used in numpy tests. I just refactored this pull request differently though.

jwass:
@mwtoews I decided to tackle this slightly differently now.

The new sub-geometry test case is always run regardless of whether numpy is available. If numpy is available, then `test_int_types` gets appended with the numpy integer types.

Rather than copy the same test case, I started a `MultiGeometryTestCase` that has the base logic for testing sub geometry access. All the Multi-part geometry test cases now inherit from it. Pretty simple class for now, but hopefully will form the base for any more shared logic among the multi-part geometry tests.

More thoughts/objections?

sgillies:
@jwass this fixes #266, right?

jwass:
Yes.
If all good, I'll squash it if preferred.

sgillies:
@jwass No need, I see what's going on. :+1:

Issue #268, opened 2015-04-24
Preparing Polygon Changes Contains Results


bixb0012:
Start with a simple polygon and call contains with an empty polygon:
```python
>>> from shapely import wkt
>>> from shapely import prepared
>>> polygon = wkt.loads('POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))')
>>> polygon_empty = wkt.loads('POLYGON EMPTY')
>>> polygon.contains(polygon_empty)
False
```
So far, so good.  The issue I am running into is the result after preparing the simple polygon:
```python
>>> polygon_prepared = prepared.prep(polygon)
>>> polygon_prepared.contains(polygon_empty)
True
```
Expected behavior?  Bug?

Thanks.

bixb0012:
Forgot to include:
```python
>>> sys.version
'2.7.8 (default, Jun 30 2014, 16:03:49) [MSC v.1500 32 bit (Intel)]'
>>> shapely.__version__
'1.5.6'
```

snorfalorpagus:
I can reproduce this peculiarity using your code. However, the equivalent C seems to behave as expected (returning "false" for both). A bug in Shapely?

```
#include <stdio.h>

#include "geos_c.h"

int main() {
    GEOSWKTReader *reader;
    GEOSGeometry* polygon1;
    GEOSGeometry* polygon2;
    const GEOSPreparedGeometry* prepared1;
    
    initGEOS(NULL, NULL);
    
    reader = GEOSWKTReader_create();
    
    polygon1 = GEOSWKTReader_read(reader, "POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))");
    polygon2 = GEOSWKTReader_read(reader, "POLYGON EMPTY");
    
    char a = GEOSContains(polygon1, polygon2);
    if (a == 0) {
        printf("false\n");
    } else if(a == 1) {
        printf("true\n");
    } else {
        printf("error\n");
    }
    
    prepared1 = GEOSPrepare(polygon1);
    
    char b = GEOSPreparedContains(prepared1, polygon2);
    if (b == 0) {
        printf("false\n");
    } else if(b == 1) {
        printf("true\n");
    } else {
        printf("error\n");
    }
    
    return 0;
}
```

snorfalorpagus:
This may be a bug with the Windows build of `libgeos`. On OS X (homebrew) I get:

```
Python 3.4.3 (default, Apr  7 2015, 08:05:21) 
[...]

PR #269, opened 2015-04-25
Corrected restype on GEOSPrepared functions (c_int -> c_byte)

snorfalorpagus:
This fixes a bug with prepared geometries highlighted in issue #268. The `restype` was incorrectly set to `c_int` rather than `c_byte`. This caused unexpected results on Windows only, although I'm not sure why.


snorfalorpagus:
There are a few more issues I think should be addressed before this is merged:

* Predicate methods on prepared geometries are missing docstrings.
* `errcheck_predicate` should be used for prepared geometries as it is for regular geometries, but isn't currently.
* Not all predicates are implemented (e.g. `covers` is implemented for prepared geometries but not regular geometries, whereas `touches` is implemented for regular geometries but not prepared geometries).

snorfalorpagus:
I think that's everything. Ready for review.

sgillies:
@snorfalorpagus Thanks!

bixb0012:
@snorfalorpagus, thanks from me as well.  I look forward to working with these changes.

PR #270, opened 2015-04-26
Use OffsetCurve instead of SingleSidedBuffer in GEOS >= 3.3

snorfalorpagus:
This PR switches to using `GEOSOffsetCurve` instead of `GEOSSingleSidedBuffer` which was deprecated in GEOS 3.3.0. Existing code shouldn't notice any change, although the requirement that the `distance` parameter be positive has been removed; instead, if the distance is negative the `side` parameter is inverted. The `side` parameter is now optional, as it can be defined implicitly by the polarity of `distance`. This is all handled in `geos.py` to keep version specific details outside of `linestring.py`.

https://github.com/libgeos/libgeos/blob/aa57155d18f8cbb33b5090c5ea4bae349abf4db2/capi/geos_ts_c.cpp#L1800

### DIFF
--- a/shapely/ctypes_declarations.py
+++ b/shapely/ctypes_declarations.py
@@ -129,8 +129,16 @@ def prototype(lgeos, geos_version):
         lgeos.GEOSBufferWithStyle.restype = c_void_p
         lgeos.GEOSBufferWithStyle.argtypes = [c_void_p, c_double, c_int, c_int, c_int, c_double]
 
-        lgeos.GEOSSingleSidedBuffer.restype = c_void_p
-        lgeos.GEOSSingleSidedBuffer.argtypes = [c_void_p, c_double, c_int, c_int, c_double, c_int]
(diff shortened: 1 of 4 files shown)


snorfalorpagus:
@sgillies I hadn't thought much about sidedness. The reason I chose left as the default is that seems to be the default of the `GEOSOffsetCurve` function; if you give it a positive distance it creates a line on the left hand side.

https://github.com/libgeos/libgeos/blob/e95fcc7d8b9b6b58ae1a04ff889bcf62c0c0b15b/capi/geos_ts_c.cpp#L1778

Although the underlying C++ function seems not to be biased either way, as it can computer both the left and right *at the same time*.

https://github.com/libgeos/libgeos/blob/svn-trunk/src/operation/buffer/OffsetCurveBuilder.cpp#L107

However, I'd agree that having right as the default seems to fit better with the right hand rule. I don't have a strong opinion either way.

PR #274, opened 2015-05-19
Compatibility with 64 bit

thorstenkranz:
I recommend this change i.o.t. make it really compatible with 64 bit systems. 

Currently, _speedups might fail when the address for your handles becomes to large for 32-bit long. So all situations when this solution works are pure luck (which happens very often, almost always actually), but might fail under certain situations.

We observed this problem when integrating Python within a .NET-application. This led to larger pointer addresses, and thus to failing casts for 32bit long.

Remark: I cannot fully estimate the impact of this change for other operating systems / processor architectures.

### DIFF
--- a/shapely/speedups/_speedups.pyx
+++ b/shapely/speedups/_speedups.pyx
@@ -15,15 +15,15 @@ from shapely.geometry.base import geom_factory
 include "../_geos.pxi"
     
 
-cdef inline GEOSGeometry *cast_geom(unsigned long geom_addr):
+cdef inline GEOSGeometry *cast_geom(unsigned long long geom_addr):
     return <GEOSGeometry *>geom_addr
 
 
-cdef inline GEOSContextHandle_t cast_handle(unsigned long handle_addr):
+cdef inline GEOSContextHandle_t cast_handle(unsigned long long handle_addr):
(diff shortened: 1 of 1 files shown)


sgillies:
@thorstenkranz looks good to me, and Travis gives it a green light, but we're not testing across many archs. @jwass, I defer to your judgement on this. Do you any 32 bit systems to try it on?

snorfalorpagus:
Should we be using `uintptr_t` instead?

http://stackoverflow.com/questions/1845482/what-is-uintptr-t-data-type

> In C99, it is defined as "an unsigned integer type with the property that any valid pointer to void can be converted to this type, then converted back to pointer to void, and the result will compare equal to the original pointer".

thorstenkranz:
I think both ways would be o.k., but your proposal is the more correct way of doing it.

thorstenkranz:
Any news regarding this one? Or rather: any decissions?

sgillies:
@thorstenkranz I'll close this one in favor of #276 and get that merged to maint (for 1.5.8).

sgillies:
1.5.9 is up on PyPI now.

thorstenkranz:
Nice! Thank you!

Issue #294, opened 2015-07-24
Relate does not handle exceptions


mwtoews:
Here's an example with an invalid Polygon:
```python
from shapely.wkt import loads as load_wkt
a = load_wkt(
    'POLYGON ((40 100, 80 100, 80 60, 40 60, 40 100), (60 60, 80 60, 80 40, 60 40, 60 60))')
a.relate(a)  # oh no!
```
on Windows with Shapely 1.5.9, GEOS 3.4.2-CAPI-1.8.2 r3921, it shows

> Traceback (most recent call last):
> WindowsError: exception: access violation reading 0x000000000000000F

and on Debian with Shapely 1.5.8, GEOS 3.3.3-CAPI-1.7.4, it is dead with Segmentation fault.

However, it should probably raise a TopologicalError, like `a.touches(a)` does:
> shapely.geos.TopologicalError: The operation 'GEOSTouches_r' could not be performed. Likely cause is invalidity of the geometry <shapely.geometry.polygon.Polygon object at 0x7fb1e5e4f090>

I should also note that I've checked all the other binary operators, and most of them raise TopologicalError, except for `distance`, which is 0.0, and that's ok. So only `relate` is at fault.

snorfalorpagus:
It looks like GEOSRelate returns a NULL pointer on exception, which we don't check for.

https://github.com/libgeos/libgeos/blob/svn-trunk/capi/geos_c.h.in#L852

I think we need to add a check for this in `errcheck_just_free`.

mwtoews:
The issue is effectively `string_at(None)`, which I've [reported upstream](http://bugs.python.org/issue24714).

And yes, part of the fix is in `errcheck_just_free`, to check `if result.value is None`. But then somehow raise a TopologicalError for relate.

sgillies:
Closed via #303 and #304.

PR #295, opened 2015-07-27
GEOSRelate_r exceptions handled gracefully. Fix for #294.

snorfalorpagus:
This PR addresses issue #294. `GEOSRelate_r` returns a NULL pointer on exception, which isn't currently caught by Shapely resulting in a segmentation fault when an attempt is made to access the result. A wrapper has been added around the existing errcheck function, which checks for this and raises a `TopologicalError` if needed.

### DIFF
--- a/shapely/geos.py
+++ b/shapely/geos.py
@@ -414,6 +414,12 @@ def errcheck_just_free(result, func, argtuple):
     else:
         return retval
 
+def errcheck_null_exception(result, func, argtuple):
+    '''Wraps errcheck_just_free, raising a TopologicalError if result is NULL'''
+    if not result:
+        raise TopologicalError("The operation '{}' could not be performed. \
+Likely cause is invalidity of the geometry.".format(func.__name__))
+    return errcheck_just_free(result, func, argtuple)
 
 def errcheck_predicate(result, func, argtuple):
     '''Result is 2 on exception, 1 on True, 0 on False'''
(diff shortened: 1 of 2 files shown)


shapely/geos.py
```diff
@@ -414,6 +414,12 @@ def errcheck_just_free(result, func, argtuple):
     else:
         return retval
 
+def errcheck_null_exception(result, func, argtuple):
+    '''Wraps errcheck_just_free, raising a TopologicalError if result is NULL'''
+    if not result:
+        raise TopologicalError("The operation '{0}' could not be performed. \
```
sgillies: Can you deal with this long string using implicit string concat instead of a forward slash?

Also I prefer docstrings to be triple double-quotes like in the examples in https://www.python.org/dev/peps/pep-0257.

sgillies:
I'm all in favor of this. Thanks, @snorfalorpagus!

mwtoews:
Yup, these were essentially the changes I was thinking of.

shapely/geos.py
snorfalorpagus: Sure. I'm aware of this convention in Shapely, although the little group of functions around that section don't follow it. I might as well correct them all while I'm in there.

sgillies:
@snorfalorpagus my fault for branching before merging this, but can you open a new PR from your branch with maint-1.5 as the target branch? I'll merge it as soon as you do.

Issue #57, opened 2013-06-21
Polygonize errors: cuts, dangles and invalid rings


pconstr:
It would be very useful for Polygonize to report cuts, dangles and invalid rings.
It looks like this is available in the C API.

PR #301, opened 2015-08-17
Allow tests to work with NumPy 1.4.1

mwtoews:
I regularly use a CentOS 6 server with NumPy 1.4.1 (released 22 Apr 2010), which is [the current version](http://mirror.centos.org/centos/6/os/x86_64/Packages/). This PR drops the minimum-supported NumPy version and has a few modifications to enable py.test to pass with speedups enabled.

Here is the specific rational for the changes:
 * The `copy` keyword for `astype()` was introduced in 1.7.0 (e.g. see [ndarray.astype(t)](http://docs.scipy.org/doc/numpy-1.6.0/reference/generated/numpy.ndarray.astype.html#numpy.ndarray.astype) from Numpy 1.6.0)
 * Before 1.6.0, `copy()` only supported positional arguments, and `order='f'` would raise "TypeError: copy() takes no keyword arguments"; see [bug #2178](https://github.com/numpy/numpy/issues/2178)
 * The modification to the cdef bool result array is for "ValueError: unknown dtype code in numpy.pxd (0)", which is a fix [similar to this one on SO](http://stackoverflow.com/a/29431979/327026). The issue could be similar to [pandas #57](https://github.com/pydata/pandas/issues/57) fixed by Numpy 1.5.1.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,5 @@
 setuptools
-Numpy>=1.8.0
+Numpy>=1.4.1
 Cython>=0.19
 descartes==1.0.1
 packaging
--- a/shapely/vectorized/_vectorized.pyx
+++ b/shapely/vectorized/_vectorized.pyx
@@ -75,10 +75,12 @@ cdef _predicated_elementwise(geometry, x, y, predicate fn):
     if x.shape != y.shape:
         raise ValueError('X and Y shapes must be equivalent.')
 
-    x_1d = x.astype(np.float64, copy=False).ravel()
-    y_1d = y.astype(np.float64, copy=False).ravel()
+    if x.dtype != np.float64:
+        x = x.astype(np.float64)
+    if y.dtype != np.float64:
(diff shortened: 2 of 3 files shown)


sgillies:
@mwtoews LGTM, but I'd like to be certain that we're not switching back to deprecated usage. Can you check on `copy()`?

mwtoews:
I don't believe there is any deprecated usage for copy(), except that it has always had one positional argument `order`. Here's what I get, showing the Numpy bug:
```
$ python -W all
Python 2.6.6 (r266:84292, Jan 22 2014, 09:42:36)
[GCC 4.4.7 20120313 (Red Hat 4.4.7-4)] on linux2
>>> import numpy as np
>>> x = np.array([[1,2,3],[4,5,6]])
>>> x.copy('f')
array([[1, 2, 3],
       [4, 5, 6]])
>>> x.copy(order='f')
TypeError: copy() takes no keyword arguments
```
which was fixed by NumPy 1.6.0

PR #308, opened 2015-08-25
Fix tests on windows:

mindw:
* there are no .so files in win32
* resetlocale() raises exception on windows http://bugs.python.org/issue10466


sgillies:
Thanks!

PR #313, opened 2015-08-26
Don't configure root logger

### LINKED ISSUE #312: Please refrain from calling logging.basicConfig
From here https://github.com/Toblerity/Shapely/blob/master/shapely/libgeos.py#L22

Else for why, b/c this screw up with everyone logging setup, changing root logger should never be done by a library :)

mouadino:
fixes #312


sgillies:
@mouadino thank you, this is very welcome. I've made the change in the maint-1.5 branch (for 1.5.12) and have also merged that to master (for 1.6+).

Issue #209, opened 2014-12-03
Geometries no longer hashable in Python 3


jwass:
After 1bf2b40b36077ad3bc122372216e483d87a2af52 , which implements `__eq__()` for geometries, geometries are no longer hashable in Python 3. This arose from testing GeoPandas with the latest Shapely. A test case performs a DataFrame merge on the geometry column, which requires hashable values.

The relevant new info in Python 3 from https://docs.python.org/3/reference/datamodel.html#object.__hash__:
"A class that overrides `__eq__()` and does not define `__hash__()` will have its `__hash__()` implicitly set to `None`"
(compare to Python 2 docs https://docs.python.org/2/reference/datamodel.html#object.__hash__)

We could:
- Set `__hash__ = (parent class).__hash__` on any object that implements `__eq__()` to return to the old behavior. Are there any side effects if `x == y` but `hash(x) != hash(y)`?
- Argue that geometries shouldn't ever have been hashable as they are technically mutable (although I'd bet there are few applications that do this). This would make it a Pandas/GeoPandas problem (and you probably shouldn't be merging on the geometry column anyway). If so, we should explicitly set `__hash__` to `None` to backport the behavior to Python 2. And note that it will break any code that hashed geometries previously, such as using geometries as keys to a dictionary
- Make our own custom hash function for geometries that's probably some combination of the geometry type and coordinates - still has the mutability issue.
- Anything else I haven't thought of or don't know about.

I think I'd vote for the first one, but I'm not completely sold.

danielwe:
I happened to bump into this issue today. As explained by @demosdemon, the current behavior is inconsistent. From the python docs (https://docs.python.org/3/reference/datamodel.html#object.__hash__):
> \_\_hash\_\_() should return an integer. The only required property is that objects which compare equal have the same hash value; it is advised to somehow mix together (e.g. using exclusive or) the hash values for the components of the object that also play a part in comparison of objects.

and

> If a class defines mutable objects and implements an \_\_eq\_\_() method, it should not implement \_\_hash\_\_(), since the implementation of hashable collections requires that a key’s hash value is immutable (if the object’s hash value changes, it will be in the wrong hash bucket).

Thus, shapely should either
- embrace mutablility as a property of geometries, and set `__hash__ = None` in the relevant (base) classes, or
- define geometries as immutable, and develop suitable custom hash methods.

I don't think creating custom hash methods has to be difficult. For example, the `BaseGeometry` class could perhaps get away with something like
```python
[...]

PR #320, opened 2015-09-06
Make geometries unhashable

sgillies:
Shapely's geometries are mutable, but we're providing a hash function. These two features are inconsistent. Rather than remove mutability (for now) we'll remove the hashability.

See #209

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -270,7 +270,7 @@ class BaseGeometry(object):
     def __ne__(self, other):
         return not self.__eq__(other)
 
-    __hash__ = object.__hash__
+    __hash__ = None
 
     # Array and ctypes interfaces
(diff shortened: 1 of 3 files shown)


sgillies:
@jwass I'm revisiting the change I made in #211, which seems wrong. I'd love your assessment on the effects of this PR on GeoPandas et al.

jwass:
Will do.

jwass:
@sgillies See geopandas/geopandas#221.

As far as GeoPandas goes, I think this only affects anyone that has a merge/join that includes the geometry column in the list of fields to join. My guess is it's rare to have done so intentionally but I proposed additional functionality that would explicitly support that operation, so they won't be left out in the cold.

Once geopandas/geopandas#222 is merged, then the GeoPandas tests will pass against this change.

Issue #324, opened 2015-10-02
Problems with importing Shapely after import of another module that dlopens GEOS


sgillies:
Originally reported in https://github.com/Toblerity/Fiona/issues/276. This feels like a problem with ctypes. TODO: 

- [x] research on ctypes and dlopen and why you can reliably import shapely after fiona on Linux, but not on some OS X systems.
- [x] experiment with cffi instead of ctypes.

sgillies:
Data point: the bug (empty geometry collection from unary union) *does not* show up when Fiona and Shapely access the very same dylib. It *does* show up when each of them accesses differently *delocated* copies of the same dylib.

sgillies:
Closed via #327.

PR #327, opened 2015-10-04
Restore shapely/geos.py to state at 1.5.9

### LINKED ISSUE #325: GEOS shared object location and loading regressions
I think we've made the situation worse in releases 1.5.10+. For example, if you have GEOS installed on your OS X system, and `pip install` one of the batteries-included wheels from PyPI, the system's GEOS .dylib is going to be loaded instead of the wheel's .dylib.

I propose to largely roll back the changes made to setup.py, shapely/geos.py, and shapely/libgeos.py since 1.5.9. Then I will replace shapely/libgeos.py with a module that is only ever imported by setup.py. I'll feel a lot better if `geos-config` is never executed except when we're running setup.py.

sgillies:
And move shapely/libgeos.py to shapely/_buildcfg.py. Now it's only ever imported by setup.py when building distributions.

Closes #325.

Status: 30%


sgillies:
This PR also fixes #324. The script in https://github.com/Toblerity/Fiona/issues/276 succeeds now, no matter whether you import fiona or shapely first. If fiona has already been imported, we try to reuse a DLL we find in its `.dylibs` directory in shapely.

@snorfalorpagus @mwtoews I would very much appreciate your :eyes: on this. Getting a 1.5.13 release out soon would be a huge relief for me. This issue has been bugging me for tooooo long.

/cc @schwehr because you've recently filed a related issue.

sgillies:
And now we reuse any GEOS DLL that we can find already loaded on OS X, not just when the shapely module has its own `.dylibs`.

sgillies:
Barring objections, I'll merge this after lunch and release 1.5.13.

PR #334, opened 2015-10-20
Fix shapely version check

nerdcha:
With the current master, running `python setup.py install` gives

    TypeError: unorderable types: Version() >= tuple()

on the shapely version check, but this change seems to fix it. That's on Python 3.4.


sgillies:
Thanks!

PR #342, opened 2015-11-13
the os package has no __file__ attribute when frozen using cx_freeze.

danaukes:
When you try to freeze shapely, an exception is thrown because this particular piece of code doesnt' work when frozen.  I've tested this on a windows machine in both the frozen and unfrozen versions, and it seems to work.

### DIFF
--- a/shapely/geos.py
+++ b/shapely/geos.py
@@ -131,7 +131,7 @@ elif sys.platform == 'win32':
     try:
         egg_dlls = os.path.abspath(os.path.join(os.path.dirname(__file__),
                                    "DLLs"))
-        wininst_dlls = os.path.abspath(os.__file__ + "../../../DLLs")
+        wininst_dlls = os.path.normpath(os.path.abspath(sys.executable+'../../DLLS'))
         original_path = os.environ['PATH']
         os.environ['PATH'] = "%s;%s;%s" % \
             (egg_dlls, wininst_dlls, original_path)


sgillies:
@danaukes I'm hesitant to replace the old code, which was working for the contributor, when we have no tests for the new code (or the old, for that matter). Would you be willing to add a test that runs on Travis using cx_freeze?

danaukes:
I've never used travis, and am not sure how to begin.  cx_freeze creates
executable installers for python projects, so making a test will involve
creating a test project that uses shapely, creating an executable, and then
running it in a way that loads and runs some shapely methods while catching
any errors that result in that process.  I can't imagine that is your
vanilla test case for travis, and as much as I'd like to learn, I'm not
going to have a chance to sit down and work on it for a while.  The issue
is that because it is a system level package, the OS package does not carry
all its attributes along with it when "frozen" into an executable, causing
the error I mentioned.

Dan

On Fri, Nov 13, 2015 at 6:43 PM, Sean Gillies <notifications@github.com>
wrote:

> @danaukes
> <https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_danaukes&d=CwMCaQ&c=WO-RGvefibhHBZq3fL85hQ&r=yjnTLiBlKr4W189mDVx3e7kaEB8U4ZXO5uHigeERe7g&m=dfIblaqt_9cocOdJlJKBVs3AjhAVLEir3V83uRL-TZw&s=Sjkc5TtXRzAMgbOppDxTFUjtKHBD5E-5Peg5sJCjKtI&e=>
> I'm hesitant to replace the old code, which was working for the
> contributor, when we have no tests for the new code (or the old, for that
> matter). Would you be willing to add a test that runs on Travis using
> cx_freeze?
>
> —
> Reply to this email directly or view it on GitHub
> <https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_Toblerity_Shapely_pull_342-23issuecomment-2D156613292&d=CwMCaQ&c=WO-RGvefibhHBZq3fL85hQ&r=yjnTLiBlKr4W189mDVx3e7kaEB8U4ZXO5uHigeERe7g&m=dfIblaqt_9cocOdJlJKBVs3AjhAVLEir3V83uRL-TZw&s=KVntX2Mjf1c-w1F4N-rWOjw12kiLXBj2NNVKi_w7LmY&e=>
> .
>



-- 
Dan Aukes
danaukes@seas.harvard.edu

Please note that my work email address will be changing to danaukes@asu.edu
this winter.  Please update your address book.

snorfalorpagus:
I think this would require writing a simple cx_Freeze application, building and testing it on Travis. To do this you need a `setup.py` (see https://cx-freeze.readthedocs.org/en/latest/distutils.html#distutils) and a simple application script which could be as simple as `from shapely.geometry import Point; print(Point(1,2).wkt)`. The compiled version then gets called from Travis after the usual call to pytest. Presumably Travis catches errors by executable returning non-zero so this should be enough I think. The main point is that it's not just a standard test with pytest. Where should this little application live? `test_cxfreeze` subfolder? Or perhaps in an `examples` folder, where it could serve two purposes?

sgillies:
It seems there's some consensus in https://bugs.python.org/issue21736 that using `__file__` like we (specifically I) do in Shapely is wrong. I think I'll be able to demonstrate that the above patch works without having to explicitly test with cx_Freeze.

sgillies:
Any of you on Windows tried my suggestion :arrow_up: yet?

danaukes:
I tested on my own machine, and your suggestion works, but unfortunately am not able to spend the time to figure out how to get a test set up to freeze a shapely exe in travis and subsequently test it.

sgillies:
Superceded by #376

PR #348, opened 2015-12-21
Filter empty geometries in STRtree to avoid segfaults. Fixes #345.

### LINKED ISSUE #345: STRtree with empty geometries segfaults
When constructing an STRtree with an empty geometry, any query against that will cause a segfault as GEOS attempts to dereference null bounds for the empty geometry. It seems like empty geometries should be skipped?

Example:

```python
>>> from shapely.geometry import Polygon
>>> from shapely.strtree import STRtree
>>> polys = [ Polygon() ]
>>> tree = STRtree(polys)
>>> tree.query(Polygon([(0,0),(1,1),(2,0),(0,0)]))
Segmentation fault (core dumped)
```

(Tested with Python 2.7.6 on Ubuntu 14.04.3 with Shapely 1.5.13)

snorfalorpagus:
This PR fixes the issue with empty geometries and STRtree in #345.

I think I've done the right thing aiming this at the maint-1.5 branch but correct me if I've done it wrong?

### DIFF
--- a/shapely/strtree.py
+++ b/shapely/strtree.py
@@ -31,6 +31,8 @@ class STRtree:
     """
 
     def __init__(self, geoms):
+        # filter empty geometries out of the input
+        geoms = [geom for geom in geoms if not geom.is_empty]
         self._n_geoms = len(geoms)
         # GEOS STRtree capacity has to be > 1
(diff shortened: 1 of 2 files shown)


sgillies:
Perfect. Thanks, @snorfalorpagus!

PR #349, opened 2015-12-21
cx_Freeze

snorfalorpagus:
As discussed in #342, a test for cx_Freeze and Shapely. *Work in progress!*

I'm not sure what's going on with the 2.x builds not working.

Also not sure if this is the right thing to be doing. Where do we draw the line? What about py2exe and PyInstaller? Not sure if this should be included at all.


sgillies:
@snorfalorpagus let's not tie Shapely to cx_Freeze. I'm going to look for another solution. More comments in #342.

PR #350, opened 2015-12-22
trim=False in .wkt property

### LINKED ISSUE #339: Inconsistency between geometry.wkt and wkt.dumps(geometry)
Correct me if I am wrong, but I think there is still an inconsistency in the wkt/shapely-object/wkt roundtrip.

If I do not set ```WKTWriter.defaults['trim'] = False``` and then say:

```
from shapely import geometry
line = geometry.LineString([(13.51039642756912, 52.598912814414675), (13.525173800277184, 52.60620240344557)])
print line.wkt
print wkt.dumps(line)
```

I am getting two different results:

```
LINESTRING (13.51039642756912 52.59891281441467, 13.52517380027718 52.60620240344557) 
LINESTRING (13.5103964275691197 52.5989128144146747, 13.5251738002771837 52.6062024034455717)
```

A side question: Why is it that even in the wkt.dumps version of the result, the value of the original wkt coordinate has some added digits (52.598912814414675 --> 52.5989128144146747) ?

sgillies:
Closes #339

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -348,9 +348,9 @@ class BaseGeometry(object):
         return geom_to_wkt(self)
 
     @property
-    def wkt(self, **kw):
+    def wkt(self):
         """WKT representation of the geometry"""
-        return WKTWriter(lgeos, **kw).write(self)
+        return WKTWriter(lgeos, trim=False).write(self)
 
     @property
     def wkb(self):


sgillies:
@mwtoews you okay with this?

mwtoews:
@sgillies the consequence of this change would show only the untrimmed output (e.g. `POINT (1.0000000000000000 2.0000000000000000)`), without any option to change it to trimmed (`POINT (1 2)`).

A better option is to either change the [`WKTWriter.defaults['trim']` value] (https://github.com/Toblerity/Shapely/blob/0ed992fdd84386755da4c2bc9c260f25f0399bb0/shapely/geos.py#L159), or to recommend to end users to change the default themselves.
```python
from shapely import geos
geos.WKTWriter.defaults['trim'] = False
print(Point(1, 2))  # POINT (1.0000000000000000 2.0000000000000000)
```
Also, the default for `wkt.dumps` is already `trim=False`, so users that require full precision could just use that. I suppose that I'd rather not see excessive default outputs from the wkt property.

sgillies:
Closing this one up. I think we're fine.

Issue #338, opened 2015-10-27
Constructing Empty Polygon Creates Empty Geometry Collection, or something else...


bixb0012:
Looking at issue #332 got me stuck down another rabbit hole.  I haven't had time to fully look into it, but something is off with constructing empty geometries.
```python
>>> from shapely.geometry import Polygon, GeometryCollection
>>> from shapely.wkt import loads
>>> empty_poly_wkt = loads('POLYGON EMPTY')
>>> empty_poly_wkt.wkt
'POLYGON EMPTY'
>>> isinstance(empty_poly_wkt, Polygon)
True
>>> isinstance(empty_poly_wkt, GeometryCollection)
False
>>> 
>>> empty_poly_none = Polygon()  #or Polygon(None)
>>> empty_poly_none.wkt
'GEOMETRYCOLLECTION EMPTY'
>>> isinstance(empty_poly_none, Polygon)
True
>>> isinstance(empty_poly_none, GeometryCollection)
False
>>> 
>>> empty_poly_wkt.wkb
'\x01\x03\x00\x00\x80\x01\x00\x00\x00\x00\x00\x00\x00'
>>> empty_poly_none.wkb
'\x01\x07\x00\x00\x00\x00\x00\x00\x00'
>>> 
```
It is clear that constructing an empty geometry from well-known text creates a different geometry than constructing an empty geometry using the default class constructor.  The results from well-known text seem correct while results from the default class constructor seems off, i.e., an empty polygon shouldn't have a WKT of empty geometry collection.

sgillies:
@bixb0012 Shapely doesn't currently distinguish between a Polygon that has an empty set of points and any other kind of geometry with an empty set of points. To me, the distinction isn't meaningful. If you have no points – nothing – what's the useful distinction between a point-shaped nothingness and a polygon-shaped nothingness?

I settled on `'GEOMETRYCOLLECTION EMPTY'` as the most generic nothingness, but I wouldn't be opposed to a PR that had more specifically typed nothingnesses that better matched other programs.

bixb0012:
Regardless of the broader philosophical debate/question over emptiness and empty geometries, defaulting to empty geometry collections for all empty geometries appears to create inconsistencies or discrepancies between certain functions/checks.

If I can find some time over this holiday season, I may take you up on that PR suggestion.

snorfalorpagus:
I'm reopening this as I've done some exploration of the issue and I think it's worth further discussion.

[...]

PR #352, opened 2015-12-23
Improvements to handling of empty geometries

snorfalorpagus:
*For discussion - see #338*.

This PR moves to different representations of empty geometries in WKT/WKB, and also includes a fix for comparison of empty Polygon geometries. The tests include checking roundtrip of all geometries with the `pickle` module (which is currently broken - see #338), and tests for the empty Polygon comparison issue.

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -165,7 +165,13 @@ class JOIN_STYLE(object):
     mitre = 2
     bevel = 3
 
-EMPTY = deserialize_wkb(a2b_hex(b'010700000000000000'))
+EMPTY_LINESTRING = deserialize_wkb(a2b_hex(b'010200000000000000'))
+EMPTY_POLYGON = deserialize_wkb(a2b_hex(b'010300000000000000'))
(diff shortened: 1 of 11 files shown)


sgillies:
@snorfalorpagus everything looks good to me, but before I merge I'd like to see the intent of this PR clearly spelled out.

Currently, Shapely is consistent with PostGIS, or at least with 2010 PostGIS as [Mateusz wrote](http://mateusz.loskot.net/posts/2010/02/26/sqlgeometry-and-point-empty-in-wkb/):

> A word about how PostGIS behaves. PostGIS reports GEOMETRYCOLLECTION EMPTY, regardless of actual type of input EMPTY geometry.

In other words: what exactly is the win from changing our empty representation?

snorfalorpagus:
I'm leaning towards leaving things the way they are, because of the issue with there being no representation of `POINT EMPTY` in WKB, i.e. to keep consistency with PostGIS. I think what I'd like to do is pull the fix for comparing empty polygons out of this PR, and adding some comments into the manual about this issue and the reason for the decision.

PR #355, opened 2016-01-11
Fix for equals comparison with empty polygons

snorfalorpagus:
This PR fixes a bug triggered when comparing the equality of empty Polygons. The accompanying test illustrates the issue. The problem is that an empty polygon doesn't have any coordinates, so `self.exterior.coords` in the existing algorithm falls over.


sgillies:
:pray:

Issue #354, opened 2016-01-11
Minimum area bounding rectangle


Sklavit:
Please, add computation of the minimum area bounding rectangle.
Possible implementations can be found at:
http://stackoverflow.com/questions/13542855/python-help-to-implement-an-algorithm-to-find-the-minimum-area-rectangle-for-gi?rq=1
https://github.com/dbworth/minimum-area-bounding-rectangle/tree/master/python
https://geidav.wordpress.com/2014/01/23/computing-oriented-minimum-bounding-boxes-in-2d/

snorfalorpagus:
The algorithm given by whuber in this GIS.SE answer looks nice to me: http://gis.stackexchange.com/a/22934/12420. Only 12 lines and no trig functions. With the existing convex hull method provided by GEOS this should be possible to implement.

georgeouzou:
I think computing the mbr is already accomplished by geometry.envelope by geos

snorfalorpagus:
@georgeouzou The geometry.envelope method only calculates the MBR with sides parallel to the coordinate axis. These methods find the smallest rectangle, with rotation.

snorfalorpagus:
It looks like GEOS already has an algorithm for this. It should be trivial to implement.

```
extern GEOSGeometry GEOS_DLL *GEOSMinimumRotatedRectangle(const GEOSGeometry* g);
```

https://github.com/libgeos/libgeos/blob/a291393068a45150313140d57972d9c914e21f21/capi/geos_c.h.in#L1475

I'll do this when I find a little time.

georgeouzou:
@Sklavit  I just submitted PR #361 that adds the new functionality to compute the rotated minimum rectangle. I thought this might be of interest to you.

sgillies:
Merged #361

BebeSparkelSparkel:
Hey I think this is great to add. Unfortunately I already wrote the code for [Minimum Bounding Box](https://github.com/BebeSparkelSparkel/MinimumBoundingBox) but I also use the [Smallest Enclosing Circle](https://github.com/BebeSparkelSparkel/SmallestEnclosingCircle) and it would be nice if that could be added.

Lastly, I noticed that the bounds method does not cache the result and if called a lot it slows down the execution time a lot. It would be nice if the bounds method, rotated minimum rectangle, and if added smallest enclosing circle could have their results cached to save execution time.

PR #361, opened 2016-02-12
Add minimum_rotated_rectangle function.

georgeouzou:
I tried adding the functionality requested in #354  to enhance the base geometry object. 
This function enables the computation of a minimum bounding rectangle that is not constrained to be parallel to the axes but can be rotated.
The solution is based on the algorithm proposed in this [link](http://gis.stackexchange.com/questions/22895/how-to-find-the-minimum-area-rectangle-for-given-points/22934#22934) that was provided by @snorfalorpagus

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -2,15 +2,18 @@
 """
 
 import sys
+import math
 from warnings import warn
 from binascii import a2b_hex
 from ctypes import pointer, c_size_t, c_char_p, c_void_p
+from itertools import islice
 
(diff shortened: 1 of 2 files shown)


georgeouzou:
mbr
Here is an example of the results of the function in matplotlib.

sgillies:
@georgeouzou this is great! I really appreciate the clear code and tests :clap:

georgeouzou:
My pleasure to contribute to such an awesome library :)

PR #365, opened 2016-02-18
Implemented GEOSSharedPaths.

snorfalorpagus:
This PR implements the GEOSSharedPaths operation, which arrived in GEOS 3.3.x.

The function finds shared parts between two lineal geometries.

I've implemented this as `shapely.ops.shared_paths` rather than `LineString.shared_paths`. An argument could be made for either I guess.


sgillies:
:rocket:

PR #366, opened 2016-02-18
Implemented GEOSHausdorffDistance

snorfalorpagus:
This PR implements GEOHausdorffDistance. In simple terms the hausdorff distance is the furthest distance between any point on one geometry to any point on another.

https://en.wikipedia.org/wiki/Hausdorff_distance

Available in GEOS 3.2.x.


sgillies:
:pray:

Issue #293, opened 2015-07-23
ST_Split


snorfalorpagus:
PostGIS implements a function called `ST_Split`, which can split a polygon by a line (as well as line by line, or line by point): http://postgis.net/docs/ST_Split.html

Is this something we'd like to be able to do with Shapely?

I think there is an issue with the fact PostGIS is licence under GPL. Given that PostGIS also uses GEOS it might be difficult to write anything that wasn't too close to a straight copy, even though the algorithm isn't especially complicated.

sgillies:
@snorfalorpagus sure, I welcome functions like this in Shapely. Don't copy the PostGIS code and all is fine. Testing that the process is reversible using Shapely's `union` would be good to do.

snorfalorpagus:
We don't currently wrap `GEOSRelatePatternMatch`. Any suggestions on where this should go?

`g1.relate_match(g2, '1********') == True` ?

snorfalorpagus:
Relate was implemented 36b0e8b24541de916f48791734eacb27ddb71d4e, but I've not found the time to finish `ST_Split` yet.

georgeouzou:
I could implement the function, if you have not the time @snorfalorpagus

snorfalorpagus:
@georgeouzou Sure, give it a go!

https://gist.github.com/snorfalorpagus/46ef62d515e28f0a9e9f

That's as far as I got.

snorfalorpagus:
Closed by #371.

PR #371, opened 2016-03-09
Add shapely.ops.split function.

georgeouzou:
This branch provides the functionality for splitting a geometry by another geometry as requested on #293,
based on the implementation of ST_Split used in PostGIS. 
I added the function **shapely.ops.split**, some tests and the documentation.


sgillies:
:bow:

PR #382, opened 2016-04-24
PyInstaller's OS X-specific bug fix

giumas:
PyInstaller's OSX-specific bug fix that add an additional fallback path to geos_c library using _MEIPASS related to https://github.com/HDFGroup/hdf-compass/issues/79


sgillies:
:+1:

PR #386, opened 2016-05-24
Raise PicklingError when attempting to pickle a prepared geometry.

snorfalorpagus:
It's not possible to pickle a prepared geometry. It doesn't look like there is any way to get the original representation of the geometry from the prepared one from GEOS, so I think it's better to explicitly forbid pickling. The following code demonstrates the issue, causing a segmentation fault when the unpickled instance tries to access the GEOS geometry that has been destroyed.

```
import pickle
p1 = Point(1,2)
p2 = prep(p1)
data = pickle.dumps(p2)
del(p2)
p3 = pickle.loads(data)
p3.intersects(p1)  # segmentation fault
```

(Why you'd want to pickle a prepared geometry I don't know, but better safe than sorry?)

### DIFF
--- a/shapely/prepared.py
+++ b/shapely/prepared.py
@@ -4,6 +4,7 @@ Support for GEOS prepared geometry operations.
 
 from shapely.geos import lgeos
 from shapely.impl import DefaultImplementation, delegated
+from pickle import PicklingError
 
 
 class PreparedGeometry(object):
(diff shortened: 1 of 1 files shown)


snorfalorpagus:
A similar exception should probably also be raised for `STRTree` for which I can see a reason to want to pickle.

PR #391, opened 2016-07-07
Build manylinux1 wheels

rouge8:
This patch uses Docker and [auditwheel][] to build [manylinux1][] wheels for Python 2.6/3.3+. These wheels bundle GEOS into the wheel, similar to the OS X wheels, so Shapely can be `pip install`d on most Linuxes without GEOS or even a compiler.

All of the wheels are built with one step: `./build-linux-wheels.sh`.  This will build a Docker image based on the official manylinux images containing GEOS and will build all of the wheels and output them in `wheels/`. Note that this only builds x86_64 wheels, but it could be easily extended to building i686 wheels using the quay.io/pypa/manylinux1_i686 Docker image.

[auditwheel]: https://github.com/pypa/auditwheel
[manylinux1]: https://github.com/pypa/manylinux

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+# The wheel-building image only depends on requirements-dev.txt. Ignore all
+# other files.
+*
+.*
+!requirements-dev.txt
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ docs/modules.txt
 .idea/
 *.pyd
 *.pdb
+wheels/
--- a/Dockerfile.wheels
+++ b/Dockerfile.wheels
@@ -1,0 +1,23 @@
+FROM quay.io/pypa/manylinux1_x86_64
+
+ENV GEOS_VERSION 3.5.0
+
+# Install geos
+RUN mkdir -p /src \
+    && cd /src \
+    && curl -f -L -O http://download.osgeo.org/geos/geos-$GEOS_VERSION.tar.bz2 \
+    && tar jxf geos-$GEOS_VERSION.tar.bz2 \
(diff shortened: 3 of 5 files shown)


rouge8:
I've only actually tested the `cp27-cp27mu` wheel; this started as an experiment but came together pretty quickly. Happy to answer any questions or make some changes.

sgillies:
@rouge8 🙇 I promise to take a look at this over the weekend. I've been looking into manylinux for Fiona and Rasterio as well.

sgillies:
@rouge8 I'm happy to say the `Shapely-1.6.dev0-cp34-cp34m-manylinux1_x86_64.whl` worked on Ubuntu 14.04. I'm going to merge this and upload the dev0 dists to PyPI for the adventurous to try.

rouge8:
Awesome!

durden:
This is great, just learned about audit wheel from these changes.  I have a question about building the whl without docker.  I have a CentOS machine setup with geos installed and built the wheel with this:

`python setup.py bdist_wheel`

Then tried to run `auditwheel repair` on the resulting whl and it doesn't detect any *.so dependencies.  `auditwheel show` also doesn't detect any.  Is there some special flag to make sure the whl requires libgeos that I'm missing?

Thanks!!

sgillies:
@durden If you don't have Numpy and Cython installed when you build the Shapely wheel, you won't build the speedups extension module and that's the only module that links GEOS. See https://github.com/sgillies/frs-wheel-builds/blob/master/build-linux-wheels.sh for how I'm building the wheels with speedups.

durden:
@sgillies Ah, OK that makes sense.  Thanks for the help!

PR #392, opened 2016-07-18
Polygon.from_bounds()

geowurster:
Ready for review.

To support patterns like:

```python
import fiona

with fiona.open('input.shp') as src:
    bounds = Polygon.from_bounds(*src.bounds)
```

### DIFF
--- a/shapely/geometry/polygon.py
+++ b/shapely/geometry/polygon.py
@@ -333,6 +333,15 @@ class Polygon(BaseGeometry):
             'stroke-width="{0}" opacity="0.6" d="{1}" />'
             ).format(2. * scale_factor, path, fill_color)
 
+    @classmethod
+    def from_bounds(cls, xmin, ymin, xmax, ymax):
+        """Construct a `Polygon()` from spatial bounds."""
+        return cls([
+            (xmin, ymin),
+            (xmin, ymax),
+            (xmax, ymax),
+            (xmax, ymin)])
+
 
 class PolygonAdapter(PolygonProxy, Polygon):
 
--- a/tests/test_polygon.py
+++ b/tests/test_polygon.py
@@ -226,5 +226,17 @@ class PolygonTestCase(unittest.TestCase):
         self.assertEqual(polygon1, polygon2)
         self.assertNotEqual(None, polygon_empty1)
 
+    def test_from_bounds(self):
+        xmin, ymin, xmax, ymax = -180, -90, 180, 90
+        coords = [
+            (xmin, ymin),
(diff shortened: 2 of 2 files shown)


sgillies:
@geowurster 🙇

geowurster:
@sgillies + @snorfalorpagus: I have been thinking about how this was implemented and I think this would be better:

```python
bounds = (-180, -90, 180, 90)
p = Polygon.from_bounds(bounds)
```

instead of: `Polygon.from_bounds(*bounds)`, which is how the classmethod introduced in this PR works.

The way we typically represent spatial bounds across Fiona, Shapely, and Rasterio is as a `(xmin, ymin, xmax, ymax)` tuple rather than 4 individual values, so our spatial bounds are typically an object that _happens_ to usually be a tuple.

Moving spatial bounds around as a single object leaves us with cleaner function signatures too, for instance this function treats spatial bounds as 4 independent values and would be difficult to use:

```python
def f(value, xmin=-180, ymin=-90, xmax=180, ymax=90, *args, **kwargs):
```

This is much clearer and easier to use:

```python
def f(value, bounds=(-180, -90, 180, 90), *args, **kwargs):
```

Thoughts?

snorfalorpagus:
We could of course accept both forms. There are arguments for both. If I
had to choose one I think I'd lean towards the tuple. The keyword arguments
could imply that they aren't all required, which isn't true.

geowurster:
@snorfalorpagus Unless you have a specific example I don't see any benefit in supporting both forms in this specific case.  Good point about the keyword arguments.

sgillies:
@geowurster @snorfalorpagus precedent for supporting both forms exists in `Point()`:

```python
>>> Point(1,1).wkt
'POINT (1 1)'
>>> Point((1,1)).wkt
'POINT (1 1)'
>>> Point(*(1,1)).wkt
'POINT (1 1)'
```

but I'd like to see this method follow the similar method in Rasterio: https://github.com/mapbox/rasterio/blob/master/rasterio/windows.py#L163 and have a `from_bounds(left, bottom, top, right, ...)` signature.

snorfalorpagus:
@sgillies To confirm, you're suggesting to keep it as-merged (accepting 4 arguments) but rename "xmin, ymin, xmax, ymax" to "left, bottom, right, top"?

sgillies:
Hi @snorfalorpagus, it's great to be just 1 timezone away from you!

4 arguments, right. Renaming them isn't important as they are positional args and the names aren't part of the API. As long as the documentation is consistent, I think we'll be fine.

PR #398, opened 2016-08-06
Polygon.from_bounds() accepts a tuple rather than 4 positional arguments

geowurster:
Ready for review @snorfalorpagus or @sgillies.

We pass spatial bounds around as a `(xmin, ymin, xmax, ymax)` almost everywhere across Fiona, Rasterio, and Shapely, so this class method should not be an exception.  See https://github.com/Toblerity/Shapely/pull/392#issuecomment-237355773 for background.

### DIFF
--- a/shapely/geometry/polygon.py
+++ b/shapely/geometry/polygon.py
@@ -334,8 +334,19 @@ class Polygon(BaseGeometry):
             ).format(2. * scale_factor, path, fill_color)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


sgillies:
@geowurster in Rasterio we have individual positional arguments https://github.com/mapbox/rasterio/blob/master/rasterio/windows.py#L163. I think it's so easy to unpack a tuple using `*bounds` that these forms are pretty much equivalent.

Issue #395, opened 2016-07-28
Port polylabel algorithm


sgillies:
I'd love to see the algorithm of https://github.com/mapbox/polylabel in Shapely.

bixb0012:
Ostensibly, given the name, the purpose of porting over [polylabel](https://github.com/mapbox/polylabel) would be for finding an interior point to place labels.  I am not intimately familiar with all of the GEOS/JTS algorithms, but doesn't the existing `representative_point()` suffice?  Using the first test case from [polylabel](https://github.com/mapbox/polylabel) , the representative point from GEOS is fairly close to the label point from polylabel.

snorfalorpagus:
There is quite a nice description of how the algorithm used by `representative_point` works here:

http://gis.stackexchange.com/a/76563

The only guarantee of the "point on surface" function is that the point it produces will fall inside the polygon. This is in contrast to the centroid, which for certain polygons (e.g. a "C" shape) will fall outside. However, it is by no means the most visually appealing point for labelling.

The polylabel algorithm approximates the pole of inaccessibility, which is a useful function in general as I don't think GEOS provides this natively.

bixb0012:
Thanks for the link.

I agree that JTS/GEOS doesn't have a single function for calculating the pole of inaccessibility, but I have seen it done through a series of steps using existing functionality.  That said, I can't recall the specifics at the moment, but it involved densify and Voronoi diagrams.

Having a pole of inaccessibility function in Shapely would be handy, I can think of a couple uses, but I don't know if using such a function would provide significantly better label locations than `representative_point`.

allanadair:
Is anyone working on this? If not, I would be interested in giving it a try.

bixb0012:
@allanadair, I have/had the interest, I just can't quite seem to find the time in the next few weeks.  I say go for it.

sgillies:
@allanadair before the 1.6 release we'll need to figure out how to use the tolerance value to get the best label points. See in the image below the difference between using the default (1.0) and a value of 20.0 where the polygon is roughly 20x20 units in size.

![polylabel](https://cloud.githubusercontent.com/assets/33697/17696623/e6f642c6-63af-11e6-8cdc-c57d2d818a1a.png)

sgillies:
This is great, and so are Shapely's SVG representations in Jupyter notebooks.

[...]

PR #399, opened 2016-08-08
Attempt at porting Polylabel algorithm (#395)

allanadair:
First attempt at porting Polylabel algorithm from https://github.com/mapbox/polylabel (#395). I would appreciate advice/guidance on contributing to Shapely since this is pretty much my first code contribution.

A few things to note:

1. Performance is not as good as I would like. Ideas?
2. The original test data from mapbox/polylabel produced pretty dirty polygons, so I did not include the original data in the test fixture. I just grabbed a dense polygon of Lake Superior.
3. The original algorithm uses a maximum priority queue. Python provides a minimal priority queue through `heapq`, which required some logic modifications.
4. I used preexisting Shapely functionality where possible instead of porting identical functionality from the JavaScript mapbox/polylabel implementation.
5. I had several assumptions while porting this. For example, I didn't include support for 3D.
6. I'm not sure if I put source files in the correct locations. Please let me know. I also wasn't sure if the algorithm should be imported to `shapely/geometry/polygon.py`.
7. I didn't add any documentation, but I would like to if/when this addition is accepted.


snorfalorpagus:
Thanks @allanadair. This is looking good!

I haven't tested it, but reading the code it looks like a faithful translation of the original.

Regarding performance, the first thing to do is some profiling! Don't try optimising any further until you know where your bottlenecks are. There is pleanty of documentation about the Internet on how to do this. In particular, take a look at the standard library's cProfile (https://docs.python.org/3.4/library/profile.html) and also line_profiler (https://github.com/rkern/line_profiler)

I would expect this kind of algorithm to live in `shapely.ops`, e.g.:

```
from shapely.ops import polylabel
```

However, given the amount of code I would suggest you keep the actual code where it is (`shapely.algorithms`) and just put an import in `ops`.

sgillies:
@allanadair this is looking good! Will you add a test of a non-convex polygon ("C" shape or something like that) and assert that the label is within the polygon?

shapely/algorithms/polylabel.py
```diff
@@ -0,0 +1,115 @@
...
+
+        # max distance to polygon within a cell
+        self.max_distance = self.distance + h * 1.4142135623730951  # sqrt(2)
+
+    # rich comparison operators for sorting in minimum priority queue
+    def __lt__(self, other):
+        # inverted for minimum priority queue
+        if self.max_distance < other.max_distance:
+            return False
+        return True
```
sgillies: @allanadair how about something like `return self.max_distance >= other.max_distance` instead, and similarly in the other methods?

shapely/algorithms/polylabel.py
```diff
@@ -0,0 +1,99 @@
...
+    if not polygon.is_valid:
+        raise TopologicalError('Invalid polygon')
+    minx, miny, maxx, maxy = polygon.bounds
+    cell_size = min(maxx - minx, maxy - miny)
+    h = cell_size / 2.0
+    cell_queue = []
+
+    # First best cell approximation is one constructed from the centroid
+    # of the polygon
+    best_cell = Cell(polygon.centroid.x, polygon.centroid.y, 0, polygon)
```
snorfalorpagus: This calculates the centroid of the input polygon twice.

```
x, y = polygon.centroid.coords[0]
best_cell = Cell(x, y, 0, polygon)
```

allanadair: Thanks @snorfalorpagus

shapely/algorithms/polylabel.py
```diff
@@ -0,0 +1,100 @@
...
+        # split the cell into quadrants
+        h = cell.h / 2.0
+        heappush(cell_queue,
+                 Cell(cell.centroid.x - h, cell.centroid.y - h, h, polygon))
+        heappush(cell_queue,
+                 Cell(cell.centroid.x + h, cell.centroid.y - h, h, polygon))
+        heappush(cell_queue,
+                 Cell(cell.centroid.x - h, cell.centroid.y + h, h, polygon))
+        heappush(cell_queue,
+                 Cell(cell.centroid.x + h, cell.centroid.y + h, h, polygon))
```
sgillies: @allanadair we can eliminate more centroid calculations here, too.

sgillies:
@allanadair we generally reserve "precision" for a number of decimal places of precision. Can you change it to "tolerance" and explain in the docstring what it means for the function? Do you think It is possible to come up with a sensible default for the tolerance based on any property of the input geometry? Its area or bounding box for example?

shapely/algorithms/polylabel.py
allanadair: @sgillies In this case `cell.centroid` is actually a shapely `Point` geometry. I recently renamed the `Cell.point` property to `Cell.centroid` because I thought it increased readability. Do you think that I should rename that property to something else?

sgillies: @allanadair ah, I just misread the code. Using `x` and `y` would save several attribute lookups (`.` isn't free), but that's not a big speedup.

allanadair:
@sgillies Do we have a style guide for docstrings? I saw some docstrings that look like numpydoc.

I'm not sure about a sensible default as of right now. I know that when the tolerance is too high that we will just take a longer path to returning the input geometry's centroid (first best_cell). I will give it some more thought and I'd love some more feedback from others.

sgillies:
@allanadair I may have been overthinking things. The JS polylabel method has a default tolerance of 1.0 and we should follow suit.

A numpy style docstring would be excellent. There are a few of them in Shapely's newer code, like https://github.com/Toblerity/Shapely/blob/master/shapely/ops.py#L280-L303. 

After this, is it ready to be merged?

shapely/algorithms/polylabel.py
allanadair: I will add `x` and `y` properties back to the Cell class.

allanadair:
@sgillies I think it's ready to be merged.

I didn't verify the formatting of my docstrings because I'm not sure how we are building the docs. I used your example and [NumPy](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) as a reference.

sgillies:
🙏 @allanadair.

PR #404, opened 2016-08-16
Instancing geometries with an empty list returns an empty geom.

### LINKED ISSUE #397: Creating empty polygon from empty list fails
Empty polygons can be created using

```
polygon = geometry.Polygon()
polygon.is_empty
# True
```

However, creating an empty polygon from an empty list of vertices, `polygon = geometry.Polygon([])`, throws the following error

>   File "<stdin>", line 1, in <module>
>   File "[..]/python2.7/site-packages/shapely/geometry/polygon.py", line 229, in __init__
>     self._geom, self._ndim = geos_polygon_from_py(shell, holes)
>   File "[..]/python2.7/site-packages/shapely/geometry/polygon.py", line 508, in geos_polygon_from_py
>     geos_shell, ndim = geos_linearring_from_py(shell)
>   File "[..]/python2.7/site-packages/shapely/geometry/polygon.py", line 450, in geos_linearring_from_py
>     n = len(ob[0])
> IndexError: list index out of range

It would be nice if this would also just return an empty polygon.

I use shapely version 1.5.16 in case this matters.

snorfalorpagus:
This PR enables the creation of empty geometries, by passing a zero-length list (or other iterable), e.g:

```line = LineString([])```

Fixes #397.

### DIFF
--- a/shapely/geometry/linestring.py
+++ b/shapely/geometry/linestring.py
@@ -91,7 +91,9 @@ class LineString(BaseGeometry):
     # Coordinate access
     def _set_coords(self, coordinates):
         self.empty()
-        self._geom, self._ndim = geos_linestring_from_py(coordinates)
+        ret = geos_linestring_from_py(coordinates)
+        if ret is not None:
+            self._geom, self._ndim = ret
 
     coords = property(BaseGeometry._get_coords, _set_coords)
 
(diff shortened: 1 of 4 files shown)


sgillies:
🙏

PR #414, opened 2016-09-14
Fix 'defualt' typo, replace with 'default'

sebastic:
The spelling error was noticed while reviewing the changes in 1.6a1 as part of the Debian package update.


sgillies:
🙏 @sebastic

PR #428, opened 2016-11-17
minor formatting consistency

kannes:
Three times ``geoms`` vs one `geoms`
Three times ``list()`` vs one :func:`list`


sgillies:
🙏 @kannes

Issue #432, opened 2016-11-19
Support for Python 2.6


snorfalorpagus:
I noticed in the Travis logs the following message:

> DEPRECATION: Python 2.6 is no longer supported by the Python core team, please upgrade your Python. A future version of pip will drop support for Python 2.6

It looks like Python 2.6 is no longer supported. https://www.python.org/download/releases/2.6.9/

Should we drop (explicit) support for this in the next release? I think this mostly just involves dropping 2.6 from the CI. It might still work, but it's not something we should put any effort into maintaining.

sgillies:
Let's do remove 2.6 from the build matrix. 3.3 is coming up on the end of its life, too, https://docs.python.org/devguide/#status-of-python-branches.

Issue #431, opened 2016-11-18
Shapely throws error on linemerge when numpy not present


grant-humphries:
**Environment details:**
Ubuntu 16.04
GEOS 3.5.0-1 (libgeos-dev)
Python 2.7.12 (python-dev)
Shapely 1.5.17

Hi folks,

Shapely is able to find GEOS and install (I'm using buildout to fetch python dependencies), but I don't have numpy and the below error message was thrown when I called linemerge.  I'd prefer not require numpy to run my code just to keep project dependencies minimal.

```
Traceback (most recent call last):
  File "./bin/rlis2osm", line 22, in <module>
    sys.exit(rlis2osm.main.main())
  File "/home/ubuntu/rlis2osm/rlis2osm/main.py", line 228, in main
    dissolver.dissolve_ways(paths.combined, paths.dissolved)
  File "/home/ubuntu/rlis2osm/rlis2osm/dissolve.py", line 40, in dissolve_ways
    dissolve_geom = linemerge(geom_list)
  File "/home/ubuntu/rlis2osm/eggs/Shapely-1.5.17-py2.7-linux-x86_64.egg/shapely/ops.py", line 113, in linemerge
    result = lgeos.GEOSLineMerge(source._geom)
  File "/home/ubuntu/rlis2osm/eggs/Shapely-1.5.17-py2.7-linux-x86_64.egg/shapely/geometry/proxy.py", line 33, in _geom
    self.__geom__, n = self.factory(self.context)
  File "/home/ubuntu/rlis2osm/eggs/Shapely-1.5.17-py2.7-linux-x86_64.egg/shapely/geometry/multilinestring.py", line 134, in geos_multilinestring_from_py
    geom, ndims = linestring.geos_linestring_from_py(obs[l])
  File "/home/ubuntu/rlis2osm/eggs/Shapely-1.5.17-py2.7-linux-x86_64.egg/shapely/geometry/linestring.py", line 197, in geos_linestring_from_py
    ob = required(ob)
  File "/home/ubuntu/rlis2osm/eggs/Shapely-1.5.17-py2.7-linux-x86_64.egg/shapely/coords.py", line 31, in required
    return numpy.require(ob, numpy.float64, ["C", "OWNDATA"])
NameError: global name 'numpy' is not defined
```

snorfalorpagus:
I hadn't thought much about the numpy dependency as I use it so much for other projects.

[...]

PR #433, opened 2016-11-19
Changes to travis.yml and support for running without numpy

snorfalorpagus:
This is a WIP but could benefit from some comments.

- Tests running shapely on Travis without numpy (#431).
- Fixes a bug preventing shapely running without numpy (also #431).
- Removes support for Python 2.6 on Travis (#432).
- Expands the build matrix used by Travis to be more verbose. This gives more control over what is tested. There is now a `SPEEDUPS=0/1` and `NUMPY=0/1` variable. This could also provide an `AUTODOC=0/1` variable for use in #337.
- The test requirements for test_vectorized.py have been modified. Previously they ran if numpy was present. We distribute .c files for speedups.pyx and vectorized.pyx, in case the user doesn't have cython installed. I think it's more likely that the user doesn't have a C compiler installed, in which case the .c files aren't much use. The test now checks if the speedups module was build (as this was easy to check with `speedups.available` and only runs the tests if this is true.
- In order to prevent setuptools pulling in numpy as a dependency when it isn't desired it was necessary to add a couple of lines to `setup.py`. I don't like this very much but can't think of a better way to do it, other than a find/replace of setup.py in Travis.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -176,10 +176,9 @@ long_description = readme + '\n\n' + credits + '\n\n' + changes
 
 
 extra_reqs = {
-    'test': ['pytest', 'pytest-cov', 'numpy>=1.4.1', 'packaging']
+    'test': ['pytest', 'pytest-cov', 'packaging'],
+    'vectorized': ['numpy>=1.4.1'],
 }
-if os.getenv('NUMPY') == '0':
-    extra_reqs['test'].remove('numpy>=1.4.1')
 extra_reqs['all'] = list(it.chain.from_iterable(extra_reqs.values()))
 
 
--- a/setup.py
+++ b/setup.py
@@ -178,6 +178,8 @@ long_description = readme + '\n\n' + credits + '\n\n' + changes
 extra_reqs = {
     'test': ['pytest', 'pytest-cov', 'numpy>=1.4.1', 'packaging']
 }
+if os.getenv('NUMPY') == '0':
+    extra_reqs['test'].remove('numpy>=1.4.1')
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -178,6 +178,8 @@ def version(self):
 extra_reqs = {
     'test': ['pytest', 'pytest-cov', 'numpy>=1.4.1', 'packaging']
 }
+if os.getenv('NUMPY') == '0':
+    extra_reqs['test'].remove('numpy>=1.4.1')
```
sgillies: @snorfalorpagus instead, how about we add an extra requirements item: `'vectorized': ['numpy>=1.4.1']`? If we've already installed numpy, then nothing changes.

sgillies:
I like this work a lot @snorfalorpagus 👍

setup.py
snorfalorpagus: That makes sense to me. Should it still be controlled using an environment varaible?

sgillies: No, let's remove the NUMPY env var, and remove numpy from the test requirements. This means that on Travis we'll need to install `shapely[all]`.

sgillies: No, not unless they installed `shapely[vectorized]` or `shapely[all]`. `pip install shapely` won't get any of the extra requirements.

sgillies:
All through with this one @snorfalorpagus? I'll merge later if so.

snorfalorpagus:
Yes it's ready to go. I added a line to the pytest header which makes it easy to check if numpy was installed when the tests were run.

sgillies:
🙏 @snorfalorpagus

PR #434, opened 2016-11-20
Improvements to documentation figures

snorfalorpagus:
This PR adds support for Python 3.x rendering of the documentation figures (as mentioned in #337). It also removes some code duplication by moving common functions into `figures.py`. More could probably be done, but this will do for now.

### DIFF
--- a/docs/code/buffer.py
+++ b/docs/code/buffer.py
@@ -2,7 +2,7 @@ from matplotlib import pyplot
 from shapely.geometry import LineString
 from descartes import PolygonPatch
 
-from figures import SIZE, BLUE, GRAY
+from figures import SIZE, BLUE, GRAY, set_limits
 
 def plot_line(ax, ob):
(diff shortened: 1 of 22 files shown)


sgillies:
@snorfalorpagus could you attach a screenshot or two for me? I won't have time to pull your changes and make the docs today.

snorfalorpagus:
> @snorfalorpagus could you attach a screenshot or two for me? I won't have time to pull your changes and make the docs today.
In theory none of the figures have changed visually. I checked this while I was going along, but it's probably worth doing a final check. I'll do that this evening.

snorfalorpagus:
I've been through and corrected any differences. There is one outstanding difference between my rendering and the one currently on the website - subplot b) of parallel_offset.py has a slightly different geometry. I think this might be due to a different GEOS version?

Original:

parallel_offset-1

New:

parallel_offset_new

PR #436, opened 2016-12-02
Move access to coords via array protocol entirely to speedups

sgillies:
The final touches on removing our accidental hard dependency on Numpy.


sgillies:
I've checked and `python -m pytest --without-speedups` passes when numpy is not installed.

PR #438, opened 2016-12-03
Fix creation of 3D multipoint from array. Closes #437.

### LINKED ISSUE #437: MultiPoint constructor drops z coords
I've noticed that when trying to create a MultiPoint instance with numpy arrays the z array gets ignored for some operations such as intersections.

I think the problem is that constructor doesn't properly register the z coordinates when creating in the pointers to the numpy arrays - the points seems to get properly created when passing a Python list but there is obviously a big performance hit for lots of points.

I think this demonstrates the problem:
```
import numpy as np
from shapely.geometry import MultiPoint

pts = np.vstack([np.arange(20)*0.1, np.arange(20)*0.2, np.arange(20)])
multi_point = MultiPoint(pts.T)

print(multi_point.has_z)  # False

multi_point = MultiPoint(pts.T.tolist())

print(multi_point.has_z)  # True
```

snorfalorpagus:
This PR fixes a bug in the creation of MultiPoints from numpy arrays with 3 dimensions.

This bug has been present since at least Shapely 1.2, but this PR is only aimed at the maint-1.5 branch. I hope that's the right thing to do?

### DIFF
--- a/shapely/geometry/multipoint.py
+++ b/shapely/geometry/multipoint.py
@@ -181,7 +181,7 @@ def geos_multipoint_from_py(ob):
         subs = (c_void_p * m)()
 
         for i in range(m):
-            geom, ndims = point.geos_point_from_py(cp[n*i:n*i+2])
+            geom, ndims = point.geos_point_from_py(cp[n*i:n*i+n])
             subs[i] = cast(geom, c_void_p)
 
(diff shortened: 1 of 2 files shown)


snorfalorpagus:
Closed in favour of a PR for maint-1.6.

Issue #439, opened 2016-12-03
Speeding up Rotate and Translate


BebeSparkelSparkel:
I'm calling rotate and then translate directly after on a polygon and I'm wondering if there is a way to increase the speed of this.
Is there a way to combine the the two operations so that the rotation and translation can be performed at the some time?

snorfalorpagus:
The `affine_transform` function has a cython implementation. If you're not already you should make sure you have shapely installed with the speedups submodule. You can check this with:

```
>>> import shapely.speedups
>>> shapely.speedups.available
True
```

Both the rotate and translate methods use this function, so it's being called twice (which is inefficient). What is needed is to combine the transformation matrix for both operations, which can be done by multiplying the matrices together.

It's not currently possible to access the matrices created by Shapely. I propose that we could separate the code that generates the matrices from the actual call to affine_transform. A function could then be created to multiply the matrices to create aggregated transformations. It would look something like this:

```
def rotate(geom, ...):
    matrix = create_rotate_matrix(...)
    return affine_transform(geom, matrix)

def create_rotate_matrix(angle, origin, use_radians=False):
    ...
    return matrix

def multiply(*args):
    for matrix in args: ...
    return aggregated_matrix
```

All the code you need is in `shapely/affinity.py` if you want to give this a go. The module is very well commented. A PR would be welcome.

https://github.com/Toblerity/Shapely/blob/7b09eb777f688d9c404b0f20b80864c8e51c446e/shapely/affinity.py

BebeSparkelSparkel:
So I've come up with a solution that I think will work check out pull request [https://github.com/Toblerity/Shapely/pull/440](url)

I'm getting a lot of integration errors that I could use some help with though.

BebeSparkelSparkel:
Ok I've fixed the integration errors

PR #440, opened 2016-12-05
Improving affine_transform functions from issue #439

BebeSparkelSparkel:
Created the affine_matrix_builder class in affinity.py to allow affine matrices to be combined.
Affine transforms can now be applied with `geom_transformed = affine_matrix_builder(Point(1,1)).rotate(90,(0,0)).translate(20, 30).transform()`
It also automatically determines when to combine the affine matrices or when to apply the last transform and store the current transform.

Tests for the class have been added in test_affinity.py 

Documentation has been updated for the class in manual.rst

Runtime of some affine transforms has decreased by 30%

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ wheels/
 ignore/
 
 venv/
+*.DS_Store
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -1859,6 +1859,7 @@ preserved or supported by 3D affine transformations.
 .. class:: shapely.affinity.affine_matrix_builder(geom, matrix=None)
 
   Class for building affine transformation matrices. Can speed up transformations because it can reduces the number of transformations needed by combining the affine matrices when appropriate.
+  Class implemented by William Rusnack, github.com/BebeSparkelSparkel, williamrusnack@gmail.com
 
 .. method:: shapely.affinity.affine_matrix_builder.affine_transform(matrix)
   Look applies what is described in function shapely.affinity.affine_transform above.
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -1856,6 +1856,41 @@ preserved or supported by 3D affine transformations.
       0 & 0 & 0 & 1
     \end{bmatrix}
 
+.. class:: shapely.affinity.affine_matrix_builder(geom, matrix=None)
+
+  Class for building affine transformation matrices. Can speed up transformations because it can reduces the number of transformations needed by combining the affine matrices when appropriate.
+
+.. method:: shapely.affinity.affine_matrix_builder.affine_transform(matrix)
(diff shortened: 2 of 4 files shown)


BebeSparkelSparkel:
@snorfalorpagus 
I needed to include numpy in affinity.py but I'm not sure were to add numpy in setup.py

BebeSparkelSparkel:
@coveralls I'm looking at the [continuous integration error report of the failed 3.5 logs lines 365-367](https://travis-ci.org/Toblerity/Shapely/jobs/181421038) and the code shown in the log does not match the current commit version of the file. Is there a way to check and make sure it is using the latest commit.

snorfalorpagus:
I'm certainly no expect with this. Could you explain why some transformations can't be combined? I'm referring to this comment:

> Holds the safe relationships of when two affine matricies can be combined.

BebeSparkelSparkel:
If a rotation then a translation is applied then the affine matrices can be combined and will create the same result as when a rotation and then a translation is applied to the object. However, if a translation then a rotation is applied then the combined affine matrices will have a different result than if if the translation and rotation affine matrices are applied separately. Here's an example where the order of the transformation matters.
```python
>>> from shapely.affinity import rotate, translate
>>> 
>>> print(
...   translate(
...     geom=rotate(
...       geom=Point(1,1),
...       angle=90,
...       origin=(0,0)
...     ),
...     xoff=1,
...     yoff=1
...   )
... )
POINT (0 2)
>>> 
>>> print(
...   rotate(
...     geom=translate(
...       geom=Point(1,1),
...       xoff=1,
...       yoff=1
...     ),
...     angle=90,
...     origin=(0,0)
...   )
... )
POINT (-2 2)
>>> 
```
If the order isn't preserved and the affine matrices are always combined the result will always be the case where the rotate then the translate is applied, Point (0 2). Meaning if using the combine matrices method it would be impossible to translate then rotate a geometric object. So, users would have to consider what cases they could combine the matrices and when they would have to apply the matrices separately.
There are some more specific cases where the matrices can be combined like when the some transforms have the same origin, but I haven't investigated those cases enough to have added them to the affine_matrix_builder._combine_matrices method.

BebeSparkelSparkel:
@snorfalorpagus 
Does that make sense or do you have a way to preserve the transformation order when combining the affine matrices?

BebeSparkelSparkel:
Do you plan on pulling this?

sgillies:
@BebeSparkelSparkel I haven't been able to make the time for the review this deserves, but soon, I promise!

BebeSparkelSparkel:
Ok thanks!

sgillies:
@snorfalorpagus @mwtoews I'd be super grateful 🙏 if either of you could join me in reviewing this. I think it's good and mainly needs some test restructuring, but I may have overlooked something.

PR #464, opened 2017-02-06
Updates README.rst with svg badge

sobolevn:


sgillies:
Why @sobolevn?

sobolevn:
Because it looks untidy on retina displays.


Just compare these two: the one the left in png, the on the right is svg.

PR #475, opened 2017-03-23
Add support for reading/passing along CRS info

kirkhansen:
If a CRS is present on the context object when using `shape` or `asShape`, let's just read in the CRS for them.
This also adds the CRS to the `__geo_interface__` property (if present).
There were also some code style / import fixes and some other pep8 stuff done to some of the files along the way.

Comments / concerns welcome!

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -219,6 +219,22 @@ class BaseGeometry(object):
     # a reference to the so/dll proxy to preserve access during clean up
     _lgeos = lgeos
 
+    def __init__(self, *args, **kwargs):
+        # set the crs if there is one.
+        try:
+            self.crs = kwargs['crs']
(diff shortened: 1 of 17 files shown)


sgillies:
@kirkhansen Thank you for the offer, your code is fine (higher quality than much of Shapely's original code, even), but I will continue to decline to add crs as a feature to Shapely. This is entirely up to applications. The functions like `make_point()`, etc, also increase the surface area of the Shapely API. I prefer to keep it small.

You are certainly right to want Shapely to have better docstrings! This is something that we'll continue to chip away at over time.

kirkhansen:
Why is it the application's responsibility to copy the CRS attribute over?
I'm not suggesting that shapely _use_ the CRS attribute, only that it doesn't remove it when provided.

PR #477, opened 2017-04-05
Fix bad assert in test_numpy in test_linestring.py

zware:
Misuse of `assert` like a function resulted in a warning from pytest, this fixes it and also removes trailing whitespace from the file.


sgillies:
Thanks @zware!

This may have slipped in during our transition to use of py.test, but consistency with the rest of the test suite is a good thing.

PR #478, opened 2017-04-05
Add __repr__ to BaseGeometry

zware:
This adds a nice repr to BaseGeometry.  This is useful to me because I have unit tests that compare fairly large dicts that include Point or MultiPolygon objects; when some non-geometric object differs, unittest spits out a line-by-line diff of the repr of the dicts which points out "differences" between the geometric objects due to their different memory addresses.  This silences those false differences.

Using the wkt does make for an extra-large repr in some cases, but I haven't figured out a nice way to reduce that while still making it possible to tell whether two objects are equal by looking at their reprs.  If anyone has a better idea than using the wkt, I'd be happy to implement it :)

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -237,6 +237,10 @@ class BaseGeometry(object):
     def __str__(self):
         return self.wkt
 
+    def __repr__(self):
+        name = getattr(self.__class__, '__qualname__', self.__class__.__name__)
+        return '<{} object representing {}>'.format(name, self.wkt)
+
     # To support pickling
     def __reduce__(self):
         return (self.__class__, (), self.wkb)
--- a/tests/test_point.py
+++ b/tests/test_point.py
@@ -13,6 +13,7 @@ class LineStringTestCase(unittest.TestCase):
         self.assertEqual(p.y, 2.0)
         self.assertEqual(p.coords[:], [(1.0, 2.0)])
         self.assertEqual(str(p), p.wkt)
+        self.assertEqual(repr(p), '<Point object representing %s>' % p.wkt)
         self.assertFalse(p.has_z)
         with self.assertRaises(DimensionError):
             p.z
@@ -21,6 +22,7 @@ class LineStringTestCase(unittest.TestCase):
         p = Point(1.0, 2.0, 3.0)
         self.assertEqual(p.coords[:], [(1.0, 2.0, 3.0)])
         self.assertEqual(str(p), p.wkt)
+        self.assertEqual(repr(p), '<Point object representing %s>' % p.wkt)
         self.assertTrue(p.has_z)
         self.assertEqual(p.z, 3.0)
 


sgillies:
@zware what do you think about the documentation at https://docs.python.org/3/reference/datamodel.html#object.__repr__? That says to me that `"Point(0.0, 0.0)"` may be the least surprising representation of `Point(0.0, 0.0)`.

zware:
I agree, but is there a nice general way to implement that in `BaseGeometry.__repr__` that covers `Point`, `Polygon`, `MultiPolygon`, et al?  Otherwise `__repr__` would need to be implemented separately in each class (which is doable, but more work).

snorfalorpagus:
This works well for simple geometries like Point, but I'd be worried about larger geometries like LineString and Polygon which could easily fill a screen.

zware:
True; my particular situation includes Polygons with 360 vertices, which can and does fill most of a screen.  This doesn't bother me, but I can certainly see how it would be annoying in other situations :).

Another option would be to include a hash of the wkt instead of the wkt itself, but that's fairly useless for seeing what any differences actually are.  If extra-large reprs aren't acceptable, I'd prefer this option.

Yet another option is to truncate after ~100 characters, but that means larger geometries that are mostly the same but do have subtle differences may wind up with the same repr, which I'd rather avoid.

sgillies:
I like what Numpy does.

```python
>>> import numpy
>>> numpy.ones((1000, 1000))
array([[ 1.,  1.,  1., ...,  1.,  1.,  1.],
       [ 1.,  1.,  1., ...,  1.,  1.,  1.],
       [ 1.,  1.,  1., ...,  1.,  1.,  1.],
       ...,
       [ 1.,  1.,  1., ...,  1.,  1.,  1.],
       [ 1.,  1.,  1., ...,  1.,  1.,  1.],
       [ 1.,  1.,  1., ...,  1.,  1.,  1.]])
```

We could do something like this, elliding coordinates as necessary so that `Polygon(shell=[(0.0, 0.0), ...], holes=[[...]])` fits in < 80 chars.

zware:
That has the problem of large, subtly-different geometries winding up with the same repr, though.

sgillies:
Let's not concern ourselves with extra large representations. Python doesn't, see for example

```
>>> x = tuple(0 for i in range(1000))
>>> x
(0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0)
```

I do want `__repr__()` to serve a different purpose from `__str__()`, along the lines of what's written at https://docs.python.org/3/reference/datamodel.html#object.__repr__. I'm also concerned about user code potentially calling `repr(geometry)` when it should be calling `geometry.wkt` and then having to support that code for a long time. In other words, I'm 👎 on returning WKT from `repr(geometry)`.

I realize that it's a bit of work to implement `__repr__()` for all the classes. If you are unable or unwilling, you can close this PR and I'll follow up with a new one. Okay?

zware:
I'm unfortunately not savvy enough with the details of each type to implement individual `__repr__`s, so I'll have to leave that to you.

I think this PR does satisfy the constraints given in your last message (different than `__str__`, not bare `wkt`, don't care about length), but it does miss the nice property of `eval(repr(x)) == x`.  If you'd rather keep that property, feel free to close this with no hard feelings from me.  I'm just looking forward to nice reprs one way or the other :)

sgillies:
Closing. We'll keep our existing reprs.

PR #479, opened 2017-04-07
Cartopytrace

marqh:
from https://github.com/SciTools/cartopy/issues/805

it was discussed about bringing https://github.com/SciTools/cartopy/blob/master/lib/cartopy/trace.pyx into shapely

I have researched this and I can confirm that this is fine to do.

Regarding Licensing
@QuLogic : you are the only contributor not already covered by the copyright, so approval from you to relicense your code from LGPL to BSD is required in order for this PR to be merged.  A comment on this PR is sufficient

@sgillies I have included trace.pyx and updated the source license header with a snippet of history, I hope that is reasonable.  I have also removed the cartopy function, to avoid circular referencing. I have altered the setup.py but i haven't tested all of this for all builds, so please check you are happy

If this or a similar approach is adopted and then included it in a release, I can refactor cartopy to use this code from shapely and (hopefully) remove the issue from https://github.com/SciTools/cartopy/issues/805 in the next cartopy release

### DIFF
--- a/shapely/trace.pyx
+++ b/shapely/trace.pyx
@@ -1,0 +1,125 @@
+# (C) British Crown Copyright 2011 - 2016, Met Office
+#
+# This file is part of cartopy.
+#
+# cartopy is free software: you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# cartopy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with cartopy.  If not, see <https://www.gnu.org/licenses/>.
+
+"""
+This module pulls together _trace.cpp, proj.4, GEOS and _crs.pyx to implement a function
+to project a LinearRing/LineString. In general, this should never be called manually,
+instead leaving the processing to be done by the :class:`cartopy.crs.Projection`
+subclasses.
+"""
+
+from libc.stdint cimport uintptr_t as ptr
+
+
+cdef extern from "geos_c.h":
+    ctypedef void *GEOSContextHandle_t
+    ctypedef struct GEOSGeometry:
+        pass
+
+from cartopy._crs cimport CRS
+
+
+import shapely.geometry as sgeom
+from shapely.geos import lgeos
+
+
+cdef extern from "proj_api.h":
+    ctypedef void *projPJ
+
+
+cdef extern from "_trace.h":
+    cdef cppclass Interpolator:
+        pass
+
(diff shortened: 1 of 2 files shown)


shapely/trace.pyx
```diff
@@ -0,0 +1,67 @@
+# (C) British Crown Copyright 2011 - 2017, Met Office
+#
+# This file was part of cartopy.  It is now part of shapely
+#
+# Shapely is licensed under the BSD 3-clause "New" or "Revised" License
+
+"""
+This module pulls together _trace.cpp, proj.4, and GEOS.
```
QuLogic: `_trace.cpp` is not part of Shapely, no?

shapely/trace.pyx
```diff
@@ -0,0 +1,67 @@
...
+
+import shapely.geometry as sgeom
+from shapely.geos import lgeos
+
+
+cdef extern from "proj_api.h":
+    ctypedef void *projPJ
+
+
+cdef extern from "_trace.h":
```
QuLogic: Shouldn't this entire block be removed as well?

shapely/trace.pyx
```diff
@@ -0,0 +1,67 @@
...
+    ctypedef void *GEOSContextHandle_t
+    ctypedef struct GEOSGeometry:
+        pass
+
+
+import shapely.geometry as sgeom
+from shapely.geos import lgeos
+
+
+cdef extern from "proj_api.h":
```
QuLogic: Could probably be removed if the next block is removed as well.

QuLogic:
Re-licensing is fine by me.

sgillies:
I'm on vacation for the next week. Will look at this when I get back.

marqh:
> _trace.cpp is not part of Shapely, no?

I see this point now. I'm either doing too much, or not enough

@QuLogic @sgillies do you think it the case that the 'trace' module is not required or wanted by shapely, just the lower level

`cdef GEOSContextHandle_t get_geos_context_handle():`

`cdef GEOSGeometry *geos_from_shapely(shapely_geom) except *:`
and
`cdef shapely_from_geos(GEOSGeometry *geom):` 
 
that might better live in shapely?

Or, should the `_trace.cpp` and `_trace.h` also live within the shapely code base?


in https://github.com/SciTools/cartopy/issues/805
the aim is given as:
Move all of cartopy's GEOS dependency into Shapely

The extension code in trace.pyx is rather small and could be moved to Shapely and deleted from cartopy.  

is trace.pyx the only thing we need to work with to ensure this?
are there hooks in
https://github.com/SciTools/cartopy/blob/master/lib/cartopy/_trace.cpp

sgillies:
@marqh I'm slow to realize that the cartopy module we're proposing to move to Shapely also involves PROJ.4. I'm not willing to make PROJ.4 a requirement for Shapely, so I think we're blocked until there's a Cartopy refactor that unmixes the GEOS and PROJ.4 concerns.

QuLogic:
@sgillies I don't believe this PR has been cleaned up completely, and I don't think that Proj.4 is actually required. They're just unused definitions.

sgillies:
It looks like this PR is stuck. I'm going to close it. We can always reopen when we get unstuck.

Issue #484, opened 2017-04-21
Unable to load libgeos on CentOS when using PyInstaller


durden:
The `so` loading doesn't work when running inside PyInstaller even though PyInstaller correctly adds libgeos to the build.

## Operating system

CentOS 6.4 and most likely newer versions

## Shapely version and provenance

1.6b4 installed via whl

durden:
There is special code in `geometry.py` to handle anaconda, but not `frozen`.

PR #485, opened 2017-04-21
Correctly find libgeos when running under PyInstaller with Linux, Fix…

durden:
…es #484


durden:
Not sure if there is a better way to do this. I followed the idea used to find the `so` from the whl file.

davidh-ssec:
I ran in to this issue as well and followed the same workaround I used for my mac app here but with linux .so files:
https://github.com/pyinstaller/pyinstaller/issues/1792

I would really appreciate this fix though.

PR #487, opened 2017-05-03
Enable Shapely to work in gevent without reinitializing the geos bindings for every greenlet.

rowillia:
Gevent is a popular coroutine based networking library that uses greenlets
to provide users a synchronous API while allowing for cooperative
multitasking.  Gevent accomplishes this through monkey patching the stdlib
to swap out blocking functions with non-blocking variants, including
threading.

If monkey patched, Gevent will replace `threading.local` with an
implementation that makes all thread local variables greenlet local
instead.  In the context of shapely, this means all objects that inherit
from LGEOSBase will become greenlet local instead of thread local.
Effectively, this causes servers that utilize both gevent and shapely to
re-initialize the geos bindings every request.

Since LGEOSBase inherits from `threading.local` to prevent concurrent
access to the GEOS C API, this is not relevant in a monkey patched gevent
process which will only ever run one thread at a time.  In this case, we
inherit from `object` instead of `threading.local`


.travis.yml
```diff
@@ -58,20 +83,24 @@ before_install:
   - pip install pip setuptools --upgrade
   # if building with speedups install cython
   - if [ "$SPEEDUPS" == "1" ]; then pip install --install-option="--no-cython-compile" cython; fi
+  # if building with speedups install cython
```
snorfalorpagus: Update comment.

.travis.yml
```diff
@@ -8,45 +8,70 @@ matrix:
       env:
         - SPEEDUPS=1
         - NUMPY=1
+        - GEVENT=0
+        - GEVENT_PATCHED=0
```
snorfalorpagus: What is the need for two environment variables?

shapely/geos.py
```diff
@@ -555,7 +554,38 @@ def errcheck_predicate(result, func, argtuple):
...
+
+    If monkey patched, Gevent will replace `threading.local` with an
+    implementation that makes all thread local variables greenlet local
+    instead.  In the context of shapely, this means all objects that inherit
+    from LGEOSBase will become greenlet local instead of thread local.
+    Effectively, this causes servers that utilize both gevent and shapely to
+    re-initialize the geos bindings every request.
+
+    Since LGEOSBase inherits from `threading.local` to prevent concurrent
+    access to the GEOS C API, this is not relevant in a monkey patched gevent
```
snorfalorpagus: I suggest opening with the comment about `threading.local` and move onto gevent - gevent is the exception.

snorfalorpagus:
This looks sensible to me, but I don't have any experience with gevent so can't comment any further.

.travis.yml
rowillia: @snorfalorpagus I wanted to be able to test both gevent being present but `patch_all` not being invoked _and_ gevent being present and `patch_all` is invoked.

.travis.yml
rowillia: Done

shapely/geos.py
rowillia: Done

rowillia:
Thanks for the feedback @snorfalorpagus !  Let me know if there's anything else I can take care of.

.travis.yml
```diff
@@ -58,10 +83,14 @@ before_install:
   - pip install pip setuptools --upgrade
   # if building with speedups install cython
   - if [ "$SPEEDUPS" == "1" ]; then pip install --install-option="--no-cython-compile" cython; fi
+  # if runnign with gevent installed
```
mxr: s/gn/ng

rowillia:
Thanks for taking a look @sgillies !  

Users of gevent know that the _very_ first thing they must do, before importing any other module, is enable monkey patching.  In fact, most of the WSGI services that support gevent do it for you (https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/ggevent.py#L57-L65).  If shapely is imported first, the monkey patch won't be active and the original behavior will apply.

This is part of what makes modifying invoking the monkey patch hard...usually it's the WSGI server that does it first.

rowillia:
ping @sgillies !  Any other feedback?  A "This is crazy, gevent is crazy, and you are crazy so I'm closing this PR" is also fine ;)

shapely/geos.py
```diff
@@ -555,7 +553,32 @@ def errcheck_predicate(result, func, argtuple):
...
+    process which will only ever run one thread at a time.  In this case, we
+    inherit from `object` instead of `threading.local`
+    """
+    result = threading.local
+    try:
+        from gevent import monkey
+        if 'threading' in monkey.saved:
+            result = object
+    except Exception:
```
sgillies: @rowillia how about catching the most specific error(s) possible? `ImportError` or whatever.

sgillies: @rowillia just following up, I've got this one change request blocking merge of the PR.

PR #512, opened 2017-08-28
Remove some unused imports

jameslamb:
Thank you for this incredibly useful module! `shapely` is as fun to use as it is powerful.

I was reading through the source code tonight to try to gain a deeper understanding of what's going on, and found a few unused imports in some modules. Please consider this PR to remove them.


sgillies:
🙏 @jameslamb

Issue #513, opened 2017-08-28
Add an Empty class to geometry.base


sgillies:
The GeoPandas project needs an empty geometry type to pass to `fillna()`. While `BaseGeometry()` will work, it couples GeoPandas to Shapely internals. A class like

```python
class Empty(BaseGeometry):
    pass
```

would be better. It doesn't even need much of an implementation.

See #510 for background.

sgillies:
Done!

Issue #510, opened 2017-08-22
Failing geopandas tests related to empty collection


jorisvandenbossche:
With the release of 1.6.0, the geopandas tests started failing (we should really test with shapely master, or at least with beta/rc versions, sorry!)

I didn't look yet into detail what is going wrong, but I could already pin it down to:

```
In [6]: from shapely.geometry import Point

In [7]: from geopandas import GeoSeries

In [8]: s = GeoSeries([Point(1, 1), None])

In [9]: s.fillna(Point())
Out[9]: 
0                 POINT (1 1)
1    GEOMETRYCOLLECTION EMPTY
dtype: object

In [10]: import shapely

In [11]: shapely.__version__
Out[11]: '1.5.16'
```

vs

```
In [7]: from shapely.geometry import Point

In [8]: from geopandas import GeoSeries

In [9]: s = GeoSeries([Point(1, 1), None])

In [10]: s.fillna(Point())
Out[10]: 
0    POINT (1 1)
1           None
dtype: object

In [11]: import shapely

In [12]: shapely.__version__
Out[12]: '1.6.0'
```

So the `fillna` method is no longer filling with an empty collection. 

I have to further drill down into the fillna implementation to see where the difference happens, and what actual shapely behaviour changed(will try to do that tomorrow). But already putting this issue, in case anybody has an idea where it could be coming from.

jorisvandenbossche:
I could already drill it down to a non-geopandas reproducible example, but with numpy. Apparently, `np.putmask` no longer wants to put an empty geometry inside an array:

```
In [2]: import numpy as np

In [3]: from shapely.geometry import Point

In [4]: a = np.array([Point(1, 1), None], dtype=object)

In [5]: mask = np.array([False, True])

In [6]: fill_value = Point()

In [7]: print(fill_value)
GEOMETRYCOLLECTION EMPTY

In [9]: np.putmask(a, mask, fill_value)

In [10]: a
Out[10]: 
array([<shapely.geometry.point.Point object at 0x7f6299f1ae10>,
       <shapely.geometry.point.Point object at 0x7f6299f202e8>], dtype=object)

In [13]: shapely.__version__
Out[13]: '1.5.16'
```

vs 

```
In [4]: a = np.array([Point(1, 1), None], dtype=object)

In [5]: mask = np.array([False, True])

In [6]: fill_value = Point()

In [7]: print(fill_value)
GEOMETRYCOLLECTION EMPTY

[...]

PR #514, opened 2017-08-29
Add EmptyGeometry class

jdmcbr:
This provides an explicit means of creating an empty geometry object. Motivated by desire for a clear empty geometry type to use in na operations in geopandas. Thanks @sgillies for laying this out in #513, and @jorisvandenbossche for motivation in #510.

I went with `EmptyGeometry`, rather than `Empty`, thinking it made the context slightly clearer when used outside of `shapely`. I created an `__init__` mainly for the docstring, while retaining the `BaseGeometry` docstring for the class; maybe that's not necessary though?

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -943,6 +943,12 @@ class HeterogeneousGeometrySequence(GeometrySequence):
         return g
 
 
+class EmptyGeometry(BaseGeometry):
+    def __init__(self):
+        """Create an empty geometry."""
+        BaseGeometry.__init__(self)
+
+
 def _test():
     """Test runner"""
     import doctest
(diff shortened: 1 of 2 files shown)


sgillies:
@jdmcbr this looks just right to me. Anything else to add?

jdmcbr:
@sgillies No, if you're happy with it as is, I don't think there's anything else I'd like to add now.

sgillies:
@jdmcbr thanks!

leblancfg:
Question. Should perhaps geometry classes constructed from None-type values
```
import numpy as np

p = Point(np.nan, np.nan)
```
or even

```
p = Point(-45.2, np.nan)
```
return an `EmptyGeometry` instead of `POINT(nan, nan)`?

sgillies:
No, I don't think we should mix those cases.

PR #521, opened 2017-09-08
patch to find conda geos first

ocefpaf:
@sgillies I hope this is OK for your coding standards. This PR modifies `geos.py` to find the `conda` `geos` lib first, if `shapely` is in a conda env, and then tries to find the system libs.

This should not affect `wheels` or `frozen`  packages. The only pathological case I can imagine is if someone installed `shapely` with `conda` but expects it to find the system libs, which is a really bad idea but still possible if someone uninstalls the `conda` `geos`.

xref: https://github.com/conda-forge/shapely-feedstock/pull/28
ping @dopplershift and @Chilipp who are interested in this.

### DIFF
--- a/shapely/geos.py
+++ b/shapely/geos.py
@@ -71,12 +71,13 @@ if sys.platform.startswith('linux'):
         if len(geos_pyinstaller_so) == 1:
             _lgeos = CDLL(geos_pyinstaller_so[0])
             LOG.debug("Found GEOS DLL: %r, using it.", _lgeos)
+    elif os.getenv('CONDA_PREFIX', ''):
+        # conda package.
+        _lgeos = CDLL(os.path.join(sys.prefix, 'lib', 'libgeos_c.so'))
     else:
         alt_paths = [
             'libgeos_c.so.1',
             'libgeos_c.so',
-            # anaconda
-            os.path.join(sys.prefix, "lib", "libgeos_c.so"),
         ]
         _lgeos = load_dll('geos_c', fallbacks=alt_paths)
     free = load_dll('c').free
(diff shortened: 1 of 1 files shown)


davidh-ssec:
@ocefpaf Any idea if these changes are a problem for frozen environments based on a conda install? We are currently discussing some issues here to figure exactly what's going on with my installation issues: https://github.com/pyinstaller/pyinstaller/issues/2834

ocefpaf:
> @ocefpaf Any idea if these changes are a problem for frozen environments based on a conda install?

The check for `'frozen'` happens after the check for a conda env, so I guess this PR could be the culprit.

We can move the check order b/c checking for frozen first won't affect the check for a conda env. I can try to do that if you really want to support frozen environments based on a conda install. However, to be honest I don't really se why one would do that, but my applications are limited and I may be wrong.

davidh-ssec:
I see that you have this same patch on the conda-forge build so this could be the issue that I'm seeing.

As for why someone would do this, I think you need to get more users who aren't programmers ;). I have a GUI application that needs to be installed by non-programmers. My GUI application needs to work on Mac, Linux, and OSX and depends on a lot of GDAL/GEOS, PyQt4, and OpenGL libraries so installing and updating these myself for all 3 platforms would be impossible. Conda with conda-forge is by far the easiest solution I've found (even with the complications of getting gdal to behave) for doing this. We provide a custom conda channel for users who are comfortable installing that way, but an all-in-one installer with pyinstaller (with inno setup on windows) for everyone else.

ocefpaf:
> As for why someone would do this

Don't worry. As long as there is a why (even if I understand it or not) we should fix it.
In order to avoid the noise here I'll work on the patch in conda-forge and, if that works for you, I'll port it here. Is that OK?

davidh-ssec:
I'm checking it on my Windows VM now. There may be more than one issue or more than one directory that needs to be looked at for these libraries. Since it is only really noticeable when the application is frozen, it's kind of out of scope for conda-forge's tests and shapely's basic tests. I'll respond here or make a PR if/when I figure out a fix.

ocefpaf:
> It looks like CONDA_PREFIX isn't set in the frozen environment anyway.

If that is the case than the current setup should not affect your configuration :confused:

PR #523, opened 2017-09-17
Fix finding libgeos on windows in pyinstaller frozen application

davidh-ssec:
In a PyInstaller frozen application the libraries are placed in the same directory as the main executable (`bin`). PyInstaller creates a `sys._MEIPASS` attribute to point to this directory. Without this change running an application from anywhere other than the `bin` directory results in an exception for not finding the geos library.

I previously thought fixing this would require a change to @ocefpaf's conda-based changes in #521 but it seems like `CONDA_PREFIX` isn't set in a pyinstaller frozen application.

### DIFF
--- a/shapely/geos.py
+++ b/shapely/geos.py
@@ -129,7 +129,9 @@ elif sys.platform == 'win32':
         try:
             egg_dlls = os.path.abspath(
                 os.path.join(os.path.dirname(__file__), 'DLLs'))
-            if hasattr(sys, "frozen"):
+            if hasattr(sys, '_MEIPASS'):
+                wininst_dlls = sys._MEIPASS
+            elif hasattr(sys, "frozen"):
                 wininst_dlls = os.path.normpath(
                     os.path.abspath(sys.executable + '../../DLLS'))
             else:


durden:
Doesn't pyinstaller set sys.frozen and sys._MEIPASS? If so, why is frozen now the second option checked? Will this break things for non conda-forge installs?

davidh-ssec:
@durden Yes pyinstaller sets both. This has nothing to do with conda or conda-forge even though I thought that was the original issue. The issue with conda-forge is specific to pyinstaller and the PR mentioned above for that issue.

I could have put the MEIPASS check and usage inside the `if frozen` block, but it would have been redundant and added even more clutter to the code. I assumed that the frozen check was for some other python freezing technology that does put the geos libraries in a `DLLs` directory. There doesn't seem to be a consistent usage between the platform checks for `fallbacks` versus `PATH` versus something else so I put the code where it made the most sense to me.

durden:
Won't your change break anyone using sys.frozen now? The sys.frozen line will never be run because it's an elif and _MEIPASS always exists. Since both of those attributes always exist an elif doesn't seem right but maybe I'm missing something.

davidh-ssec:
@durden Maybe I'm missing something but _MEIPASS *only* exists for pyinstaller bundled environments from what I know and what google tells me. If that is not the case then of course I agree my changes won't work the way we want.

durden:
I think you're right, but sys.frozen also exists for pyinstaller. So this change breaks anyone using pyinstaller that has their libraries stored in the location added by the sys.frozen line. The change gives preference to your conda forge paths and not the ones previously used in the if frozen block.

davidh-ssec:
@durden This has *NOTHING* to do with conda or conda-forge. This is *after* the libraries have already been collected by pyinstaller and placed in one single directory.

PyInstaller has one (and only one) default location where it places library (binary) files. On Windows systems this is the same as the _MEIPASS directory. So if an environment has _MEIPASS specified THERE IS NO OTHER LOCATION that pyinstaller would have placed libraries.

If _MEIPASS does not exist and sys.frozen *does* exist then the application was frozen with something other than pyinstaller. In this case we will enter the second if statement and the non-pyinstaller software will need to place their frozen libraries in the `DLLs` folder mentioned in that chunk code.

> So this change breaks anyone using pyinstaller that has their libraries stored in the location added by the sys.frozen line.

This ^ is not possible. Pyinstaller only has *one* location that it places libraries and it has nothing to do with whether the original environment was conda-based or not. Your own pyinstaller hook does this.

durden:
Ok sounds good, didn't mean to belabor the point just trying to make sure what was going on.

shapely/geos.py
```diff
@@ -129,7 +129,9 @@ def load_dll(libname, fallbacks=None, mode=DEFAULT_MODE):
         try:
             egg_dlls = os.path.abspath(
                 os.path.join(os.path.dirname(__file__), 'DLLs'))
-            if hasattr(sys, "frozen"):
+            if hasattr(sys, '_MEIPASS'):
+                wininst_dlls = sys._MEIPASS
+            elif hasattr(sys, "frozen"):
```
sgillies: There are two other tests for `frozen` in this module. Do we need the same for those?

davidh-ssec: @sgillies The simplest test application would be a `test.py` with:

```
from shapely import geos
```

Then using pyinstaller (currently need the hook in github master iirc) run `pyinstaller test.py`. This should create `dist/test` directory with a `test` binary inside. Running this should run to import shapely and then return 0.

As for the other frozen tests, in my experience the others are fine the way they are. The OSX one already uses MEIPASS as a backup. The linux one seems to do the equivalent of MEIPASS, since in my experience MEIPASS is set to the directory of the interpreter/executable which is also the directory where all bundled libraries are placed.

davidh-ssec: @sgillies Were you thinking this would be added to the automated travis testing? Do you want me to try to add that? It seems a little odd to do it in shapely's repository, but it also seems unreasonable for pyinstaller to test a ton of the possible packages that someone could try to bundle.

sgillies: @davidh-ssec I would love to have such a test for Travis. To be honest, I'm concerned more about the untested code that exists in Shapely now (having taken on some freezing responsibilities) than I am about the burden on the pyinstaller project, but I'm happy to pick up some of that burden.

sgillies: @davidh-ssec Hi there, I don't have the time to learn about pyinstaller and write the tests, and don't want to commit an untested feature. Are you interested in doing it? Have you found another solution?

davidh-ssec: I will try to find the time to do it. There is a way to test this I think, but like I said it'll seem odd being in the shapely testing. I guess there isn't another way.

I haven't found a different way of doing this I just had built my application using my own forks and then stepped away from that project for a while. I'll try to find some time to do this this week.

austinorr: @sgillies I've been faced with the same issue as @davidh-ssec and arrived at the same solution that he did for building a shapely dependent exe on windows with PyInstaller. In order for you to be comfortable merging these 3 lines, would we have to build this test in AppVeyor in order for `sys.platform == 'win32'` to ever be `True` in the test case? Would we also need to build the test exe for Darwin systems too since those lines don't have a test?

It sounds like checking for and testing all use/build cases is a lot of work that could be avoided by letting the user set/overwrite the path to the dependencies they want `shapely` find. This solution would also be more resilient to changes that are outside Shapely's control, like conventions chosen by conda and PyInstaller that may change with each point release.
pseudocode:
```
import sys
import shapely
shapely.geos.dlls(sys._MEIPASS)
```
I'm a newcomer to this conversation so I don't know if this is feasible (but I am aware that this would need to be its own separate issue/PR and would require a good deal more time to implement). If this type of pattern is possible, it could be fully tested on Travis and eliminate the complicated maintenance required for supporting corner cases like custom installs and bundled executables built by 3rd party compilers. In the meantime, or if this pattern is not possible, maybe we can merge the _MEIPASS lines for windows as a 'non-supported or tested corner case courtesy' rather than treating it as a 'new feature' since windows-only lines are hard to test in Travis, and loading from `sys._MEIPASS` is already provided (though not tested) for Darwin users.

sgillies: @austinorr I think your suggestion is the ultimately the right way to do this and that a better-designed shapely module would invoke such a method (with defaults) right before the first use of the GEOS DLL.

That's more work than I'm able to do or oversee at the moment, and so I'm inclined to merge this PR if @snorfalorpagus agrees and put off any rewrite of shapely.geos to a future version.

snorfalorpagus:
This looks OK to me.

I agree that `shapely.geos` needs a refactor, but it's quite an undertaking. In the mean time this PR will help.

davidh-ssec:
Awesome thanks. @sgillies Any idea when 1.7 will be ready? Looks like the milestone is only 40% done. Thanks again.

Issue #526, opened 2017-10-04
User manual about shapely.geometry.Polygon is misleading


michelbl:
(Shapely version 1.6.1 installed from PyPI using pip)

User manuel about Polygons (https://shapely.readthedocs.io/en/latest/manual.html#polygons) defines the constructor with parameters `exterior` and `interiors`. However the true parameters are `shell` and `holes` (https://github.com/Toblerity/Shapely/blob/master/shapely/geometry/polygon.py#L208)

I think the problem lies in https://github.com/Toblerity/Shapely/blob/master/docs/manual.rst#polygons (https://raw.githubusercontent.com/Toblerity/Shapely/master/docs/manual.rst)

Also, documentation in the user manual and in the source file seems redundant (https://github.com/Toblerity/Shapely/blob/master/shapely/geometry/polygon.py#L198 vs https://shapely.readthedocs.io/en/latest/manual.html#polygons)

michelbl:
Oops this issue is a duplicate of #500

sgillies:
I'd like to keep the interior and exterior attributes and bring the constructor in sync. In 1.7 we could add interior and exterior keyword arguments and mark the future deprecation of shell and holes, using them as aliases for the former.

Issue #500, opened 2017-07-07
Polygon's docs __init__ arguments mismatch 


BebeSparkelSparkel:
On the [docs website](http://toblerity.org/shapely/manual.html#polygons) it says to created a polygon with the arguments **exterior** and **interiors** `class Polygon(exterior[, interiors=None])` but in [Shapely/shapely/geometry/polygon.py](https://github.com/Toblerity/Shapely/blob/master/shapely/geometry/polygon.py#L189) the arguments are actually **shell** and **holes** `def __init__(self, shell=None, holes=None):`.

sgillies:
@BebeSparkelSparkel thanks for flagging this!

sgillies:
Resolved by #527.

PR #527, opened 2017-10-04
Fix user manual (geometry.Polygon)

michelbl:
Connected to #526 and #500
Here I fix the wrong definition of the Polygon constructor in the user manual.

Also, the documentation about the "Attributes" (https://github.com/Toblerity/Shapely/blob/master/shapely/geometry/polygon.py#L196) seems to me a bit unclear and redundant with the user manual. However I don't know the policy of this repo about documentation.

### DIFF
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -442,7 +442,7 @@ constructor parameter.
 Polygons
 --------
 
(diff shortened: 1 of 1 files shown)


sgillies:
Thanks @michelbl !

Issue #524, opened 2017-09-19
ValueError: LineStrings must have at least 2 coordinate tuples


Charlesfox1:
A code I am working returns "ValueError: LineStrings must have at least 2 coordinate tuples" 
when I pass this:

splitted = shapely.ops.split(line, points) 

...where line is a multilinestring object and points a multipoint object, as described in the attached zipfile (POINT_fails.csv, LINE_fails.csv). Any idea why this is happening? The same code 'works' on the version of the information tagged with underscore 'works'. .

If it helps, I am picking up and handling these objects from pandas with the shapely.wkt.loads functionality. I know I am passing two shapely objects to the function to being with. 

[example data.zip](https://github.com/Toblerity/Shapely/files/1315207/example.data.zip)

Full error trace:

```
Traceback (most recent call last):
  File "Network_clean.py", line 202, in <module>
    iteration_roads, iteration_junctions = RoadCleanup(Phase_1_roads, Phase_1_junctions, 1)
  File "Network_clean.py", line 194, in RoadCleanup
    splitted = shapely.ops.split(line, points)
  File "C:\Users\charl\Anaconda2\lib\site-packages\shapely\ops.py", line 421, in split
    return GeometryCollection([i for part in geom.geoms for i in SplitOp.split(part, splitter).geoms])
  File "C:\Users\charl\Anaconda2\lib\site-packages\shapely\ops.py", line 442, in split
    return GeometryCollection(split_func(geom, splitter))
  File "C:\Users\charl\Anaconda2\lib\site-packages\shapely\ops.py", line 387, in _split_line_with_multipoint
    new_chunks.extend(SplitOp._split_line_with_point(chunk, pt))
  File "C:\Users\charl\Anaconda2\lib\site-packages\shapely\ops.py", line 363, in _split_line_with_point
    LineString(coords[:i+1]),
  File "C:\Users\charl\Anaconda2\lib\site-packages\shapely\geometry\linestring.py", line 49, in __init__
    self._set_coords(coordinates)
  File "C:\Users\charl\Anaconda2\lib\site-packages\shapely\geometry\linestring.py", line 94, in _set_coords
    self._geom, self._ndim = geos_linestring_from_py(coordinates)
  File "C:\Users\charl\Anaconda2\lib\site-packages\shapely\geometry\linestring.py", line 258, in geos_linestring_from_py
    "LineStrings must have at least 2 coordinate tuples")
ValueError: LineStrings must have at least 2 coordinate tuples
```

[...]

PR #530, opened 2017-10-21
Fix for #524, splitting closed ring with starting point

snorfalorpagus:
This PR fixes an issue identified in #524.

Test case here: https://github.com/Toblerity/Shapely/compare/master...snorfalorpagus:fix524?expand=1#diff-5f3469e182096a44abd37c7568afc200R94

Fix here: https://github.com/Toblerity/Shapely/compare/master...snorfalorpagus:fix524?expand=1#diff-2e25b147950af44f159ee507a4dee69dR381

I'm not 100% sure why this works...

### DIFF
--- a/shapely/ops.py
+++ b/shapely/ops.py
@@ -257,7 +257,7 @@ def transform(func, geom):
 
 def nearest_points(g1, g2):
     """Returns the calculated nearest points in the input geometries
-    
+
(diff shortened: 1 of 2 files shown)


sgillies:
@snorfalorpagus would you be willing to change this PR to be against the maint-1.6 branch so we can make a 1.6.2 release?

PR #531, opened 2017-10-22
Raise TypeError if Point is init with > 3 args

### LINKED ISSUE #525: Point initialisation ignores extra arguments
The `Point` class can be initialised with a number of different arguments: 2 coords, 3 coords, an existing Point, and more. I came across the following case where the arguments are invalid, but no exception is raised. Instead they are silently dropped.

```python
>>> p = Point(1,2,3,4,5,6,7) # this should error!
>>> p.coords[:]
[(1.0, 2.0)]
```

I think this should raise a `TypeError`.

snorfalorpagus:
Closes #525.


sgillies:
Thanks @snorfalorpagus !

PR #533, opened 2017-10-24
Fix for #524, splitting closed ring with starting point

snorfalorpagus:
Same as #530, aimed at `maint-1.6`. Sorry I can't get the hang of these maintenance branches.


sgillies:
Resolves #524

PR #538, opened 2017-10-31
Don't double-free in STRtree's __del__ if it is called multiple times.

zerebubuth:
There are some circumstances in which CPython can call `__del__` on an object twice. For me, that was happening when running a script under [`python-flamegraph`](https://github.com/evanhempel/python-flamegraph). This is because it runs a separate thread which is walking the stack, causing it to keep a reference to the `STRtree` object and attempt to delete it twice.


sgillies:
I made a trial switch of base to maint-1.6, which I'd like to do when you're finished.

zerebubuth:
Nothing to add, thanks! Merge at your convenience.

PR #544, opened 2017-11-17
Let user set colors for GeometryCollection elements in Jupyter Notebook

invisibleroads:
```
from shapely.geometry import GeometryCollection, Point

GeometryCollection([
    Point(0, 0),
    Point(1, 1),
], colors=['red', 'blue'])
```

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -856,9 +856,6 @@ class BaseMultipartGeometry(BaseGeometry):
         """
         if self.is_empty:
             return '<g />'
-        if hasattr(self, 'colors') and self.colors:
-            return '<g>' + ''.join(p.svg(scale_factor, c) for p, c in zip(
-                self, self.colors)) + '</g>'
         if color is None:
             color = "#66cc99" if self.is_valid else "#ff3333"
         return '<g>' + ''.join(p.svg(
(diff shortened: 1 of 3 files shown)


invisibleroads:
@minrk

sgillies:
@invisibleroads this is a neat idea, but I can't support adding a colors attribute to geometries. I've been saying "no" to proposals for an srid attribute for a long time, so it's not that I'm specifically against colors. I don't want to complicate the classes more than they already are or encumber them with new visualization features and responsibilities.

invisibleroads:
Heh that's true, if you accept this pull request then you might get inundated by feature requests for customized visualization which will complicate the original scope of shapely as an optimized GEOS wrapper.

I moved the class to a [separate package](https://github.com/invisibleroads/geotable/blob/master/geotable/__init__.py) and renamed the class to ColorfulGeometryCollection.

Thank you very much for your work on shapely and other spatial packages. We use it extensively in our [algorithms for planning off-grid electricity infrastructure](https://crosscompute.com/n/JrbhgevrX9nf2wr2tF7iZZdQfJr7I8A0/_/Logs/Assemble%20Algorithm%20(1st%20Implementation)%2020171115).

https://crosscompute.com/n/JrbhgevrX9nf2wr2tF7iZZdQfJr7I8A0

sgillies:
@invisibleroads thanks for understanding! That looks like a good approach to me.

PR #548, opened 2017-11-26
Fixed bug in unpickling of LinearRing. Closes #547.

### LINKED ISSUE #547: Serialization issue with LinearRing

When pickling a LinearRing and trying to build a Polygon from it, the Polygon is not initialized, with a trace from geos: `Shell is not a LinearRing`. Attempting to run boolean operations results in `ValueError: Null geometry supports no operations`.

```python
def test_pickle_linear_ring():
    import pickle
    from shapely.geometry import Polygon, LinearRing

    ring = LinearRing([(1, 1), (2, 1), (2, 2), (1, 2), (1, 1)])
    assert Polygon(ring).difference(Polygon(ring)).is_empty

    pickled = pickle.loads(pickle.dumps(ring))
    assert Polygon(pickled).difference(Polygon(pickled)).is_empty
```

## Operating system
Fedora 27, x86_64

- `Shapely[vectorized]==1.6.2.post1` installed from PyPI using pip
- geos 3.6.1
- Python 3.6.3

snorfalorpagus:
This PR fixes the issue identified in #547.

Geometries are pickled using the WKB format, which does not differentiate between LINESTRING and LINEARRING objects. When a LinearRing is pickled/unpickled the result is a Shapely LinearRing with a GEOS LineString inside, which can lead to issues later. This change works around the problem by copying the coordinate sequence of the geometry after it's read from WKB into the correct geometry type. It's slightly frustrating that we have to clone the sequence, but it's required due to GEOS's ownership model.

### DIFF
--- a/shapely/geometry/polygon.py
+++ b/shapely/geometry/polygon.py
@@ -71,6 +71,15 @@ class LinearRing(LineString):
 
     coords = property(_get_coords, _set_coords)
 
+    def __setstate__(self, state):
+        """WKB doesn't differentiate between LineString and LinearRing so we
+        need to move the coordinate sequence into the correct geometry type"""
+        super(LinearRing, self).__setstate__(state)
+        cs = lgeos.GEOSGeom_getCoordSeq(self.__geom__)
+        cs_clone = lgeos.GEOSCoordSeq_clone(cs)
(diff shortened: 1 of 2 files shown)


sgillies:
3 builds failed due to a transient server error. I restarted them and they have all passed.

PR #573, opened 2018-03-26
shape() method can parse empty geojson

PPACI:
In the current code, if you try to save and parse an empty polygon, the shape() method will result in an IndexError.

```python
from shapely.geometry import Polygon, shape, mapping
shape(mapping(Polygon()))

  File "...", line 2910, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-3-305ea9043881>", line 1, in <module>
    shape(mapping(Polygon()))
  File ".../python3.5/site-packages/shapely/geometry/geo.py", line 35, in shape
    return Polygon(ob["coordinates"][0], ob["coordinates"][1:])
IndexError: tuple index out of range 
```

This pull request aim to resolve this issue with a simple Try/Except

### DIFF
--- a/shapely/geometry/geo.py
+++ b/shapely/geometry/geo.py
@@ -32,7 +32,10 @@ def shape(context):
     elif geom_type == "linestring":
         return LineString(ob["coordinates"])
     elif geom_type == "polygon":
-        return Polygon(ob["coordinates"][0], ob["coordinates"][1:])
+        try:
+            return Polygon(ob["coordinates"][0], ob["coordinates"][1:])
+        except IndexError:
+            return Polygon()
     elif geom_type == "multipoint":
         return MultiPoint(ob["coordinates"])
     elif geom_type == "multilinestring":
--- a/shapely/geometry/geo.py
+++ b/shapely/geometry/geo.py
@@ -32,10 +32,10 @@ def shape(context):
     elif geom_type == "linestring":
         return LineString(ob["coordinates"])
     elif geom_type == "polygon":
-        try:
-            return Polygon(ob["coordinates"][0], ob["coordinates"][1:])
-        except IndexError:
+        if not ob["coordinates"]:
             return Polygon()
+        else:
+            return Polygon(ob["coordinates"][0], ob["coordinates"][1:])
     elif geom_type == "multipoint":
         return MultiPoint(ob["coordinates"])
     elif geom_type == "multilinestring":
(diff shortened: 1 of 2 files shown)


shapely/geometry/geo.py
```diff
@@ -32,7 +32,10 @@ def shape(context):
     elif geom_type == "linestring":
         return LineString(ob["coordinates"])
     elif geom_type == "polygon":
-        return Polygon(ob["coordinates"][0], ob["coordinates"][1:])
+        try:
+            return Polygon(ob["coordinates"][0], ob["coordinates"][1:])
+        except IndexError:
+            return Polygon()
```
sgillies: This will work, yes, but instead I think we should avoid the IndexError and do something like this instead

```
elif geom_type == "polygon":
    if not ob["coordinates"]:
        return Polygon()
    else:
        ...
```

PPACI: the problem is that `ob["coordinates"]` exist.
in fact the dict is like `{"coordinates":()}`. But we could like 
```python
elif geom_type == "polygon":
    if len(ob["coordinates"])==0:
        return Polygon()
```

But then i personally find it simpler, and a bit more pythonic but that's totally subjective, to have a try except statement.

PPACI: fixed

PPACI:
I've added the "assertIsNotNone" unit test.

tests/test_mapping.py
```diff
@@ -7,6 +7,10 @@ def test_point(self):
         m = mapping(Point(0, 0))
         self.assertEqual(m['type'], 'Point')
         self.assertEqual(m['coordinates'], (0.0, 0.0))
+        
+    def test_empty_polygon():
```
sgillies: This method is missing its `self` arg. I expect it to pass after that @PPACI .

tests/test_mapping.py
```diff
@@ -8,7 +8,7 @@ def test_point(self):
         self.assertEqual(m['type'], 'Point')
         self.assertEqual(m['coordinates'], (0.0, 0.0))
         
-    def test_empty_polygon():
+    def test_empty_polygon(self):
         """Empty polygons will round trip without error"""
         self.assertIsNotNone(mapping(Polygon()))
```
sgillies: The only thing missing now is `from shapely.geometry import Polygon` at the top of the file.

sgillies:
Great, thanks @PPACI !

PR #593, opened 2018-05-28
Added srid keyword to wkb.dumps. Closes #592.

### LINKED ISSUE #592: Implement wkb.dumps(geom, srid=XXX)
In #132 @sgillies suggested an `srid` keyword argument to the `wkb.dumps` function.

Currently dumps has a `include_srid` keyword argument, which defaults to False. If set to True it does work as intended, although there is no way in Shapely to set the SRID of a geometry without poking the GEOS internals (GEOSSetSRID/GEOSGetSRID).

Are you still amenable to an `srid=4326` style argument to dumps? If so, does it need to clone the geometry before exporting, otherwise the function will not be idempotent (although not in any way visible from Shapely).

snorfalorpagus:
Added the `srid` keyword argument to `wkb.dumps` to include an SRID in WKB data, as suggested in #592.

I've added some tests for WKB in `test_wkb.py` (to match `test_wkt.py`). There are already some in `test_persist.py` which I've left alone. Perhaps this could be expanded in the future, but it's not essential.

An observation is that `pickle` uses `wkb.dumps` to store data. This defaults to not exporting the SRID (even if one is present in the geometry after loading with `loads`. This means that the round-trip for a geometry with an SRID isn't perfect. However I think this is acceptable as generally we're ignoring the SRID in Shapely.

### DIFF
--- a/shapely/wkb.py
+++ b/shapely/wkb.py
@@ -1,7 +1,8 @@
 """Load/dump geometries using the well-known binary (WKB) format
 """
 
-from shapely import geos
+from shapely.geos import WKBReader, WKBWriter, lgeos
+from shapely.geometry.base import geom_factory
 
 # Pickle-like convenience functions
 
@@ -9,7 +10,7 @@ def loads(data, hex=False):
     """Load a geometry from a WKB byte string, or hex-encoded string if
     ``hex=True``.
     """
-    reader = geos.WKBReader(geos.lgeos)
+    reader = WKBReader(lgeos)
     if hex:
         return reader.read_hex(data)
     else:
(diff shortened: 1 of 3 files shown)


shapely/wkb.py
```diff
@@ -20,12 +21,29 @@ def load(fp, hex=False):
...
+        Spatial reference system ID to include in the output. The default value
+        means no SRID is included.
+    **kw : kwargs
+        See available keyword output settings in ``shapely.geos.WKBWriter``."""
+    if srid is not None:
+        # clone the object and set the SRID before dumping
+        geom = lgeos.GEOSGeom_clone(ob._geom)
+        lgeos.GEOSSetSRID(geom, srid)
+        ob = geom_factory(geom)
+        kw["include_srid"] = True
```
sgillies: 👍

PR #595, opened 2018-05-28
Raise an exception when preparing an already prepared geometry.

### LINKED ISSUE #577: preparing an already prepared geometry causes a segfault

I'd expect either an error to be raised or for prep to just return the input geometry.

import shapely.geometry, shapely.prepared
shapely.prepared.prep(shapely.prepared.prep(shapely.geometry.Point(1,2)))

## Operating system

Ubuntu linux 17.10, python 2.7.14

1.6.4.post1 installed from pypi using pip

snorfalorpagus:
Closes #577.

Also converted test_prepared.py to use pytest syntax.

### DIFF
--- a/shapely/errors.py
+++ b/shapely/errors.py
@@ -31,3 +31,7 @@ class TopologicalError(ShapelyError):
 
 class PredicateError(ShapelyError):
     """A geometric predicate has failed to return True/False."""
+
+
+class GeometryAlreadyPreparedError(ShapelyError):
+    """Attempted to prepare an already prepared geometry"""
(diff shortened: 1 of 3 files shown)


shapely/errors.py
```diff
@@ -31,3 +31,7 @@ class TopologicalError(ShapelyError):
 
 class PredicateError(ShapelyError):
     """A geometric predicate has failed to return True/False."""
+
+
+class GeometryAlreadyPreparedError(ShapelyError):
+    """Attempted to prepare an already prepared geometry"""
```
sgillies: @snorfalorpagus how about a more generic GeometryPreparationError ?

sgillies: Thinking about this a little more, I think we could just return the object without an error if it is an instance of PreparedGeometry, analogous to the result of `int(4)` or `float(3.14)`.

PR #597, opened 2018-05-31
construct LineString but ~LinearRing from Points

gdmcbain:
```python
 data = [Point(0, 0), (1, 0), (2, 1)]
 LineString(data)
```
works but
```python
LinearRing(data)
```
raises
> TypeError: object of type 'Point' has no len()


drnextgis:
Since https://github.com/Toblerity/Shapely/pull/732 is merged we can close that PR.

PR #610, opened 2018-07-17
Add a runtime Python version check in setup.py

mwtoews:
While (accidentally) attempting to build Shapely on an old computer with Python 2.6, I found an incomprehensible error:
```
  File "setup.py", line 167, in <module>
    fp.write(str(shapely_version))
  File "/home/mtoews/src/Shapely/_vendor/packaging/version.py", line 252, in __str__
    parts.append(".post{}".format(self._version.post[1]))
ValueError: zero length field name in format
```
This PR provides a clearer message:
> RuntimeError: Shapely requires Python 2.7, >=3.4

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -75,6 +75,11 @@ from distutils.errors import CCompilerError, DistutilsExecError, \
 
 from _vendor.packaging.version import Version
 
+# Ensure minimum version of Python is running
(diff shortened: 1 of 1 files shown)


sgillies:
Because Python 2.6 requires `".post{0}".format(self._version.post[1])`. Okay, this makes sense to me.

PR #611, opened 2018-07-18
raise ValueError if LinearRing is made with invalid LineString

### LINKED ISSUE #609: Crash with invalid LineString -> LinearRing
```python
In [1]: import shapely.geometry as sgeom

In [2]: ls = sgeom.LineString([(0, 0), (0, 0), (0, 0)])

In [3]: lr = sgeom.LinearRing(ls)
IllegalArgumentException: Invalid number of points in LinearRing found 3 - must be 0 or >= 4

In [4]: lr.is_ccw
[1]    16446 segmentation fault  ipython
```
I see this on my Mac with Shapely 1.6.4 from conda-forge. I've also had the problem reported from users on other OSes.

mwtoews:
Fixes #609

### DIFF
--- a/shapely/_geos.pxi
+++ b/shapely/_geos.pxi
@@ -43,6 +43,7 @@ cdef extern from "geos_c.h":
     char GEOSHasZ_r(GEOSContextHandle_t, GEOSGeometry *) nogil
     char GEOSisRing_r(GEOSContextHandle_t, GEOSGeometry *) nogil
     char GEOSisClosed_r(GEOSContextHandle_t, GEOSGeometry *) nogil
+    char GEOSisValid_r(GEOSContextHandle_t, GEOSGeometry *) nogil
 
 
 cdef GEOSContextHandle_t get_geos_context_handle():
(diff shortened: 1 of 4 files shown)


sgillies:
@mwtoews this change contradicts https://shapely.readthedocs.io/en/stable/manual.html#LinearRing. I'm not sure I want to support a new mode for the constructor. I think I could get behind a `LinearRing.from_line()` class method.

mwtoews:
@sgillies it looks like a LineString constructor mode for LinearRings were added nearly four years ago for #170 in 45ee253db42616cc4dcf7ffa93010a0f2e039429 by @snorfalorpagus

I'd rather have a class method too, but we'd need to deprecate this constructor mode. And ideally address the invalid object bug reported in #609 (which I can partially reproduce, without a segmentation fault).

Also, I just realised that you can create invalid LinearRings (if you want to), so the invalid check in this PR is not the correct solution.

sgillies:
@mwtoews how about this for a plan: add a classmethod, and use it within the existing API? Also, I'd help update the documentation.

mwtoews:
@sgillies sounds good. I can revise this in a separate PR, but it may take the patience of perhaps a few weeks.

PR #621, opened 2018-08-02
Added Python 3.7 to Travis 

expobrain:


.travis.yml
```diff
@@ -1,18 +1,33 @@
...
-  - "3.6"
+cache:
+  - pip
+  - apt
 
 env:
   - "TRAVIS_SPEEDUP_OPTS=--with-speedups"
   - "TRAVIS_SPEEDUP_OPTS=--without-speedups"
 
+python:
```
snorfalorpagus: Why isn't this as simple as adding 3.7 to the existing list?

snorfalorpagus:
@expobrain Adding Python 3.7 to Travis is a good idea. However the `.travis.yaml` has changed quite a bit since the copy in `maint-1.6`. Can you open another PR with the change against `master`?

https://github.com/Toblerity/Shapely/blob/master/.travis.yml

expobrain:
@snorfalorpagus sure! Give some time to check out `master` and update the configuration

PR #623, opened 2018-08-08
Improved STRtree docs

kannes:
Untested in terms of RST rendering!


sgillies:
@kannes I won't be able to fully review until I'm done traveling, but I like what I see. Thanks!

kannes:
Hm, considering the references culminating on https://github.com/Toblerity/Shapely/issues/618 it might be a good idea to put the monkey-patching example to somewhere more general. On the other hand this could invite people to use it often which is *not* a good idea. The only reason I like its usage here is that the STRTree is "lossy" (if the order of your geometries was significant) and this is the only way of saving some kind of reference.

sventech:
This is very helpful, thanks! One key feature missing from SRTree as initially documented is the ability to determine an ID or other metadata of the shapes that match. This monkey patching example makes the index more clearly useful. As you say, although it is common in Ruby it is not a Pythonic procedure.

docs/manual.rst
```diff
@@ -2358,13 +2359,13 @@ constructor to create an R-tree that you can query with another geometric object
 
   `New in version 1.4.0`.
 
-The `query` method on `STRtree` returns a list of all geometries in the tree that
-intersect the provided geometry argument. If you want to match geometries of a
-more specific spatial relationship (eg. crosses, contains, overlaps), consider
-performing the query on the R-tree, followed by a manual search through the
-returned subset using the desired binary predicate.
+The `query` method on `STRtree` returns a list of all geometries in the tree whose
```
sgillies: @kannes what would you think about "which have extents that intersect" ?

docs/manual.rst
```diff
@@ -2373,13 +2374,35 @@ add or remove geometries.
...
+      >>> points_numbered = []
+      >>> for i, point in enumerate(points):
+      >>>     point = deepcopy(point)  # let's not mutate the originals
+      >>>     point.i = i
+      >>>     points_numbered.append(point)
+      >>> tree = STRtree(points_numbered)
+      >>> result = tree.query(query_geometry)
+      >>> [(o.i, o.wkt) for o in result]
+      [(1, 'POINT (1 1)'), (2, 'POINT (2 2)'), (3, 'POINT (3 3)')]
+
```
sgillies: @kannes I think this is a fine example and a necessary one because shapely objects cannot currently be hashed and so can't serve as keys for an auxiliary dict. This is not quite the same thing as "monkey patching", so let's just say "attributes can be added to shapely objects and they will be saved in the STRtree.

docs/manual.rst
kannes: I like it!

docs/manual.rst
kannes: I think I got that term from you a long time ago, I have no strong feelings about it (though monkeys in trees are a nice sight). It feels a bit ugly to do things like this, so having a "ugly" name for the process seemed nice to me.

kannes:
Yay!

You are welcome to change anything as you wish, your experience is much more in-depth. I won't have time to truly have a look of this too soon , so you have my "go ahead" in any case.

kannes:
Oh and absolutely no time to do changes and resolve conflicts, sorry.

PR #627, opened 2018-08-11
Added Python 3.7 to Travis

expobrain:


.travis.yml
```diff
@@ -38,6 +38,24 @@ matrix:
       env:
         - SPEEDUPS=0
         - NUMPY=0
+    - python: "3.7"
+      sudo: required
```
sgillies: @expobrain thanks for the contribution! Why `sudo: required`? Is xenial not available otherwise?

I'm largely 👎 on requiring sudo for tests. We should be testing in an environment as close to normal use as possible.

PR #650, opened 2018-10-22
Added convenience method `anchored_box()` to geo.py

awiebe:
The `anchored_box()` method provides a standard way to specify an anchor vertex for boxes, which is used in many image processing tools.  A corner, edge centre, or rectangle centroid may be specified as the anchor point.

### DIFF
--- a/shapely/geometry/geo.py
+++ b/shapely/geometry/geo.py
@@ -18,6 +18,33 @@ def box(minx, miny, maxx, maxy, ccw=True):
         coords = coords[::-1]
     return Polygon(coords)
 
+def anchored_box(anchor='nw',x,y,width,height,ccw=True):
+    """Return a rectangular polygon extruded from the specified anchor point.
+    The anchor point can be a corner or the centre of an edge in relation to the centre point, or the centre itself.
+    Anchor options from left to right, top to bottom are: nw,n,ne,w,c,e,sw,s,se . (Defaults to North-West)
+    """
+    if anchor =='nw':
+        return box(x,y-height,x+width,y,ccw)
+    elif anchor =='w':
+        return box(x,y-height/2.0,x+width,y+height/2.0,ccw)
+    elif anchor =='sw':
+        return box (x,y,x+width,y+height,ccw)
+    elif anchor =='n':
+        return box(x-width/2.0,y-height,x+width/2.0,y,ccw)
+    elif anchor =='c':
+        return box(x-width/2.0,y-height/2.0,x+width/2.0,y+height/2.0,ccw)
+    elif anchor == 's':
+        return box(x-width/2.0,y,x+width/2.0,y+height,ccw)
+    elif anchor == 'ne':
+        return box(x-width,y-height,x,y,ccw)
+    elif anchor == 'e':
+        return box(x-width,y-height/2.0,x,y+height/2.0,ccw)
+    elif anchor =='se':
+        return box(x-width,y,x,y+height,ccw)
+    else:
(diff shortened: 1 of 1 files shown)


shapely/geometry/geo.py
```diff
@@ -18,6 +18,33 @@ def box(minx, miny, maxx, maxy, ccw=True):
...
+    """Return a rectangular polygon extruded from the specified anchor point.
+    The anchor point can be a corner or the centre of an edge in relation to the centre point, or the centre itself.
+    Anchor options from left to right, top to bottom are: nw,n,ne,w,c,e,sw,s,se . (Defaults to North-West)
+    """
+    if anchor =='nw':
+        return box(x,y-height,x+width,y,ccw)
+    elif anchor =='w':
+        return box(x,y-height/2.0,x+width,y+height/2.0,ccw)
+    elif anchor =='sw':
+        return box (x,y,x+width,y+height,ccw)
```
snorfalorpagus: Remove the space after `box`.

shapely/geometry/geo.py
```diff
@@ -18,6 +18,33 @@ def box(minx, miny, maxx, maxy, ccw=True):
         coords = coords[::-1]
     return Polygon(coords)
 
+def anchored_box(anchor='nw',x,y,width,height,ccw=True):
```
snorfalorpagus: Arguments should have a space after the comma, e.g.

```
def anchored_box(anchor="nw", x, y, width, height, ccw=True):
```

shapely/geometry/geo.py
```diff
@@ -18,6 +18,33 @@ def box(minx, miny, maxx, maxy, ccw=True):
...
+    """Return a rectangular polygon extruded from the specified anchor point.
+    The anchor point can be a corner or the centre of an edge in relation to the centre point, or the centre itself.
+    Anchor options from left to right, top to bottom are: nw,n,ne,w,c,e,sw,s,se . (Defaults to North-West)
+    """
+    if anchor =='nw':
+        return box(x,y-height,x+width,y,ccw)
```
snorfalorpagus: Same comment re spaces between arguments for all of these lines.

shapely/geometry/geo.py
```diff
@@ -18,6 +18,33 @@ def box(minx, miny, maxx, maxy, ccw=True):
...
+    elif anchor == 's':
+        return box(x-width/2.0,y,x+width/2.0,y+height,ccw)
+    elif anchor == 'ne':
+        return box(x-width,y-height,x,y,ccw)
+    elif anchor == 'e':
+        return box(x-width,y-height/2.0,x,y+height/2.0,ccw)
+    elif anchor =='se':
+        return box(x-width,y,x,y+height,ccw)
+    else:
+        raise ValueError('Unknown anchor point')
```
snorfalorpagus: Include the expected values in this exception.

snorfalorpagus:
Thanks for this @awiebe. It looks like a useful function.

I've added some inline comments - all minor.

Could you also add a unit test for this, and an entry to the manual?

shapely/geometry/geo.py
snorfalorpagus: Also, see the error on Travis. You have a non-default argument (x, y, etc.) following a default argument (anchor). The easiest thing is to just remove the default. If we do keep it we'll have to argue about if the default should be NW or SW. ;)

sgillies:
I'm going to close this one as it's coming up on a year with no follow up.

PR #658, opened 2018-11-19
Fixing bug in ops.substring

shongololo:
If `min_dist` is equal to 0 (i.e. the first coordinate) then the loop immediately breaks.

This means that internal coordinates are skipped because only `(start_point.x, start_point.y)` and `(end_point.x, end_point.y)` are written to the new `vertex_list`.

By changing the less than to a less-than-equal sign the loop behaves as intended.

### DIFF
--- a/shapely/ops.py
+++ b/shapely/ops.py
@@ -516,9 +516,9 @@ def substring(geom, start_dist, end_dist, normalized=False):
     
     vertex_list = [(start_point.x, start_point.y)]
     coords = list(geom.coords)
-    for i, p in enumerate(coords):
+    for i, p in coords:
         pd = geom.project(Point(p))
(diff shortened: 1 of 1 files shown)


shongololo:
Fixes #659

sgillies:
@shongololo I see what you mean. Can you add a test to tests/test_substring.py to confirm the fix? Something like this after 70 should suffice.

```
    @unittest.skipIf(geos_version < (3, 2, 0), 'GEOS 3.2.0 required')
    def test_return_substring_min_dist-zero(self):
        """Confirm fix of #658"""
        self.assertTrue(substring(self.line2, 0.0, 7.0).equals(LineString(([3.0, 0.0], [3.0, 6.0], [4.0, 6.0]))))
```

Issue #659, opened 2018-11-19
ops.substring omits internal coordinates if starting with first coordinate


songololo:
## Expected behavior and actual behavior.

`ops.substring` omits internal coordinates if starting with first coordinate. Pull Requests #658 and #660 (tests) submitted.

## Steps to reproduce the problem.
```
from shapely.geometry import LineString, Point

start = 0
end = 0.75
normalized = True
geom = LineString([[0, 0], [0.5, 0.25], [1, 1]])

start_point = geom.interpolate(start, normalized)
end_point = geom.interpolate(end, normalized)

min_dist = min(0, 0.75)
max_dist = max(0, 0.75)
if normalized:
    min_dist *= geom.length
    max_dist *= geom.length

# incorrect behaviour:
vertex_list = [geom.coords[0]]
coords = list(geom.coords)
for p in coords:
    pd = geom.project(Point(p))
    if pd < min_dist:
        pass
    elif min_dist < pd < max_dist:
        vertex_list.append(p)
    else:
        break
vertex_list.append((end_point.x, end_point.y))
# intermediate coordinates are missing:
print(vertex_list)  # [(0.0, 0.0), (0.7974782908817447, 0.696217436322617)]

# correct behaviour:
vertex_list = [geom.coords[0]]
coords = list(geom.coords)
for p in coords:
    pd = geom.project(Point(p))
    if pd <= min_dist:
        pass
    elif min_dist < pd < max_dist:
        vertex_list.append(p)
    else:
        break
vertex_list.append((end_point.x, end_point.y))
# intermediate coordinates are now present:
print(vertex_list)  # [(0.0, 0.0), (0.5, 0.25), (0.7974782908817447, 0.696217436322617)]
```

## Operating system

Not applicable.

## Shapely version and provenance

1.7a1

albin3:
Found same issue, it seems has been fixed. Looking forward to release a new version. 😊

PR #660, opened 2018-11-20
add ops.substring test for edge case: endpoint geoms with internal coords

shongololo:
Adding edge case (no pun intended...) to test that internal coordinates are preserved when cutting to either end of line.
Addressed by #658 and #659


shongololo:
Note that the tests will fail against the master branch until #658 is merged.

PR #670, opened 2019-01-16
DOC: switch cascaded_union and unary_union

jorisvandenbossche:
I was reading the docs and saw that `cascaded_union` is still more prominently explained than `unary_union`, and I was assuming you should now prefer to use `unary_union` over `cascaded_union`, so therefore I switched them in order (long explanation for `unary_union`, and then short one for `cascaded_union` referring to the other, instead of the other way around now).

But then I saw in the code that for more recent versions of GEOS, they are actually simply an alias, and `cascaded_union` is not using the old cascaded union implementation any more. So maybe `unary_union` is not necessarily "recommended" over `cascaded_union`?

### DIFF
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -1506,7 +1506,7 @@ boundaries.
 .. note::
 
   :meth:`union` is an expensive way to find the cumulative union
-  of many objects. See :func:`shapely.ops.cascaded_union` for a more effective
+  of many objects. See :func:`shapely.ops.unary_union` for a more effective
   method.
 
 Constructive Methods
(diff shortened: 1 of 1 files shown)


docs/manual.rst
```diff
@@ -2095,20 +2095,20 @@ efficient than accumulating with :meth:`union`.
...
 
+.. function:: shapely.ops.cascaded_union(geoms)
+
+  Returns a representation of the union of the given geometric objects.
+
   .. note::
 
-     In 1.2.16 :func:`shapely.ops.cascaded_union` is superceded by
+     Since 1.2.16 :func:`shapely.ops.cascaded_union` is superceded by
      :func:`shapely.ops.unary_union` if GEOS 3.2+ is used. The unary union
```
jorisvandenbossche: For GEOS 3.3+ this note doesn't seem to be fully correct, as `cascaded_union` is from then on aliased to `unary_union` instead of using the older implementation

sgillies: True. I think I'll let it go for now.

Issue #668, opened 2019-01-07
Feature request: Add STRtree.nearest function


FuriousRococo:
## Note
Since Shapely support no more than GEOS 3.4.0 right now, it will be great to add other functions like **GEOSSTRtree_nearest** in GEOS 3.6.0. So I was trying to add this function by myself and meet some problems.

## Problems
GEOS source code is like:
```
const GEOSGeometry *
GEOSSTRtree_nearest (geos::index::strtree::STRtree *tree,
                     const geos::geom::Geometry *g)
{
    return GEOSSTRtree_nearest_r( handle, tree, g);
}

const void* GEOSSTRtree_nearest_generic(GEOSSTRtree *tree,
                                        const void* item,
                                        const GEOSGeometry* itemEnvelope,
                                        GEOSDistanceCallback distancefn,
                                        void* userdata)
{
    return GEOSSTRtree_nearest_generic_r( handle, tree, item, itemEnvelope, distancefn, userdata);
}
```

Then I added code in strtree.py, ctypes_declarations.py and test.py and tried to run the test but got the error **Segmentation fault: 11**
**strtree.py**
```
from shapely.geometry.base import geom_factory
class STRtree:
    def nearest(self, geom):
        if self._n_geoms == 0:
            return None

        return geom_factory(lgeos.GEOSSTRtree_nearest(self._tree_handle, geom._geom))
```
**ctypes_declarations.py**
```
def prototype(lgeos, geos_version):
    if geos_version >= (3, 6, 0):
        lgeos.GEOSSTRtree_nearest.argtypes = [c_void_p, c_void_p]
        lgeos.GEOSSTRtree_nearest.restype = c_void_p
```
**test.py**
```
from shapely.geometry import Point, Polygon
from shapely.strtree import STRtree

a = []
a.append(Polygon([(1,0),(2,0),(2,1),(1,1)]))
a.append(Polygon([(0,2),(1,2),(1,3),(0,3)]))
a.append(Polygon([(-1.5,0),(-2.5,0),(-2.5,-1),(-1.5,-1)]))
tree = STRtree(a)
tree.nearest(Point(0,0))
```

I thought my problem is about data processing between the ctypes and geometry, so I tried to reimplement to **shared_path** function but still got something wrong.
```
from shapely.geometry import LineString
from shapely.geometry.base import geom_factory
from ctypes import CDLL, c_void_p

[...]

PR #672, opened 2019-01-25
Add nearest function to STRtree (issue #668)

FuriousRococo:
Address #668 by adding GEOSSTRtree_nearest_generic function from GEOS library.


FuriousRococo:
@sgillies Please review the test_strtree_nearest.py file.

PR #673, opened 2019-01-31
Fixing doc typo in matrix math for affine_transform().

vpradeep07:
In the documentation for `affine_transform()`, the matrix multiplication math don't match the rest of the equations in the docs and also don't match the implementation:
https://github.com/Toblerity/Shapely/blob/5d18af283e485a427d121e05fbf8e9968db4a569/shapely/speedups/_speedups.pyx#L480-L482

This PR updates the matrix equations in the documentation and code comments to be consistent with the other docs and existing code implementation.

### DIFF
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -1766,12 +1766,20 @@ preserved or supported by 3D affine transformations.
   which represents the augmented matrix:
 
   .. math::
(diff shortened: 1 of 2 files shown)


sgillies:
@vpradeep07 thanks for the PR, but it looks like we got the fix in another one, so I'll close this one.

PR #680, opened 2019-02-21
Travis CI: Upgrade from Python 3.4 --> 3.5

cclauss:
23 days until Python 3.4 end of life.  https://docs.python.org/devguide/index.html#branchstatus

Also, [Travis are now recommending removing the __sudo__ tag](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).

"_If you currently specify __sudo: false__ in your __.travis.yml__, we recommend removing that configuration_"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: python
-sudo: false
 cache:
   - pip
   - apt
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,5 +1,4 @@
 language: python
-sudo: false
```
sgillies: 👍

PR #682, opened 2019-03-01
Fix misorderded points in substrings

cquest:
When start_dist > end_dist, the first/last points where disordered


sgillies:
@cquest would you be willing to add a test for this to our test suite? tests/test_substring.py would be the file.

sgillies:
I'm going to merge this now and we'll follow up with a test later.

Issue #695, opened 2019-04-01
WKT loads finish with Segmentation fault for positive integers for python2


tbicr:
## Please note

If you are reporting an installation or module import issue, please note that
this project only accepts reports about problems with packages downloaded from
the Python Package Index. Conda users should take issues to one of the
following trackers:

- https://github.com/ContinuumIO/anaconda-issues/issues
- https://github.com/conda-forge/shapely-feedstock

## Expected behavior and actual behavior.

`shaply.wkt.loads ` never should down with `Segmentation fault` and just raise `TypeError`, `UnicodeError`, `WKTReadingError` or other for incorrect input.

For python3 you will get `AttributeError: 'int' object has no attribute 'encode'`, but `TypeError` more appropriate there.

## Steps to reproduce the problem.

    import shapely.wkt
    shapely.wkt.loads(4)

## Operating system

Ubuntu 16.04.6 LTS

## Shapely version and provenance

1.6.4.post2 installed to venv from PyPI using pip for python 2.7.13

snorfalorpagus:
I can reproduce this. It looks like the behaviour is different on Python 2 vs Python 3.

```
Python 2.7.15rc1 (default, Nov 12 2018, 14:31:15) 
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from shapely.wkt import loads
>>> loads(4)
[1]    1483 segmentation fault (core dumped)  python
```

```
Python 3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from shapely.wkt import loads
>>> loads(4)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/snorf/Desktop/venv36/lib/python3.6/site-packages/shapely/wkt.py", line 10, in loads
    return geos.WKTReader(geos.lgeos).read(data)
  File "/home/snorf/Desktop/venv36/lib/python3.6/site-packages/shapely/geos.py", line 250, in read
    text = text.encode('ascii')
AttributeError: 'int' object has no attribute 'encode'
```

tbicr:
Fixed

PR #700, opened 2019-04-07
#695: fix segfault and unicode errors for non text parameter wkt loads

tbicr:
fix for https://github.com/Toblerity/Shapely/issues/695

### DIFF
--- a/shapely/geos.py
+++ b/shapely/geos.py
@@ -21,6 +21,11 @@ from .errors import WKBReadingError, WKTReadingError, TopologicalError, Predicat
 # Add message handler to this module's logger
 LOG = logging.getLogger(__name__)
 
+if sys.version_info[0] >= 3:
+    text_types = str
+else:
+    text_types = (str, unicode)
+
 # Find and load the GEOS and C libraries
 # If this ever gets any longer, we'll break it into separate modules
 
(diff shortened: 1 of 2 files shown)


shapely/geos.py
```diff
@@ -257,6 +257,8 @@ def __del__(self):
 
     def read(self, text):
         """Returns geometry from WKT"""
+        if not isinstance(text, (bytes, str) if sys.version_info[0] >= 3 else (str, unicode)):
+            raise TypeError("Only bytes or string types can be processed.")
```
sgillies: The problem here isn't Shapely's WKT reader, it's ctypes' c_char_p, which interprets integers as pointers to memory. A `if isinstance(text, int): raise TypeError(...)` will suffice – we only need to intercept the int case. See https://bugs.python.org/issue1701409 for more details.

tbicr: in this case there are another issue for python3 https://github.com/Toblerity/Shapely/blob/master/shapely/geos.py#L261 which will raise `AttributeError`

sgillies: I agree 👍

tbicr: so what the best approach for this bug fixing:
1. use byte/string instance parameter check?
2. use int parameter check + add AttributeError handling and TypeError reraising?

sgillies: I'm sorry that I wasn't clear: number 2, please.

tbicr: one more question, for python 3 `bytes` type doesn't allowed now, should be this case covered or only unicode string should be allowed?

tbicr: added bytes handling for python3

shapely/geos.py
```diff
@@ -257,7 +257,9 @@ def __del__(self):
 
     def read(self, text):
         """Returns geometry from WKT"""
-        if sys.version_info[0] >= 3:
+        if isinstance(text, int):
+            raise TypeError("Only bytes or string types can be processed.")
+        if sys.version_info[0] >= 3 and isinstance(text, str):
             text = text.encode('ascii')
         geom = self._lgeos.GEOSWKTReader_read(self._reader, c_char_p(text))
```
sgillies:
```suggestion
        c_string = c_char_p(text)
        geom = self._lgeos.GEOSWKTReader_read(self._reader, c_string)
```

shapely/geos.py
tbicr: updated

PR #715, opened 2019-05-17
Replace cascaded with unary union in docs and more

kannes:
- Brief info about what happens with different kinds of geometries (via
https://geos.osgeo.org/doxygen/classgeos_1_1operation_1_1geounion_1_1UnaryUnionOp.html)
- Extend doc example with validity of MultiPolygon
- Demote cascaded_union from being worthy of mention
- Fix GEOS version in note
- Update plot generation from cascaded to unary
- Update examples from cascaded to unary
- "Hypier" and more abstract section title

### DIFF
--- a/docs/code/unary_union.py
+++ b/docs/code/unary_union.py
@@ -1,6 +1,6 @@
 from matplotlib import pyplot
 from shapely.geometry import Point
-from shapely.ops import cascaded_union
+from shapely.ops import unary_union
 from descartes import PolygonPatch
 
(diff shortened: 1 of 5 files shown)


kannes:
Fixes https://github.com/Toblerity/Shapely/issues/667 and is much better than https://github.com/Toblerity/Shapely/commit/524f26fdc9934de99bc5279863df9b560c36a8e3 :clown_face: (no seriously!)

PR #728, opened 2019-06-08
PROPOSAL: add array_transformer that uses the array interface for transformations

snowman2:
I was tinkering around and had this idea. This could be an issue, but I thought it would be nice to see the code and make comments on it, suggest improvements, etc. In most cases this has ~2x speedup for transformations. Minor, but still something.

image

If this is something of interest, just let me know. It will need tests, etc... If not, that's fine too and definitely understandable.


sgillies:
Interesting! I'll have to look closer.

snowman2:
I think this will be superseded by #782

PR #731, opened 2019-06-15
Exterior of an empty polygon is equivalent to LinearRing()

drnextgis:
Fix for https://github.com/Toblerity/Shapely/issues/712 as was suggested by @sgillies.


sgillies:
Thanks @drnextgis !

PR #733, opened 2019-06-16
Add support of MultiPolygon and GeometryCollection to 'orient' function

drnextgis:
Fix for https://github.com/Toblerity/Shapely/issues/635.

### DIFF
--- a/shapely/geometry/polygon.py
+++ b/shapely/geometry/polygon.py
@@ -9,7 +9,6 @@ if sys.version_info[0] < 3:
 from ctypes import c_void_p, cast, POINTER
 import weakref
 
-from shapely.algorithms.cga import signed_area
 #from shapely.coords import required
 from shapely.geos import lgeos
 from shapely.geometry.base import BaseGeometry, geos_geom_from_py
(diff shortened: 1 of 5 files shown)


shapely/geometry/polygon.py
```diff
@@ -388,22 +387,6 @@ def asPolygon(shell, holes=None):
...
-    else:
-        rings.append(list(ring.coords)[::-1])
-    for ring in polygon.interiors:
-        if signed_area(ring)/s <= 0.0:
-            rings.append(ring)
-        else:
-            rings.append(list(ring.coords)[::-1])
-    return Polygon(rings[0], rings[1:])
-
-
```
sgillies: If we `from shapely.ops import orient` in polygon.py, we won't break any code that depends on importing it from here.

drnextgis: Actually adding that import causes circular dependency issue. I've implemented your suggestion in a little different way.

pglotov:
Shouldn't it've been added to the docs?

sgillies:
Docs added in 053df24.

vpipkt:
@drnextgis This saved me a ton of headache today! 🍰

PR #734, opened 2019-06-17
Fix invalid scape sequence warning

zetaatlyft:
Made changes in Docstrings to avoid "invalid escape sequence" DeprecationWarnings with Python 3.6

### DIFF
--- a/shapely/affinity.py
+++ b/shapely/affinity.py
@@ -6,7 +6,7 @@ __all__ = ['affine_transform', 'rotate', 'scale', 'skew', 'translate']
 
 
 def affine_transform(geom, matrix):
-    """Returns a transformed geometry using an affine transformation matrix.
+    r"""Returns a transformed geometry using an affine transformation matrix.
 
     The coefficient matrix is provided as a list or tuple with 6 or 12 items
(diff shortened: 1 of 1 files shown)


mwtoews:
This could be easier fixed by declaring the docstrings as literal strings, by prefixing with an `r`:
```python
def foo():
    r"""Foo

    \whatever
    """
    pass
````

zetaatlyft:
> This could be easier fixed by declaring the docstrings as literal strings, by prefixing with an `r`:
> 
> ```python
> def foo():
>     r"""Foo
> 
>     \whatever
>     """
>     pass
> ```

thank you for the comment @mwtoews. I added the suggestion.

sgillies:
Are the backticks in the docstrings the root problem? I'd be willing to remove them. Putting extra modifiers on the docstrings seems a little odd, but if I saw some prior art I would happily concede.

jorisvandenbossche:
> Are the backticks in the docstrings the root problem?

Yes, I suppose so. In pandas we eg have some docstrings with a latex equation, and then we also use the "raw" string to deal with this.

Issue #553, opened 2018-01-01
GEOS aborting with "Assertion failed" on OS X, discussion and work around


sgillies:
I've spent a bit of free time looking more deeply into the reports of GEOS aborting in the AbstractSTRtree.cpp module (either in `query()` or `itemsTree()`) on OS X when fiona is imported before shapely in a Python script.

Here are the conditions for a program that will unexpectedly abort:

1. Any version of Python on OS X.
2. Fiona and Shapely binary wheels for macosx downloaded from PyPI. These wheels include libgeos_c.dylib and libgeos-3.6.2.dylib files.
3. A script that imports fiona before shapely and which triggers the creation of a GEOS STR R-tree. Calling `ops.unary_union` with a number of lines or polygons will suffice.

Here's an example program:

```python
import fiona

from shapely.geometry import Point
from shapely.geos import geos_version
from shapely.ops import unary_union

print(geos_version)

SHAPES = [Point(i, i).buffer(1.5) for i in range(20)]
for i, s in enumerate(SHAPES):
    print(i, s.is_valid)
    print(i, s.wkt)
    print(i, s.area, s.length)
    print(i, s.intersects(Point(10, 10).buffer(8.0)))

union = unary_union(SHAPES)
print(union)
```

I've run the script under lldb with a modified version of GEOS 3.6.2 that prints some debugging output before calling `assert()`.

```
...
19 7.057234103728362 9.420993470864257
19 False
Assertion failed: (!static_cast<bool>("should never be reached")), function itemsTree, file AbstractSTRtree.cpp, line 379.
dynamic_cast<AbstractNode*>: 0x0dynamic_cast<ItemBoundable*>: 0x0childBoundable: 0x100498510Process 17968 stopped
* thread #1: tid = 0x606ba9, 0x00007fff8d415f06 libsystem_kernel.dylib`__pthread_kill + 10, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x00007fff8d415f06 libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill:
->  0x7fff8d415f06 <+10>: jae    0x7fff8d415f10            ; <+20>
    0x7fff8d415f08 <+12>: movq   %rax, %rdi
    0x7fff8d415f0b <+15>: jmp    0x7fff8d4107cd            ; cerror_nocancel
    0x7fff8d415f10 <+20>: retq
```

[...]

PR #735, opened 2019-06-20
Load GEOS only once in the wheel case

sgillies:
Towards resolving #553.

### DIFF
--- a/shapely/geos.py
+++ b/shapely/geos.py
@@ -26,6 +26,7 @@ if sys.version_info[0] >= 3:
 else:
     text_types = (str, unicode)
 
+
 # Find and load the GEOS and C libraries
 # If this ever gets any longer, we'll break it into separate modules
 
@@ -71,8 +72,7 @@ if sys.platform.startswith('linux'):
         _lgeos = CDLL(geos_whl_so[0])
         LOG.debug("Found GEOS DLL: %r, using it.", _lgeos)
     elif hasattr(sys, 'frozen'):
-        geos_pyinstaller_so = glob.glob(os.path.join(sys.prefix,
-                                                    'libgeos_c-*.so.*'))
+        geos_pyinstaller_so = glob.glob(os.path.join(sys.prefix, 'libgeos_c-*.so.*'))
         if len(geos_pyinstaller_so) == 1:
             _lgeos = CDLL(geos_pyinstaller_so[0])
             LOG.debug("Found GEOS DLL: %r, using it.", _lgeos)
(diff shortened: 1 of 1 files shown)


shapely/geos.py
```diff
@@ -71,8 +72,7 @@ def load_dll(libname, fallbacks=None, mode=DEFAULT_MODE):
         _lgeos = CDLL(geos_whl_so[0])
         LOG.debug("Found GEOS DLL: %r, using it.", _lgeos)
     elif hasattr(sys, 'frozen'):
-        geos_pyinstaller_so = glob.glob(os.path.join(sys.prefix,
-                                                    'libgeos_c-*.so.*'))
+        geos_pyinstaller_so = glob.glob(os.path.join(sys.prefix, 'libgeos_c-*.so.*'))
```
sgillies: Fixed a nit.

shapely/geos.py
```diff
@@ -93,9 +93,16 @@ def load_dll(libname, fallbacks=None, mode=DEFAULT_MODE):
...
     geos_whl_dylib = os.path.abspath(os.path.join(os.path.dirname(
         __file__), '.dylibs/libgeos_c.1.dylib'))
+
     if os.path.exists(geos_whl_dylib):
-        _lgeos = CDLL(geos_whl_dylib)
-        LOG.debug("Found GEOS DLL: %r, using it.", _lgeos)
+        handle = CDLL(None)
+        if hasattr(handle, "initGEOS_r"):
+            LOG.debug("GEOS already loaded")
+            _lgeos = handle
```
sgillies: I did not realize this about CDLL, but called with `None` it gets all the currently loaded symbols in the process. We can discover whether libgeos has already been loaded, such as via libgdal in the case where fiona or rasterio is imported before shapely. If it has, we don't try to load the wheel's included dynamic library.

shapely/geos.py
```diff
@@ -265,7 +272,7 @@ def read(self, text):
         if not isinstance(text, text_types):
             raise TypeError("Only str is accepted.")
         if sys.version_info[0] >= 3:
-             text = text.encode()
+            text = text.encode()
```
sgillies: Fixed another nit.

sgillies:
This PR adds some logic to dodge double-loading of different geos shared libraries on OS  X. I don't believe we need to do this in other cases. The problem of #553 doesn't affect our Linux wheels. The problem doesn't exist in a conda environment where we only have one geos shared lib. It's only when we mix fiona or rasterio wheels with our shapely wheel that we can be loading two different library files.

sgillies:
I built wheels last night https://travis-ci.com/sgillies/shapely-wheels/builds/116234380 and am having success. Unless there are objections I'd like to merge this soon and cut a 1.7a2 release for shapely.

PR #762, opened 2019-08-14
In linux env add fallback load dll for libc.so.6

clefrks:
Allow fallback for loading libc.so.6 
i'm having issue in mock for linux environment to build this package but with allowing fallback for libc.so.6 i was able to pass the build

  File "setup.py", line 80, in <module>
    from shapely._buildcfg import geos_version_string, geos_version, \
  File "/builddir/build/BUILD/Shapely-1.6.4.post2/shapely/_buildcfg.py", line 168, in <module>
    free = load_dll('c').free
  File "/builddir/build/BUILD/Shapely-1.6.4.post2/shapely/_buildcfg.py", line 161, in load_dll
    libname, fallbacks or []))
OSError: Could not find library c or load any of its variants []

### DIFF
--- a/shapely/_buildcfg.py
+++ b/shapely/_buildcfg.py
@@ -165,7 +165,7 @@ if sys.platform.startswith('linux'):
     if not lgeos:
         lgeos = load_dll('geos_c',
                          fallbacks=['libgeos_c.so.1', 'libgeos_c.so'])
-    free = load_dll('c').free
+    free = load_dll('c', fallbacks=['libc.so.6']).free
     free.argtypes = [c_void_p]
     free.restype = None
 


sgillies:
I'm reluctant to do this because I've never seen it necessary with any version of glibc. Can you tell me more about your linux environment?

clefrks:
Its failing to build in mock environment which is building the source in chroot. I'm including the spec files that i'm using to build it. 

im using Mock version: 1.2.14

Below is the spec file used to build
[Shapely.log](https://github.com/Toblerity/Shapely/files/3532552/Shapely.log)

andreasbaumann:
I can reproduce the same problem with cura (which uses python-shapely) on Archlinux32.
/usr/lib/libc.so is a linker script:

```
/* GNU ld script
   Use the shared library, but some functions are only in
   the static library, so try that secondarily.  */
OUTPUT_FORMAT(elf32-i386)
GROUP ( /usr/lib/libc.so.6 /usr/lib/libc_nonshared.a  AS_NEEDED ( /usr/lib/ld-linux.
so.2 ) )
```
I'm not sure, if load_dll is reading this information.

Setting it to:

```
    free = load_dll('c', fallbacks=['/usr/lib/libc.so.6']).free
```

makes the module build.

andreasbaumann:
The same issue exists when running the tests in geos.py.
Also for instance the sunos5 version already has an explicit fallback to libc.so.6 there, so
it is not a new Linux thing..

sgillies:
Sorry for the delay, all!

Issue #394, opened 2016-07-24
OSX (10.10.5) fails to load GEOX framework


mromanie:
I'm running OS X 10.10.5 and python, GEOS and shapely from MacPorts:

>>> from shapely.geometry import Point
Failed `CDLL(/Library/Frameworks/GEOS.framework/Versions/Current/GEOS)`
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/shapely/geometry/__init__.py", line 4, in <module>
    from .base import CAP_STYLE, JOIN_STYLE
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/shapely/geometry/base.py", line 9, in <module>
    from shapely.coords import CoordinateSequence
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/shapely/coords.py", line 8, in <module>
    from shapely.geos import lgeos
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/shapely/geos.py", line 98, in <module>
    free = load_dll('c').free
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/shapely/geos.py", line 61, in load_dll
    libname, fallbacks or []))
OSError: Could not find lib c or load any of its variants [].

I've also tried GEOS from homebrew and kyngchaos' framework (http://www.kyngchaos.com/software/frameworks), but the result is always the same.

Could you, please, advise? Thanks!!

sgillies:
> OSError: Could not find lib c or load any of its variants [].

I'm so sorry, but I don't know how to help if ctypes can not find libc on your system.

mromanie:
Many thanks for the answer! It turns out that I needed to unset the environment variable DYLD_FALLBACK_LIBRARY_PATH, after which everything works just fine (the package is brilliant, by the way!). Maybe my experience can be of use to others, as well.

douglowe:
Your experience is definitely useful to others. I had the same problem (on OSX too), and the recommended solution of unsetting DYLD_FALLBACK_LIBRARY_PATH has worked for me too - thanks!

PR #765, opened 2019-08-20
Add fallback path for libc.musl

jkschin:
With reference to https://github.com/Toblerity/Shapely/issues/394, this was the same problem that happened to me. The root cause is because `find_library("c")` returns something empty for certain linux environments. Searching through some of these forums, it looks like this won't be fixed. As such, we have to accommodate this on the package side. 

My environment only had the following `libc` libraries:
```
./lib/libc.musl-x86_64.so.1
./usr/glibc-compat/lib/libc.so.6
./usr/glibc-compat/lib/libc.so
```

The fix below adds the path for `./lib/libc.musl-x86_64.so.1`. I've tested this and it solved the issue.

### DIFF
--- a/shapely/geos.py
+++ b/shapely/geos.py
@@ -87,8 +87,7 @@ if sys.platform.startswith('linux'):
         _lgeos = load_dll('geos_c', fallbacks=alt_paths)
     # Necessary for environments with only libc.musl
     c_alt_paths = [
-        'libc.musl-x86_64.so.1',
-        os.path.join(sys.prefix, "lib", "libc.musl-x86_64.so.1")
+        'libc.musl-x86_64.so.1'
     ]
     free = load_dll('c', fallbacks=c_alt_paths).free
     free.argtypes = [c_void_p]
--- a/shapely/geos.py
+++ b/shapely/geos.py
@@ -85,7 +85,12 @@ if sys.platform.startswith('linux'):
             'libgeos_c.so',
         ]
         _lgeos = load_dll('geos_c', fallbacks=alt_paths)
-    free = load_dll('c').free
+    # Necessary for environments with only libc.musl
+    c_alt_paths = [
+        'libc.musl-x86_64.so.1',
+        os.path.join(sys.prefix, "lib", "libc.musl-x86_64.so.1")
+    ]
+    free = load_dll('c', fallbacks=c_alt_paths).free
     free.argtypes = [c_void_p]
     free.restype = None
 


jkschin:
For more context, the problem is with `ctypes.util.find_library` as mentioned [here](https://gitlab.alpinelinux.org/alpine/aports/issues/4512).

jkschin:
Just looked through the PRs and I realized it's very similar to https://github.com/Toblerity/Shapely/pull/762/files. Tagging @sgillies.

snorfalorpagus:
For a bit more context, musl-libc is an alternative to glibc and is used by the Alpine Linux distribution commonly used in Docker images.

```
$ docker run -it alpine:latest
/ # apk update
fetch http://dl-cdn.alpinelinux.org/alpine/v3.10/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.10/community/x86_64/APKINDEX.tar.gz
v3.10.2-2-g10fc00b3db [http://dl-cdn.alpinelinux.org/alpine/v3.10/main]
v3.10.2-1-g93e3be01c5 [http://dl-cdn.alpinelinux.org/alpine/v3.10/community]
OK: 10335 distinct packages available
/ # apk add python3
(1/11) Installing libbz2 (1.0.6-r7)
(2/11) Installing expat (2.2.7-r0)
(3/11) Installing libffi (3.2.1-r6)
(4/11) Installing gdbm (1.13-r1)
(5/11) Installing xz-libs (5.2.4-r0)
(6/11) Installing ncurses-terminfo-base (6.1_p20190518-r0)
(7/11) Installing ncurses-terminfo (6.1_p20190518-r0)
(8/11) Installing ncurses-libs (6.1_p20190518-r0)
(9/11) Installing readline (8.0.0-r0)
(10/11) Installing sqlite-libs (3.28.0-r0)
(11/11) Installing python3 (3.7.3-r0)
Executing busybox-1.30.1-r2.trigger
OK: 69 MiB in 25 packages
/ # python3
Python 3.7.3 (default, May  3 2019, 11:24:39) 
[GCC 8.3.0] on linux
>>> from ctypes.util import find_library
>>> find_library("c")
>>> find_library("libc.musl-x86_64.so.1")
'libc.musl-x86_64.so.1'
```

The binary wheels for Shapely available on PyPI can't be used on Alpine because of their dependency on glibc and because setup.py can't find libc it can't be installed from source either:

```
/ # pip3 install shapely
Collecting shapely
  Downloading https://files.pythonhosted.org/packages/a2/fb/7a7af9ef7a35d16fa23b127abee272cfc483ca89029b73e92e93cdf36e6b/Shapely-1.6.4.post2.tar.gz (225kB)
    100% |████████████████████████████████| 235kB 6.0MB/s 
    Complete output from command python setup.py egg_info:
    Failed `CDLL(libgeos_c.so.1)`
    Failed `CDLL(libgeos_c.so)`
      File "/tmp/pip-install-yqcoc3sm/shapely/setup.py", line 80, in <module>
        from shapely._buildcfg import geos_version_string, geos_version, \
      File "/tmp/pip-install-yqcoc3sm/shapely/shapely/_buildcfg.py", line 167, in <module>
        fallbacks=['libgeos_c.so.1', 'libgeos_c.so'])
      File "/tmp/pip-install-yqcoc3sm/shapely/shapely/_buildcfg.py", line 161, in load_dll
        libname, fallbacks or []))
    OSError: Could not find library geos_c or load any of its variants ['libgeos_c.so.1', 'libgeos_c.so']
```

Adding this additional path seems like a reasonable workaround to me.

jkschin:
@snorfalorpagus, if this looks good to you, shall we merge this?

jkschin:
I don't have a strong opinion on this but since `geos_c` was hard coded, I would think that adding extra environment variables to control this would increase the complexity of the package. We might end up with many SO questions and work arounds like #394, which don't seem to address the root cause.

If a future user knew how to set environment variables for this, it would be easy for that person to also find the place to PR and fix it for the majority.

snorfalorpagus:
I'm content with this PR as-is. If there is appetite for an environment variable to control this it can be done in another issue/PR.

jkschin:
@snorfalorpagus when will a new version be released? Is there a work around for the docker containers in the mean time?

PR #784, opened 2019-10-28
Cleanup References

### LINKED ISSUE #617: docs: Remove unused references
manual.rst has some references at the bottom.

I could not find where e.g. "_Understanding spatial relations" or "_record" were used at all. Someone should take a look at that list at its bottom and clean up what's not needed anymore. While at it, best also homogenize how URLs are linked throughout the whole document. Sometimes it is direct URL, sometimes those named thingies.

I might do this _some day_™.

chadwhawkins:
* Removes unused references at the end of `manual.rst`
* Updates some URLs to use https
* Fixes a footnote reference (previously unreferenced)
* Switched one link to use a reference instead but the remaining links seemed fine as is

I noticed that `design.rst` is not used. Should that file be removed? 

Also, the ToC has an entry for `API Documentation <modules>` that goes unused, it looks like because `make apidocs` is still outputting txt files. Should that get updated to rst or should that entry just get removed?

Resolves #617

### DIFF
--- a/CREDITS.txt
+++ b/CREDITS.txt
@@ -68,6 +68,6 @@ Additional help from:
 * Sandro Santilli, Mateusz Loskot, Paul Ramsey, et al (GEOS Project)
 
 Major portions of this work were supported by a grant (for Pleiades_) from the
-U.S. National Endowment for the Humanities (http://www.neh.gov).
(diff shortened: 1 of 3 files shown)


kannes:
Nice! Thanks so much @chadwhawkins :fireworks:

PR #786, opened 2019-10-31
Guard for empty geometries in affinity functions

### LINKED ISSUE #763: Rotating empty geometries gives errors
Right now rotating empty geometries results in errors:
```python
from shapely.affinity import rotate
from shapely.geometry import (LineString, LinearRing, 
                              MultiLineString, MultiPoint, 
                              MultiPolygon, Polygon, Point)

objs = [LineString, LinearRing, MultiLineString, 
        MultiPoint, MultiPolygon, Polygon, Point]
for obj in objs:
    try:
        rotate(obj(), 45)
        print('successful rotation')
    except (ValueError, IndexError) as err:
        print(f'Failed to rotate empty {obj}:\n'
              f'{type(err).__name__}: {err}\n')
```
output:
```
Failed to rotate empty <class 'shapely.geometry.linestring.LineString'>:
ValueError: not enough values to unpack (expected 4, got 0)

Failed to rotate empty <class 'shapely.geometry.polygon.LinearRing'>:
ValueError: not enough values to unpack (expected 4, got 0)

Failed to rotate empty <class 'shapely.geometry.multilinestring.MultiLineString'>:
ValueError: not enough values to unpack (expected 4, got 0)

Failed to rotate empty <class 'shapely.geometry.multipoint.MultiPoint'>:
ValueError: not enough values to unpack (expected 4, got 0)

Failed to rotate empty <class 'shapely.geometry.multipolygon.MultiPolygon'>:
ValueError: not enough values to unpack (expected 4, got 0)

Failed to rotate empty <class 'shapely.geometry.polygon.Polygon'>:
ValueError: not enough values to unpack (expected 4, got 0)

Failed to rotate empty <class 'shapely.geometry.point.Point'>:
IndexError: list index out of range
```
Shouldn't the result of a rotation of an empty geometry be also an empty geometry?

---

For the context, I was writing a test where I generated a `MultiPolygon`, rotated it and returned resulting parts in a list:
```python
# code to generate a random MultiPolygon
return list(rotate(multipolygon, angle).geoms)
```
But, as sometimes the generated geometry was empty, this code resulted in a `ValueError`. So, to fix it currently I write:
```python
return [] if multipolygon.is_empty else list(rotate(multipolygon, angle).geoms)
```
which I find too cumbersome.

Do you think we could change the current behavior so that rotated empty geometry would be also an empty geometry?

---

chadwhawkins:
`affine_transform` checks if empty but the functions in affinity will raise an Exception before reaching that check as noted in #763. This resolves #763

### DIFF
--- a/shapely/affinity.py
+++ b/shapely/affinity.py
@@ -254,6 +254,8 @@ def translate(geom, xoff=0.0, yoff=0.0, zoff=0.0):
         | 0  0  1 zoff |
         \ 0  0  0   1  /
     """
+    if geom.is_empty:
+        return geom
     matrix = (1.0, 0.0, 0.0,
               0.0, 1.0, 0.0,
               0.0, 0.0, 1.0,
--- a/shapely/affinity.py
+++ b/shapely/affinity.py
@@ -151,6 +151,8 @@ def rotate(geom, angle, origin='center', use_radians=False):
         xoff = x0 - x0 * cos(r) + y0 * sin(r)
         yoff = y0 - x0 * sin(r) - y0 * cos(r)
     """
+    if geom.is_empty:
+        return geom
     if not use_radians:  # convert from degrees
         angle *= pi/180.0
     cosp = cos(angle)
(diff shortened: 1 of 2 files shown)


chadwhawkins:
@mwtoews thanks for the review. I had the test in there already for `translate` but added the explicit `is_empty` check as requested. Should be good to go now.

PR #794, opened 2019-11-22
Avoid in-place multiplication of affinity parameters

### LINKED ISSUE #793: Affinity rotate changes angle in place
## version/platform
All from Anaconda main channel
* Shapely 1.6.4
* Geo's 3.8.0
* Python 3.7 x64
* Windows 10 x64

When applying affinity rotate to line string with an angle in degrees, afterwards the variable containing angle has been permanently changed to radians.

If angle is already in radians, and `use_radians=True` then works as expected

```python
from shapely import affinity
L = 1.6
P = 5
tracker_theta = 129.7  # degrees
tracker1 = LineString([(-L/2, 0), (L/2, 0)])
tracker1rot = affinity.rotate(
    tracker1, tracker_theta)
# tracker_theta has been permanently converted to radians
tracker2 = LineString([(P-L/2, 0), (P+L/2, 0)])
tracker2rot = affinity.rotate(
    tracker2, tracker_theta)

# tracker2rot is incorrect because use_radians is false by default, but tracker theta had been converted in place unexpectedly
```

## Operating system

(For example, Mac OS X 10.12.3)

(For example, 1.6b4 installed from PyPI using pip)

mwtoews:
Fixes #793, and similar potential issue with skew.

### DIFF
--- a/shapely/affinity.py
+++ b/shapely/affinity.py
@@ -154,7 +154,7 @@ def rotate(geom, angle, origin='center', use_radians=False):
     if geom.is_empty:
         return geom
     if not use_radians:  # convert from degrees
-        angle *= pi/180.0
+        angle = angle * pi/180.0
     cosp = cos(angle)
     sinp = sin(angle)
     if abs(cosp) < 2.5e-16:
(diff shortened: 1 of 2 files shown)


sgillies:
Great, thanks @mwtoews !

PR #806, opened 2019-12-10
Implement GEOSBufferParams to support single sided buffers.

akadouri:
Attempting to add support for single sided buffers raised in this issue: https://github.com/Toblerity/Shapely/issues/727

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -531,7 +531,7 @@ class BaseGeometry(object):
 
     def buffer(self, distance, resolution=16, quadsegs=None,
                cap_style=CAP_STYLE.round, join_style=JOIN_STYLE.round,
-               mitre_limit=5.0, single_side=0):
+               mitre_limit=5.0, single_sided=False):
         """Returns a geometry with an envelope at a distance from the object's
         envelope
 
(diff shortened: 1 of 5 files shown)


shapely/geometry/base.py
```diff
@@ -531,7 +531,7 @@ def _transformed_rects():
 
     def buffer(self, distance, resolution=16, quadsegs=None,
                cap_style=CAP_STYLE.round, join_style=JOIN_STYLE.round,
-               mitre_limit=5.0):
+               mitre_limit=5.0, single_side=0):
```
sgillies: My only suggestion is to add a mention of this new keyword argument in the docstring. The docstring predates my discovery of numpydoc and lacks structure, but somewhere at the end after the mentions of other arguments would be fine.

akadouri: Thanks for the quick review.

Forgot about the docstring. I didn't know about numpydoc, it was already somewhat close so I reformatted the docstring to match that style.

shapely/geometry/base.py
```diff
@@ -531,29 +531,38 @@ def _transformed_rects():
...
```
jorisvandenbossche: `single_sided` instead of `single_side` as keyword? (seems more natural to me, and matches the GEOS API)

I think it might also be nice to make this False/True in the user API, instead of the 0/1 .

akadouri: Good catch on the typo, updated it to `single_sided` to match the GEOS API.

I didn't realize the value was actually a boolean - I thought it was as integer from looking at [geos_c.cpp](https://github.com/libgeos/geos/blob/master/capi/geos_c.cpp#L1441-L1445) with > 0 indicating a right buffer and < 0 referring to left. 🤦‍♂ 

Updated the code and comments match the [GEOS docs](https://geos.osgeo.org/doxygen/classgeos_1_1operation_1_1buffer_1_1BufferParameters.html#ae2a7d92145547fb6e2852c41c5353700). Also added a test for left & right buffers.

PR #812, opened 2019-12-28
bool(geom) shall be the logical complement of geom.is_empty

### LINKED ISSUE #754: Inconsistent bool value of empty geometries
The boolean value of an "empty" geometry (`bool(geom)`) depends on how the geometry is created. 

Empty `Polygon` and `EmptyGeometry` return True, while an empty `GeometryCollection` returns False.

```
In [1]: import shapely

In [2]: shapely.__version__
Out[2]: '1.6.4.post2'

In [3]: from shapely.geometry import Polygon, GeometryCollection

In [4]: from shapely.geometry.base import BaseGeometry, EmptyGeometry

In [5]: p1 = Polygon([])

In [6]: p1
Out[6]: <shapely.geometry.polygon.Polygon at 0x7f1319e719e8>

In [7]: print(p1)
GEOMETRYCOLLECTION EMPTY

In [8]: bool(p1)
Out[8]: True

In [9]: p2 = p1.buffer(0)   # not fully sure how to create such a POLYGON EMPTY manually

In [11]: p2
Out[11]: <shapely.geometry.polygon.Polygon at 0x7f131960c7f0>

In [12]: print(p2)
POLYGON EMPTY

In [13]: bool(p2)
Out[13]: True

In [14]: p3 = GeometryCollection()

In [15]: p3
Out[15]: <shapely.geometry.collection.GeometryCollection at 0x7f13038d74e0>

In [16]: print(p3)
GEOMETRYCOLLECTION EMPTY

In [17]: bool(p3)
Out[17]: False

In [18]: p4 = EmptyGeometry()

In [19]: p4
Out[19]: <shapely.geometry.base.EmptyGeometry at 0x7f1314fa91d0>

In [20]: print(p4)
GEOMETRYCOLLECTION EMPTY

In [21]: bool(p4)
Out[21]: True
```

(shapely installed with conda from conda-forge on Ubuntu 18.04, Python 3.7)

I think this should be consistent, ideally. 
Should we change the behaviour of GeometryCollection to match Polygon, LineString and EmptyGeometry ?

sgillies:
Resolves #754

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,8 @@ Changes
 Next
 ----
 
+- Implement ``__bool__()`` for geometry base classes so that ``bool(geom)``
+  returns the logical complement of ``geom.is_empty`` (#754).
 - Remove assertion on the number of version-like strings found in the GEOS
   version string. It could be 2 or 3.
 
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -230,6 +230,9 @@ class BaseGeometry(object):
         self._is_empty = True
         self.__geom__ = val
 
+    def __bool__(self):
+        return self.is_empty is False
+
     def __del__(self):
         self.empty(val=None)
         self.__p__ = None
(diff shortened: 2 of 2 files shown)


sgillies:
I'm going to merge this one for 1.7a3. If it causes heartburn we can revert before the first beta release.

PR #843, opened 2020-02-18
Problems with VScode users and multiples environments

arubiales:
## Virtual Environment VScode problem

I have multiples different environments and diferents IDEs. When I was in VScode without conda, I had the following error:

```

from MTSPTW.algorithm.clustering import Clustering...
OSError                                   Traceback (most recent call last)
c:\Users\alberto.rubiales\Desktop\projects\MTSPTW\prueba_test.py in 
----> 1 from MTSPTW.algorithm.clustering import Clustering

c:\Users\alberto.rubiales\Desktop\projects\MTSPTW\MTSPTW\algorithm\clustering.py in 
      2 from pandas import concat, DataFrame
      3 from sklearn.cluster import KMeans
----> 4 from geopandas import datasets, read_file
      5 from shapely.geometry import Point
      6 import matplotlib.cm as color_maps

c:\Users\alberto.rubiales\Desktop\projects\MTSPTW\venv-MTSPTW\lib\site-packages\geopandas\__init__.py in 
----> 1 from geopandas.geoseries import GeoSeries  # noqa
      2 from geopandas.geodataframe import GeoDataFrame  # noqa
      3 from geopandas.array import _points_from_xy as points_from_xy  # noqa
      4 
      5 from geopandas.io.file import read_file  # noqa

c:\Users\alberto.rubiales\Desktop\projects\MTSPTW\venv-MTSPTW\lib\site-packages\geopandas\geoseries.py in 
      8 
      9 from pyproj import CRS, Transformer
---> 10 from shapely.geometry.base import BaseGeometry
     11 from shapely.ops import transform
     12 

c:\Users\alberto.rubiales\Desktop\projects\MTSPTW\venv-MTSPTW\lib\site-packages\shapely\geometry\__init__.py in 
      2 """
      3 
----> 4 from .base import CAP_STYLE, JOIN_STYLE
      5 from .geo import box, shape, asShape, mapping
      6 from .point import Point, asPoint

c:\Users\alberto.rubiales\Desktop\projects\MTSPTW\venv-MTSPTW\lib\site-packages\shapely\geometry\base.py in 
     16 
     17 from shapely.affinity import affine_transform
---> 18 from shapely.coords import CoordinateSequence
     19 from shapely.errors import WKBReadingError, WKTReadingError
     20 from shapely.geos import WKBWriter, WKTWriter

c:\Users\alberto.rubiales\Desktop\projects\MTSPTW\venv-MTSPTW\lib\site-packages\shapely\coords.py in 
      6 from ctypes import byref, c_double, c_uint
      7 
----> 8 from shapely.geos import lgeos
      9 from shapely.topology import Validating
     10 

c:\Users\alberto.rubiales\Desktop\projects\MTSPTW\venv-MTSPTW\lib\site-packages\shapely\geos.py in 
    143     if os.getenv('CONDA_PREFIX', ''):
    144         # conda package.
--> 145         _lgeos = CDLL(os.path.join(sys.prefix, 'Library', 'bin', 'geos_c.dll'))
    146     else:
    147         try:

~\AppData\Local\Programs\Python\Python37\lib\ctypes\__init__.py in __init__(self, name, mode, handle, use_errno, use_last_error)
    362 
    363         if handle is None:
--> 364             self._handle = _dlopen(self._name, mode)
    365         else:
    366             self._handle = handle

OSError: [WinError 126] The specified module could not be found
```

In VScode without Conda when you execute code in jupyter interactive console with a virtual environment and you have other environments with conda  ```os.getenv('CONDA_PREFIX', '')``` is True, and the code give and error, because your actual environment is not in conda, so you need to change the doble condition.

**I have been researching and I think that this behaviour occurs because the Interactive Jupyter Python console of VScode works internally with Conda.**  But don't have conda-meta folder.

I have tested my solution in Virtual Environments with and without conda, and both work.

my current setup is:

* Windows: 10 64bits win32
* VScode: 1.42
* Python: 3.7.6
* Conda versions: 3.5.2 and 4.8.1

Yes, I have tested in both Conda versions.

any help or additional information you need, please let me know!

### DIFF
--- a/shapely/geos.py
+++ b/shapely/geos.py
@@ -62,8 +62,7 @@ def load_dll(libname, fallbacks=None, mode=DEFAULT_MODE):
                 libname, fallbacks or []))
 
 _lgeos = None
-def 
-:
+def exists_conda_env():
     """Does this module exist in a conda environment?"""
     return os.path.exists(os.path.join(sys.prefix, 'conda-meta'))
 
--- a/shapely/geos.py
+++ b/shapely/geos.py
@@ -140,7 +140,7 @@ elif sys.platform == 'darwin':
     free.restype = None
 
 elif sys.platform == 'win32':
-    if os.getenv('CONDA_PREFIX', ''):
+    if os.getenv('CONDA_PREFIX', '') and os.path.exists(os.path.join(sys.prefix, 'conda-meta')):
         # conda package.
         _lgeos = CDLL(os.path.join(sys.prefix, 'Library', 'bin', 'geos_c.dll'))
     else:
--- a/shapely/geos.py
+++ b/shapely/geos.py
@@ -62,7 +62,9 @@ def load_dll(libname, fallbacks=None, mode=DEFAULT_MODE):
                 libname, fallbacks or []))
 
 _lgeos = None
-is_conda = os.path.exists(os.path.join(sys.prefix, 'conda-meta'))
+def exists_conda_env():
+    """Does this module exist in a conda environment?"""
+    return os.path.exists(os.path.join(sys.prefix, 'conda-meta'))
 
 
 if sys.platform.startswith('linux'):
--- a/shapely/geos.py
+++ b/shapely/geos.py
@@ -62,6 +62,8 @@ def load_dll(libname, fallbacks=None, mode=DEFAULT_MODE):
                 libname, fallbacks or []))
 
 _lgeos = None
+is_conda = os.path.exists(os.path.join(sys.prefix, 'conda-meta'))
+
 
 if sys.platform.startswith('linux'):
     # Test to see if we have a wheel repaired by 'auditwheel' containing its
@@ -76,7 +78,7 @@ if sys.platform.startswith('linux'):
         if len(geos_pyinstaller_so) == 1:
             _lgeos = CDLL(geos_pyinstaller_so[0])
             LOG.debug("Found GEOS DLL: %r, using it.", _lgeos)
-    elif os.getenv('CONDA_PREFIX', ''):
+    elif is_conda:
         # conda package.
         _lgeos = CDLL(os.path.join(sys.prefix, 'lib', 'libgeos_c.so'))
     else:
@@ -107,7 +109,7 @@ elif sys.platform == 'darwin':
             _lgeos = CDLL(geos_whl_dylib)
             LOG.debug("Found GEOS DLL: %r, using it.", _lgeos)
 
-    elif os.getenv('CONDA_PREFIX', ''):
+    elif is_conda:
         # conda package.
         _lgeos = CDLL(os.path.join(sys.prefix, 'lib', 'libgeos_c.dylib'))
     else:
@@ -140,7 +142,7 @@ elif sys.platform == 'darwin':
     free.restype = None
 
 elif sys.platform == 'win32':
-    if os.getenv('CONDA_PREFIX', '') and os.path.exists(os.path.join(sys.prefix, 'conda-meta')):
+    if is_conda:
         # conda package.
         _lgeos = CDLL(os.path.join(sys.prefix, 'Library', 'bin', 'geos_c.dll'))
     else:
--- a/shapely/geos.py
+++ b/shapely/geos.py
@@ -62,7 +62,8 @@ def load_dll(libname, fallbacks=None, mode=DEFAULT_MODE):
                 libname, fallbacks or []))
 
 _lgeos = None
-def exists_conda_env():
+def 
+:
     """Does this module exist in a conda environment?"""
     return os.path.exists(os.path.join(sys.prefix, 'conda-meta'))
 
@@ -80,7 +81,7 @@ if sys.platform.startswith('linux'):
         if len(geos_pyinstaller_so) == 1:
             _lgeos = CDLL(geos_pyinstaller_so[0])
             LOG.debug("Found GEOS DLL: %r, using it.", _lgeos)
-    elif is_conda:
+    elif exists_conda_env():
         # conda package.
         _lgeos = CDLL(os.path.join(sys.prefix, 'lib', 'libgeos_c.so'))
     else:
@@ -111,7 +112,7 @@ elif sys.platform == 'darwin':
             _lgeos = CDLL(geos_whl_dylib)
             LOG.debug("Found GEOS DLL: %r, using it.", _lgeos)
 
-    elif is_conda:
+    elif exists_conda_env():
         # conda package.
         _lgeos = CDLL(os.path.join(sys.prefix, 'lib', 'libgeos_c.dylib'))
     else:
@@ -144,7 +145,7 @@ elif sys.platform == 'darwin':
     free.restype = None
 
 elif sys.platform == 'win32':
-    if is_conda:
+    if exists_conda_env():
         # conda package.
         _lgeos = CDLL(os.path.join(sys.prefix, 'Library', 'bin', 'geos_c.dll'))
     else:


sgillies:
I'm hesitant to merge this without review by someone who is more familiar with Windows, conda, and VScode. @mwtoews or @jorisvandenbossche can either of you take a look at this?

jorisvandenbossche:
I am familiar with conda, but not with windows .. (although I don't know if it is different on this aspect).

I can confirm though that all my conda environments have a `conda-meta` directory.

> In VScode without Conda when you execute code in jupyter interactive console with a virtual environment and you have other environments with conda os.getenv('CONDA_PREFIX', '') is True, 

@arubiales any idea whether this is expected, or maybe rather a bug on VSCode's side?

arubiales:
Yes! sorry for not developing the explanation further @jorisvandenbossche .

As I said is an error produced when you use VScode **and execute in VScode's interactive Jupyter console.**

**If you execute the code in Windows terminal, everything works perfectly.**

The problem is because:
1. The Interactive Jupyter Console, uses a reduce conda version environent to execute the code like a Jupyter Notebook

2. After uses the packages of my Virtual environment to show me the output.

But as I said, I'm not in a Conda environment. I'm just in a normal Python vitual Environent, so, the condition ```os.getenv('CONDA_PREFIX', '')``` is ```True``` and gives you the Conda enviroment of the Interactive Jupyter Console, but there`s no shapely module in this environment, because is not your real environment.

The VScode's Jupyter Interactive Console environment don't have the ```conda-meta``` directory and I saw that Conda versions: 3.5.2 and 4.8.1 have it. That's why I implemented this kind of solution.

I Also attach you [here](https://code.visualstudio.com/docs/python/jupyter-support-py) information about VScode Jupyter Interactive Console.

mwtoews:
Hi, yes I do have these tools on a Windows system, and am able to re-create the issue. But I also see this is a general issue that I can also recreate on Linux (e.g.):
```bash
# works fine, since CONDA_PREFIX is essentially undefined (it is a zero-length string)
$ CONDA_PREFIX= /usr/bin/python3 -c "from shapely import geos"
# here is trouble
$ CONDA_PREFIX=foo /usr/bin/python3 -c "from shapely import geos"
  File "/home/user/src/Shapely/shapely/geos.py", line 81, in <module>
    _lgeos = CDLL(os.path.join(sys.prefix, 'lib', 'libgeos_c.so'))
  File "/usr/lib/python3.6/ctypes/__init__.py", line 348, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: /usr/lib/libgeos_c.so: cannot open shared object file: No such file or directory
```
The issue is when `CONDA_PREFIX` is set to a non-zero string, but a non-conda Python interpreter is used.

The suggested PR ([based on this answer?](https://stackoverflow.com/a/47610844)) does fix the issue for both conda and non-conda environments, but I wonder if a better fix would be to create a single variable somewhere near the top of geos.py:

```python
is_conda = os.path.exists(os.path.join(sys.prefix, 'conda-meta'))
```
and replace the three instances of `os.getenv('CONDA_PREFIX', '')` with `is_conda`. And yes, I do see a `conda-meta` in all of my anaconda/minconda environments, so this looks like a better test than checking the environment variable.

jorisvandenbossche:
I think your `is_conda` proposal sounds good.

Just wondering now: what if a user installs shapely with pip in a conda environment? (on the other platforms, it seems we first check for a wheel before checking for a conda environment, but not so on windows)

sgillies:
@jorisvandenbossche pip installing an extension module into a conda environment can't work and I think we shouldn't try to accommodate this. Ideally, pip would always execute with --no-binary from a conda environment, but in a less ideal world I think we need to address this with more documentation.

mwtoews:
I see mixed success with pip install within a conda environment.

With Linux, conda and pip packed versions of shapely load different libgeos_c libraries, which is expected:
```
conda install --force-reinstall --yes shapely
python -c "from shapely import speedups; assert speedups.enabled"
python -c "from shapely import geos, __version__; print('{}: {}'.format(__version__, geos._lgeos._name))"
# 1.6.4.post2: /home/user/miniconda3/envs/py37/lib/libgeos_c.so.1
pip install -I shapely
python -c "from shapely import speedups; assert speedups.enabled"
python -c "from shapely import geos, __version__; print('{}: {}'.format(__version__, geos._lgeos._name))"
# 1.7.0: /home/user/miniconda3/envs/py37/lib/python3.7/site-packages/shapely/.libs/libgeos_c-5031f9ac.so.1.13.1
pip uninstall --yes shapely
conda install --force-reinstall --yes shapely
```
With Windows, both conda and pip package versions of shapely reference the same conda packaged geos_c.dll, which is unexpected:
```
conda install --force-reinstall --yes shapely
python -c "from shapely import speedups; assert speedups.enabled"
python -c "from shapely import geos, __version__; print('{}: {}'.format(__version__, geos._lgeos._name))"
# 1.6.4.post2: C:\Users\user\AppData\Local\Continuum\anaconda3\Library\bin\geos_c.dll
pip install -I shapely
python -c "from shapely import speedups; assert speedups.enabled"
python -c "from shapely import geos, __version__; print('{}: {}'.format(__version__, geos._lgeos._name))"
# 1.7.0: C:\Users\user\AppData\Local\Continuum\anaconda3\Library\bin\geos_c.dll
pip uninstall --yes shapely
conda install --force-reinstall --yes shapely
```
(I'm slightly surprised the speedups are enabled, but who knows what's going on.)

So there's yet more work to be done with geos.py, which I'm willing to do with a subsequent PR.

---

@arubiales are you able do some simple modifications to this PR? As described above, just check `is_conda` once, then replace the three instances of `os.getenv('CONDA_PREFIX', '')` with this better check.

arubiales:
Sorry @mwtoews I've been very busy this week

Yes, I think that ```is_conda``` is a better aproach! I'm going to do the changes to this PR

thank you all for the work!

anntzer:
While you're working on this part of the code, note that there is another problematic case: even when in a conda env, users can install the package from PyPI instead (this is explicitly supported per https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/pip-interoperability.html).  So even if CONDA_PREFIX is set, one may need to use the non-conda path to locate the dll.

On Windows, it appears that the following version works for me:
```python
    _conda_dll_path = os.path.join(sys.prefix, 'Library', 'bin', 'geos_c.dll')
    if os.getenv('CONDA_PREFIX', '') and os.path.exists(_conda_dll_path):  # or is_conda, after this PR
        # conda package.
        _lgeos = CDLL(_conda_dll_path)
    else:
        # continue on non-conda codepath
```
(I can also make a separate PR after this one is merged, if you prefer.)

sgillies:
@anntzer see my comment above. You can't pip install a shapely binary wheel into a conda environment and expect it to work. We don't have to support this scenario.

anntzer:
Currently, pip installing numpy, scipy, matplotlib, or most other packages of the scientific python stack works just fine (both on Linux and on Windows: this has been my normal approach for years); conda also explicitly supports that.

This doesn't mean that shapely absolutely *needs* to follow the same path, but it would be a pity not to, given that this is in fact quite easy to do so and would *not* "destabilize your environment".

jorisvandenbossche:
> pip installing an extension module into a conda environment can't work and I think we shouldn't try to accommodate this.

@sgillies in general, this works fine, and also for shapely it does work on Linux. 

First, I think many users do this not on purpuse, having installed Anaconda to get python, and then finding online guidelines to install shapely as `pip install shapely`. But I have even spoken with people that work like this on purpose (use conda just to create and manage envs, and then use pip to install everything)

I personally also don't see a reason why not to support this for shapely (although I personally would never recommend it, but there will always be users that simply do it, and getting confusing errors in that case is not ideal IMO). 
On linux, shapely explicitly supports this by first checking if it is a wheel before checking it is in a conda-env. Is there a reason that approach would not work on Windows? (or is there no easy way to determine it's a wheel on windows?)

anntzer:
> But I have even spoken with people that work like this on purpose (use conda just to create and manage envs, and then use pip to install everything)

Indeed, this is my case.

> Is there a reason that approach would not work on Windows? (or is there no easy way to determine it's a wheel on windows?)

I believe the snippet I posted should be sufficient (it works for me locally, at least...), although you can probably play even more tricks using e.g. importlib-metadata to list the files installed with the package and lookup the dll in them; see e.g. the implementation starting at https://github.com/dhermes/bezier/pull/208/files#diff-ef65eef0dcf3d4deda75ab42c6866ebfR52.

arubiales:
>But I have even spoken with people that work like this on purpose (use conda just to create and manage envs, and then use pip to install everything)

Of Course, I have conda to manage the environments and pip to install it. Even sometimes you have librarys that gives you errors with pip and not with conda and vice versa. 

Anyways, anything I can do or change please say to me!

shapely/geos.py
```diff
@@ -62,6 +62,8 @@ def load_dll(libname, fallbacks=None, mode=DEFAULT_MODE):
                 libname, fallbacks or []))
 
 _lgeos = None
+is_conda = os.path.exists(os.path.join(sys.prefix, 'conda-meta'))
```
sgillies: What would you think about defining a function instead @arubiales ? Use of a function will make this easier to test and we'll be able to avoid checking the filesystem in some cases.

```suggestion
def exists_conda_env():
    """Does this module exist in a conda environment?"""
    return os.path.exists(os.path.join(sys.prefix, 'conda-meta'))
```

arubiales: Ohhh, I have acepted but we need to implement the function in the three checks! I'm going to change it!

Thanks.

shapely/geos.py
```diff
@@ -140,7 +144,7 @@ def load_dll(libname, fallbacks=None, mode=DEFAULT_MODE):
     free.restype = None
 
 elif sys.platform == 'win32':
-    if os.getenv('CONDA_PREFIX', ''):
+    if exists_conda_env():
```
jorisvandenbossche: Can you add an additional `and ..` check here that the file actually exists? 
That should solve the problem of "wheel in a conda env" (at least for the case where the conda geos package is not installed ..)

sgillies: I don't quite understand @jorisvandenbossche . Which file, the wheel's GEOS DLL?

I still feel like we're setting users up for failure by enabling them to `conda install gdal; pip install shapely` and then expect these to work together in the same python program. But not using the DLL in the wheel would be the worst thing that could happen, for sure.

jorisvandenbossche: This branch sets the path to `os.path.join(sys.prefix, 'Library', 'bin', 'geos_c.dll')`, without checking that this actually exists (which is indeed not necessarily the case if installed with pip). So I would personally at least check that the file actually exists.

hobu: > I still feel like we're setting users up for failure by enabling them to `conda install gdal; pip install shapely` and then expect these to work together in the same python program. 

Agree 100%. They should use conda into install all three – GEOS, GDAL, and Shapely. The spread of `is CONDA_PREFIX` allover the codebase is a bit ugly and symptomatic of code smell on the Conda side, not a failure of any of the libraries.

anntzer: If you really, really don't want to support pip-installation into conda envs (even though my personal experience is that it just works with the patch above), please at least consider checking if the dll/so file exists and emitting a clearer error if it doesn't ("you appear to be in a conda environment but some files are missing from shapely, please note that shapely does not support pip installs in conda environments" or something similar).

jorisvandenbossche: We already enable this for linux and mac, so rephrasing my earlier question: is there a reason that on Windows this would be less robust? 

People are going to do this anyway. First, because there are legitimate reasons to do so (eg using conda only as environment manager, installing nightly wheels if they would be available (packages like numpy, scikit-learn, pandas provide those, and those are regularly installed in conda envs), ..). And second because people are simply not aware, and the internet is full of instructions to do `pip install shapely`. 
So personally, I think we should either try to ensure it works properly, or either try to raise an informative error message. But for both options, we will need to detect this case.

sgillies: Right, let's continue to isolate this case.

If shapely didn't dlopen libgeos (via ctypes) we wouldn't have this problem. This is what makes the flaky shapely/gdal/fiona + libgeos combination different from, for example, the stable numpy/scipy + libgfortran combination. I don't blame the conda design at all.

We're planning a rewrite of shapely to eliminate ctypes and which will delete all the code we're discussing now. Let's try to make the best of a bad situation and solve this particular issue. I'm going to stop opposing and complaining about pip installing shapely wheels into a conda env. In return, I'm going to rely on the community for helping in supporting users who do this.

Issue #840, opened 2020-02-14
shapely.geometry.shape() can't parse coordinates in numpy arrays


nrweir:
## Expected behavior and actual behavior.

We've been using the `shapely.geometry.shape()` function to create geometries of multiple different types from a json string in [`solaris.utils.geo._reduce_geom_precision()`](https://github.com/CosmiQ/solaris/blob/master/solaris/utils/geo.py#L600). We pass coordinate pairs in as 2D `numpy` arrays. When the geometry corresponds to a polygon (see example below), we're suddenly getting errors that seem to boil down to problems with truthiness testing of multi-element `numpy` arrays [here](https://github.com/Toblerity/Shapely/blob/master/shapely/geometry/geo.py#L35). I _think_ this is the result of a recent deprecation of this type of testing in `numpy` arrays, though I'm not 100% sure.

My question: should I be avoiding using arrays for coordinates, or should I make a PR to help parse arrays more effectively?

## Steps to reproduce the problem.

```python

from shapely.geometry import shape
from numpy import array

geom_dict = {
    'type': 'Polygon',
    'coordinates': array(
        [[[ 0.,  3.],
        [ 8.,  8.],
        [ 6., 21.],
        [ 5., 29.],
        [19., 38.],
        [18., 58.],
        [ 0., 54.],
        [ 0.,  3.]]]
    )
}

shape(geom_dict)
```

Error message
```
Traceback (most recent call last):
  File "<stdin>", line 8, in <module>
  File "/opt/miniconda3/envs/solaris_021/lib/python3.7/site-packages/shapely/geometry/geo.py", line 35, in shape
    if not ob["coordinates"]:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
## Operating system

MacOS 10.15.3

## Shapely version and provenance

1.6.4 or 1.7.0 (occurs in both) installed from conda-forge. python 3.7.6.

sgillies:
I'd be happy to see numpy arrays supported. It might not be much more than testing `if len(ob["coordinates"]) == 0` instead of `if not ob["coordinates"]`. A PR would be very welcome!

nrweir:
Awesome! I'll get that in.

PR #850, opened 2020-02-21
Fixing empty-ness test for coords - accommodate numpy

nrweir:
Per #840, this updates `shapely.geometry.geo.shape()`'s `Polygon` creation functionality accommodate NumPy arrays in `ob['coordinates']`. It required an extra `if`/`elif` to check for either `None` or empty lists/arrays because [truth-testing empty numpy arrays is now deprecated and throws an error](https://github.com/numpy/numpy/issues/9583).

I tested the fix with the `solaris` functionality that uses numpy arrays as coordinates and found it works.

### DIFF
--- a/shapely/geometry/geo.py
+++ b/shapely/geometry/geo.py
@@ -32,7 +32,7 @@ def shape(context):
     elif geom_type == "linestring":
         return LineString(ob["coordinates"])
     elif geom_type == "polygon":
-        if not ob["coordinates"]:
+        if len(ob["coordinates"]) == 0:
             return Polygon()
         else:
             return Polygon(ob["coordinates"][0], ob["coordinates"][1:])
--- a/shapely/geometry/geo.py
+++ b/shapely/geometry/geo.py
@@ -32,7 +32,9 @@ def shape(context):
     elif geom_type == "linestring":
         return LineString(ob["coordinates"])
     elif geom_type == "polygon":
-        if len(ob["coordinates"]) == 0:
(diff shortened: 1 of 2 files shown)


nrweir:
Argh. I don't know why my edits are blowing up file encodings. I'll see if I can fix this...at this point, though, I think code-wise things are in good shape.

koshy1123:
Ooh - I just saw this after making my PR which is addressing the same issue I think: https://github.com/Toblerity/Shapely/pull/852

nrweir:
Definitely very similar, @koshy1123! I just checked your changes and I think a couple of them would still cause problems with numpy arrays - e.g. the `_is_coordinates_empty` function [here](https://github.com/Toblerity/Shapely/pull/852/files#diff-aba84314caf408dcf74ac88b10b38f2eR14) would say a `numpy` array wasn't empty. But your solution also covers a lot of stuff that I didn't here - all I addressed was polygons (to resolve #840). We could combine the two PRs?

koshy1123:
Ah you're right, you're specifically trying it for numpy arrays - happy to incorporate your change in mine if you're okay with it.

nrweir:
Sure sounds good. Once you've added it in, I can go through and make sure it works for numpy arrays!

koshy1123:
All done @nrweir - PTAL: https://github.com/Toblerity/Shapely/pull/852
(credits in the commit for your test)

nrweir:
Closing in favor of #852, which incorporates this functionality.

PR #851, opened 2020-02-21
Fix polylabel for polygons with holes

### LINKED ISSUE #817: Polylabel bug

Holes are not correctly accounted for in shapely's implementation of the polylabel algorithm. 

```
from shapely.algorithms.polylabel import polylabel
poly1=Polygon(shell=[(0,0),(10,0),(10,10),(0,10),(0,0)][::-1], holes=[[(2,2),(6,2),(6,6),(2,6),(2,2)]])
import geopandas
gdf2 = geopandas.GeoDataFrame(geometry=[poly1])
gdf2.plot()
label_location = polylabel(poly1, tolerance=0.05) 
print(label_location.x)
print(label_location.y)
```

The output of this script is: 

```
6.25
3.75
```

But the correct position is something close to 7.65625, 7.65625. This can be verified by looking at a plot of this polygon 

image

Maybe I am doing something wrong in the definition of the polygon but I have tried many different ways of instantiating this polygon and I always get the same result.

## Operating system

Linux

Tested with shapely version 1.6 and 1.7a3

snorfalorpagus:
The existing implementation of `polylabel` does not properly account for interior holes in the polygon, as the original algorithm does.

 Closes #817.

### DIFF
--- a/shapely/algorithms/polylabel.py
+++ b/shapely/algorithms/polylabel.py
@@ -47,7 +47,9 @@ class Cell(object):
         boundary.
         """
         inside = polygon.contains(self.centroid)
-        distance = self.centroid.distance(LineString(polygon.exterior.coords))
+        distance = self.centroid.distance(polygon.exterior)
+        for interior in polygon.interiors:
+            distance = min(distance, self.centroid.distance(interior))
         if inside:
             return distance
         return -distance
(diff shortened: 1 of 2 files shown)


sgillies:
Great @snorfalorpagus !

PR #852, opened 2020-02-22
Handle for empty list/list of lists in 'coordinates'

koshy1123:
Fixes https://github.com/Toblerity/Shapely/issues/837
Fixes https://github.com/Toblerity/Shapely/issues/840

Two questions:
- I wasn't sure about more tests, but I'm not opposed to adding a `test_shape.py`
- Does this logic also need to support `asShape`? It wasn't super clear to me how that should work.

### DIFF
--- a/shapely/geometry/geo.py
+++ b/shapely/geometry/geo.py
@@ -19,7 +19,7 @@ else:
 
 
 def _is_coordinates_empty(coordinates):
-    """Helper to identify if list or all nested lists of coordinates are empty"""
+    """Helper to identify if coordinates or subset of coordinates are empty"""
 
     if coordinates is None:
         return True
--- a/shapely/geometry/geo.py
+++ b/shapely/geometry/geo.py
@@ -14,7 +14,7 @@ from .collection import GeometryCollection
 def _is_coordinates_empty(coordinates):
     """Helper to identify if list or all nested lists of coordinates are empty"""
     if isinstance(coordinates, (list, tuple)):
-        if not coordinates:
+        if len(coordinates) == 0:
             return True
         for sub_coordinates in coordinates:
             return _is_coordinates_empty(sub_coordinates)
--- a/shapely/geometry/geo.py
+++ b/shapely/geometry/geo.py
@@ -1,7 +1,7 @@
 """
 Geometry factories based on the geo interface
 """
-
+import numpy as np
 from .point import Point, asPoint
 from .linestring import LineString, asLineString
 from .polygon import Polygon, asPolygon
@@ -13,10 +13,12 @@ from .collection import GeometryCollection
 
 def _is_coordinates_empty(coordinates):
     """Helper to identify if list or all nested lists of coordinates are empty"""
-    if isinstance(coordinates, (list, tuple)):
+    if isinstance(coordinates, (list, tuple, np.ndarray)):
         if len(coordinates) == 0:
             return True
         return all(map(_is_coordinates_empty, coordinates))
+    elif coordinates is None:
+        return True
     else:
         return False
 
--- a/shapely/geometry/geo.py
+++ b/shapely/geometry/geo.py
@@ -11,12 +11,42 @@ from .multipolygon import MultiPolygon, MultiPolygonAdapter
 from .collection import GeometryCollection
 
 
+def _is_coordinates_empty(coordinates):
+    """Helper to identify if list or all nested lists of coordinates are empty"""
+    if isinstance(coordinates, (list, tuple)):
+        if not coordinates:
+            return True
+        for sub_coordinates in coordinates:
+            return _is_coordinates_empty(sub_coordinates)
+    else:
+        return False
+
+
+def _empty_shape_for_no_coordinates(geom_type):
+    """Return empty counterpart for geom_type"""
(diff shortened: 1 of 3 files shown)


nrweir:
Looks great on my end @koshy1123, thanks!

shapely/geometry/geo.py
```diff
@@ -1,7 +1,7 @@
 """
 Geometry factories based on the geo interface
 """
-
+import numpy as np
```
sgillies: @koshy1123 numpy is still an optional dependency of shapely, so we need to work on adding support for arrays in a way that doesn't import numpy here.

shapely/geometry/geo.py
```diff
@@ -11,13 +11,44 @@
 from .collection import GeometryCollection
 
 
+def _is_coordinates_empty(coordinates):
+    """Helper to identify if list or all nested lists of coordinates are empty"""
+    if isinstance(coordinates, (list, tuple, np.ndarray)):
+        if len(coordinates) == 0:
```
sgillies: To avoid making numpy a dependency here, what would you think about something like 

```suggestion
        if coordinates is None:
            return True
        else:
            try:
                coords_len = len(coords)
                ...
            except TypeError:
                return False
```

sgillies: Changes needed to following lines of course.

shapely/geometry/geo.py
```diff
@@ -11,13 +11,44 @@
...
+            return True
+        return all(map(_is_coordinates_empty, coordinates))
```
sgillies: Now that I think about it, the use of all() here makes this method relatively expensive. any(), used in the opposite sense, would return faster. What would you think about turning this into an is_coordinates_not_empty() function?

koshy1123: Yeah, I had a similar thought. I thought it had to be done this way to check if _all_ the contents are empty. My first implementation just stopped on the first non-empty content. See this commit for example: https://github.com/Toblerity/Shapely/pull/852/commits/36abf1ab2132c844d4c02553d8d716538c7ed192

The issue I struggled with was what the shape should be if only one component was empty? For example, an empty point in a polygon?  Should it error out downstream as it does currently? 
```
>>> data = {"type": "Polygon", "coordinates": [[(), (1, 1), (1, 0)]]}
>>> geom = shape(data)
  File "/usr/local/lib/python3.7/site-packages/shapely/geometry/geo.py", line 38, in shape
    return Polygon(ob["coordinates"][0], ob["coordinates"][1:])
  File "/usr/local/lib/python3.7/site-packages/shapely/geometry/polygon.py", line 242, in __init__
    ret = geos_polygon_from_py(shell, holes)
  File "/usr/local/lib/python3.7/site-packages/shapely/geometry/polygon.py", line 506, in geos_polygon_from_py
    ret = geos_linearring_from_py(shell)
  File "/usr/local/lib/python3.7/site-packages/shapely/geometry/polygon.py", line 442, in geos_linearring_from_py
    assert (n == 2 or n == 3)
AssertionError
```

koshy1123: But, +1 to inverting the logic. I had the check of "stop when you hit one empty", but I can do the check of "stop when you hit one non-empty"

shapely/geometry/geo.py
koshy1123: Ah, got it - sure! I'll take your suggestion from the review

shapely/geometry/geo.py
sgillies: @koshy1123 I had to take a few minutes to remind myself why fiona emits `[[]]` for the coordinates of a null geometry. It's by accident, not by design, that we see `[[]]` and not just `[]`. There's no special significance to the inner empty list and I think that in the future fiona may very well emit `[]` for all null/empty geometries or even just emit `None` for the entire geometry.

A mix of empty and filled coordinate pairs (or trios) within a coordinate sequence is not something we should try to smooth over. That's invalid data. I'm 100% sure that fiona and geopandas will never return such data.

This means that you can count on seeing `[]` in the case of empty/null points, linestrings, linearrings, or `[[]]` in the case of empty/null polygons, multipoints, multilinestrings, multipolygons, or geometry collections.

In light of this, I think the use of all() can remain. Does that makes sense?

koshy1123: Got it 👍  Thanks for the deep dive! `all()` should be a fast compute in both then:
- if the coordinates field is non empty, the first non empty set will make it `False`
- if the coordinates field is empty, i.e, one of `[]` or `[[]]`  it's a fast compute to `True`

shapely/geometry/geo.py
koshy1123: I went with an import check instead. Does that work? PTAL: https://github.com/Toblerity/Shapely/pull/852/commits/bc3b0a72646e2cfd763db3061240aaa16b219c6d

koshy1123: Also, makes me think the Travis CI builds for `NUMPY=0` aren't working correctly? They passed for my previous commit.

sgillies: I'm okay with the import check. It's possible that there's been a regression in testing with numpy, I haven't looked closely at that in a while.

Issue #854, opened 2020-02-23
Docs absent for STRTree.nearest()


Jeremiah-England:
Looking through the change log for 1.7.0 I noticed that STRTree now supports nearest neighbor searches (issue #668). 

I also noticed that @kannes has been working on the STRTree docs for issue #618. There was a merge attempt early January. However, there were conflicts and nothing further happened: https://github.com/Toblerity/Shapely/pull/623. ¯\\\_(ツ)_/¯  

This seems like a good first issue and I'm interested in getting involved with shapely. So maybe next weekend I could fix the conflicts with https://github.com/Toblerity/Shapely/pull/623 and add the docs for nearest neighbors, if that's OK with @kannes. 

It looks like this would also close #845?

link to function code: [STRTree.nearest()](https://github.com/Toblerity/Shapely/blob/master/shapely/strtree.py#L64-L84)

kannes:
I am not "OK" with that, but rather I would happily jump with happiness if you did that :))

sgillies:
@Jeremiah-England count on me for help too!

sgillies:
Resolved by #857.

Issue #845, opened 2020-02-20
Bug: shapely.strtree.STRtree returns false positives


NavinF:
## Expected behavior and actual behavior.

Expected: "The query method on STRtree returns a list of all geometries in the tree that intersect the provided geometry argument." as documented: https://shapely.readthedocs.io/en/latest/manual.html#str-packed-r-tree

Reality: `STRtree.query(x)` sometimes returns objects that don't intersect `x`. Worse yet, it always returns a non-empty list when certain rectangle-shaped regions of the search space are queried. 

## Steps to reproduce the problem.

1. Create STRtree  from buffered random lines
2. Query STRtree  with another buffered line and assert result[0].intersects(queried_line)
3. Wait for assertion to fail

I'll code a minimum repro later if there is interest. For now, here's my workaround that first queries the R-tree and then double-checks the R-tree query results by brute forcing `intersects` to test if a line segment intersects any obstacles. This solved my problem for now, but kills the log(n) time complexity that I'll need later 

```
def drc_check(self, a :Tuple, b :Tuple, obstacles :STRtree) -> bool:
  """
  Returns True if the (a,b) trace is far enough from obstacles to
  meet design rules
  """
  new_trace_string = LineString([a,b]).buffer(self.trace_width/2 + self.trace_spacing/2)
  intersecting_obstacles = obstacles.query(new_trace_string)
  for o in intersecting_obstacles:
    if o.intersects(new_trace_string):
      return False
    else:
      pass #False positive in STRtree :(
  return True
```

## Operating system

Linux

## Shapely version and provenance

1.7.0 installed from PyPI using pip on Python 3.6

sgillies:
This is a documentation bug. The query method returns objects which have *extents* that intersect with the extent of the given geometry. So, yes, there will be false positives of this kind. We will fix the documentation.

NavinF:
Ah, thanks for confirming. Lemme tack on a performance bug/note to this.

I profiled one run of my program and found that checking for false positives with `intersects` takes 318 seconds whereas all 354836 calls to `strtree.py:50(query)` take 9.2 seconds.  

Oddly enough, some instances of STRtree only return a non-empty list when the query is close to the obstacles. In those cases, double-checking R-tree queries is quick and acceptable. 

Other instances of STRtree consistently return a non-empty list if i query anywhere inside large rectangles covering 1/4 of my search space even though the obstacles (buffered LineString) cover a minuscule area of the search space.  

[...]

Issue #618, opened 2018-07-30
docs: Clarification of STRTree documentation


kannes:
I would like to (slightly) improve the documentation for the STRTree.

http://shapely.readthedocs.io/en/stable/manual.html#str-packed-r-tree says 
> Shapely provides an interface to the query-only GEOS R-tree packed using the Sort-Tile-Recursive algorithm. Pass a list of geometry objects to the STRtree constructor to create an R-tree that you can query with another geometric object.

The example then shows two queries that return Geometry objects.

I was hoping I could use STRTree just like the rtree module: passing index numbers to be able to match geometries to (Fiona) features but that seems not to be the case. STRTree simply takes a list of Geometry objects and on a query, returns the intersecting Geometry objects it knows about. There is no order or indices. Is that correct?

If so, I would just add one sentences about what the query method returns. For class descriptions (e.g. Point) the attributes are presented in prose, but if this should get a new `.. method::` entry, just say so. :)

sgillies:
@kannes we're not flexible right now about what gets stored, but we do store the geometry object as well as all its attributes. You can add attributes to a geometry object after it has been created because these geometry objects store their attributes in a `__dict__` member (`geom.myattr` gets read as `geom.__dict__['myattr']`). We're not going to change this behavior of Shapely any time soon, so the following code should work at least until a hypothetical Shapely 2.0:

```
>>> from shapely.strtree import STRtree
>>> from shapely.geometry import Point
>>> pt = Point(0.0, 0.0)
>>> pt.name = 'foo'
>>> tree = STRtree([pt])
>>> tree.query(Point(1.0, 1.0).buffer(2.0))
[<shapely.geometry.point.Point object at 0x109254208>]
>>> results = tree.query(Point(1.0, 1.0).buffer(2.0))
>>> respt = results[0]
>>> respt
<shapely.geometry.point.Point object at 0x109254208>
>>> respt.name
'foo'
```

There is some related discussion in #615.

kannes:
Thanks, I will update this

kannes:
Done!

But I screwed up because Git is confusing and now there is earlier stuff in my branch... https://github.com/Toblerity/Shapely/compare/master...kannes:strtree-docs?expand=1
I will try to remove the two other commits or "rebase to master" or whatever to fix this...

A graphical representation of the example geometries and query geometry would be great to further show how it returns geometries whose extents intersect.

Not sure if I should suggest using deepcopy to avoid mutating the original objects.

[...]

PR #857, opened 2020-03-01
Strtree docs

Jeremiah-England:
I've update the docs for the STRtree in manual.rst. This should close these issues (#854, #845, #618) and one other pull request (https://github.com/Toblerity/Shapely/pull/623). Note that I have not done anything with the docstrings. I think I would like to work on that in a separate merge later. Maybe waiting to close those issues until that is done as well is a good idea. 

Also note that I have changed the work around for getting indexes to use an auxiliary dictionary keyed on geometry memory IDs instead of adding an index attribute to the geometries themselves (or a deep copy if you don't want to mutate.) 

Compare: 

``` python 
>>> from copy import deepcopy
>>> 
>>> points = [Point(i, i) for i in range(10)]
>>> tree = STRtree(points)
>>>
>>> points_numbered = []
>>> for i, point in enumerate(points):
>>>     point = deepcopy(point)  # let's not mutate the originals
>>>     point.i = i
>>>     points_numbered.append(point)
>>> tree = STRtree(points_numbered)
>>> result = tree.query(Point(2, 2).buffer(1.0))
>>> [(o.i, o.wkt) for o in result]
[(1, 'POINT (1 1)'), (2, 'POINT (2 2)'), (3, 'POINT (3 3)')]
```  
vs 

``` python 
>>> points = [Point(i, i) for i in range(10)]
>>> tree = STRtree(points)
>>>
>>> index_by_id = dict((id(pt), i) for i, pt in enumerate(points))
>>> [(index_by_id[id(pt)], pt.wkt) for pt in tree.query(Point(2,2).buffer(1.0))]
[(1, 'POINT (1 1)'), (2, 'POINT (2 2)'), (3, 'POINT (3 3)')]
```

This appears to be robust even in the case where the list grows, which you might expect to change memory location (id) of the geometries.  Observe:

``` python
>>> points = [Point(-1, -1)]
>>> original_id = id(points[0])
>>> for i in range(1000000):
>>>     points.append([Point(i, i)])
>>>     if id(points[0]) != original_id:
>>>         print(f"Element id changed from {original_id} to {id(points[0])}")
>>>         break
>>> original_id == id(points[0])
True
```

The RST rendering has been tested. Results are available here: https://jeremiah-england.github.io/shapely/build/manual.html#str-packed-r-tree  

Any review and feedback will be appreciated.

### DIFF
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -2410,62 +2410,61 @@ STR-packed R-tree
 Shapely provides an interface to the query-only GEOS R-tree packed using the
 Sort-Tile-Recursive algorithm. Pass a list of geometry objects to the STRtree
 constructor to create a spatial index that you can query with another geometric
-object. The returned geometries are the originals, not copies.
+object. Query-only means that once created, the `STRtree` is immutable. You
+cannot add or remove geometries.
 
 .. class:: strtree.STRtree(geometries)
 
   The `STRtree` constructor takes a sequence of geometric objects.
 
-  These are copied and stored in the R-tree.
+  These are copied and stored in the R-tree. However, references are kept to
+  the original geometries and queries return these, not copies.
 
   `New in version 1.4.0`.
 
-  Query-only means that once created, the `STRtree` is immutable. You cannot
-  add or remove geometries.
+  .. method:: strtree.query(goem)
 
-.. method:: strtree.query(goem)
-
-  Returns a list of all geometries in the `strtree` whose extents intersect the
-  extents of `goem`. This means that a subsequent search through the returned
-  subset using the desired binary predicate (eg. intersects, crosses, contains,
-  overlaps) may be necessary to further filter the results according to their
-  specific spatial relationships.
-
-  .. code-block:: pycon
-
-    >>> from shapely.strtree import STRtree
-    >>> points = [Point(i, i) for i in range(10)]
-    >>> tree = STRtree(points)
-    >>> query_geom = Point(2,2).buffer(0.99)
-    >>> [o.wkt for o in tree.query(query_geom)]
-    ['POINT (2 2)']
-    >>> query_geom = Point(2, 2).buffer(1.0)
-    >>> [o.wkt for o in tree.query(query_geom)]
-    ['POINT (1 1)', 'POINT (2 2)', 'POINT (3 3)']
-    >>> [o.wkt for o in tree.query(query_geom) if o.intersects(query_geom)]
-    ['POINT (2 2)']
-
-  .. note::
-    To get the original indexes of the query results, create an auxiliary
-    dictionary. But use the geometry `ids` as keys since the shapely geometries
-    themselves are not hashable.
+    Returns a list of all geometries in the `strtree` whose extents intersect the
+    extents of `goem`. This means that a subsequent search through the returned
+    subset using the desired binary predicate (eg. intersects, crosses, contains,
+    overlaps) may be necessary to further filter the results according to their
+    specific spatial relationships.
 
     .. code-block:: pycon
 
-      >>> index_by_id = dict((id(pt), i) for i, pt in enumerate(points))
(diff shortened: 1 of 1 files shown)


docs/manual.rst
```diff
@@ -2300,9 +2300,9 @@ The :func:`~shapely.ops.substring` function in `shapely.ops` returns a line segm
     Negative distance values are taken as measured in the reverse
     direction from the end of the geometry. Out-of-range index
     values are handled by clamping them to the valid range of values.
-    
+
     If the start distance equals the end distance, a point is being returned.
-
```
sgillies: 👍

docs/manual.rst
```diff
@@ -2409,37 +2409,62 @@ STR-packed R-tree
 
 Shapely provides an interface to the query-only GEOS R-tree packed using the
 Sort-Tile-Recursive algorithm. Pass a list of geometry objects to the STRtree
-constructor to create an R-tree that you can query with another geometric object.
+constructor to create a spatial index that you can query with another geometric
+object. Query-only means that once created, the `STRtree` is immutable. You
+cannot add or remove geometries.
```
sgillies: 👍

docs/manual.rst
```diff
@@ -2409,37 +2409,62 @@ STR-packed R-tree
...
 
 .. class:: strtree.STRtree(geometries)
 
   The `STRtree` constructor takes a sequence of geometric objects.
 
-  These are copied and stored in the R-tree.
+  These are copied and stored in the R-tree. However, references are kept to
+  the original geometries and queries return these, not copies.
```
sgillies: Let's delete the original text, which is in fact erroneous, and say something like "references to these geometric objects are stored in the R-tree".

docs/manual.rst
```diff
@@ -2409,37 +2409,62 @@ STR-packed R-tree
...
-intersect the provided geometry argument. If you want to match geometries of a
-more specific spatial relationship (eg. crosses, contains, overlaps), consider
-performing the query on the R-tree, followed by a manual search through the
-returned subset using the desired binary predicate.
+  .. method:: strtree.query(goem)
 
-Query-only means that once created, the R-tree is immutable. You cannot 
-add or remove geometries.
+    Returns a list of all geometries in the `strtree` whose extents intersect the
+    extents of `goem`. This means that a subsequent search through the returned
```
sgillies:
```suggestion
    extents of `geom`. This means that a subsequent search through the returned
```

docs/manual.rst
```diff
@@ -2409,37 +2409,62 @@ STR-packed R-tree
...
+    subset using the desired binary predicate (eg. intersects, crosses, contains,
+    overlaps) may be necessary to further filter the results according to their
+    specific spatial relationships.
```
sgillies: 👍

docs/manual.rst
```diff
@@ -2409,37 +2409,62 @@ STR-packed R-tree
...
+      .. code-block:: pycon
+
+        >>> index_by_id = dict((id(pt), i) for i, pt in enumerate(points))
+        >>> [(index_by_id[id(pt)], pt.wkt) for pt in tree.query(Point(2,2).buffer(1.0))]
+        [(1, 'POINT (1 1)'), (2, 'POINT (2 2)'), (3, 'POINT (3 3)')]
+
+
+  .. method:: strtree.nearest(geom)
+
+    Returns the nearest goemetry in `strtree` to `geom`.
```
sgillies:
```suggestion
    Returns the nearest geometry in `strtree` to `geom`.
```

docs/manual.rst
Jeremiah-England: af89f54000e95ad99095e64b92082fc751fd4aab

Jeremiah-England:
@sgillies I've fixed the problems you mentioned. Let me know if you find any others. Thanks.

kannes:
Great work!

Issue #785, opened 2019-10-28
parallel_offset 'right' return wrong value


lichunr:
## Please note

If you are reporting an installation or module import issue, please note that
this project only accepts reports about problems with packages downloaded from
the Python Package Index. Conda users should take issues to one of the
following trackers:

- https://github.com/ContinuumIO/anaconda-issues/issues
- https://github.com/conda-forge/shapely-feedstock

## Expected behavior and actual behavior.
 create a LineString, and parallel_offset to right by 1
l = LineString([(0, 0), (0, 2)])
l = l.parallel_offset(1, 'right')
print(l) # display LINESTRING (1 2, 1 0)
l = l.parallel_offset(1, 'right')
print(l) # display LINESTRING (0 0, 0 2)

expected value: LINESTRING(2 0, 2 2)

(parallel_offset 'left' has no such issue)

## Steps to reproduce the problem.

(like shown above)

## Operating system

Mac OS X 10.15 Catalina . (also on windows 10)

## Shapely version and provenance

1.6b4 installed from PyPI using pip
geos 3.8 (also the same on geos 3.7.3)

LostFan123:
This happens because the right offset changes the orientation of the `LineString`. I don't know if this behavior is intended or not, but it would be good to document it because it confused me as well when I first encountered it.

lichunr:
@LostFan123 agree the change of orientation, though 'left' has no such issue...

Jeremiah-England:
It always reverses right side offsets but not left side offsets. There's [a mention of it in the change log](https://shapely.readthedocs.io/en/latest/project.html#id8) because of issue #284. 

The change log states that the right side reversal was 'noted' but I can't find it in the docs. It should probably be put in.

jorisvandenbossche:
The change in orientation, although documented on in our docs, was an implementation detail of GEOS. And starting with GEOS >= 3.11, the implementation will now always try to preserve the orientation, instead of changing it. 

This has been clarified in the docs in https://github.com/shapely/shapely/pull/1537

Issue #746, opened 2019-07-17
parallel_offset returns MultiString with missing parts


xathis:
Hi,

I'm noticing strange behavior in the parallel_offset method.

## Expected behavior and actual behavior.

parallel_offset should return a single LineString in the following example but returns a MultiLineString with a part missing.
Works correctly for some other distances or side='left' or when removing one of the points.

![incorrect](https://user-images.githubusercontent.com/175649/61411115-ea191400-a8e5-11e9-883e-c63fac1fa45c.png) incorrect behavior for distance=10
![correct](https://user-images.githubusercontent.com/175649/61411114-ea191400-a8e5-11e9-834f-5b96509fd950.png) correct behavior for distance=8

## Steps to reproduce the problem.
```python
import shapely.wkt
from matplotlib import pyplot as plt
from shapely.geometry import LineString

road = shapely.wkt.loads('LINESTRING ('
                         '424367.2267078362 4513099.023495375,'
                         ' 424370.9561004792 4513098.974304576,'
                         ' 424380.476033291 4513098.832676926,'
                         ' 424382.1039646821 4513098.816053259'
                         ')')
plt.plot(*road.coords.xy, marker='o')

line = road.parallel_offset(distance=10, side='right')

if isinstance(line, LineString):
    line = [line]

for l in line:
    plt.plot(*l.coords.xy, marker='o')

plt.gca().set_aspect('equal')
plt.show()
```

## Operating system

Mac OS X 10.14.5

## Shapely version and provenance

1.6.4.post1 installed from Conda

sgillies:
Thanks for the report. The parallel offset method in GEOS has been buggy, that's for sure.

LostFan123:
Also, sometimes it adds additional points to a `LineString`:
```python
from shapely import wkt

x = wkt.loads('LINESTRING (0.0000000000000000 0.0000000000000000, 0.2496337890625000 0.4992675781250000)')
x.parallel_offset(1e-10, side='left').wkt
# LINESTRING (8.94427190999916e-11 -4.47213595499958e-11, -8.94427190999916e-11 4.47213595499958e-11, 0.2496337889730573 0.4992675781697213, 0.2496337891519427 0.4992675780802787)
```

---
Shapely version: 1.6.4.post1, installed from conda.

guyanf:
When I use postgis , sometimes there are problems.

Jeremiah-England:
[...]

PR #862, opened 2020-03-10
Parallel offset docs

Jeremiah-England:
Additions: 
  1. Mention right side reversal (#785)  
  2. Break up parameter descriptions into paragraphs  
  3. Italicize parameters (``)  
  4. Link buffer method where the docs for *resolution* parameter are outsourced to it.  
  5. Improved explanation of the *mitre_limit* parameter.
  6. Warn that a MultiLineString may be returned where a simple LineString is expected (#746)

Reservations:  

  1. Breaking up the parameter descriptions into paragraphs makes it consistent. But maybe harder to read because of some very short paragraphs? 
  2. Perhaps a warning about #746 is too heavy.  

Could be better: 
  1. An example of how to post-process in the case of a MultiLineString where a LineString is expected to get the LineString you want. Perhaps there is a simple solution to this that could rely on predictable ordering of the LineStrings within the returned MultiLineString? 

Edit: rst rendering checks out.

### DIFF
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -1676,19 +1676,24 @@ linestring feature (right).
   Returns a LineString or MultiLineString geometry at a distance from the
   object on its right or its left side.
 
-  Distance must be a positive float value. The side parameter may be 'left' or
-  'right'. Left and right is determined by following the direction of given
-  geometric points of the LineString. The resolution of the offset around
-  each vertex of the object is parameterized as in the buffer method.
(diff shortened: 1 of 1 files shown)


Jeremiah-England:
@sgillies Thanks for the quick review! Changed the warning to a note. 

Also, about changing to numpydoc standards. I think that's an excellent idea. This is a perfect example, actually, because the right side reversal was already mentioned in the docstring.

PR #867, opened 2020-03-14
strtree docstrings

Jeremiah-England:
Here are some docstrings for the strtree module. They're in the numpydoc style. Even though shapely does not do that yet, @sgillies mentioned that sometime he would like to take them in that direction. I guess my thinking is that it can be done incrementally. You can checkout the auto generated documentation from them here: https://jeremiah-england.github.io/shapely/test-numpydocs/docs/build/index.html 

Let me know if you see any issues. One question that may to come up is that ugly
```rst
.. autoclass:: STRtree
    :members:
```  
in the middle of the module docstring. I'm auto generating the class documentation in the middle of the module docstring so the reference goes at the end. I'm not super attached to that if someone wants me to change it though. 

doctest and rst check out.

### DIFF
--- a/shapely/strtree.py
+++ b/shapely/strtree.py
@@ -1,17 +1,49 @@
+"""
+Index geometry objects for efficient lookup of nearby or
+nearest neighbors. Home of the `STRtree` class which is
+an interface to the query-only GEOS R-tree packed using
+the Sort-Tile-Recursive algorithm [1]_. 
+
+References
+----------
+  .. [1]  Leutenegger, Scott & Lopez, Mario & Edgington, Jeffrey. (1997).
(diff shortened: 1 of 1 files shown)


shapely/strtree.py
```diff
@@ -1,35 +1,74 @@
...
     True
     >>> polys[1] in result
     True
     >>> polys[2] in result
     False
-    >>> # Test empty tree
-    >>> s = STRtree([])
-    >>> s.query(query_geom)
+
+    Bahavior if an `STRtree` is created empty:
```
sgillies: Fix for a typo:

```suggestion
    Behavior if an `STRtree` is created empty:
```

Jeremiah-England:
Thanks for taking the time to review! I committed the suggested fix for the typo.

PR #870, opened 2020-03-19
Add Minimum Clearance implementation

tomplex:
This pull request exposes the "minimum clearance" functionality added in GEOS 3.6 to Shapely geometries as a unary property. I've never contributed here before and I'm sure there are more implications than I saw to adding support for a newer version of GEOS, so if this is a no-go (or there's a ton more work to be done to fully support that) please let me know, but I'm happy to put in the work to get it done. 

I've wanted for a while to add something back to this library which I've gotten great use out of, so I hope this is something that fits.

Thanks for your time!


tests/test_minimum_clearance.py
```diff
@@ -0,0 +1,33 @@
...
+
+import math
+
+import pytest
+
+from shapely.wkt import loads as load_wkt
+from shapely.geos import geos_version
+
+
+pytestmark = pytest.mark.skipif(geos_version < (3, 6, 0), reason='GEOS 3.6.0 required')
```
tomplex: It looks like the tests are being skipped even on versions of GEOS >= 3.6.0. Did I set this up correctly?

tests/test_minimum_clearance.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def test_point():
```
sgillies: @tomplex I've checked out your branch and when I modify it to decorate the new test functions as in the suggestion below, the skipping seems to work as intended. Note: I have GEOS 3.8.0, so I also modified `pytestmark` to skip if the version < 3.9.0 in my test.

```suggestion
@pytestmark
def test_point():
```

tomplex:
Hey @sgillies,

I modified the code the explicitly use the decorator and it seems the tests were still being skipped in Travis, though I was able to run / skip them on my local by playing with the required version. 

That still didn't get the tests to stop always skipping in CI, so I switched from `pytest` to `unittest` to see if that helps. Travis is taking a bit to start the run so I'll check back a bit later.

As an aside, the [pytest docs](http://doc.pytest.org/en/latest/skipping.html#skip-all-test-functions-of-a-class-or-module) suggest that creating a marker called `pytestmark` at the module level should enforce those rules for an entire file, so I guess it's good that the marker wasn't working regardless of how it was applied, but that's not helpful as to why it wasn't working. I'll try to take a closer look.

sgillies:
@tomplex I think you're doing everything right. Makes me think that our test environment is misconfigured and we're not using the version of GEOS we think we are. I'll dig into that today.

tomplex:
@sgillies, I made `pytest` show print statements so we could see the diagnostic info in the `test/__init__.py`. I checked every job in the matrix and they're all running with `geos` version 3.5.0, or at least as far as Shapely is concerned.

tomplex:
I also added a check of the `geos-config --version` which shows us that cached versions of GEOS are correct, so it means that the wires are getting crossed somewhere in the `pip install`, and we're not installing against the specified version properly.

tomplex:
Also, it looks like during the `pip install` step, we're using the right `geos-config` - so I'm not sure where we're going wrong. Maybe we're using the wrong Python to run tests?

tomplex:
hey @sgillies, I think I've gotten to the root of this, sort of. It looks like in all cases we're referring to the right version of GEOS, passing it into `pip` / the `setup.py` file correctly. I think the issue is that Shapely is not consistently referencing the proper version of GEOS. I included more info including steps to reproduce in #872.

sgillies:
@tomplex in https://github.com/Toblerity/Shapely/pull/873/files I've made some changes to help shapely use the correct GEOS and assert that it is doing so.

tomplex:
@sgillies  should this request be pointing at `master` or a different branch? Seems like you merged the changes for Travis into `maint-1.7`, so should I be pointing at that branch instead?

sgillies:
@tomplex if you're comfortable rebasing it on maint-1.7, please do 🙏 . If not, no sweat, I'm going to squash merge this PR and then I could cherry-pick the commit after.

tomplex:
No problem! I can do that as soon as I "get home" from work this evening.

tomplex:
@sgilles I think I mucked up the history a teensy bit but if you're gonna squash it anyway then it shouldn't make a difference. Thanks again for taking a look a this!

tomplex:
Crap, I accidentally carried over a change in the `__init__.py` regarding the Shapely version...no idea where that came from.

sgillies:
Fantastic. Thanks @tomplex !

PR #873, opened 2020-03-24
Set library paths in our CI environment

sgillies:
In https://github.com/Toblerity/Shapely/pull/870 we're seeing signs that we're not using the GEOS versions we expect when we run the tests.


sgillies:
Going to push an experimental commit to disable the env variables and see what version is printed.

sgillies:
I'm going to merge this into the master branch now.

Issue #804, opened 2019-12-09
Improve docs for shape and mapping functions


sgillies:
The docstrings need to provide more detail about the parameters and return values.

tomplex:
Released in 1.7.1

PR #876, opened 2020-03-26
Improve docstrings for shape, asShape, and mapping

tomplex:
per #804.

### DIFF
--- a/shapely/geometry/geo.py
+++ b/shapely/geometry/geo.py
@@ -76,15 +76,15 @@ def shape(context):
     -------
     Create a Point from GeoJSON, and then create a copy using __geo_interface__.
 
-      >>> context = {'type': 'Point', 'coordinates': [0, 1]}
-      >>> geom = shape(context)
-      >>> geom.type == 'Point'
-      True
-      >>> geom.wkt
-      'POINT (0 1)'
-      >>> geom2 = shape(geom)
-      >>> geom == geom2
-      True
+    >>> context = {'type': 'Point', 'coordinates': [0, 1]}
+    >>> geom = shape(context)
+    >>> geom.type == 'Point'
+    True
+    >>> geom.wkt
+    'POINT (0 1)'
+    >>> geom2 = shape(geom)
+    >>> geom == geom2
+    True
     """
     if hasattr(context, "__geo_interface__"):
         ob = context.__geo_interface__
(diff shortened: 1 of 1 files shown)


shapely/geometry/geo.py
```diff
@@ -123,5 +168,19 @@ def asShape(context):
 
 
 def mapping(ob):
-    """Returns a GeoJSON-like mapping"""
+    """
+    Returns a GeoJSON-like mapping from an object which
+    implements __geo_interface__
```
jorisvandenbossche: I would be explicit here that this includes shapely's own geometries (since I suppose that is probably the main usage of this function in practice), like ".. from a Geometry or any object which implements `__geo_interface__`"

shapely/geometry/geo.py
```diff
@@ -60,8 +60,31 @@ def box(minx, miny, maxx, maxy, ccw=True):
...
+    context :
+        a GeoJSON-like dict, which provides a "type" member describing the type
+        of the geometry and "coordinates" member providing a list of coordinates,
+        or an object which implements __geo_interface__.
+
+    Example
+    -------
+    Create a Point from GeoJSON, and then create a copy using __geo_interface__.
+
+      >>> context = {'type': 'Point', 'coordinates': [0, 1]}
```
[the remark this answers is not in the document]
tomplex: Thanks for the note, somehow I've managed to avoid heavily using RST so far so I copied from some other docstrings I found in Shapely.

shapely/geometry/geo.py
tomplex: Great point, I added that clarification.

tomplex:
Thanks for the notes @jorisvandenbossche!

shapely/geometry/geo.py
```diff
@@ -90,8 +113,30 @@ def shape(context):
...
+    Parameters
+    ----------
+    context :
+        a GeoJSON-like dict, which provides a "type" member describing the type
+        of the geometry and "coordinates" member providing a list of coordinates,
+        or an object which implements __geo_interface__.
+
+    Example
+    -------
+    Create a Point from GeoJSON, and then create a copy using __geo_interface__.
```
tomplex: Before this gets merged I need to alter this description - I copied it from the example for `shape` and then proceeded to not update it.

shapely/geometry/geo.py
```diff
@@ -90,8 +113,31 @@ def shape(context):
...
+    stored in the context, and changes to them will be reflected in the
+    returned geometry object.
+
+    Parameters
+    ----------
+    context :
+        a GeoJSON-like dict, which provides a "type" member describing the type
+        of the geometry and "coordinates" member providing a list of coordinates,
+        or an object which implements __geo_interface__.
+
```
sgillies: @tomplex shape is almost always better than asShape (see https://github.com/Toblerity/Shapely/issues/124) and we're going to start warning about asShape in the next version. What would you think about adding the following?

```suggestion

Notes
-----
The Adapter classes returned by this function trade performance for reduced storage of coordinate values. In general, the shape() function should be used instead.

```

shapely/geometry/geo.py
```diff
@@ -123,5 +169,19 @@ def asShape(context):
...
-    """Returns a GeoJSON-like mapping"""
+    """
+    Returns a GeoJSON-like mapping from a Geometry or any
+    object which implements __geo_interface__
+
+    Parameters
+    ----------
+    ob :
+        An object which implements __geo_interface__.
+
```
sgillies:
```suggestion

Returns
-------
dict

```

shapely/geometry/geo.py
tomplex: No problem! I didn't know that myself, so I'm glad you mentioned it.

tomplex:
Thanks for the suggestions! I also added "Returns" to the other docstrings to keep it consistent.

Issue #869, opened 2020-03-17
Failure while converting geojson to wkt


rohanil:
## Expected behavior and actual behavior.

shapely throws error while converting geojson to wkt. The geojson can be visualised in map as polygon.

## Steps to reproduce the problem.

```
from shapely import geometry

def geojson_to_wkt(geojson):
    area = geometry.asShape(geojson)
    return "SRID=4326;%s" % (area.wkt)

[...]

PR #877, opened 2020-03-27
Raise a descriptive error when a geometry has mixed dimensions.

tomplex:
There was some unguarded access to the third item in a coordinate in the speedups module which led to unhelpful errors when creating geometries from geojson which had mixed dimensionality (#869). This change should make it so that any time a user tries to create a geometry which might fail due to mixed dimensions, they'll get an error message telling them that.


tests/test_create_inconsistent_dimensionality.py
```diff
@@ -20,19 +21,34 @@
...
     with pytest.raises(ValueError) as exc:
         wkt = asShape(geojson).wkt
         assert exc.match("Inconsistent coordinate dimensionality")
 
 
+@mark.parametrize('constructor, args', direct_cases)
+def test_create_directly(constructor, args):
+    with pytest.raises(ValueError) as exc:
+        geom = constructor(*args)
+        assert exc.match("Inconsistent coordinate dimensionality")
```
sgillies: This assert needs to come out of the `with` block, otherwise it will never be reached.

shapely/speedups/_speedups.pyx
```diff
@@ -384,7 +388,10 @@ def geos_linearring_from_py(ob, update_geom=None, update_ndim=0):
             dy = coords[1]
             dz = 0
             if n == 3:
-                dz = coords[2]
+                try:
+                    dz = coords[2]
+                except IndexError:
+                    raise ValueError("Inconsistent coordinate dimensionality")
```
sgillies: 👍

tests/test_create_inconsistent_dimensionality.py
```diff
@@ -0,0 +1,55 @@
...
+When a "context" passed to shape/asShape has a coordinate
+which is missing a dimension we should raise a descriptive error.
+
+When we use mixed dimensions in a WKT geometry, the parser strips
+any dimension which is not present in every coordinate.
+"""
+
+
+import pytest
+from pytest import mark
```
sgillies: I've gotten in the habit of following the usage in the pytest docs. Would you be okay with always using the fully qualified name like `@pytest.mark.parametrize()`?

tomplex: Sure, no problem.

tests/test_create_inconsistent_dimensionality.py
tomplex: Duh. Thanks.

tomplex:
Thanks for the review!!

Issue #833, opened 2020-02-02
Add Voronoi Diagram


sgillies:
It's in the GEOS C API: https://github.com/libgeos/geos/blob/master/capi/geos_ts_c.cpp#L3223.

jorisvandenbossche:
This was added in https://github.com/Toblerity/Shapely/pull/881 by @tomplex, so closing.

PR #881, opened 2020-03-31
Add GEOSVoronoiDiagram to shapely

tomplex:
Per #833. 

I thought it made sense to keep Delauney / Voronoi close together in the code since they're similar in use and right next to each other in the CAPI, but if you disagree I'm happy to move it to wherever.

I couldn't nail down for sure exactly when VoronoiDiagram was added to the CAPI. [This](https://github.com/libgeos/geos/blob/master/NEWS#L167) entry was the best I could find indicating release version, so I went with 3.5.0.

### DIFF
--- a/shapely/ops.py
+++ b/shapely/ops.py
@@ -204,8 +204,14 @@ def voronoi_diagram(geom, envelope=None, tolerance=0.0, edges=False):
     """
     func = lgeos.methods['voronoi_diagram']
     envelope = envelope._geom if envelope else None
-    gc = geom_factory(func(geom._geom, envelope, tolerance, int(edges)))
-    return [g for g in gc.geoms]
+    result = geom_factory(func(geom._geom, envelope, tolerance, int(edges)))
+    try:
+        return [g for g in result.geoms]
+    except AttributeError:
+        # if the returned result is just one geometry (e.g. if your
+        # source has two points and you ask for edges-only)
+        # then we'll get back a single geom instead of a collection.
+        return [result]
 
 
 class ValidateOp(object):
--- a/shapely/ops.py
+++ b/shapely/ops.py
@@ -197,6 +197,14 @@ def voronoi_diagram(geom, envelope=None, tolerance=0.0, edges=False):
     -------
         List of geometries representing the Voronoi regions.
 
+    Notes
+    -----
+        The tolerance `argument` can be finicky and is known to cause the
+        algorithm to fail in several cases. If you're using `tolerance`
+        and getting a failure, try removing it. The test cases in
+        tests/test_voronoi_diagram.py show more details.
+
+
     References
     ----------
     [1] https://en.wikipedia.org/wiki/Voronoi_diagram
@@ -204,7 +212,14 @@ def voronoi_diagram(geom, envelope=None, tolerance=0.0, edges=False):
     """
     func = lgeos.methods['voronoi_diagram']
     envelope = envelope._geom if envelope else None
-    result = geom_factory(func(geom._geom, envelope, tolerance, int(edges)))
+    try:
+        result = geom_factory(func(geom._geom, envelope, tolerance, int(edges)))
+    except ValueError:
+        errstr = "Could not create Voronoi Diagram with the specified inputs."
+        if tolerance:
+            errstr += " Try running again with default tolerance value."
+        raise ValueError(errstr)
+
     try:
         return [g for g in result.geoms]
     except AttributeError:
--- a/shapely/ops.py
+++ b/shapely/ops.py
@@ -172,6 +172,14 @@ def triangulate(geom, tolerance=0.0, edges=False):
     gc = geom_factory(func(geom._geom, tolerance, int(edges)))
     return [g for g in gc.geoms]
 
+
+def voronoi_diagram(geom, envelope=None, tolerance=0.0, edges=False):
+    func = lgeos.methods['voronoi_diagram']
+    e = envelope._geom if envelope else None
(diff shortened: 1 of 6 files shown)


tests/test_voronoi_diagram.py
```diff
@@ -0,0 +1,101 @@
...
+@requires_geos_35
+def test_from_polygon_with_enough_tolerance():
+    poly = load_wkt('POLYGON ((0 0, 0.5 0, 0.5 0.5, 0 0.5, 0 0))')
+    regions = voronoi_diagram(poly, tolerance=1.0)
+
+    assert len(regions) == 2
+
+
+@requires_geos_35
+def test_from_polygon_without_enough_tolerance():
```
tomplex: This test and the one below it drive some un-intuitive behavior with regards to the `tolerance` argument. The first case is when we specify a tolerance to the function which is, I guess, "not enough"? I'm not necessarily sure how the tolerance value is associated with snapping (I need to look at the source C code more) but the C API returns a null geometry, which throws an error in the `geometry` factory function. 

The case below is perhaps more interesting: if we specify a tolerance and we don't have any floating point coordinates, then we also get a null geometry back. 

Either way, we could catch the generic error about null values and raise an error which is generally more descriptive, but I wonder if it's worth calling this behavior out specifically in an error, or just by adding more information about it to the docstrings / documentation.

shapely/ops.py
```diff
@@ -179,6 +179,63 @@ def triangulate(geom, tolerance=0.0, edges=False):
     gc = geom_factory(func(geom._geom, tolerance, int(edges)))
     return [g for g in gc.geoms]
 
+
+def voronoi_diagram(geom, envelope=None, tolerance=0.0, edges=False):
+    """
+    Constructs a Voronoi Diagram [1] from the given geometry.
```
Jeremiah-England: Most of this is over my head but it looks really nice. I think you need a underscore after the references in numpydocs if we want it to build with hyperlinks eventually. Like this: `[1]_`. And you'll also want to add a `.. ` before each reference down in the References section.  

See point number 14 here: https://numpydoc.readthedocs.io/en/latest/format.html

shapely/ops.py
```diff
@@ -179,6 +179,63 @@ def triangulate(geom, tolerance=0.0, edges=False):
...
+    Returns a list of geometries.
+
+    Parameters
+    ----------
+        geom: the input geometry whose vertices will be used to calculate
```
Jeremiah-England: This format is generally used for numpydocs: 

```rst 
param_name : param_type, default_value/'optional' if needed
    parameter description
```

shapely/ops.py
```diff
@@ -179,6 +179,63 @@ def triangulate(geom, tolerance=0.0, edges=False):
...
+            of the computation. A tolerance of 0.0 specifies that no
+            snapping will take place.
+        edges: If False (default), return Polygon regions. Else, return only
+            edges e.g. LineStrings.
+
+    GEOS documentation can be found at [2]
+
+    Returns
+    -------
+        List of geometries representing the Voronoi regions.
```
Jeremiah-England: In numpydocs you can include the return type right after the `Returns` header without any indentation. 

```
Returns
---------
list 
    geometries representing the Voronoi regions
``` 

There's a good example here: https://numpydoc.readthedocs.io/en/latest/example.html

shapely/ops.py
```diff
@@ -179,6 +179,63 @@ def triangulate(geom, tolerance=0.0, edges=False):
...
+
+    Notes
+    -----
+        The tolerance `argument` can be finicky and is known to cause the
+        algorithm to fail in several cases. If you're using `tolerance`
+        and getting a failure, try removing it. The test cases in
+        tests/test_voronoi_diagram.py show more details.
```
Jeremiah-England: The content of the `Notes` section doesn't need to be indented. 

I hope you don't mind the nitpicking! These do look really nice. : )

shapely/ops.py
tomplex: I'll defer to @sgillies on this, I'm happy to add the extra formatting if the docstrings will be built into documentation, but if they're mostly used for inline help (like `help(voronoi_diagram)`) then the extra formatting might not be helpful. Either way thanks for the suggestion! =)

shapely/ops.py
tomplex: This project is my first time using numpy docstrings, so thanks for the clarification!

shapely/ops.py
tomplex: No problem!! I'd rather do it right. Thanks for the review!

jorisvandenbossche:
The current return value is a list of geometries. Could it also be considered to return the actual GeometryCollection (as it is returned by GEOS)? The user could then still get the list by doing `result.geoms` themselves.

I assume that for users that pass a single geometry, the current API returning a list is easier (and also more consisitent with the existing `triangulate`, which is a good reason of course ..). But returning a geometry collection would make it easier to have it compatible with the PyGEOS version.

tomplex:
@jorisvandenbossche, though I personally agree that a `GeometryCollection` feels like a better option since that's what being returned natively by GEOS, I think one of the best parts of Shapely is the consistency and therefore intuitiveness of the API, so I went with a list of geometries. I'd be happy to discuss more if you feel strongly, especially because it seems like there will be a closer union of Shapely and PyGeos in the future, so it makes sense to take such things into account.

tomplex:
@sgillies should I rebase this onto master?

sgillies:
@tomplex yes, please. 

Returning a GeometryCollection would be more efficient and I'm in favor :+1:

Issue #56, opened 2013-06-07
Python 3 support


ojdo:
Are there any plans to add Python 3 support in the forseeable future?

PR #884, opened 2020-04-01
Drop Python 2.7 and Python 2 related logic and functions

### LINKED ISSUE #743: Plan for dropping Python 2 support
Python 2 has great for many of us. But after two decades, the [sun is setting on Python 2 support](https://python3statement.org/), which will [not be maintained past 2020](https://www.python.org/dev/peps/pep-0373/).

I'll propose the following. The next release for shapely is version 1.7, which currently supports both Python 2 and 3. This support should continue for the `maint-1.7` branch for this release series. However, after this release, the `master` and development efforts for the 1.8 milestone should require Python 3.5+.

How does this seem to everyone?

mwtoews:
* Minimum Python version 3.5
* Simplify/remove if-blocks that handled Python 2 or 3 logic
* No more izip or xrange
* Set cython: language_level=3
* Rearrange setup() kwargs for distutils or setuptools

Closes #743

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,24 +61,13 @@ import re
 import shutil
 import subprocess
 import sys
-try:
-    # If possible, use setuptools
-    from setuptools import setup
-    from setuptools.extension import Extension
-    from setuptools.command.build_ext import build_ext as distutils_build_ext
-except ImportError:
-    from distutils.core import setup
-    from distutils.extension import Extension
-    from distutils.command.build_ext import build_ext as distutils_build_ext
+from setuptools import setup
+from setuptools.extension import Extension
+from setuptools.command.build_ext import build_ext as distutils_build_ext
 from distutils.errors import CCompilerError, DistutilsExecError, \
     DistutilsPlatformError
 
 from _vendor.packaging.version import Version
-
-# Ensure minimum version of Python is running
-py_version = sys.version_info[0:2]
-if not (py_version == (2, 7) or py_version >= (3, 4)):
-    raise RuntimeError('Shapely requires Python 2.7, >=3.4')
 
 # Get geos_version from GEOS dynamic library, which depends on
 # GEOS_LIBRARY_PATH and/or GEOS_CONFIG environment variables
@@ -110,10 +99,7 @@ class GEOSConfig(object):
             raise OSError("Could not find geos-config script")
         if stderr and not stdout:
             raise ValueError(stderr.strip())
-        if sys.version_info[0] >= 3:
-            result = stdout.decode('ascii').strip()
-        else:
-            result = stdout.strip()
+        result = stdout.decode('ascii').strip()
         log.debug('%s %s: %r', self.cmd, option, result)
         return result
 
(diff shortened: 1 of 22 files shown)


setup.py
```diff
@@ -193,7 +185,7 @@ def version(self):
 setup_args = dict(
     name                = 'Shapely',
     version             = str(shapely_version),
-    requires            = ['Python (>=2.7)', 'libgeos_c (>=3.3)'],
+    requires            = ['Python (>=3.5)', 'libgeos_c (>=3.3)'],
```
jorisvandenbossche, line 188: Should probably add a "python_requires" field to the setup args (see https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires), that will make pip not try to install it it (and run into the runtime error defined above)

mwtoews: I've now added this, but only if we are using setuptools' setup(). The kwargs are different than for distutils' setup(). I'll explain further in the main thread ...

mwtoews:
Note that shapely's setup.py currently uses two packaging utilities:

1. setuptools - a popular package, widely used for most modern Python projects
2. distutils - a Python built-in standard package

I've tested to ensure that both packing methods work on Linux with `python setup.py sdist` and `python setup.py bdist`. There was a discussion on setuptools vs distutils way back in #56 and it was decided to keep supporting distutils. Are there any benefits of supporting a distutils fallback today?

sgillies:
@mwtoews I'm inclined to go all in on setuptools.

jorisvandenbossche:
Yes, I don't think there are any reasons for Shapely to still support distutils. As far as I know, the recommended tool is setuptools nowadays.

setup.py
```diff
@@ -61,25 +61,14 @@
...
-    from distutils.command.build_ext import build_ext as distutils_build_ext
+from setuptools import setup
+from setuptools.extension import Extension
+from setuptools.command.build_ext import build_ext as distutils_build_ext
 from distutils.errors import CCompilerError, DistutilsExecError, \
     DistutilsPlatformError
 
 from _vendor.packaging.version import Version
 
-# Ensure minimum version of Python is running
```
mwtoews, line 78: this logic/exception is handled by setuptools with `python_requires` kwarg, so is no longer needed

mwtoews:
I've amended this commit to also drop distutils

setuptools is already listed in requirements-dev.txt

sgillies:
@mwtoews @jorisvandenbossche is there any reason to keep supporting Python 3.5 in version 1.8.0?

mwtoews:
Probably no more reason today. [PEP 478](https://www.python.org/dev/peps/pep-0478/) says "Python 3.5 has now reached its end-of-life and has been retired" (September 5, 2020), so it's probably best to shelve it here too.

sgillies:
Right. A year ago, when 1.8.0 seemed close, 3.5 was still around, but not now. I will document the change, remove 3.5 from the build matrix, and unblock #1132.

jorisvandenbossche:
@sgillies I already have a PR to remove Python 3.5, but targeted to the shapely-2.0 branch. I can easily rebase that onto master.

sgillies:
@jorisvandenbossche I updated the master branch earlier today, forgot to follow up here.

Issue #878, opened 2020-03-28
DOC: shapely pyproj examples missing always_xy=True


BrendanMD:
Given that proj may now flip axes and that Shapely expects x-y order, I suggest adding always_xy=True to the pyproj examples with a note in the text. This bit me on the ass while trying to reproject to gnomonic.
`<< project = pyproj.Transformer.from_proj(proj_in, proj_out).transform`
`>> project = pyproj.Transformer.from_proj(proj_in, proj_out, always_xy=True).transform`

tomplex:
Released in 1.7.1.

jorisvandenbossche:
This was fixed by https://github.com/shapely/shapely/pull/885

Issue #826, opened 2020-01-30
DOC: update the pyproj transform example for deprecated syntax


jorisvandenbossche:
The example using pyproj in `transform` (https://github.com/Toblerity/Shapely/pull/783) is now raising deprecation warnings with the latest pyproj release:

```
In [2]: pyproj.Proj(init='epsg:4326')
/home/joris/miniconda3/envs/geo-dev/lib/python3.8/site-packages/pyproj/crs.py:77: FutureWarning: '+init=<authority>:<code>' syntax is deprecated. '<authority>:<code>' is the preferred initialization method.
  return _prepare_from_string(" ".join(pjargs))
Out[2]: Proj('+proj=longlat +datum=WGS84 +no_defs', preserve_units=True)
```

So we should update that to use it without `init=` (and could maybe also use `CRS` instead)

tomplex:
`pyproj` 2.1 actually introduced new functionality to do this with the [`Transformer`](https://pyproj4.github.io/pyproj/stable/api/transformer.html) API, and specifically calls out using `Proj` in the [gotcha's section](https://pyproj4.github.io/pyproj/stable/gotchas.html#proj-not-a-generic-latitude-longitude-to-projection-converter) as not the right thing when you're converting geographic coords to projected coords (which, for me at least, is almost always the case). `Transformer` seems to be the preferred API going forward. Since many folks are probably copying the example right from the docs, should they be updated to use this newer `Transformer` API instead?

sgillies:
I'm out of time to work on this. Would be grateful for a PR from a current pyproj user 🙏

tomplex:
Released in 1.7.1

jorisvandenbossche:
This was fixed by https://github.com/shapely/shapely/pull/885

Issue #789, opened 2019-11-02
shapely.ops.transform docs should state how it determines function arguments


brsr:
`shapely.ops.transform(func, geom)` takes a function `func(x,y)` that can accept scalar values or iterables for `x` and `y`. It determines which kind of function by trying to call the function on an iterable. If the function throws `TypeError`, it assumes it only accepts scalars and calls it on scalars.

This isn't explained [in the current version of the docs](https://shapely.readthedocs.io/en/stable/manual.html#shapely.ops.transform). It makes sense once you know what it's doing, but it seems like magic if you don't. It would be worthwhile to add a line to the docs explaining that `transform` figures out if `func` can accept iterables by seeing whether it throws `TypeError` or not.

tomplex:
Released in 1.7.1

jorisvandenbossche:
As noted above, this issue was fixed already in 1.7.1, so closing.

PR #885, opened 2020-04-01
Update the docs and docstrings for transform

tomplex:
This should address both #878 and #826. The docs now reflect the updated API in `pyproj >= 2.1`, demonstrate use of `Transformer` and `CRS`; show clearly and call out the necessity of `always_xy` per #878; and include the classic example using older versions of `pyproj`. I also changed the order of the examples to show the more current behavior of `pyproj` first.

### DIFF
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -2024,7 +2024,7 @@ If using `pyproj>=2.1.0`, the preferred method to project geometries is:
     utm = pyproj.crs.CRS('epsg:26913')
 
     project = pyproj.Transformer.from_crs(wgs84, utm, always_xy=True).transform
-    utm_point = transform(project, wgs84_point)
+    utm_point = transform(project, wgs84_pt)
 
 Otherwise, a partially applied transform function from pyproj also satisfies the requirements
 for `func`.
@@ -2044,7 +2044,7 @@ for `func`.
         wgs84,
         utm)
 
-    utm_point = transform(project, wgs84_point)
+    utm_point = transform(project, wgs84_pt)
 
 It is important to note that in the two examples above, the `always_xy` kwarg is required as Shapely only supports coordinates in X,Y
 order, and in PROJ 6 the WGS84 CRS uses the EPSG-defined Lat/Lon coordinate order instead of the expected Lon/Lat.
@@ -2066,7 +2066,7 @@ If using `pyproj < 2.1`, use syntax like:
         wgs84,
         utm)
 
-    utm_point = transform(project, wgs84_point)
+    utm_point = transform(project, wgs84_pt)
 
 Lambda expressions such as the one in
 
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -2019,8 +2019,8 @@ If using `pyproj>=2.1.0`, the preferred method to project geometries is:
 
     wgs84_pt = Point(-72.2495, 43.886)
 
-    wgs84 = pyproj.CRS('epsg:4326')
-    utm = pyproj.CRS('epsg:32618')
+    wgs84 = pyproj.CRS('EPSG:4326')
+    utm = pyproj.CRS('EPSG:32618')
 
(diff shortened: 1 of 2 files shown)


docs/manual.rst
```diff
@@ -2007,30 +2007,66 @@ For example, here is an identity function applicable to both types of input
...
+    import pyproj
+    import pyproj.crs
+
+    from shapely.geometry import Point
     from shapely.ops import transform
+
+    wgs84_pt = Point(-72.2495, 43.886)
+
+    wgs84 = pyproj.crs.CRS('epsg:4326')
+    utm = pyproj.crs.CRS('epsg:32618')
```
jorisvandenbossche, line 2024:
```suggestion
    wgs84 = pyproj.CRS('epsg:4326')
    utm = pyproj.CRS('epsg:32618')
```

(unlike the API docs indicate, the CRS object is available top-level, and the preferred way to use it IMO)

shapely/ops.py
```diff
@@ -211,11 +211,14 @@ def id_func(x, y, z=None):
 
       project = partial(
           pyproj.transform,
-          pyproj.Proj(init='epsg:4326'),
-          pyproj.Proj(init='epsg:26913'))
+          pyproj.Proj('epsg:4326', always_xy=True),
+          pyproj.Proj('epsg:26913', always_xy=True))
```
jorisvandenbossche, line 215: I would personally update it here to use CRS as well (which I think is almost always to prefer above Proj, nowadays)

docs/manual.rst
```diff
@@ -2007,30 +2007,66 @@ For example, here is an identity function applicable to both types of input
...
+
+    project = pyproj.Transformer.from_crs(wgs84, utm, always_xy=True).transform
+    utm_point = transform(project, wgs84_pt)
+
+Otherwise, a partially applied transform function from pyproj also satisfies the requirements
+for `func`.
```
jorisvandenbossche, line 2030: I am not fully sure that it is best to show both. As I think most people should use the above approach, also showing the below might only be confusing. Or maybe at least give a stronger indication the first is preferred?

tomplex: Yeah I think you're right, I'm going to remove the second example and not even bother showing a "not preferred" way to do it.

docs/manual.rst
tomplex: I missed that, thanks for the heads up!

tomplex:
Thanks for the review @jorisvandenbossche! I appreciate the feedback.

docs/manual.rst
```diff
@@ -2007,30 +2007,45 @@ For example, here is an identity function applicable to both types of input
...
+    from shapely.ops import transform
 
-    project = partial(
-        pyproj.transform,
-        proj_in,
-        proj_out)
+    wgs84_pt = Point(-72.2495, 43.886)
+
+    wgs84 = pyproj.CRS('epsg:4326')
+    utm = pyproj.CRS('epsg:32618')
```
sgillies, line 2023: Can we capitalize EPSG in the two lines above @tomplex? CRS authorities are generally capitalized in GDAL and PROJ usage. In combination with "init", on the other hand, we do need to use lower case only because "epsg" is literally a filename.

shapely/ops.py
```diff
@@ -203,19 +203,20 @@ def id_func(x, y, z=None):
...
 
-      from functools import partial
       import pyproj
 
-      project = partial(
-          pyproj.transform,
-          pyproj.Proj(init='epsg:4326'),
-          pyproj.Proj(init='epsg:26913'))
+      wgs84 = pyproj.CRS('epsg:4326')
+      utm = pyproj.CRS('epsg:32618')
```
sgillies, line 211: "EPSG" here too.

tomplex:
I added an extra paragraph regarding #789 because it seemed relevant and I was in in the docs anyway. If you want me to remove it and add it as its own item instead, that's fine.

Issue #888, opened 2020-04-02
Error when importing geos in MacOSX: libc does not exists 


pradal:
NB: the issue have been reported first as a conda-forge [issue](https://github.com/conda-forge/shapely-feedstock/issues/64). 
The error can be reproduced either on anaconda, conda or pip. 

## Expected behavior and actual behavior.

`from shapely import geos` fail.
The error is:

```Traceback (most recent call last): 
    File "<stdin>", line 1, in <module> 
    File "/Users/pradal/miniconda3/envs/shapeely/lib/python3.8/site-packages/shapely/geos.py", line 113, in <module> 
        free = load_dll('c').free 
    File "/Users/pradal/miniconda3/envs/shapeely/lib/python3.8/site-packages/shapely/geos.py", line 54, in load_dll 
        raise OSError( 
OSError: Could not find lib c or load any of its variants [].
```

## Steps to reproduce the problem.
`from shapely import geos`

Everything works if I change (in geos.py, line 113)

`free = load_dll('c').free`

by

1. `free = load_dll('System').free`
or
2. `free = load_dll('cxx').free` (
or even
3. `free=CDLL(None).free

See issue [64 on shapely-feedstock](https://github.com/conda-forge/shapely-feedstock/issues/64). 
> `libc` equivalent of macOS is `libSystem`, so `load_dll('c')` doesn't work.
> This is an issue upstream and not a packaging issue.

## Operating system

Mac OS X 10.15.1
No homebrew or macport installed.

## Shapely version and provenance

1.7.0

sgillies:
We are testing on Mac OS X, but only version 10.13: https://travis-ci.com/github/shapely/shapely-wheels/jobs/292629466. Thanks for suggesting `CDLL(None)`, that seems like an approach that would work across platforms.

pradal:
Do you want a PR?

sgillies:
@pradal sure, if you've got the time and interest! Can you aim the PR at our maint-1.7 branch?

pradal:
I have made the modification  in the PR #891  only for darwin, because I can not test other systems.
But it can easily generalised on other *nix systems.

ambientlight:
@sgillies: This issue still exist in `1.7.0`, as I missing something?

sgillies:
@ambientlight the fix is in the 1.7 branch but we haven't yet made a release that has the fix. Sorry about that.

camden-lowrance:
[...]

PR #891, opened 2020-04-05
Load libc on darwin in any cases

pradal:
Per #888 

`load_dll('c').free` fail on Darwin, without MacPort or Homebrew, because libc is not found on my system.
By default, on Mac, `libc` is provided by `libSystem.dylib`.

A robust solution is to replace `load_dll('c').free` by `CDLL(None).free` because Python is already using libc.

>     CDLL(None) internally calls dlopen(NULL), and as the dlopen
>     manpage says, "If filename is NULL, then the returned handle is for the
>     # main program". This way we can let the linker do the work to figure out
>     # which libc Python is actually using.
>

### DIFF
--- a/shapely/geos.py
+++ b/shapely/geos.py
@@ -135,7 +135,11 @@ elif sys.platform == 'darwin':
             ]
         _lgeos = load_dll('geos_c', fallbacks=alt_paths)
 
-    free = load_dll('c').free
+    # ctypes.CDLL(None) internally calls dlopen(NULL), and as the dlopen
+    # manpage says, "If filename is NULL, then the returned handle is for the
+    # main program". This way we can let the linker do the work to figure out
+    # which libc Python is actually using.
+    free = CDLL(None).free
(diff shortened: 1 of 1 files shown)


shapely/geos.py
```diff
@@ -135,7 +135,11 @@ def load_dll(libname, fallbacks=None, mode=DEFAULT_MODE):
             ]
         _lgeos = load_dll('geos_c', fallbacks=alt_paths)
 
-    free = load_dll('c').free
+    # ctypes.CDLL(None) internally calls dlopen(NULL), and as the dlopen
+    # manpage says, "If filename is NULL, then the returned handle is for the
+    # main program". This way we can let the linker do the work to figure out
+    # which libc Python is actually using.
+    free = CDLL(None).free
```
sgillies, line 142: 👍

adamjstewart:
I just ran into this same issue on macOS. Shouldn't this be merged to `master` as well?

PR #892, opened 2020-04-05
Add a macosx job

sgillies:


sgillies:
Nevermind. Travis-CI doesn't support python language builds on osx 🙄

rbuffat:
> The current hack-ish way to get a macOS machine with a specific Python version is to ask for a specific Xcode image, using the osx_image tag, which you know comes preinstalled with the Python version you want to use.

https://blog.travis-ci.com/2019-08-07-extensive-python-testing-on-travis-ci

Issue #830, opened 2020-01-31
'NoneType' object has no attribute 'GEOSSTRtree_destroy'


lexelby:
##  Expected behavior and actual behavior.

Shapely should dispose of resources without exceptions on program exit.

In https://github.com/inkstitch/inkstitch, we occasionally see this error:

```
Exception AttributeError: "'NoneType' object has no attribute 'GEOSSTRtree_destroy'" in <bound method STRtree.del of <shapely.strtree.STRtree instance at 0x0A56A328>> ignored
```

This is happening in a background thread (I think), hence the "ignored".  I believe it happens when the program exits while an STRtree instance is allocated, but I'm not sure.

This looks rather similar to #473.  Perhaps a similar solution would apply?

lexelby:
BTW, Shapely is incredibly awesome, and Ink/Stitch would not be what it is without it!

PR #897, opened 2020-04-15
Guard against calling methods on lgeos after it's been cleaned up

tomplex:
Fix for #830.

I couldn't really replicate this issue on macos, but [this merge request](https://github.com/Toblerity/Shapely/pull/528) fixed a very similar issue in basically the same way, and [`PreparedGeometry`](https://github.com/Toblerity/Shapely/blob/de4838e102dcfd54aadc8af03ed33def918334e6/shapely/prepared.py#L36) has a guard against the same thing, so it seems like it is a known situation where shapely's geos handle has been disposed of before python has cleaned up other Shapely objects.

For completeness I checked all the other `__del__` methods in Shapely and it looks like this is the last unguarded one. 

I switched the tests for `STRtree` from `unittest` to `pytest` because it seems like tests are moving in that direction and I thought I'd do it while I was in there, but if that's not cool I'm happy to switch them back.

### DIFF
--- a/tests/test_strtree.py
+++ b/tests/test_strtree.py
@@ -2,7 +2,8 @@ import gc
 
 from shapely.strtree import STRtree
 from shapely.geometry import Point, Polygon
-from shapely import geos
+
+from shapely import strtree
 
 from tests.conftest import requires_geos_342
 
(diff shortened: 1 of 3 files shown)


tests/test_strtree.py
```diff
@@ -1,69 +1,81 @@
...
+@requires_geos_342
+def test_safe_delete():
+    tree = STRtree([])
+
+    _lgeos = strtree.lgeos
+    strtree.lgeos = None
+
+    del tree
+
+    strtree.lgeos = _lgeos
```
sgillies, line 81: 👏

PR #898, opened 2020-04-16
Raise an error on creation of MultiPoint or LineString with empty components

### LINKED ISSUE #836: Inconsistent behavior when passing empty Points to MultiPoint
The [documentation on `MultiPoint`](https://shapely.readthedocs.io/en/latest/manual.html#MultiPoint) says that I can pass to it a sequence of `Point`s. But it doesn't warn us against passing empty `Point`s. Interestingly enough, if I pass an empty `Point` as the first element, it will throw an `AssertionError`:
```python
>>> from shapely.geometry import Point, MultiPoint
>>> MultiPoint([Point()])
```
```none
AssertionError                            Traceback (most recent call last)
<ipython-input-30-8942eaebd9a2> in <module>
----> 1 MultiPoint([Point()])

~\Miniconda3\lib\site-packages\shapely\geometry\multipoint.py in __init__(self, points)
     56             pass
     57         else:
---> 58             self._geom, self._ndim = geos_multipoint_from_py(points)
     59 
     60     def shape_factory(self, *args):

~\Miniconda3\lib\site-packages\shapely\geometry\multipoint.py in geos_multipoint_from_py(ob)
    162     except TypeError:
    163         n = ob[0]._ndim
--> 164     assert n == 2 or n == 3
    165 
    166     # Array of pointers to point geometries

AssertionError: 
```
```python
>>> MultiPoint([Point(), Point(1, 2)])
```
```none
AssertionError                            Traceback (most recent call last)
<ipython-input-31-08a0b83c889a> in <module>
----> 1 MultiPoint([Point(), Point(1, 2)])

~\Miniconda3\lib\site-packages\shapely\geometry\multipoint.py in __init__(self, points)
     56             pass
     57         else:
---> 58             self._geom, self._ndim = geos_multipoint_from_py(points)
     59 
     60     def shape_factory(self, *args):

~\Miniconda3\lib\site-packages\shapely\geometry\multipoint.py in geos_multipoint_from_py(ob)
    162     except TypeError:
    163         n = ob[0]._ndim
--> 164     assert n == 2 or n == 3
    165 
    166     # Array of pointers to point geometries

AssertionError: 
```
But if the empty `Point` goes after a nonempty `Point`, then it works fine:
```python
>>> MultiPoint([Point(1, 2), Point()])
>>> # no error here
```

But the resulting object will throw an exception after trying to get its wkt:
```python
>>> MultiPoint([Point(1, 2), Point()]).wkt
```
```none
OSError                                   Traceback (most recent call last)
<ipython-input-38-c645b8b312e6> in <module>
----> 1 MultiPoint([Point(1, 2), Point()]).wkt

~\Miniconda3\lib\site-packages\shapely\geometry\base.py in wkt(self, **kw)
    365     def wkt(self, **kw):
    366         """WKT representation of the geometry"""
--> 367         return WKTWriter(lgeos, **kw).write(self)
    368 
    369     @property

~\Miniconda3\lib\site-packages\shapely\geos.py in write(self, geom)
    361         if geom is None or geom._geom is None:
    362             raise ValueError("Null geometry supports no operations")
--> 363         result = self._lgeos.GEOSWKTWriter_write(self._writer, geom._geom)
    364         text = string_at(result)
    365         lgeos.GEOSFree(result)

OSError: exception: access violation reading 0x0000000000000000
```
I assume there needs to be a guard against a presence of empty points in the passed sequence, and the docs should be updated too.

---

tomplex:
Closes #836.

As of Shapely 1.7, asking for the WKT of a MultiPoint or -LineString with an empty component results in a segfault. I figured it made more sense to error in that case and let the user take care of the issue - realistically I can't see much utility in having a MultiPoint or -LineString with an empty component anyway, and I don't think you can create such a geometry through WKT, WKB, or GeoJSON, so it doesn't make sense to allow it through the geometry object constructor, either.

### DIFF
--- a/shapely/geometry/multilinestring.py
+++ b/shapely/geometry/multilinestring.py
@@ -132,6 +132,10 @@ def geos_multilinestring_from_py(ob):
     # add to coordinate sequence
     for l in range(L):
         geom, ndims = linestring.geos_linestring_from_py(obs[l])
+
+        if lgeos.GEOSisEmpty(geom):
+            raise ValueError("Can't create MultiLineString with empty component")
+
         subs[l] = cast(geom, c_void_p)
             
     return (lgeos.GEOSGeom_createCollection(5, subs, L), N)
--- a/shapely/geometry/multilinestring.py
+++ b/shapely/geometry/multilinestring.py
@@ -8,6 +8,7 @@ if sys.version_info[0] < 3:
 
 from ctypes import c_void_p, cast
 
+from shapely.errors import EmptyPartError, DimensionError
 from shapely.geos import lgeos
 from shapely.geometry.base import BaseMultipartGeometry, geos_geom_from_py
 from shapely.geometry import linestring
@@ -134,7 +135,7 @@ def geos_multilinestring_from_py(ob):
         geom, ndims = linestring.geos_linestring_from_py(obs[l])
 
         if lgeos.GEOSisEmpty(geom):
-            raise ValueError("Can't create MultiLineString with empty component")
+            raise EmptyPartError("Can't create MultiLineString with empty component")
 
         subs[l] = cast(geom, c_void_p)
             
--- a/shapely/geometry/multilinestring.py
+++ b/shapely/geometry/multilinestring.py
@@ -8,7 +8,7 @@ if sys.version_info[0] < 3:
 
 from ctypes import c_void_p, cast
 
-from shapely.errors import EmptyPartError, DimensionError
+from shapely.errors import EmptyPartError
 from shapely.geos import lgeos
 from shapely.geometry.base import BaseMultipartGeometry, geos_geom_from_py
 from shapely.geometry import linestring
--- a/shapely/errors.py
+++ b/shapely/errors.py
@@ -31,3 +31,7 @@ class TopologicalError(ShapelyError):
 
 class PredicateError(ShapelyError):
     """A geometric predicate has failed to return True/False."""
+
+
+class EmptyPartError(ShapelyError):
+    """An error signifying an empty part was encountered when creating a multi-part."""
--- a/shapely/geometry/multipoint.py
+++ b/shapely/geometry/multipoint.py
@@ -8,6 +8,7 @@ if sys.version_info[0] < 3:
 
 from ctypes import byref, c_double, c_void_p, cast
 
+from shapely.errors import EmptyPartError, DimensionError
 from shapely.geos import lgeos
 from shapely.geometry.base import (
     BaseMultipartGeometry, exceptNull, geos_geom_from_py)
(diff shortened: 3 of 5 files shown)


tomplex:
I can't for the life of me figure out why adding 4 (tested) lines of code reduced test coverage by half a percent.

shapely/geometry/multilinestring.py
```diff
@@ -132,6 +132,10 @@ def geos_multilinestring_from_py(ob):
     # add to coordinate sequence
     for l in range(L):
         geom, ndims = linestring.geos_linestring_from_py(obs[l])
+
+        if lgeos.GEOSisEmpty(geom):
+            raise ValueError("Can't create MultiLineString with empty component")
```
sgillies, line 137: @tomplex what would you think about a more shapely-specific exception here? Like a new shapely.errors.EmptyPartError?

I got this idea from Brett Slatkin: https://books.google.com/books?id=bTUFCAAAQBAJ&pg=PA185&lpg=PA185&dq=slatkin+python+base+exception&source=bl&ots=GahcQjraMD&sig=ACfU3U15-chIf93EZAganyYHVhWGQpI_EA&hl=en&sa=X&ved=2ahUKEwiG_tj4korpAhUOAp0JHUleDVEQ6AEwBXoECAsQAQ#v=onepage&q=slatkin%20python%20base%20exception&f=false

tomplex: That would be fine with me, but you mentioned in #836 you'd rather this be a warning and skip the empty geometry instead of `raise`-ing. Would you prefer if I re-worked this to do that instead?

sgillies: After more thought, I think an exception is the right approach. GEOS is going to crash otherwise, so an exception only helps.

tomplex:
While I'm in here, in `multilinestring.py` and `multipoint.py` there are `ValueError`s which are raising for invalid coordinate dimensionality, when we have `DimensionError` in `shapely.errors`. Should those be changed (here, or in another PR)? Or would an error *change* need a warning of some sort because it could be breaking?

tomplex:
Also, I'd be happy to move the relevant tests to `pytest` if that's desirable, but I don't want to include excess or unnecessary change.

sgillies:
@tomplex let's deal with the existing exceptions in a different PR. Would be a breaking change, a small one.

tomplex:
I think this is functionally "done" but I'm still not sure why coverage was reduced. I'll try and take a closer look this evening.

sgillies:
Don't sweat the little coverage changes, it's not entirely reliable. There's an issue with using assertRegex with python 2.7 though.

tomplex:
Can't believe I missed that...I'll fix that this evening.

sgillies:
🎉

PR #900, opened 2020-04-22
Deprecate mutability of Point, LineString, LinearRing geometries

jorisvandenbossche:
xref https://github.com/Toblerity/Shapely/issues/782/ (Shapely 2.0).

Example of deprecating mutability this for Point:

- Raise a warning in the `coords` setter
- Splitted the tests to more easily isolate the parts that are deprecated (to now filter the warnings, and later to remove in Shapely 2.0)
- Added one test that actually asserts a warning is raised

If that workflow looks OK, I can do the same for the other geometry types.


jorisvandenbossche:
Any feedback on this PR? 
I know the RFC still needs to be further discussed / approved, but to the extent possible, I would like to already move forward with some of the (less controversial) code changes (I could also target a separate branch than master until the RFC is actually approved?)

tests/test_point.py
```diff
@@ -2,6 +2,11 @@
 from shapely.geometry import Point, asPoint
 from shapely.errors import DimensionError
 
+import pytest
+
+
+shapely20_deprecated = pytest.mark.filterwarnings("ignore:Setting:FutureWarning")
```
tomplex, line 8: Having written a couple of these `mark`ers for specifying required GEOS versions, I was wondering if we should put them in a common place (even just in `conftest.py`) so they don't need to be re-declared in every file in which they're used. Especially for one like this, which will likely get re-used many times in the coming months.

tomplex: It was lost on me at the time of commenting that the `:Setting:` part in the middle is referencing the text of the deprecation warning. Should there be a standard message that can be used everywhere deprecations are occurring?

jorisvandenbossche: > Should there be a standard message that can be used everywhere deprecations are occurring?

Yes, that sounds a good idea to have all messages that will be related to "Shapely 2.0 deprecations" have a consistent structure. That makes filtering them out like the above easier with a single pytest marker, and is also more recognizable for users.

(we could actually also go for a warning subclass, for example sphinx has "RemovedInSphinx4Warning": https://github.com/sphinx-doc/sphinx/blob/776ec7957c3407a5ed75dc4369a0042d91698bd1/sphinx/deprecation.py#L21, but that can also easily be adopted later if we want to do so)

> I was wondering if we should put them in a common place (like conftest.py) so they don't need to be re-declared in every file in which they're used. 

Yes, certainly! It was just because I was only editing a single file here, but once I do the same for the other geometry classes, we certainly need to move it to a general place (and conftest.py seems appropriate for that with pytest)

tomplex: I quite like the idea of a `RemovedInShapely20` warning subclass. Would make it much easier for users to filter out all of those warnings with a single go, if they'd like to.

jorisvandenbossche:
@sgillies are you OK with the approach here? If so, I can update this PR to cover all geometry types.

(although RFC 1 is not yet officially approved, I hope I can start on some of the non-controversial parts, like deprecating mutability)

shapely/geometry/point.py
```diff
@@ -128,6 +136,10 @@ def bounds(self):
     # Coordinate access
 
     def _set_coords(self, *args):
+        warnings.warn(
+            "Setting the 'coords' to mutate a Geometry in place is deprecated,"
+            " and will not be possible any more in Shapely 2.0",
+            FutureWarning, stacklevel=2)
```
sgillies, line 142: @jorisvandenbossche what would you think about adding a new ShapelyDeprecationWarning as recommended in
https://www.python.org/dev/peps/pep-0565/#additional-use-case-for-futurewarning ?

> For library and framework authors that want to ensure their API compatibility warnings are more reliably seen by their users, the recommendation is to use a custom warning class that derives from DeprecationWarning in Python 3.7+, and from FutureWarning in earlier versions.

tests/test_point.py
sgillies: I should have scrolled down here before I recommended the new class above :)

sgillies:
@jorisvandenbossche I'm in favor of this :+1: Let's do consider a new shapely warning class.

jorisvandenbossche:
I added a

jorisvandenbossche:
Also a workflow question: do I continue here in this PR with doing the same for the other geometry types, or do you prefer to keep that for a separate PR(s)?

sgillies:
@jorisvandenbossche it's all the same to me, one or multiple PRs as you prefer.

jorisvandenbossche:
OK, I added deprecation warnings for linestrings and linearrings as well (I think that's all, since collections' coords are not directly mutable (only their parts), and the same for polygons and its rings)

So this should be ready for review / merge

shapely/errors.py
```diff
@@ -31,3 +31,10 @@ class TopologicalError(ShapelyError):
 
 class PredicateError(ShapelyError):
     """A geometric predicate has failed to return True/False."""
+
+
+class ShapelyDeprecationWarning(FutureWarning):
+    """
+    Warning for features that will be removed or behaviour that will be
+    changed in a future release.
+    """
```
sgillies, line 40: :+1:

shapely/geometry/linestring.py
```diff
@@ -40,7 +42,9 @@ def __init__(self, coordinates=None):
         """
         BaseGeometry.__init__(self)
         if coordinates is not None:
-            self._set_coords(coordinates)
+            ret = geos_linestring_from_py(coordinates)
+            if ret is not None:
+                self._geom, self._ndim = ret
```
sgillies, line 47: @jorisvandenbossche I appreciate how we will be able to just delete _set_coords in the future. Thank you!

tests/__init__.py
```diff
@@ -24,3 +26,8 @@
 if lgeos.geos_version >= (3, 3, 0):
     # Remove any WKT writer defaults to pass tests for all versions of GEOS
     WKTWriter.defaults = {}
+
+
+shapely20_deprecated = pytest.mark.filterwarnings(
+    "ignore::shapely.errors.ShapelyDeprecationWarning"
+)
```
sgillies, line 33: Then in our own tests we should convert these warnings to errors?

jorisvandenbossche:
> Then in our own tests we should convert these warnings to errors?

It seems you deleted this comment (or I don't find it back in the github UI), but yes, we could do that, to ensure that we are explicit in the tests about what is deprecated (and don't miss a consequence elsewhere in the package). 
The `@shapely20_deprecated` marks tests we know are testing isolated deprecated behaviour, and then none of the other tests should raise this warning. And we can let pytest detect if that still happens.

jorisvandenbossche:
And thanks for merging! 
I will continue with some of the other deprecations (I think the adapter/proxy classes are also quite uncontroversial to already start with deprecating?)

sgillies:
Excellent! I'm really excited about getting a 1.8.0 out. I've got time this afternoon to review tickets and do some release planning. The sooner, the better. Maybe we can even skip a 1.7.1.

PR #912, opened 2020-05-14
Add note that orient() will orient holes in clockwise.

adishavit:


docs/manual.rst
```diff
@@ -986,7 +987,6 @@ example will be shown for each.
   >>> LinearRing([(1,0), (1,1), (0,0)]).is_ccw
   True
 
-A ring with an undesired orientation can be reversed like this:
```
sgillies, line 989: @adishavit can you restore this line please or explain why you removed it?

Otherwise, this is a very welcome addition, thanks!

adishavit: Huh! I don’t know why this got erased. Will do.

PR #915, opened 2020-05-19
Fix segfault when using unpickled STRtree in another process

kveretennicov:
GEOS STRtree handle was not excluded when pickling Python STRtree. When unpickled in another process it would become a dangling pointer and segfault.


kveretennicov:
@sgillies would you prefer this to target `maint-1.7` instead of `master`?

sgillies:
Yes, please @kveretennicov !

kveretennicov:
Done, @sgillies :+1:

sgillies:
Thanks very much @kveretennicov !

PR #920, opened 2020-05-26
update overlaps docs to more closely match Wikipedia / JTS

### LINKED ISSUE #908: Doc inaccuracy in definition of .overlaps()
This regards to the doc text for `geom.overlaps(other)` in the public documentation [here](https://shapely.readthedocs.io/en/latest/manual.html#object.overlaps)

The doc text says:

> Returns True if the objects intersect (see above) but neither contains the other.

I'm testing this with two polygons `a` and `b` where  `a.touches(b) == True`.

Ex:

```python
ln [3]: import shapely.wkt                                                                                                                                                                
In [4]: a = shapely.wkt.loads("POLYGON ((1 1, 1 4, 4 4, 4 1, 1 1))")                                                                                                                      
In [5]: b = shapely.wkt.loads("POLYGON ((2 0, 2 1, 3 1, 3 0, 2 0))")                                                                                                                      
In [6]: a.touches(b)                                                                                                                                                                      
Out[6]: True
In [7]: a.intersects(b)                                                                                                                                                                   
Out[7]: True
In [8]: b.intersects(a)
Out[8]: True
In [9]: a.contains(b)                                                                                                                                                                       
Out[9]: False
In [10]: b.contains(a)                                                                                                                                                                       
Out[10]: False
In [11]: a.overlaps(b)                                                                                                                                                                    
Out[11]: False
```
This contradicts the doc text which says that `a.overlaps(b)` should be `True`, given that `a` and `b` do intersect and that neither contains the other. 

I suspect this is an inaccuracy in the docs.

tomplex:
Closes #908.

### DIFF
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -1223,8 +1223,8 @@ interior point in common.
 
 .. method:: object.overlaps(other)
 
-  Returns ``True`` if the objects have more than one but not all points in common,
-  they have the same dimension, and the intersection of the interiors of the geometries
+  Returns ``True`` if the geometries have more than one but not all points in common,
+  have the same dimension, and the intersection of the interiors of the geometries
   has the same dimension as the geometries themselves.
 
 .. method:: object.touches(other)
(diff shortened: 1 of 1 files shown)


sgillies:
@tomplex thank you!

Issue #910, opened 2020-05-09
Add documentation about the wkt precision


cyang-kth:
Currently, there are two ways to get the wkt of a geometry. 

The direct reading of `wkt` property of a geometry object returns a trimmed version of the geometry while the `dumps` function of a geometry returns a non-trimmed (high precision) of a geometry. 

https://github.com/Toblerity/Shapely/blob/c613520bc4a1fce667a18ac14ef95de62677c093/shapely/geometry/base.py#L363-L365

https://github.com/Toblerity/Shapely/blob/4433a841d379044bdca0a891f8fe79c6eed4e06d/shapely/wkt.py#L17-L22

Should this feature be documented in the official manual?

tomplex:
Yes, I definitely feel this should be documented, especially considering with the property there's no real way to control the output, unlike when you use the serialization functions in the `wkt` module.

apiszcz:
This may be related.

https://gis.stackexchange.com/questions/94856/wkt-precision-rounding-using-shapely

sgillies:
@apiszcz thanks for the suggestion!

tomplex:
@sgillies, this was resolved in 1.7.1.

PR #921, opened 2020-05-26
Update docs re: WKT serialization

tomplex:
Adds some more detail to docs around WKT serializiation, addressing #910. Also adds docstrings to `wkt.dump` and `wkt.dumps` to explicitly list available keyword arguments.

### DIFF
--- a/shapely/wkt.py
+++ b/shapely/wkt.py
@@ -6,21 +6,76 @@ from shapely import geos
 # Pickle-like convenience functions
 
 def loads(data):
-    """Load a geometry from a WKT string."""
+    """
+    Load a geometry from a WKT string.
+
+    Parameters
+    ----------
+    data :
+        A WKT string
+
+    Returns
+    -------
+    Shapely geometry object
+    """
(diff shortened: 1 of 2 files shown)


shapely/wkt.py
```diff
@@ -6,21 +6,77 @@
...
-    """Dump a WKT representation of a geometry to a string.
+    """
+    Dump a WKT representation of a geometry to a string.
 
-    See available keyword output settings in ``shapely.geos.WKTWriter``.
+    Parameters
+    ----------
+    ob :
+        A geometry object of any type to be dumped to WKT.
+    trim :
```
jorisvandenbossche, line 47:
```suggestion
    trim : bool, default False
```

shapely/wkt.py
```diff
@@ -6,21 +6,77 @@
...
+        Remove excess decimals from the WKT.
+    rounding_precision (GEOS 3.3+):
+        Round output to the specified number of digits
```
jorisvandenbossche, line 50: Maybe mention that the default is to print the full precision?

PR #922, opened 2020-05-27
Document set-like properties

### LINKED ISSUE #903: Document set-like properties
See #901 for background. The properties in particular: https://github.com/Toblerity/Shapely/blob/808fd466b8c7e07dc635bdba5e700efd0b200984/shapely/geometry/base.py#L261-L274.

tomplex:
Addressing #903. I feel like this is a good start, but I'm not sure if the intention of the issue was also to add docstrings to the methods? Happy to do that too if it's desired.


sgillies:
Thanks again @tomplex !

PR #926, opened 2020-06-06
Deprecate asShape (CachingGeometryProxy adapter classes)

jorisvandenbossche:
Deprecating the adapter classes that create geometry-like proxy objects with coordinates stored in external objects (xref https://github.com/shapely/shapely-rfc/pull/1/).

I again started with the Point geometry to check if the workflow looks good. 

There are however multiple ways to create those objects. For points, you have the general `asShape` and `asPoint`, but in principle also the `PointAdapter` class constructor (but I don't know whether those are regarded as public API?). 

The way I did it now is to add a deprecation warning to both `asShape` and `asPoint`, so the message can be specific. Bu there are certainly other ways to do it as well.

### DIFF
--- a/shapely/geometry/geo.py
+++ b/shapely/geometry/geo.py
@@ -1,7 +1,11 @@
 """
 Geometry factories based on the geo interface
 """
-from .point import Point, asPoint
+import warnings
+
+from shapely.errors import ShapelyDeprecationWarning
+
+from .point import Point, _asPoint
 from .linestring import LineString, asLineString
 from .polygon import Polygon, asPolygon
 from .multipoint import MultiPoint, asMultiPoint
@@ -153,6 +157,11 @@ def asShape(context):
     >>> poly.intersects(point)
     False
     """
+    warnings.warn(
+        "The 'asShape()' adapter is deprecated and will be removed in Shapely 2.0. "
+        "Use the 'shape()' function instead.",
+        ShapelyDeprecationWarning, stacklevel=2)
+
     if hasattr(context, "__geo_interface__"):
         ob = context.__geo_interface__
     else:
@@ -164,7 +173,7 @@ def asShape(context):
         raise ValueError("Context does not provide geo interface")
 
     if geom_type == "point":
-        return asPoint(ob["coordinates"])
+        return _asPoint(ob["coordinates"])
     elif geom_type == "linestring":
         return asLineString(ob["coordinates"])
     elif geom_type == "polygon":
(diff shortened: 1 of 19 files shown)


shapely/geometry/geo.py
```diff
@@ -164,7 +173,7 @@ def asShape(context):
         raise ValueError("Context does not provide geo interface")
 
     if geom_type == "point":
-        return asPoint(ob["coordinates"])
+        return _asPoint(ob["coordinates"])
```
jorisvandenbossche, line 176: This is to avoid a duplicate warning coming from `asPoint` (the alternative is to only warn in `asPoint` with a more general warning message, and that warning will the also be seen when calling `asShape`)

sgillies:
Excellent, @jorisvandenbossche . I can't think of a reason not to eliminate the adapter classes in 2.0. There's a lot of code there that we won't want to maintain anymore and it only exists to support a kind of mutability.

How about warning from the adapter constructors, mentioning deprecation of the class *and* the factory functions (asShape, asPoint, &c)?

jorisvandenbossche:
> How about warning from the adapter constructors, mentioning deprecation of the class and the factory functions (asShape, asPoint, &c)?

OK, doing a warning only in the class + a general message so it is valid for all possible ways to construct it sounds good.

jorisvandenbossche:
Updated the warning message and location. If that looks good, then will do the same for the other geometry types.

tests/test_geointerface.py
```diff
@@ -71,6 +74,61 @@ def test_geointerface(self):
         self.assertIsInstance(shape, MultiPolygonAdapter)
         self.assertEqual(len(shape.geoms), 1)
 
+    def test_geointerface(self):
```
jorisvandenbossche, line 77: I copied the test above, and adapted it to use `shape` instead of `asShape`, as there didn't seem a dedicated test for `shape` for all different geometry types.

tests/test_ndarrays.py
```diff
@@ -46,6 +49,36 @@ def test_polygon(self):
             [(1.0, 3.0), (1.0, 4.0), (2.0, 4.0), (2.0, 3.0), (1.0, 3.0)]
         )
 
+    @unittest.skipIf(not numpy, 'numpy not installed')
+    def test_multipoint(self):
+        arr = numpy.array([[1.0, 1.0, 2.0, 2.0, 1.0], [3.0, 4.0, 4.0, 3.0, 3.0]])
```
jorisvandenbossche, line 54: Also here, copied the tests above but adapted using the normal constructor, to ensure we keep tests for this functionality when the adapter tests are removed.

jorisvandenbossche:
@sgillies ok, updated for all geometry types. This is ready for review / merge.

sgillies:
Excellent!

PR #930, opened 2020-06-09
Fix makevalid GEOS version

tomplex:
I used > instead of >= for the GEOS version, resulting in a bug. Sorry about that.


sgillies:
Thanks @tomplex !

Issue #944, opened 2020-07-14
Python 3.9 support


expeditioneer:
It installs fine with python 3.7 and python 3.8 but it fails with python 3.9 on gentoo. See [Gentoo PR](https://github.com/gentoo/gentoo/pull/16283)

sgillies:
Thanks for the heads up @expeditioneer. I scanned the build output in the Gentoo PR and it looks like this is largely an issue between Cython and Python 3.9.

mwtoews:
Currently requirements-dev.txt specifies `cython==0.29.14`, which could cause issues with newer versions of Python (e.g. [scan this](https://cython.readthedocs.io/en/latest/src/changes.html) for "3.9")

Inevitably, we'll need to add `3.9-dev` to the CI test matrix.

austinulfers:
```
Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "C:\Users\austi\AppData\Local\Temp\pip-install-d8c90ger\shapely\setup.py", line 85, in <module>
        from shapely._buildcfg import geos_version_string, geos_version, \
      File "C:\Users\austi\AppData\Local\Temp\pip-install-d8c90ger\shapely\shapely\_buildcfg.py", line 205, in <module>
        lgeos = CDLL("geos_c.dll")
      File "C:\Python\lib\ctypes\__init__.py", line 374, in __init__
        self._handle = _dlopen(self._name, mode)
    FileNotFoundError: Could not find module 'geos_c.dll' (or one of its dependencies). Try using the full path with 
constructor syntax.
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

Getting this error when I do `pip install shapely` on windows 10 & python 3.9.0

Someone suggested installing the installer from `https://trac.osgeo.org/osgeo4w/` which didn't work either (I assume that was outdated/for a different issue). Is there a current workaround for 3.9?

sgillies:
@austinulfers we don't have wheels for Windows and 3.9 on PyPI and no time frame for doing it yet. I'm not sure whether its only a matter of extending our AppVeyor builds here https://github.com/Toblerity/Shapely/blob/maint-1.7/appveyor.yml#L40 or whether its more involved.

mwtoews:
[...]

PR #945, opened 2020-07-15
Replace Python 3.8-dev with Python 3.9-dev

mwtoews:
This should be testing with Python 3.9.0 beta 4. See [PEP 596](https://www.python.org/dev/peps/pep-0596/), where Python 3.9 is expected to be released October 2020.

See related #944

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,13 +68,13 @@ matrix:
         GEOSVERSION="3.8.1"
         SPEEDUPS=0
         NUMPY=0
-    - python: "3.8-dev"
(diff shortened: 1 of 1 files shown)


sgillies:
@mwtoews thanks! How do you explain the 3.9-dev job passing despite the problems we expected with using an oldish Cython 0.29.14?

mwtoews:
Yup, it's a curious one. Probably a bit more testing would be good to do before adding 'Programming Language :: Python :: 3.8' to setup.py, as I don't have 100% confidence for 3.9 support.

PR #951, opened 2020-07-31
Increase test coverage for GeometryCollection class

tomplex:
We're about to experience a good bit of churn with the push to Shapely 2.0 & merge with PyGEOS, so this seemed like a good time to work on increasing test coverage, especially in the geometry classes where I expect we will be trying to keep the interface and behavior identical while completely changing the internal workings. I also would like to consolidate tests to use Pytest, as it seems like most new tests have been written using that framework, and I think it'd be nice to have more consistency there.

If you're a 👍 on this, I'll keep adding to this or open another PR for more work, whatever is preferred. If not, no worries, but I wanted to check in to see what other folks thought before I did too much.

### DIFF
--- a/tests/test_collection.py
+++ b/tests/test_collection.py
@@ -30,6 +30,7 @@ def geometrycollection_geojson():
 def test_empty(geom):
     assert geom.type == "GeometryCollection"
     assert geom.type == geom.geom_type
+    assert geom.is_empty
     assert len(geom) == 0
     assert geom.geoms == []
 
@@ -83,6 +84,7 @@ def test_empty_geointerface_adapter():
     assert m.geom_type == "GeometryCollection"
     assert len(m) == 0
     assert m.geoms == []
+    assert m.is_empty
 
 
 def test_geometrycollection_adapter_deprecated(geometrycollection_geojson):
--- a/tests/test_collection.py
+++ b/tests/test_collection.py
@@ -1,7 +1,8 @@
-from . import unittest, shapely20_deprecated
+from shapely import wkt
+from . import shapely20_deprecated
 
 from shapely.errors import ShapelyDeprecationWarning
-from shapely.geometry import LineString
+from shapely.geometry import LineString, Point
 from shapely.geometry.collection import GeometryCollection
 from shapely.geometry import shape
 from shapely.geometry import asShape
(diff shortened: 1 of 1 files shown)


sgillies:
@tomplex I'm in favor, but also want to check in with @jorisvandenbossche who is doing much of the 2.0 work.

jorisvandenbossche:
Yes, also +1 on this, both for better usage of pytest as for trying to get better test coverage. 
(asking about modernizing the tests / removing some unittest specificities was on my todo list, as I noticed those issues as well doing my first shapely 2.0 PRs)

tests/test_collection.py
```diff
@@ -9,101 +10,86 @@
...
+        {"type": "LineString", "coordinates": ((2, 0), (1, 0))}
     ]}
+
+
+@pytest.mark.parametrize('geom', [
+    GeometryCollection(),
+    shape({"type": "GeometryCollection", "geometries": []}),
+    shape({"type": "GeometryCollection", "geometries": [
+        {"type": "Point", "coordinates": ()},
+        {"type": "LineString", "coordinates": (())}
```
jorisvandenbossche, line 26: Should this be considered as empty? The subparts are empty geometries, but I would expect the actual GeometryCollection to have a length of 2? 
(or are the empty subparts skipped on construction?)

jorisvandenbossche: (now, given this was in the original tests as well, that's certainly for a separate issue to discuss ;))

tomplex: Glad you brought this up! I thought it was a little funky when I wrote it. I'm happy the test was there because it demonstrated behavior I certainly would not have expected.

It feels like something we should address, but I think you're right - it should probably as a separate issue.

mwtoews:
> modernizing the tests / removing some unittest specificities was on my todo list

Yes please! The unittest way was fine 10 years ago, but I much prefer to see a modern pytest approach, as done in this PR. I have no idea if it really changes the coverage metrics (+0.1%??), but this is secondary to maintaining to modern conventions.

tests/test_collection.py
mwtoews: Empty strikes again! JTS/GEOS/PostGIS regard this geometry to be empty. There are, however, subtle differences. GEOS and JTS appear to convert this simply to 'GEOMETRYCOLLECTION EMPTY', while PostGIS keeps the empty items for WKT, but doesn't count them as geometries, e.g.:
```
db=> select ST_IsEmpty(g), ST_NumGeometries(g), ST_AsText(g), ST_AsText(ST_GeometryN(g, 1)), ST_AsText(ST_GeometryN(g, 2))
db-> from (select 'GeometryCollection(Point empty, LineString empty)'::geometry g) f;
-[ RECORD 1 ]----+-------------------------------------------------
st_isempty       | t
st_numgeometries | 0
st_astext        | GEOMETRYCOLLECTION(POINT EMPTY,LINESTRING EMPTY)
st_astext        | POINT EMPTY
st_astext        | LINESTRING EMPTY
```
Why, however, ST_NumGeometries is zero but ST_GeometryN for 1 and 2 works is suspicious.

tomplex:
> I have no idea if it really changes the coverage metrics (+0.1%??), 

Coverage increased because I added some test cases for things which weren't covered before: the `__geo_interface__` property and the `geos_geometrycollection_from_py` function with a 3-dim input. Small, but enough to get us from 79% to 92% for the file. Now all that's left uncovered here is the doctest test runner.

Either way, since it seems this is popular I'll keep on trucking down this path with added focus on updating tests to use `pytest`. Would folks prefer to see more numerous, smaller PR's, or for me to keep adding to this one? I think I'd learn towards smaller PR's but I'm generally indifferent.

tests/test_collection.py
tomplex: Weird, thanks for the example. Consistency with JTS and GEOS in these details seems nice from the perspective of someone familiar with those tools, but I wonder where we draw the line. After all, I'd expect `[[], []]` in Python to be truth-y and have a length of two, and it feels sort of wrong to change that when we're talking about what is essentially a list of geometries.

jorisvandenbossche:
I am personally in favor of smaller PRs as well (maybe updating a few files combined in a single PR is fine, but I would certainly not do a single overhaul of all test files in one PR).

tests/test_collection.py
jorisvandenbossche: It seems that GEOS also keeps the empty geometries, at least with PyGEOS I get:

```
In [39]: g = pygeos.Geometry("GEOMETRYCOLLECTION( POINT EMPTY, POINT EMPTY )") 

In [40]: g   
Out[40]: <pygeos.Geometry GEOMETRYCOLLECTION EMPTY>

In [41]: pygeos.get_num_geometries(g)
Out[41]: 2
```

So it still indicates that the collection has 2 geometries, although the WKT version just says "EMPTY" (probably because the collection is regarded as empty if all subparts are empty):

```
In [42]: pygeos.is_empty(g)   
Out[42]: True
```

mwtoews: There appears to be multiple bugs in this geometry sub-genre, such as [the PostGIS one mentioned above](https://trac.osgeo.org/postgis/ticket/4736), that [these geometries do not accurately serialize to WKT or WKB](https://trac.osgeo.org/geos/ticket/1040).

This is a minor shapely bug:
```python
from shapely import geos, wkt
g = wkt.loads("GEOMETRYCOLLECTION(LINESTRING EMPTY, POLYGON EMPTY)")
print(geos.lgeos.GEOSGetNumGeometries(g._geom))  # 2
print(len(g))  # 0
print(len(g.geoms))  # 0
```

jorisvandenbossche: I moved this discussion to a new issue: https://github.com/Toblerity/Shapely/issues/956

jorisvandenbossche:
Apart from the unrelated empty-collection discussion (opened a new issue for that -> https://github.com/Toblerity/Shapely/issues/956), this PR is ready to be merged?

tomplex:
Yes, I believe this is ready to go.

tests/test_collection.py
```diff
@@ -9,101 +10,86 @@
...
+    ]}),
+    wkt.loads('GEOMETRYCOLLECTION EMPTY'),
+])
+def test_empty(geom):
+    assert geom.type == "GeometryCollection"
+    assert geom.type == geom.geom_type
+    assert len(geom) == 0
+    assert geom.geoms == []
```
mwtoews, line 34: When #956 is addressed, the above two assertions will fail for the multi-part empty geometry. We'll get to that later however...

Suggest adding `assert geom.is_empty` here.

tests/test_collection.py
```diff
@@ -9,101 +10,86 @@
...
+
+
+@shapely20_deprecated
+def test_empty_geointerface_adapter():
+    d = {"type": "GeometryCollection", "geometries": []}
+
+    m = asShape(d)
+    assert m.geom_type == "GeometryCollection"
+    assert len(m) == 0
+    assert m.geoms == []
```
mwtoews, line 85: Also add `assert m.is_empty` here.

PR #955, opened 2020-08-05
Deprecate ctypes attribute and array interface

jorisvandenbossche:
Deprecating the ctypes attribute and the array_interface.

See https://github.com/shapely/shapely-rfc/pull/1/files#diff-554e526a1f0fe104169383c013e37a7cR360 for the rationale

The array interface is kept intact for the CoordinateSequence, so occurence of `np.array(geom)` can be replaced with `np.array(geom.coords)`. 
(we should eventually implement this differently (not using ctypes), but that can be left for later or even for Shapely 2.0)

### DIFF
--- a/tests/test_coords.py
+++ b/tests/test_coords.py
@@ -1,6 +1,9 @@
 from . import unittest, numpy
-from shapely import geometry
 
+import pytest
+
+from shapely import geometry
+from shapely.errors import ShapelyDeprecationWarning
 
 class CoordsTestCase(unittest.TestCase):
     """
@@ -34,6 +37,12 @@ class CoordsTestCase(unittest.TestCase):
             coords[::-1].tolist(),
             processed_coords.tolist()
         )
+
+
+def test_multipoint_ctypes_deprecated():
+    coords = geometry.LineString([[12, 34], [56, 78]]).coords
+    with pytest.warns(ShapelyDeprecationWarning, match="ctypes"):
+        coords.ctypes
 
 
 def test_suite():
--- a/tests/test_coords.py
+++ b/tests/test_coords.py
@@ -39,7 +39,11 @@ class CoordsTestCase(unittest.TestCase):
         )
 
 
-def test_multipoint_ctypes_deprecated():
+def test_coords_ctypes_deprecated():
+    """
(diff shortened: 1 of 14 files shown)


jorisvandenbossche:
@sgillies @tomplex this should be ready for review

tests/test_coords.py
```diff
@@ -36,5 +39,11 @@ def test_data_destriding(self):
         )
 
 
+def test_multipoint_ctypes_deprecated():
+    coords = geometry.LineString([[12, 34], [56, 78]]).coords
+    with pytest.warns(ShapelyDeprecationWarning, match="ctypes"):
+        coords.ctypes
```
sgillies, line 45: @jorisvandenbossche would the following suffice?

```suggestion
    with pytest.warns(ShapelyDeprecationWarning, match="ctypes"):
        geometry.LineString().coords.ctypes
```

I feel like the unused coordinate values are slightly distracting. Also, what would you think about a minimal docstring for the test?

tests/test_linestring.py
```diff
@@ -208,5 +218,26 @@ def test_linestring_adapter_deprecated():
...
+    with pytest.warns(ShapelyDeprecationWarning, match="array_interface"):
+        line.array_interface()
+
+
+@unittest.skipIf(not numpy, 'Numpy required')
+def test_linestring_array_interface_numpy_deprecated():
+    import numpy as np
+
+    line = LineString(((1.0, 2.0), (3.0, 4.0)))
+    with pytest.warns(ShapelyDeprecationWarning, match="array interface"):
```
sgillies, line 238: Today I learned about the `match` keyword argument! I'm going to use this more often.

tests/test_coords.py
jorisvandenbossche: Added a docstring (and also fixed the test name, that was a copy/paster error).

`geometry.LineString().coords.ctypes` actually doesn't work, because for an empty geometry the `coords` attribute returns an empty list instead of a CoordinateSequence object (to me this feels a bit like a bug, but is done on purpose in the code).

sgillies: I agree about the bug.

PR #960, opened 2020-08-11
[2.0] Remove mutability of geometries 

jorisvandenbossche:
I branched off from master, so the shapely-2.0 branch first needs to be reset to master state for this to work (the actual commit is only the last one)

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -321,11 +321,7 @@ class BaseGeometry(object):
             return []
         return CoordinateSequence(self)
 
-    def _set_coords(self, ob):
-        raise NotImplementedError(
-            "set_coords must be provided by derived classes")
-
-    coords = property(_get_coords, _set_coords)
+    coords = property(_get_coords)
 
     @property
     def xy(self):
@@ -849,10 +845,6 @@ class BaseMultipartGeometry(BaseGeometry):
                                   "provide the array interface")
 
     def _get_coords(self):
-        raise NotImplementedError("Sub-geometries may have coordinate "
-                                  "sequences, but collections do not")
-
-    def _set_coords(self, ob):
         raise NotImplementedError("Sub-geometries may have coordinate "
                                   "sequences, but collections do not")
 
--- a/tests/test_linestring.py
+++ b/tests/test_linestring.py
@@ -245,6 +245,16 @@ class LineStringTestCase(unittest.TestCase):
         self.assertEqual(a.shape[0], 0)
 
 
+def test_linestring_immutable():
+    line = LineString(((1.0, 2.0), (3.0, 4.0)))
+
+    with pytest.raises(AttributeError):
+        line.coords = [(-1.0, -1.0), (1.0, 1.0)]
+
+    with pytest.raises(TypeError):
+        line.coords[0] = (-1.0, -1.0)
+
+
 def test_linestring_adapter_deprecated():
     coords = [[5.0, 6.0], [7.0, 8.0]]
     with pytest.warns(ShapelyDeprecationWarning, match="proxy geometries"):
--- a/tests/test_linestring.py
+++ b/tests/test_linestring.py
@@ -41,16 +41,6 @@ class LineStringTestCase(unittest.TestCase):
                           'coordinates': ((1.0, 2.0), (3.0, 4.0))})
 
     @shapely20_deprecated
-    def test_linestring_mutate(self):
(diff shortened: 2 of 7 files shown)


jorisvandenbossche:
@sgillies could you rebase the `shapely-2.0` branch on top of master? (see also #962) Then the diff of this PR should be more correct.

sgillies:
@jorisvandenbossche done. You still need to rebase your own branch, right?

jorisvandenbossche:
Thnks! Switching branches back and again also seemed to trigger GitHub

jorisvandenbossche:
@sgillies @mwtoews this is ready for review/merge

shapely/geometry/base.py
```diff
@@ -338,11 +338,7 @@ def _get_coords(self):
             return []
         return CoordinateSequence(self)
 
-    def _set_coords(self, ob):
-        raise NotImplementedError(
-            "set_coords must be provided by derived classes")
```
sgillies, line 343: Bye bye fragile mutable geometries :hocho:

tests/test_linestring.py
```diff
@@ -132,16 +132,6 @@ def test_linestring(self):
                          {'type': 'LineString',
                           'coordinates': ((1.0, 2.0), (3.0, 4.0))})
 
-    @shapely20_deprecated
-    def test_linestring_mutate(self):
```
mwtoews, line 136: Perhaps rename this `test_linestring_immutable` and check that `.coord` raise TypeError.

tests/test_point.py
```diff
@@ -113,19 +113,6 @@ def test_point(self):
         self.assertEqual(p.__geo_interface__,
                          {'type': 'Point', 'coordinates': (3.0, 4.0)})
 
-    @shapely20_deprecated
-    def test_point_mutate(self):
```
mwtoews, line 117: Similar recommendation as above: rename this `test_point_immutable` and check that `.coord` raise TypeError.

mwtoews (commented):
Theses changes are welcome! I have suggestions to keep two tests to check that point and linestring geometries now have immutable `.coords` properties. It might be helpful to the reader to see a comment to say that these were mutable in previous versions. All other tests should be dropped, however.

jorisvandenbossche:
> keep (and modify) two tests to check that point and linestring geometries now have immutable .coords properties. It might be helpful to the reader to see a comment to say that these were mutable in previous versions. 

Good point, will update for that. 

Thanks for the review!

jorisvandenbossche:
Added back tests for point/linestring/linearring to ensure immutability.

We now have those errors:

```python
In [1]: from shapely.geometry import LineString

In [3]: l = LineString([(0,0), (1,1)])


In [6]: l.coords[0] = (5, 5)
TypeError                                 Traceback (most recent call last)
<ipython-input-6-41687dadde5d> in <module>
----> 1 l.coords[0] = (5, 5)

TypeError: 'CoordinateSequence' object does not support item assignment

In [7]: l.coords = [(0,0), (1,1)]
AttributeError                            Traceback (most recent call last)
<ipython-input-7-dc6ae96a6b61> in <module>
----> 1 l.coords = [(0,0), (1,1)]

AttributeError: can't set attribute
```

which are the default errors produced by Python for a property without a setter (the first one about CoordinateSequence is already existing on master, just added to the test for completeness). 

Are we OK with that test? I could also add back a setter to the `coords` property with a more custom error if we want that.

mwtoews (approved):
:tada:

sgillies:
@jorisvandenbossche I'm fine with that. Since we're going to have a warning period, a custom error won't be necessary. If these objects had custom setters, custom exceptions would help users differentiate between Shapely usage bugs and Shapely bugs/Python usage bugs. In 2.0, trying to modify a coordinate sequence is a Python usage bug, and a type or attribute error fits.

jorisvandenbossche:
OK, this should be good to go then!

sgillies:
@mwtoews @jorisvandenbossche merge when you are ready. I'm happy to let the two of you be stewards of the shapely-2.0 branch while I focus on the others.

jorisvandenbossche:
I would be happy with managing the shapely-2.0 branch (as noted in https://github.com/Toblerity/Shapely/issues/962, I would also prefer to regularly rebase it on top of master initially), but then you will need to give me commit rights ;) (I can ensure to use them wisely!)

sgillies:
@jorisvandenbossche I'd lost track of that requirement and will make it so.

jorisvandenbossche:
Thanks!

PR #963, opened 2020-08-12
TST: clean-up + more cases for Point/LineString constructor tests

jorisvandenbossche:
While refactoring the class constructors based on pygeos, I noticed that not all cases that are currently supported are fully covered by the tests (eg generator, single-element list for Point, ...). So I added some more tests, and at once cleaned up the existing constructor tests.

### DIFF
--- a/tests/test_linestring.py
+++ b/tests/test_linestring.py
@@ -42,7 +42,7 @@ def test_from_linestring():
     line = LineString(((1.0, 2.0), (3.0, 4.0)))
     copy = LineString(line)
     assert copy.coords[:] == [(1.0, 2.0), (3.0, 4.0)]
-    assert lgeos.GEOSGeomType(copy._geom).decode('ascii') == 'LineString'
+    assert copy.geom_type == 'LineString'
 
 
 def test_from_linearring():
@@ -50,7 +50,7 @@ def test_from_linearring():
     ring = LinearRing(coords)
     copy = LineString(ring)
     assert copy.coords[:] == coords
-    assert lgeos.GEOSGeomType(copy._geom).decode('ascii') == 'LineString'
+    assert copy.geom_type == 'LineString'
 
 
 def test_from_linestring_z():
@@ -58,7 +58,7 @@ def test_from_linestring_z():
     line = LineString(coords)
     copy = LineString(line)
     assert copy.coords[:] == coords
-    assert lgeos.GEOSGeomType(copy._geom).decode('ascii') == 'LineString'
+    assert copy.geom_type == 'LineString'
 
 
 def test_from_generator():
--- a/tests/test_linestring.py
+++ b/tests/test_linestring.py
@@ -2,7 +2,6 @@ from . import unittest, numpy, shapely20_deprecated
 import pytest
 
 from shapely.errors import ShapelyDeprecationWarning
-from shapely.geos import lgeos
 from shapely.geometry import LineString, asLineString, Point, LinearRing
 
 
--- a/tests/test_linestring.py
+++ b/tests/test_linestring.py
@@ -6,6 +6,109 @@ from shapely.geos import lgeos
 from shapely.geometry import LineString, asLineString, Point, LinearRing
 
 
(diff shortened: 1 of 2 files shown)


tests/test_point.py
```diff
@@ -5,6 +5,61 @@
...
+    # From single coordinate pair
+    p = Point((3.0, 4.0))
+    assert p.coords[:] == [(3.0, 4.0)]
+    p = Point([3.0, 4.0])
+    assert p.coords[:] == [(3.0, 4.0)]
+
+    # From coordinate sequence
+    p = Point([(3.0, 4.0)])
+    assert p.coords[:] == [(3.0, 4.0)]
+
```
mwtoews, line 34: I wonder if 3D sequences should be added here too (and in the following four tests too).

tests/test_point.py
```diff
@@ -5,6 +5,61 @@
...
 
+def test_from_coordinates():
+    # 2D points
+    p = Point(1.0, 2.0)
+    assert p.coords[:] == [(1.0, 2.0)]
+    assert p.has_z is False
+
+    # 3D Point
+    p = Point(1.0, 2.0, 3.0)
+    assert p.coords[:] == [(1.0, 2.0, 3.0)]
```
mwtoews, line 16: suggest adding `assert p.has_z` here.

tests/test_linestring.py
```diff
@@ -6,6 +6,109 @@
...
+    line = LineString((Point(1.0, 2.0), (2.0, 3.0), Point(3.0, 4.0)))
+    assert line.coords[:] == [(1.0, 2.0), (2.0, 3.0), (3.0, 4.0)]
+
+
+def test_from_linestring():
+    # From another linestring
+    line = LineString(((1.0, 2.0), (3.0, 4.0)))
+    copy = LineString(line)
+    assert copy.coords[:] == [(1.0, 2.0), (3.0, 4.0)]
+    assert lgeos.GEOSGeomType(copy._geom).decode('ascii') == 'LineString'
```
mwtoews, line 45: perhaps it's also time to move away from ctypes functions and use:
```suggestion
    assert copy.geom_type == 'LineString'
```
(similar applies to below)

tests/test_linestring.py
```diff
@@ -6,6 +6,109 @@
...
+def test_from_invalid_dim():
+    with pytest.raises(ValueError, match="at least 2 coordinate tuples"):
+        LineString([(1, 2)])
+
+    with pytest.raises(ValueError, match="Inconsistent coordinate dimensionality"):
+        LineString([(1, 2, 3), (4, 5)])
+
+    # TODO this does not fail
+    # with pytest.raises(ValueError, match="Inconsistent coordinate dimensionality"):
+    #     LineString([(1, 2), (3, 4, 5)]))
```
mwtoews, line 97: Good catch, see #964 for issue.

tests/test_linestring.py
jorisvandenbossche: Yes! I was assuming this was done for a reason (eg because each of the subclasses simply hardcodes its geometry type instead of calling into GEOS), but the `geom_type` attribute actually calls `GEOSGeomTypeId`, so will indeed update.

tests/test_linestring.py
jorisvandenbossche: Thanks for opening the issue

tests/test_point.py
jorisvandenbossche: Added

jorisvandenbossche:
@mwtoews Thanks for the review! Pushed some updates

tests/test_linestring.py
```diff
@@ -6,6 +6,109 @@
 from shapely.geometry import LineString, asLineString, Point, LinearRing
```
tomplex, line 6: Looks like you can remove the `lgeos` import on the line above this (github won't let me comment on that line...)

jorisvandenbossche: Indeed, thanks!

PR #974, opened 2020-08-21
CLN: remove deprecated to_wkb/to_wkt methods

jorisvandenbossche:
I noticed that those functions and methods are already deprecated for many years. So I think those can already be removed for 1.8, but of course, since we are removing a bunch of things in 2.0, we can also directly keep this removal for 2.0 as well (in which case I can edit the PR to target the shapely-2.0 branch instead of master)

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -90,46 +90,12 @@ def geom_factory(g, parent=None):
     return ob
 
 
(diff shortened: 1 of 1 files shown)


mwtoews (approved):
These can be removed for 1.8 in my opinion, as these warnings have been in place for 7 years.

PR #978, opened 2020-09-01
Replace internal usage of Multi geom iteration with .geoms attribute

jorisvandenbossche:
This should help with actually deprecating it in https://github.com/Toblerity/Shapely/pull/950


jorisvandenbossche:
cc @tomplex

sgillies (approved):

PR #979, opened 2020-09-02
CLN: remove unused doctest runners from .py files

jorisvandenbossche:
I don't think they are used at the moment, so cleaning this up. Also, renaming tests/binascii_hex.txt to tests/test_binascii_hex.txt so it gets picked up by pytest.

BTW, with `pytest shapely --doctest-modules` we could actually run the doctests on CI (but that needs some fixes first, as there are some failures at the moment)


mwtoews (approved):
Looks good. And `pytest shapely --doctest-modules` looks promising, as I only see 6 failures to address.

PR #983, opened 2020-09-16
Refactor Geometry classes to subclass the C extension type

jorisvandenbossche:
.. and use the pygeos functions for all the Geometry methods/properties. 

This is a first step for integrating PyGEOS into Shapely (https://github.com/Toblerity/Shapely/issues/962):

- For now I still use PyGEOS as an external dependency (it will be merged into Shapely in a later step, I first wanted to ensure this essential part is working out)
- This makes all the Geometry classes (Point, LineString, Polygon, etc) subclasses of the pygeos.Geometry extension type.
- This also registers those subclasses to pygeos. This means that all pygeos functions will start to return the correct shapely subclass, and not the generic pygeos.Geometry
- For most of the methods and properties of the Geometry classes, I switched to using pygeos functions (eg ` geom_factory(self.impl['difference'](self, other))` to `pygeos.difference(self, other)`). For other functionality in most other modules (eg shapely.ops etc), this is still left as is and keeps using the ctypes interface (this can be cleaned up later in a separate PR)
- A large chunk of the diff is actually the re-implementation of the Point(), Linestring(..), Polygon(..) etc constructors, as those don't directly map to the pygeos functions to create geometries from coordinates (the shapely ones are more flexible / allow more variation in input types)
- All tests are passing locally (with pygeos master), but I xfailed a bunch of tests where there are small differences or where we still need to implement something in the pygeos code to support it. But so it can be seen explicitly in the diff which ones are not yet passing.

cc @caspervdw

### DIFF
--- a/shapely/coords.py
+++ b/shapely/coords.py
@@ -54,25 +54,6 @@ class CoordinateSequence(object):
         else:
             raise TypeError("key must be an index or slice")
 
-    # def array_interface(self):
-    #     """Provide the Numpy array protocol."""
-    #     if sys.byteorder == 'little':
-    #         typestr = '<f8'
-    #     elif sys.byteorder == 'big':
-    #         typestr = '>f8'
-    #     else:
-    #         raise ValueError(
-    #             "Unsupported byteorder: neither little nor big-endian")
-    #     ai = {
-    #         'version': 3,
-    #         'typestr': typestr,
-    #         'data': self._ctypes,
-    #         }
-    #     ai.update({'shape': (len(self), self._ndim)})
-    #     return ai
-
-    # __array_interface__ = property(array_interface)
-
     def __array__(self, dtype=None):
         return self._coords
 
--- a/shapely/coords.py
+++ b/shapely/coords.py
@@ -25,18 +25,6 @@ class CoordinateSequence(object):
 
     """
 
-    # Attributes
-    # ----------
-    # _cseq : c_void_p
-    #     Ctypes pointer to GEOS coordinate sequence
-    # _ndim : int
-    #     Number of dimensions (2 or 3, generally)
-    # __p__ : object
-    #     Parent (Shapely) geometry
-    _cseq = None
-    _ndim = None
-    __p__ = None
-
     def __init__(self, coords):
         self._coords = coords
 
--- a/shapely/coords.py
+++ b/shapely/coords.py
@@ -37,41 +37,18 @@ class CoordinateSequence(object):
     _ndim = None
     __p__ = None
 
-    def __init__(self, parent):
-        self.__p__ = parent
-
-    def _update(self):
-        self._ndim = self.__p__._ndim
-        self._cseq = lgeos.GEOSGeom_getCoordSeq(self.__p__._geom)
+    def __init__(self, coords):
+        self._coords = coords
 
     def __len__(self):
-        self._update()
-        cs_len = c_uint(0)
-        lgeos.GEOSCoordSeq_getSize(self._cseq, byref(cs_len))
-        return cs_len.value
+        return self._coords.shape[0]
 
     def __iter__(self):
-        self._update()
-        dx = c_double()
-        dy = c_double()
-        dz = c_double()
-        has_z = self._ndim == 3
         for i in range(self.__len__()):
-            lgeos.GEOSCoordSeq_getX(self._cseq, i, byref(dx))
-            lgeos.GEOSCoordSeq_getY(self._cseq, i, byref(dy))
-            if has_z:
-                lgeos.GEOSCoordSeq_getZ(self._cseq, i, byref(dz))
-                yield (dx.value, dy.value, dz.value)
-            else:
-                yield (dx.value, dy.value)
+            yield tuple(self._coords[i].tolist())
 
     def __getitem__(self, key):
-        self._update()
-        dx = c_double()
-        dy = c_double()
-        dz = c_double()
         m = self.__len__()
-        has_z = self._ndim == 3
         if isinstance(key, int):
             if key + m < 0 or key >= m:
                 raise IndexError("index out of range")
@@ -79,79 +56,48 @@ class CoordinateSequence(object):
                 i = m + key
             else:
                 i = key
-            lgeos.GEOSCoordSeq_getX(self._cseq, i, byref(dx))
-            lgeos.GEOSCoordSeq_getY(self._cseq, i, byref(dy))
-            if has_z:
-                lgeos.GEOSCoordSeq_getZ(self._cseq, i, byref(dz))
-                return (dx.value, dy.value, dz.value)
-            else:
-                return (dx.value, dy.value)
+            return tuple(self._coords[i].tolist())
         elif isinstance(key, slice):
             res = []
             start, stop, stride = key.indices(m)
             for i in range(start, stop, stride):
-                lgeos.GEOSCoordSeq_getX(self._cseq, i, byref(dx))
-                lgeos.GEOSCoordSeq_getY(self._cseq, i, byref(dy))
-                if has_z:
-                    lgeos.GEOSCoordSeq_getZ(self._cseq, i, byref(dz))
-                    res.append((dx.value, dy.value, dz.value))
-                else:
-                    res.append((dx.value, dy.value))
+                res.append(tuple(self._coords[i].tolist()))
             return res
         else:
             raise TypeError("key must be an index or slice")
 
-    @property
-    def _ctypes(self):
-        self._update()
-        has_z = self._ndim == 3
-        n = self._ndim
-        m = self.__len__()
-        array_type = c_double * (m * n)
-        data = array_type()
-        temp = c_double()
-        for i in range(m):
-            lgeos.GEOSCoordSeq_getX(self._cseq, i, byref(temp))
-            data[n*i] = temp.value
-            lgeos.GEOSCoordSeq_getY(self._cseq, i, byref(temp))
-            data[n*i+1] = temp.value
-            if has_z:
-                lgeos.GEOSCoordSeq_getZ(self._cseq, i, byref(temp))
-                data[n*i+2] = temp.value
-        return data
+    # def array_interface(self):
+    #     """Provide the Numpy array protocol."""
+    #     if sys.byteorder == 'little':
+    #         typestr = '<f8'
+    #     elif sys.byteorder == 'big':
+    #         typestr = '>f8'
+    #     else:
+    #         raise ValueError(
+    #             "Unsupported byteorder: neither little nor big-endian")
+    #     ai = {
+    #         'version': 3,
+    #         'typestr': typestr,
+    #         'data': self._ctypes,
+    #         }
+    #     ai.update({'shape': (len(self), self._ndim)})
+    #     return ai
 
-    def array_interface(self):
-        """Provide the Numpy array protocol."""
-        if sys.byteorder == 'little':
-            typestr = '<f8'
-        elif sys.byteorder == 'big':
-            typestr = '>f8'
-        else:
-            raise ValueError(
-                "Unsupported byteorder: neither little nor big-endian")
-        ai = {
-            'version': 3,
-            'typestr': typestr,
-            'data': self._ctypes,
-            }
-        ai.update({'shape': (len(self), self._ndim)})
-        return ai
+    # __array_interface__ = property(array_interface)
 
-    __array_interface__ = property(array_interface)
+    def __array__(self, dtype=None):
+        return self._coords
 
     @property
     def xy(self):
         """X and Y arrays"""
-        self._update()
         m = self.__len__()
         x = array('d')
         y = array('d')
-        temp = c_double()
         for i in range(m):
-            lgeos.GEOSCoordSeq_getX(self._cseq, i, byref(temp))
-            x.append(temp.value)
-            lgeos.GEOSCoordSeq_getY(self._cseq, i, byref(temp))
-            y.append(temp.value)
+            xy = self._coords[i].tolist()
+            x.append(xy[0])
+            y.append(xy[1])
         return x, y
 
 
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -15,6 +15,8 @@ import sys
 from warnings import warn
 from functools import wraps
 
+import pygeos
+
 from shapely.affinity import affine_transform
 from shapely.coords import CoordinateSequence
 from shapely.errors import WKBReadingError, WKTReadingError
@@ -60,34 +62,20 @@ def dump_coords(geom):
         raise ValueError('Unhandled geometry type: ' + repr(geom.type))
 
 
-def geometry_type_name(g):
-    if g is None:
-        raise ValueError("Null geometry has no type")
-    return GEOMETRY_TYPES[lgeos.GEOSGeomTypeId(g)]
+class _DummyGeometry:
+    # pygeos.from_shapely expects a python object that exposes
+    # the GEOS pointer as a __geom__ attribute
+
+    def __init__(self, g):
+        self.__geom__ = g
 
 
 def geom_factory(g, parent=None):
     # Abstract geometry factory for use with topological methods below
     if not g:
         raise ValueError("No Shapely geometry can be created from null value")
-    ob = BaseGeometry()
-    geom_type = geometry_type_name(g)
-    # TODO: check cost of dynamic import by profiling
-    mod = __import__(
-        'shapely.geometry',
-        globals(),
-        locals(),
-        [geom_type],
-        )
-    ob.__class__ = getattr(mod, geom_type)
-    ob._geom = g
-    ob.__p__ = parent
-    if lgeos.methods['has_z'](g):
-        ob._ndim = 3
-    else:
-        ob._ndim = 2
-    ob._is_empty = False
-    return ob
+    ob = _DummyGeometry(g)
+    return pygeos.from_shapely(ob)
 
 
 def deserialize_wkb(data):
@@ -142,7 +130,7 @@ class JOIN_STYLE(object):
 EMPTY = deserialize_wkb(a2b_hex(b'010700000000000000'))
 
 
-class BaseGeometry(object):
+class BaseGeometry(pygeos.Geometry):
     """
     Provides GEOS spatial predicates and topological operations.
 
@@ -166,7 +154,6 @@ class BaseGeometry(object):
     #     case of a multipart geometry member.
     __geom__ = EMPTY
     __p__ = None
-    _ndim = None
     _crs = None
     _other_owned = False
     _is_empty = True
@@ -187,6 +174,10 @@ class BaseGeometry(object):
 
         self._is_empty = True
         self.__geom__ = val
+
+    @property
+    def _ndim(self):
+        return pygeos.get_coordinate_dimension(self)
 
     def __bool__(self):
         return self.is_empty is False
@@ -277,26 +268,26 @@ class BaseGeometry(object):
     # ----------------------------------------
 
     def geometryType(self):
-        return geometry_type_name(self._geom)
+        return self.geom_type
 
     @property
     def type(self):
-        return self.geometryType()
+        return self.geom_type
 
     @property
     def wkt(self):
         """WKT representation of the geometry"""
-        return WKTWriter(lgeos).write(self)
+        return pygeos.to_wkt(self)
 
     @property
     def wkb(self):
         """WKB representation of the geometry"""
-        return WKBWriter(lgeos).write(self)
+        return pygeos.to_wkb(self)
 
     @property
     def wkb_hex(self):
         """WKB hex representation of the geometry"""
-        return WKBWriter(lgeos).write_hex(self)
+        return pygeos.to_wkb(self, hex=True)
 
     def svg(self, scale_factor=1., **kwargs):
         """Raises NotImplementedError"""
@@ -343,7 +334,7 @@ class BaseGeometry(object):
     @property
     def geom_type(self):
         """Name of the geometry's type, such as 'Point'"""
-        return self.geometryType()
+        return GEOMETRY_TYPES[pygeos.get_type_id(self)]
 
     # Real-valued properties and methods
     # ----------------------------------
@@ -351,20 +342,20 @@ class BaseGeometry(object):
     @property
     def area(self):
         """Unitless area of the geometry (float)"""
-        return self.impl['area'](self)
+        return pygeos.area(self)
 
     def distance(self, other):
         """Unitless distance to other geometry (float)"""
-        return self.impl['distance'](self, other)
+        return pygeos.distance(self, other)
 
     def hausdorff_distance(self, other):
         """Unitless hausdorff distance to other geometry (float)"""
-        return self.impl['hausdorff_distance'](self, other)
+        return pygeos.hausdorff_distance(self, other)
 
     @property
     def length(self):
         """Unitless length of the geometry (float)"""
-        return self.impl['length'](self)
+        return pygeos.length(self)
 
     @property
     def minimum_clearance(self):
@@ -382,25 +373,26 @@ class BaseGeometry(object):
         collection of points. The boundary of a point is an empty (null)
         collection.
         """
-        return geom_factory(self.impl['boundary'](self))
+        return pygeos.boundary(self)
 
     @property
     def bounds(self):
         """Returns minimum bounding region (minx, miny, maxx, maxy)"""
+        # TODO return empty tuple or (nan, nan, nan, nan)
         if self.is_empty:
             return ()
         else:
-            return self.impl['bounds'](self)
+            return tuple(pygeos.bounds(self).tolist())
 
     @property
     def centroid(self):
         """Returns the geometric center of the object"""
-        return geom_factory(self.impl['centroid'](self))
+        return pygeos.centroid(self)
 
     @delegated
     def representative_point(self):
         """Returns a point guaranteed to be within the object, cheaply."""
-        return geom_factory(self.impl['representative_point'](self))
+        return pygeos.representative_point(self)
 
     @property
     def convex_hull(self):
@@ -410,12 +402,12 @@ class BaseGeometry(object):
         The convex hull of a three member multipoint, for example, is a
         triangular polygon.
         """
-        return geom_factory(self.impl['convex_hull'](self))
+        return pygeos.convex_hull(self)
 
     @property
     def envelope(self):
         """A figure that envelopes the geometry"""
-        return geom_factory(self.impl['envelope'](self))
+        return pygeos.envelope(self)
 
     @property
     def minimum_rotated_rectangle(self):
@@ -541,26 +533,11 @@ class BaseGeometry(object):
             raise ValueError(
                 'Cannot compute offset from zero-length line segment')
 
-        if 'buffer_with_params' in self.impl:
-            params = self._lgeos.GEOSBufferParams_create()
-            self._lgeos.GEOSBufferParams_setEndCapStyle(params, cap_style)
-            self._lgeos.GEOSBufferParams_setJoinStyle(params, join_style)
-            self._lgeos.GEOSBufferParams_setMitreLimit(params, mitre_limit)
-            self._lgeos.GEOSBufferParams_setQuadrantSegments(params, res)
-            self._lgeos.GEOSBufferParams_setSingleSided(params, single_sided)
-            return geom_factory(self.impl['buffer_with_params'](self, params, distance))
-
-        if cap_style == CAP_STYLE.round and join_style == JOIN_STYLE.round:
-            return geom_factory(self.impl['buffer'](self, distance, res))
-
-        if 'buffer_with_style' not in self.impl:
-            raise NotImplementedError("Styled buffering not available for "
-                                      "GEOS versions < 3.2.")
-
-        return geom_factory(self.impl['buffer_with_style'](self, distance, res,
-                                                           cap_style,
-                                                           join_style,
-                                                           mitre_limit))
+        return pygeos.buffer(
+            self, distance, quadsegs=resolution, cap_style=cap_style,
+            join_style=join_style, mitre_limit=mitre_limit,
+            single_sided=single_sided
+        )
 
     @delegated
     def simplify(self, tolerance, preserve_topology=True):
@@ -572,31 +549,29 @@ class BaseGeometry(object):
         option is used, the algorithm may produce self-intersecting or
         otherwise invalid geometries.
         """
-        if preserve_topology:
-            op = self.impl['topology_preserve_simplify']
-        else:
-            op = self.impl['simplify']
-        return geom_factory(op(self, tolerance))
+        return pygeos.simplify(
+            self, tolerance, preserve_topology=preserve_topology
+        )
 
     # Binary operations
     # -----------------
 
     def difference(self, other):
         """Returns the difference of the geometries"""
-        return geom_factory(self.impl['difference'](self, other))
+        return pygeos.difference(self, other)
 
     def intersection(self, other):
         """Returns the intersection of the geometries"""
-        return geom_factory(self.impl['intersection'](self, other))
+        return pygeos.intersection(self, other)
 
     def symmetric_difference(self, other):
         """Returns the symmetric difference of the geometries
         (Shapely geometry)"""
-        return geom_factory(self.impl['symmetric_difference'](self, other))
+        return pygeos.symmetric_difference(self, other)
 
     def union(self, other):
         """Returns the union of the geometries (Shapely geometry)"""
-        return geom_factory(self.impl['union'](self, other))
+        return pygeos.union(self, other)
 
     # Unary predicates
     # ----------------
@@ -605,44 +580,36 @@ class BaseGeometry(object):
     def has_z(self):
         """True if the geometry's coordinate sequence(s) have z values (are
         3-dimensional)"""
-        return bool(self.impl['has_z'](self))
+        return bool(pygeos.has_z(self))
 
     @property
     def is_empty(self):
         """True if the set of points in this geometry is empty, else False"""
-        return (self._geom is None) or bool(self.impl['is_empty'](self))
+        return bool(pygeos.is_empty(self))
 
     @property
     def is_ring(self):
         """True if the geometry is a closed ring, else False"""
-        return bool(self.impl['is_ring'](self))
+        return bool(pygeos.is_ring(self))
 
     @property
     def is_closed(self):
         """True if the geometry is closed, else False
 
         Applicable only to 1-D geometries."""
-        if self.geom_type == 'LinearRing':
-            return True
-        elif self.geom_type == 'LineString':
-            if 'is_closed' in self.impl:
-                return bool(self.impl['is_closed'](self))
-            else:
-                return self.coords[0] == self.coords[-1]
-        else:
-            return False
+        return bool(pygeos.is_closed(self))
 
     @property
     def is_simple(self):
         """True if the geometry is simple, meaning that any self-intersections
         are only at boundary points, else False"""
-        return bool(self.impl['is_simple'](self))
+        return bool(pygeos.is_simple(self))
 
     @property
     def is_valid(self):
         """True if the geometry is valid (definition depends on sub-class),
         else False"""
-        return bool(self.impl['is_valid'](self))
+        return bool(pygeos.is_valid(self))
 
     # Binary predicates
     # -----------------
@@ -654,23 +621,23 @@ class BaseGeometry(object):
 
     def covers(self, other):
         """Returns True if the geometry covers the other, else False"""
-        return bool(self.impl['covers'](self, other))
+        return bool(pygeos.covers(self, other))
 
     def covered_by(self, other):
         """Returns True if the geometry is covered by the other, else False"""
-        return bool(self.impl['covered_by'](self, other))
+        return bool(pygeos.covered_by(self, other))
 
     def contains(self, other):
         """Returns True if the geometry contains the other, else False"""
-        return bool(self.impl['contains'](self, other))
+        return bool(pygeos.contains(self, other))
 
     def crosses(self, other):
         """Returns True if the geometries cross, else False"""
-        return bool(self.impl['crosses'](self, other))
+        return bool(pygeos.crosses(self, other))
 
     def disjoint(self, other):
         """Returns True if geometries are disjoint, else False"""
-        return bool(self.impl['disjoint'](self, other))
+        return bool(pygeos.disjoint(self, other))
 
     def equals(self, other):
         """Returns True if geometries are equal, else False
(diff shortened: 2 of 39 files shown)


shapely/coords.py
```diff
@@ -37,121 +37,67 @@ class CoordinateSequence(object):
...
     __p__ = None
 
-    def __init__(self, parent):
-        self.__p__ = parent
-
-    def _update(self):
-        self._ndim = self.__p__._ndim
-        self._cseq = lgeos.GEOSGeom_getCoordSeq(self.__p__._geom)
+    def __init__(self, coords):
+        self._coords = coords
```
jorisvandenbossche, line 41: What exactly to do with this `CoordinateSequence` class is something still to be discussed, and I opened https://github.com/Toblerity/Shapely/issues/984 for that

shapely/geometry/collection.py
```diff
@@ -35,11 +37,18 @@ def __init__(self, geoms=None):
           >>> l = LineString([(52, -1), (49, 2)])
           >>> gc = GeometryCollection([p, l])
         """
-        BaseMultipartGeometry.__init__(self)
         if not geoms:
-            pass
-        else:
-            self._geom, self._ndim = geos_geometrycollection_from_py(geoms)
+            # TODO better empty constructor
+            return pygeos.from_wkt("GEOMETRYCOLLECTION EMPTY")
```
jorisvandenbossche, line 42: For all those constructors, I now use WKT to create an empty geometry, but see https://github.com/pygeos/pygeos/issues/149 to discuss if we need to have a better way for this

shapely/geometry/point.py
```diff
@@ -44,35 +46,65 @@ def __init__(self, *args):
...
+            geom = pygeos.points(*args)
+
+        if not isinstance(geom, Point):
+            raise ValueError("Invalid values passed to Point constructor")
+        return geom
 
     # Coordinate getters and setters
 
+    def __repr__(self):
+        return "<shapely.geometry.Point {} >".format(self.wkt)
```
jorisvandenbossche, line 90: TODO need to move this to the base class

tests/test_create_inconsistent_dimensionality.py
```diff
@@ -38,16 +40,18 @@
 def test_create_from_geojson(geojson):
     with pytest.raises(ValueError) as exc:
         wkt = shape(geojson).wkt
-    assert exc.match("Inconsistent coordinate dimensionality")
+    assert exc.match("Inconsistent coordinate dimensionality|Input operand 0 does not have enough dimensions")
```
jorisvandenbossche, line 43: We get different error message in several places (we should see to what extent we want to preserve them, sometimes they are more informative though)

tests/test_geos_err_handler.py
```diff
@@ -2,11 +2,16 @@
...
+import pygeos
+
 from shapely.geometry import LineString
 from shapely.errors import ReadingError
 from shapely.wkt import loads
 
+from tests.conftest import shapely20_todo
+
 
+@shapely20_todo  # logging is not yet implemented
```
jorisvandenbossche, line 14: This is something we need to check in pygeos how this can be possible, if it's a feature we want to preserve

tests/test_persist.py
```diff
@@ -16,6 +18,7 @@ def test_pickle(self):
         q = pickle.loads(data)
         self.assertTrue(q.equals(p))
 
+    @shapely20_todo  # big_endian is not yet implemented
```
jorisvandenbossche, line 21: In pygeos right now we didn't implement this keyword, but have a `byte_order` keyword instead (matches more closely the terminology of GEOS). The default is to use the machine's endianness, and one can use the keyword to override this.

tests/test_pickle.py
```diff
@@ -19,6 +19,9 @@ def test_pickle_round_trip(cls, coords):
     data = dumps(geom1, HIGHEST_PROTOCOL)
     geom2 = loads(data)
     assert geom2.has_z == geom1.has_z
+    # TODO(shapely-2.0) LinearRing type not preserved in roundtrip
```
jorisvandenbossche, line 22: We were just discussing this in PyGEOS as well -> https://github.com/pygeos/pygeos/pull/190 

There we thought that it might be OK enough to get a LineString back from pickling a LinearRing. However, having the pickling on the python side (still using WKB), we could more easily track the original class as well and ensure it is correctly preserved. 
However, having it in python compared to in the base C extension type has a performance penalty.

tests/test_predicates.py
```diff
@@ -43,7 +46,7 @@ def test_binary_predicate_exceptions(self):
         p2 = [(339, 207), (280, 311), (460, 138), (399, 242), (459, 277),
               (459, 415), (399, 381), (519, 311), (520, 242), (519, 173),
               (399, 450), (339, 207)]
-        self.assertRaises(TopologicalError, Polygon(p1).within, Polygon(p2))
+        self.assertRaises(pygeos.GEOSException, Polygon(p1).within, Polygon(p2))
```
jorisvandenbossche, line 49: Currently, shapely has a bunch of custom error classes in https://github.com/Toblerity/Shapely/blob/master/shapely/errors.py. But in PYGEOS, for now we just have a single GEOSException for errors that are directly raised from one of the GEOS functions.

jorisvandenbossche:
Gentle ping to get some feedback here.

mwtoews:
Thanks @jorisvandenbossche this looks great so far. I've built this PR locally, and the majority of the tests indeed pass. I'll drop a few review comments when I get spare moments...

tests/test_predicates.py
sgillies: Raising pygeos exceptions is just an intermediate state of the API then?

sgillies:
I won't be able to try this out until the weekend. Will report back then.

tests/test_predicates.py
jorisvandenbossche: Yes, the fact that it is a "pygeos" exception is intermediate state. But basically everywhere you see `pygeos` you need to mentally replace it with `shapely` (as in one of the next steps we will move pygeos into shapely and replace the pygeos imports with internal shapely imports). 

So the "real" question here is if we are, long term, fine with having only a generic `shapely.GEOSException`, or if we want to have more fine grained error *classes* (TopologicalError, WKBReadingError, WKTReadingError). 
Note that this is only about the class of the exception, also the generic GEOSException error class will still have the custom, informative error messages.

jorisvandenbossche: Opened https://github.com/Toblerity/Shapely/issues/991 to have a dedicated place to discuss this topic.

shapely/geometry/base.py
```diff
@@ -164,9 +152,7 @@ class BaseGeometry(object):
     # _other_owned : bool
     #     True if this object's GEOS geometry is owned by another as in the
     #     case of a multipart geometry member.
-    __geom__ = EMPTY
     __p__ = None
-    _ndim = None
```
caspervdw, line 169: I think that best all class attributes should be dropped in shapely 2.0. I read these as defaults for object-level attributes. These attributes add unnecessary state to the geometry objects.

As they are all private, I think they can be dropped. `_crs` and `_is_empty` could be replaced by properties calling `pygeos.get_srid` and `pygeos.is_empty` respectively if that is convenient for the implementation.

shapely/geometry/base.py
```diff
@@ -60,34 +62,20 @@ def dump_coords(geom):
...
-    ob.__class__ = getattr(mod, geom_type)
-    ob._geom = g
-    ob.__p__ = parent
-    if lgeos.methods['has_z'](g):
-        ob._ndim = 3
-    else:
-        ob._ndim = 2
-    ob._is_empty = False
-    return ob
+    ob = _DummyGeometry(g)
```
caspervdw, line 77: This seems backwards to me.
In my opinion we should never pass these pointers around as python `int` objects. Do you think we can remove the `geom_factory` completely from shapely 2.0?

caspervdw (commented):
@jorisvandenbossche  Thanks for this enormous effort of combining pygeos and shapely. I left some remarks.

shapely/geometry/base.py
jorisvandenbossche: Yes, this is certainly only temporary. I don't refactor the full of Shapely in a single PR, and thus in other parts of shapely, `geom_factory` is still used together with ctypes. So those changes here with the DummyGeometry hack is just to ensure `geom_factory` keeps working. But the end goal of Shapely 2.0 for sure means that all usage of `geom_factory` is eliminated.

shapely/geometry/base.py
jorisvandenbossche: Indeed, this needs further clean-up. I only removed the ones that I encountered explicitly while refactoring for now, but normally all others can be removed as well (they shouldn't be used anymore, or at least eventually they won't be used anymore). But so for sure, none of those needs to stay.

shapely/geometry/linestring.py
```diff
@@ -39,11 +41,37 @@ def __init__(self, coordinates=None):
...
-            ret = geos_linestring_from_py(coordinates)
-            if ret is not None:
-                self._geom, self._ndim = ret
+        if coordinates is None:
+            # empty geometry
+            # TODO better constructor
+            return pygeos.from_wkt("LINESTRING EMPTY")
+        elif isinstance(coordinates, LineString):
+            if type(coordinates) == LineString:
+                # TODO should we clone it?
```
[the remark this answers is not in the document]
jorisvandenbossche: Just checked, eg copying a tuple (an immutable object) also returns the same object. So that seems like a common pattern. Will remove the comment then, and simply return the original object.

shapely/geometry/polygon.py
```diff
@@ -176,9 +206,8 @@ class Polygon(BaseGeometry):
 
     _exterior = None
     _interiors = []
```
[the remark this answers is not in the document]
jorisvandenbossche: I am not sure what this `_interiors` attribute was used for originally, but in any case it's not used anymore now, so will simply remove.

jorisvandenbossche:
@caspervdw Thanks for the review!

jorisvandenbossche:
@caspervdw a build related question: pygeos doesn't seem to picking up the correct GEOS version (from travis ci output: *"ImportError: The shapely GEOS version (3.7.3-CAPI-1.11.3 b50468f) is incompatible with the GEOS version PyGEOS was compiled with (3.5.0-CAPI-1.9.0)"*) 
I am actually not fully sure where it is getting this GEOS version, but in any case there are multiple GEOS versions installed (and cached) on the travis image. I prepended the one to be used to the PATH, but doesn't seem to have fixed it.

jorisvandenbossche:
Note: the tests on Travis are now passing for the Python 3.8 build. For older versions only one MultiLineString.interpolate test is failing that I can't directly reproduce locally.

shapely/geometry/polygon.py
```diff
@@ -72,13 +109,17 @@ def __setstate__(self, state):
     @property
     def is_ccw(self):
         """True is the ring is oriented counter clock-wise"""
+        # TODO
```
caspervdw, line 112: See https://github.com/pygeos/pygeos/pull/201

shapely/coords.py
caspervdw: I have been reading up on https://github.com/Toblerity/Shapely/issues/984 and I tend to go with this implementation for this PR (making `self._coords` a ndarray directly)

However I think that sticking a full duplicate of `coords` to the object by default is a bit too much. Also `_coords` won't be initialized properly in all construction paths.

Within this approach (I actually like the lazy CoordinateSequency object approach better) I think it is best to use a pattern like this:

```
@property
def coords(self):
    try:
        coords = getattr(self, "_coords")
    except AttributeError:
        coords = pygeos.get_coordinates(self)
        coords.flags.writeable = False
        self._coords = coords
    return coords
```

jorisvandenbossche: > However I think that sticking a full duplicate of coords to the object by default is a bit too much. 

Note that this CoordinateSequence is not constructed by default when creating a geometry object. It will only be created once a user accesses the `geom.coords` attribute. 

Now, with the current code, each time the user does `geom.coords`, it will create a new CoordinateSequence object containing a new array, so making this "cached" is certainly a good idea (as well as making the array not writable)

tests/test_geos_err_handler.py
mwtoews: I'd say all error messages from GEOS (like the one below) should be converted to exception messages, similar to https://github.com/Toblerity/Shapely/issues/991#issuecomment-699147566 . I'd assume it would be difficult to capture an error message in pygeos, but also emit the same error message to the logger handler below. I wouldn't expect many other folks to be relying on these log capture methods.

shapely/geometry/base.py
```diff
@@ -838,32 +765,19 @@ class GeometrySequence(object):
     _ndim = None
```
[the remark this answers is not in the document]
mwtoews: `__p__` is used below, but I think this would be a good time to rename this private variable to `_parent`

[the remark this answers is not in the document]
jorisvandenbossche: Specifically for this `GeometrySequence` class, we also still need to see what to do with it (keep it or not?)

shapely/geometry/base.py
```diff
@@ -154,75 +142,42 @@ class BaseGeometry(object):
...
-                log.exception("Failed to delete GEOS geom")
+    @property
+    def _geom(self):
+        return self._ptr
 
-        self._is_empty = True
-        self.__geom__ = val
+    @property
+    def __geom__(self):
+        return self._ptr
```
mwtoews, line 161: `__geom__` is another ["dunder"](https://docs.python.org/3/reference/lexical_analysis.html#reserved-classes-of-identifiers) I think we could remove for shapely-2.0, as it's not a system-defined name. I see that this holds the same value as `_geom` and `_ptr` (from pygeos).

jorisvandenbossche: Yes, the attributes giving read access to the pointer (`_geom`, `__geom__`, and now also `_ptr`) can use a clean-up. But for now, those are kept here in any case for compatibility with the rest of shapely that is not yet adapted.

So we should probably decide which one to keep (eg `_geom`) and deprecate the other already in Shapely 1.8 (eg `__geom__`)

(might also be worth a separate issue, as it is not "solvable" for this PR)

shapely/ops.py
```diff
@@ -121,37 +122,15 @@ def cascaded_union(self, geoms):
...
-                geoms = geoms.geoms
-        except TypeError:
-            geoms = [geoms]
-            L = 1
-        subs = (c_void_p * L)()
-        for i, g in enumerate(geoms):
-            subs[i] = g._geom
-        collection = lgeos.GEOSGeom_createCollection(6, subs, L)
-        return geom_factory(lgeos.methods['cascaded_union'](collection))
+        return pygeos.union_all(geoms, axis=None)
```
mwtoews, line 125: Is pygeos calling (deprecated) `GEOSUnionCascaded_r`? This is the same return for `unary_union` below.

shapely/ops.py
```diff
@@ -121,37 +122,15 @@ def cascaded_union(self, geoms):
...
-                geoms = geoms.geoms
-        except TypeError:
-            geoms = [geoms]
-            L = 1
-        subs = (c_void_p * L)()
-        for i, g in enumerate(geoms):
-            subs[i] = g._geom
-        collection = lgeos.GEOSGeom_createCollection(6, subs, L)
-        return geom_factory(lgeos.methods['unary_union'](collection))
+        return pygeos.union_all(geoms, axis=None)
```
mwtoews, line 133: similar to comment above, is pygeos calling `GEOSUnaryUnion_r` here?

jorisvandenbossche: > similar to comment above, is pygeos calling GEOSUnaryUnion_r here?

Yes, it's a potential naming issue on the pygeos side, but `union_all` is basically `unary_union` (the main difference being that it supports an `axis` keyword in case you have a multidimensional array to union along a specific axis). But we should certainly consider addng a `unary_union` function as well to be compatible with Shapely's current naming.

tests/test_point.py
```diff
@@ -139,7 +142,9 @@ def test_coords(self):
 def test_empty_point_bounds():
     """The bounds of an empty point is an empty tuple"""
     p = Point()
-    assert p.bounds == ()
+    # TODO keep this change or not?
+    assert len(p.bounds) == 4
+    assert all(math.isnan(v) for v in p.bounds)
```
mwtoews, line 147: Why does just `point empty` have four nan bounds, and not other empty geometry types (that return an empty tuple `()`)? Otherwise, yes this idea is worth considering.

shapely/ops.py
jorisvandenbossche: > Is pygeos calling (deprecated) GEOSUnionCascaded_r? This is the same return for unary_union below.

No, it's actually calling GEOSUnaryUnion_r, and not GEOSUnionCascaded_r (GEOSUnionCascaded_r is simply not implemented in pygeos). Note that this is similar to what Shapely does: for GEOS >= 3.3, `lgeos.methods['cascaded_union']` is actually an alias of `unary_union`, only for older GEOS versions it is using GEOSUnionCascaded (but pygeos doesn't support such old GEOS version):

https://github.com/Toblerity/Shapely/blob/73626bce8e99b3f251ad874a41d97d2384e734ce/shapely/geos.py#L838

Two potential discussion items out of this:

- What's the minimal supported GEOS version we want for Shapely 2.0? Shapely starts testing for GEOS 3.5 and up (the same for pygeos), but actually has code in shapely/geos.py going back to GEOS 3.0

- Do we actually want to deprecate the python `cascaded_union` function, instead of keeping it around as alias of `unary_union`? (although this is quite independent from all shapely 2.0 / pygeos refactors, but the deprecations in 1.8 could be an opportunity to deprecate this as well, if we want. But I have the impression it is still used quite a lot)

(again might be worth turning into dedicated issues)

tests/test_point.py
jorisvandenbossche: Ah, that's an oversight / something to clean-up. For the base class method, I left the current behaviour but added a TODO comment:

```diff
     @property
     def bounds(self):
         """Returns minimum bounding region (minx, miny, maxx, maxy)"""
+        # TODO return empty tuple or (nan, nan, nan, nan)
         if self.is_empty:
             return ()
         else:
-            return self.impl['bounds'](self)
+            return tuple(pygeos.bounds(self).tolist())
```

while this property is overridden by Point, where I actually removed this empty tuple return:

```diff
    @property
    def bounds(self):
        """Returns minimum bounding region (minx, miny, maxx, maxy)"""
-        try:
-            xy = self.coords[0]
-        except IndexError:
-            return ()
-        return (xy[0], xy[1], xy[0], xy[1])
+        # TODO keep return type of tuple, or just numpy array?
+        return tuple(pygeos.bounds(self).tolist())
```

So for sure, we need to be consistent about this, whatever decision we make. 

(and actually, with the new implementation, we can remove the override in the Point subclass, as the base class implementation should work fine for Points as well)

jorisvandenbossche:
@mwtoews thanks a lot for the review / questions !

mwtoews:
> Note: the tests on Travis are now passing for the Python 3.8 build. For older versions only one MultiLineString.interpolate test is failing that I can't directly reproduce locally.

I can reproduce this locally with GEOS 3.7.1:
```python
from shapely import wkt
multiline = wkt.loads('MULTILINESTRING ((0 0, 2 0), (3 0, 3 6))')
multiline.interpolate(0).wkt

import pygeos
multiline = pygeos.Geometry('MULTILINESTRING ((0 0, 2 0), (3 0, 3 6))')
pygeos.line_interpolate_point(multiline, 0)
```
this should return `POINT(0 0)`, but pygeos returns `POINT EMPTY`. See [src/ufuncs.c](https://github.com/pygeos/pygeos/blame/7a2798ab8ef43acf59e2023b8cfcb49e6d9384d3/src/ufuncs.c#L341-L365) for the conditional implementation.

I'm not sure how to describe this issue yet, but I'm sure @caspervdw could shed some light. It seems that the logic does not support linear referencing with MutiLineStrings?

jorisvandenbossche:
@mwtoews thanks for the catch!

shapely/geometry/base.py
```diff
@@ -748,11 +679,8 @@ def interpolate(self, distance, normalized=False):
         If the normalized arg is True, the distance will be interpreted as a
         fraction of the geometry's length.
         """
-        if normalized:
-            op = self.impl['interpolate_normalized']
-        else:
-            op = self.impl['interpolate']
-        return geom_factory(op(self, distance))
+        # TODO spelling normalize(d)
+        return pygeos.line_interpolate_point(self, distance, normalize=normalized)
```
mwtoews, line 683: See https://github.com/pygeos/pygeos/pull/209

tests/test_geos_err_handler.py
jorisvandenbossche: Opened https://github.com/Toblerity/Shapely/issues/998 for a discussion on logging.

shapely/geometry/base.py
jorisvandenbossche: Opened https://github.com/Toblerity/Shapely/issues/999 for this `__geom__` / `_geom` dicussion

shapely/ops.py
jorisvandenbossche: Opened https://github.com/Toblerity/Shapely/issues/1000 for the minimum supported GEOS version and https://github.com/Toblerity/Shapely/issues/1001 for the unary_union/cascaded_union aliasing

shapely/ops.py
jorisvandenbossche: Relevant issue about `union_all` vs `unary_union` is https://github.com/pygeos/pygeos/issues/194

tests/test_point.py
jorisvandenbossche: Removed the override in `Point` class, and leaving the discussion to changy `empty.bounds` for a separate issue

jorisvandenbossche:
With the upstream fix in pygeos for `line_interpolate_point` (thanks @caspervdw!), Travis is green now ;)

jorisvandenbossche:
Appveyor is also passing, except for 1 test in the Python 3.8 build:

```
================================== FAILURES ===================================
___________________________ test_pickle_persistence ___________________________
    @requires_geos_342
    def test_pickle_persistence():
        """
        Don't crash trying to use unpickled GEOS handle.
        """
        tree = STRtree([Point(i, i).buffer(0.1) for i in range(3)])
        pickled_strtree = pickle.dumps(tree)
        unpickle_script_file_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "unpickle-strtree.py")
        proc = subprocess.Popen(
            [sys.executable, str(unpickle_script_file_path)],
            stdin=subprocess.PIPE,
        )
        proc.communicate(input=pickled_strtree)
        proc.wait()
>       assert proc.returncode == 0
E       assert 1 == 0
E        +  where 1 = <subprocess.Popen object at 0x000001F9904BD520>.returncode
tests\test_strtree.py:102: AssertionError
---------------------------- Captured stderr call -----------------------------
  File "C:\projects\shapely\tests\unpickle-strtree.py", line 8, in <module>
  File "C:\projects\shapely\shapely\geometry\__init__.py", line 4, in <module>
  File "C:\projects\shapely\shapely\geometry\base.py", line 18, in <module>
    import pygeos
  File "c:\python38-x64\lib\site-packages\pygeos\__init__.py", line 1, in <module>
    from .lib import GEOSException  # NOQA
ImportError: DLL load failed while importing lib: The specified module could not be found.
```

I am not very familiar with such Windows issues, but so within subprocess call, it is not finding the GEOS library.

mwtoews:
I've been checking the performance difference with this PR (with pygeos master) and the current shapely master, both with libgeos master '3.8.0'. These were tested in Python 3.7.7. Here is an excerpt of my findings:
```
from shapely.geometry import Point, Polygon, MultiLineString
from shapely.ops import linemerge
from shapely import wkt

# this PR
%timeit Point(1, 2)
13.6 µs ± 355 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
# shapely master
%timeit Point(1, 2)
5.78 µs ± 67.4 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

# this PR
%timeit wkt.loads('LINESTRING(2.4 4.1, 2.4 3, 3 3)')
4.66 µs ± 182 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
# shapely master
%timeit wkt.loads('LINESTRING(2.4 4.1, 2.4 3, 3 3)')
10.9 µs ± 348 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

# this PR
%timeit Point(1, 2).buffer(3)
73.6 µs ± 670 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
# shapely master
%timeit Point(1, 2).buffer(3)
49.7 µs ± 1.74 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

p1 = Point(0.0, 0.0)
p2 = Point(-1.0, -1.0)
# this PR
%timeit p1.disjoint(p2)
6.49 µs ± 63.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
# shapely master
%timeit p1.disjoint(p2)
1.86 µs ± 36.9 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

lines = MultiLineString([((0, 0), (1, 1)), ((2, 0), (2, 1), (1, 1))])
# this PR
%timeit linemerge(lines)
22.7 µs ± 1.28 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
# shapely master
%timeit linemerge(lines)
9.6 µs ± 368 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

poly = Polygon([(0,1), (1, 2), (2, 1), (1, 0), (0, 1)])
# this PR
%timeit poly.minimum_rotated_rectangle
636 µs ± 21.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
# shapely master
%timeit poly.minimum_rotated_rectangle
289 µs ± 13 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
I'm generally seeing a bit of a slowdown with pygeos compared to ctypes used with shapely master. Any ideas on why this is? Should we be concerned? (Also, did I do these tests correctly, or is there something funny on my end?)

jorisvandenbossche:
@mwtoews thanks a lot for checking this!

I profiled some of those cases, and some observations:

- The `Point()` constructor is slower because of the overhead in `_wrap_construct_ufunc`, which wraps the input numbers/arrays into a single array of the appropriate shape (https://github.com/pygeos/pygeos/blob/e475152600cd356130264c44723961a19a2d7280/pygeos/creation.py#L19-L27). For a single Point, this gives a large (*relative*)  overhead, though. 
  Using the directly exposed ufunc instead of the more user-friendly wrapper, and passing it the 2-element array, you can see the difference:

    ```
    In [7]: %timeit pygeos.points(1, 2)
    14.9 µs ± 2.61 µs per loop (mean ± std. dev. of 7 runs, 100000 loops each)
    
    In [11]: %timeit pygeos.lib.points(np.array([1, 2]))
    2.17 µs ± 114 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
    ```

  When working with larger arrays, the overhead of the wrapper is minimal, but to construct a single Point, this overhead becomes significant. So the `points()` function is simply not really optimized for this case of creating a single point. 
  If we want this, I think it is certainly possible to write a "single-point constructor" that optimizes for this use case.

* `wkt.loads` is already faster with this branch

* The buffer operation (`Point(1, 2).buffer(3)`, but only profiling the actual buffer) -> it seems that here it's the actual call into C that takes longer. So for that would need to take a closer look / use a different profiler. 

* The `p1.disjoint(p2)` case: also here the call into C is actually already slower as Shapely (eg timing `%timeit pygeos.lib.disjoint(p1, p2)` instead of the wrapper function). I suppose that here the ufunc machinery gives some overhead (also in C, it's not a plain call to GEOS, but done inside a inner loop of the ufunc. For scalars (0-dim arrays), this will always give *some* overhead. 
  In addition, some of the overhead of the python function also comes from the `multithreading_enabled` decorator. Again, this overhead is insignificant when working with arrays, but not for scalar geometries.

* `linemerge` -> this implementation is not yet changed (so I suppose the slowdown we see comes from other APIs being used inside Shapely's linemerge function)

* `poly.minimum_rotated_rectangle` -> this implementation relies on doing an `affine_transform`, and this is also something that is not yet updated to use pygeos (I would need to take a look into `affinity.py` first to be sure whether the use of pygeos will be able to make this faster, but I suppose we can use the vectorized transform functionality already included in pygeos to apply the transform on all coordinates of a Geometry, instead of the iteration through all parts and all points in python as is currently done in affinity.py's `affine_transform`. So this should be able to speed-up this functionality).
  In addition, GEOS actually has a `GEOSMinimumRotatedRectangle` since GEOS 3.6, so that is something we should add bindings for and can use that if available instead of our own implementation.

**In summary**:

So in general: yes, there is a slowdown for methods/properties of scalar Geometry objects. 
For some things (like the constructors of the different geometry classes), we can optimize this by spending some effort in having an efficient single-geometry version (this can probably be done in the class constructor itself, eg by using `pygeos.lib.points` instead of `pygeos.points`). 
But for the bulk of the ufuncs (the spatial predicates, operations, etc. I.e. those that directly wrap one of the `GEOS<..op..>_r` functions), this is the consequence of having general ufunc (array-aware) functions. This ufunc machinery will always have *some* overhead, and for single geometries this overhead can be noticeable (and seems to be larger as the benefit from not using ctypes).

But the question is then: **do we care about this small overhead for working with single geometries?** 
And to be clear, I care about performance (and where *easily* possible, we should fix this overhead). But 1) it is still about very small *absolute* numbers (eg for the `disjoint` method from 2 -> 8 µs), and 2) such a slowdown typically only becomes a problem if you are doing it *many times*. And that is exactly where pygeos/Shapely 2.0 will provide a big benefit with the array APIs (but with the caveat: the user will need to rewrite their loops to make use of the vectorized functions to get this benefit).

mwtoews:
Thanks for the thorough followup @jorisvandenbossche ! I should have picked some of these examples a bit more carefully, as indeed some are not implemented outside shapely.

I was generally suspecting there was a bit of additional overhead with ufuncs, which is expected as that's the point of pygeos. And obviously there are incredible performance gains when these constructors, operators (etc.) are used in a vectorized fashion (i.e. examples from https://caspervdw.github.io/Introducing-Pygeos/ are still valid).

Do we care about small overhead for working with single geometries? Hopefully not. Before GeoPandas, a typical workflow that required processing multiple geometries would use for-loops and lists of geometries. However, where performance is important, folks should now be migrating their workflow to use vectorized approaches.

As for specific cases of slowdowns, we can evaluate these as needed (if at all). Single geometry constructors could perhaps be investigated later. The priority for this PR is to get the expected results, which progressing really well!

Pending tasks are to look into the Windows AppVeyor fail (I could try), and I'd like to look into the `@shapely20_todo` marked tests a bit more. So hopefully another week or two??

tests/test_linear_referencing.py
```diff
@@ -57,6 +57,7 @@ def test_line2_interpolate(self):
 
     @unittest.skipIf(geos_version < (3, 2, 0), 'GEOS 3.2.0 required')
     def test_multiline_interpolate(self):
+        print(self.multiline.interpolate(0.5))
```
jorisvandenbossche, line 60: Need to clean-up

jorisvandenbossche:
> Pending tasks are to look into the Windows AppVeyor fail (I could try)

Yes, if someone could help looking into that, that would be very helpful! 

> and I'd like to look into the @shapely20_todo marked tests a bit more.

Listing them here to get an easier overview:

* `tests/test_create_inconsistent_dimensionality.py::test_create_from_wkt`

  This is about WKT parsing returning 2 or 3 dimensions when parsing an inconsistent string. The strange thing is that I can't reproduce the difference between shapely and pygeos in an *interactive console*:

    ```python
    from shapely import wkt 
    wkt_geom = 'LINESTRING (1 1 1, 2 2)' 
    geom = wkt.loads(wkt_geom)
    print(geom.wkt)
    ```

  The snippet above extracted from the test returns for me `'LINESTRING Z (1 1 1, 2 2 0)'` in an interactive console (even in an interactive console with an older, released Shapely version), but when running it the test, it gives `'LINESTRING (1.0000000000000000 1.0000000000000000, 2.0000000000000000 2.0000000000000000)'`. 
  Is there something in the test setup that might be influencing this? (also the trimming of trailing zeros is different ...)

* `tests/test_geos_err_handler.py::test_error_handler_exception/test_info_handler`

  I had already opened https://github.com/Toblerity/Shapely/issues/998 to further discuss the topic of logging.

* `tests/test_geos_err_handler.py::test_error_handler_for_bytes`

  Currently, PyGEOS accepts both strings as bytes in `from_wkt` / `wkt.loads`, while there is a test to check that passing bytes raises an error. But it is maybe fine to change this test and accept bytes as well as pygeos is doing?

* `tests/test_persist.py::test_wkb_dumps_endianness`

  Shapely has a `big_endian=True/False` keyword in ``wkb.dump(s)``, while pygeos uses the `byte_order=0/1` matching GEOS' terminology. 
  For now I skipped the tests, but it should actually be easy to adapt and forward the `big_endian` keyword as `byte_order` to the pygeos method. 
  The only remaining question long term is then if we keep it as `big_endian` keyword in Shapely, or add the `byte_order` as alias, or eventually replace it with `byte_order`?

* ``tests/test_singularity.py::PolygonTestCase::test_polygon_3``

  This case tests the creation of ` Polygon([(1,1), (1,1), (1,1)])`, something that fails with pygeos (complaining about needing 4 coordinates to create a linearring, but when the coordinates are not all equal, this is not required). 
  I think this is something we can fix in the pygeos C code, opened https://github.com/pygeos/pygeos/issues/213 for this.

Apart from those, there are a few tests that changed (eg to correct the WKT of empty geometries), and some error messages / exception class changes (and for that we have https://github.com/Toblerity/Shapely/issues/991 to discuss this)

jorisvandenbossche:
Any further comments on this? 
I would like to get this merged, so I can more easily work on follow-up PRs. In my comment above, there is a list of "skipped" tests, I think it would be fine to handle those in follow-ups (I can make issues for them to ensure we don't forget it).

Further, there is the Appveyor failure (see https://github.com/Toblerity/Shapely/pull/983#issuecomment-707706532). I don't have an easy way to reproduce/investigate this. But also, since it's about calling from subprocess, it's maybe also not a blocker for this PR.

mwtoews:
> Listing them here to get an easier overview:

Thanks, most of these are fine to mark as TODO items for now. Here are a few specific responses:

> * `tests/test_create_inconsistent_dimensionality.py::test_create_from_wkt`
>   ... The strange thing is that I can't reproduce the difference between shapely and pygeos in an _interactive console_

See https://github.com/Toblerity/Shapely/blob/86ad1031cbcceaaa615092c1af4c55b7c8efc820/tests/__init__.py#L27-L28 which resets the defaults for GEOS' WKTWriter, which has defaults `trim = False` and `rounding_precision = -1`. That's why pytest behaves different than an interactive console.

> * `tests/test_geos_err_handler.py::test_error_handler_for_bytes`
>   Currently, PyGEOS accepts both strings as bytes in `from_wkt` / `wkt.loads`, while there is a test to check that passing bytes raises an error. But it is maybe fine to change this test and accept bytes as well as pygeos is doing?

This is fine with me.

> * `tests/test_persist.py::test_wkb_dumps_endianness`
>   Shapely has a `big_endian=True/False` keyword in `wkb.dump(s)`, while pygeos uses the `byte_order=0/1` matching GEOS' terminology.
>   For now I skipped the tests, but it should actually be easy to adapt and forward the `big_endian` keyword as `byte_order` to the pygeos method.
>   The only remaining question long term is then if we keep it as `big_endian` keyword in Shapely, or add the `byte_order` as alias, or eventually replace it with `byte_order`?

I made this call while implementing WKTWriter for shapely, as the meaning of this boolean property `big_endian=True/False` is more clear than `byte_order=0/1`. However, I don't have any objections if this is made an alias to align closer to GEOS' terminology.

mwtoews:
> Further, there is the Appveyor failure (see [#983 (comment)](https://github.com/Toblerity/Shapely/pull/983#issuecomment-707706532)). I don't have an easy way to reproduce/investigate this. But also, since it's about calling from subprocess, it's maybe also not a blocker for this PR.

I debugged this directly on the AppVeyor VM, and this is an odd one. It is only an issue with Python 3.8.0 (both x86 and x64), and the issue happens when "import pygeos" is attempted outside the shapely pytest environment (including subprocess). I can break the shapely pytest environment by importing pygeos before shapely. However, I'll agree that this is not a blocker for now.

mwtoews (approved):
Thanks for the huge effort @jorisvandenbossche and pygeos team! I'll let you merge this one, as you may want to squash and merge with a concise commit message.

jorisvandenbossche:
> See https://github.com/Toblerity/Shapely/blob/86ad1031cbcceaaa615092c1af4c55b7c8efc820/tests/__init__.py#L27-L28 which resets the defaults for GEOS' WKTWriter, which has defaults `trim = False` and `rounding_precision = -1`. That's why pytest behaves different than in an interactive console.

Ah, thanks, that clarifies ;)
Would it be fine to remove that? The comment indicates this was done for being able to pass the tests for all GEOS versions, but it seems only for GEOS versions > 3.3, and now we don't test anymore with older versions.

OK, will update the few small remaining items that can easily be done here (fixing the WKT tests, endianness, etc), and then open follow-up issues for the other remaining items.

mwtoews:
> Would it be fine to remove that? The comment indicates this was done for being able to pass the tests for all GEOS versions, but it seems only for GEOS versions > 3.3, and now we don't test anymore with older versions.

This was an old hack to run the tests with the older library and compare similar WKT, so yes we can update the WKT and remove it. I can volunteer to do this.

jorisvandenbossche:
Skipped the problematic tests on Appveyor, and updated the WKT/WKB related tests (to no longer be skipped).

> I made this call while implementing WKTWriter for shapely, as the meaning of this boolean property big_endian=True/False is more clear than byte_order=0/1. However, I don't have any objections if this is made an alias to align closer to GEOS' terminology.

For now I added `big_endian=True/False` back as alias for byte_order, we can later discuss if we want to keep both as aliases or choose one long term.

jorisvandenbossche:
Indeed, already removed it ;)

tests/test_pickle.py
jorisvandenbossche: This is tracked by https://github.com/Toblerity/Shapely/issues/988

jorisvandenbossche:
OK, going to merge this, so we can further work in separate follow-up PRs. Additional comments on the changes here are of course still welcome as well, and I will open issues to track the currently remaining discussion items. Thanks all for the review!

tests/test_point.py
jorisvandenbossche: Opened https://github.com/Toblerity/Shapely/issues/1023 to discuss this

jorisvandenbossche:
I created some more follow-up issues based on the review discussions in this PR:

- Shapely 2.0: Check performance for scalar methods/attributes #1021
- Shapely 2.0: deprecate BaseGeometry() and EmptyGeometry() ? #1022
- API: return value for the bounds of an empty geometry? #1023
- BUG: failure when importing shapely/pygeos in subprocess (outside pytest) #1024

And other outstanding issues for which I already created issues earlier:

- API: exposing CoordinateSequence in Shapely 2.0? #984
- Pickling of LinearRings #988
- API: custom classes for GEOS exceptions #991
- Shapely 2.0: support for logging ? #998
- API: read-only access to GEOS pointer (`__geom__`, `_geom`, `_ptr`) #999

PR #995, opened 2020-10-02
Return empty CoordinateSequence instead of list

### LINKED ISSUE #959: LineString().coords should return an empty CoordinateSequence, not list
Reported at https://github.com/Toblerity/Shapely/pull/955#discussion_r467715843.

chadwhawkins:
Resolves #959

### DIFF
--- a/tests/test_linestring.py
+++ b/tests/test_linestring.py
@@ -206,23 +206,20 @@ class LineStringTestCase(unittest.TestCase):
 
     @unittest.skipIf(not numpy, 'Numpy required')
     def test_numpy_linestring_coords(self):
-        import numpy as np
         from numpy.testing import assert_array_equal
 
         line = LineString(((1.0, 2.0), (3.0, 4.0)))
-        expected = np.array([[1.0, 2.0], [3.0, 4.0]])
+        expected = numpy.array([[1.0, 2.0], [3.0, 4.0]])
 
         # Coordinate sequences can be adapted as well
-        la = np.asarray(line.coords)
+        la = numpy.asarray(line.coords)
         assert_array_equal(la, expected)
 
     @unittest.skipIf(not numpy, 'Numpy required')
     def test_numpy_empty_linestring_coords(self):
-        import numpy as np
-
         # Check empty
         line = LineString([])
-        la = np.asarray(line.coords)
+        la = numpy.asarray(line.coords)
 
         self.assertEqual(la.shape[0], 0)
 
--- a/tests/test_linestring.py
+++ b/tests/test_linestring.py
@@ -205,7 +205,7 @@ class LineStringTestCase(unittest.TestCase):
         assert_array_equal(la, expected)
 
     @unittest.skipIf(not numpy, 'Numpy required')
-    def test_numpy_coords(self):
+    def test_numpy_linestring_coords(self):
         import numpy as np
         from numpy.testing import assert_array_equal
 
@@ -215,6 +215,10 @@ class LineStringTestCase(unittest.TestCase):
         # Coordinate sequences can be adapted as well
         la = np.asarray(line.coords)
         assert_array_equal(la, expected)
+
+    @unittest.skipIf(not numpy, 'Numpy required')
+    def test_numpy_empty_linestring_coords(self):
+        import numpy as np
 
         # Check empty
         line = LineString([])
--- a/tests/test_linestring.py
+++ b/tests/test_linestring.py
@@ -87,6 +87,16 @@ def test_from_numpy():
     assert line.coords[:] == [(1.0, 2.0), (3.0, 4.0)]
 
 
+def test_numpy_empty_linestring_coords():
+    np = pytest.importorskip("numpy")
+
+    # Check empty
+    line = LineString([])
+    la = np.asarray(line.coords)
+
+    assert la.shape == (0,)
+
+
 def test_from_invalid_dim():
     with pytest.raises(ValueError, match="at least 2 coordinate tuples"):
         LineString([(1, 2)])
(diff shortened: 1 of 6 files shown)


sgillies:
@chadwhawkins really nice work! I'm tagging @jorisvandenbossche for review too since this would land in shapely 2.0.

jorisvandenbossche:
@chadwhawkins thanks for working on this!

@sgillies it's not something we also want to change for Shapely 1.8? (you added the 1.8 milestone on the issue)
I am a bit ambivalent on doing it already in 1.8 versus keeping the change for 2.0 (for me it is fine for 1.8 as well). But if it's for 2.0, we need to change the target branch of the PR.

tests/test_linestring.py
```diff
@@ -69,10 +70,12 @@ def test_from_generator():
 def test_from_empty():
     line = LineString()
     assert line.is_empty
+    assert isinstance(line.coords, CoordinateSequence)
     assert line.coords[:] == []
 
     line = LineString([])
     assert line.is_empty
+    assert isinstance(line.coords, CoordinateSequence)
     assert line.coords[:] == []
```
jorisvandenbossche, line 79: Could you also add a test for doing `np.array(line.coords)` for an empty LineString? 
You can add it to one of the tests below that already require numpy (and the same for Point and LinearRing)

chadwhawkins: No problem, done. Also corrected an issue with a test class name and a test constructing a LineString when it should have been a LinearRing.

tests/test_linestring.py
```diff
@@ -216,6 +216,12 @@ def test_numpy_coords(self):
         la = np.asarray(line.coords)
         assert_array_equal(la, expected)
 
+        # Check empty
+        line = LineString([])
+        la = np.asarray(line.coords)
+
+        self.assertEqual(la.shape[0], 0)
```
sgillies, line 223: @chadwhawkins would you be willing to make this assertion into a separate test function? Like a `test_empty_linestring_coords`? Test isolation is super important to us. We don't want a failure at line 217 to block executing the not related assertion at line 223.

sgillies (changes requested):
@chadwhawkins thank you very much! I have one specific change request and there are a few other places where it should be applied. cc @jorisvandenbossche

tests/test_linestring.py
sgillies: I admit that shapely is not the best example historically of isolated test cases, but we're aspiring to do better :smile:

jorisvandenbossche:
@sgillies did you see my question above(https://github.com/Toblerity/Shapely/pull/995#issuecomment-710010645) about targetting 1.8 vs 2.0 for this PR?

sgillies:
@jorisvandenbossche I missed that. Yes, I agree :+1:

chadwhawkins:
@sgillies I went through and put those new assertions into their own tests. Please let me know if that matches what you had in mind.

tests/test_linestring.py
```diff
@@ -216,6 +216,10 @@ def test_numpy_coords(self):
         la = np.asarray(line.coords)
         assert_array_equal(la, expected)
 
+    @unittest.skipIf(not numpy, 'Numpy required')
+    def test_numpy_empty_linestring_coords(self):
+        import numpy as np
```
sgillies, line 221: This line can be deleted if we use `numpy` instead of `np` on line 225.

tests/test_polygon.py
```diff
@@ -104,6 +104,10 @@ def test_linearring_coords_numpy():
     expected = np.asarray([(0.0, 0.0), (0.0, 1.0), (1.0, 1.0), (0.0, 0.0)])
     assert_array_equal(ra, expected)
 
+
+def test_numpy_empty_linearring_coords():
+    np = pytest.importorskip("numpy")
```
sgillies, line 109: Yes, this is fine, too. It seems like standardizing on one form of the other of skipping without numpy would have benefits, but I don't want to belabor this PR. I'm okay with a mix of `importorskip` and `skipif(not numpy)`. @jorisvandenbossche do you have a strong preference?

chadwhawkins: @sgillies agreed that it's certainly not ideal, basically I just tried to follow the usage elsewhere in the file. Perhaps in a separate PR I could work on transitioning all of the tests to pytest style?

jorisvandenbossche: I have a preference to use pytest whenever we write new tests. 

Doing a large conversion to pytest style might also be welcome, but might maybe be most useful if it directly refactors the tests (like splitting up more, better coverage, .. like eg what I did in https://github.com/Toblerity/Shapely/pull/963). There are also a lot of `self.assertEqual(..)`s to clean up ..

tests/test_linestring.py
```diff
@@ -202,17 +205,24 @@ def test_numpy(self):
...
+        la = numpy.asarray(line.coords)
         assert_array_equal(la, expected)
 
+    @unittest.skipIf(not numpy, 'Numpy required')
+    def test_numpy_empty_linestring_coords(self):
+        # Check empty
+        line = LineString([])
+        la = numpy.asarray(line.coords)
+
+        self.assertEqual(la.shape[0], 0)
```
jorisvandenbossche, line 224: Can you assert here the full shape instead of only the first element? Is it actually (0,) or (0, 2) ?

jorisvandenbossche, line 224:
```suggestion
        assert la.shape[0] == 0
```

(for *new* tests, I have been using pytest style assertions lately)

tests/test_polygon.py
jorisvandenbossche: BTW, specifically for numpy, it's also going to be a requied dependency for 2.0, so then we won't need those skips anymore.

chadwhawkins:
@jorisvandenbossche @sgillies I think I've addressed your comments, mind having another look?

jorisvandenbossche:
Thanks!

It can maybe be left for a follow-up, but should we ensure the shape of the coordinates array for empty geometries has a consistent dimension as for non-empty geometries (so `(0, 2)` instead of `(0,)`) ?

sgillies (approved):
Thanks!

PR #997, opened 2020-10-06
Fix doctests, run in CI

### LINKED ISSUE #992: CI: run doctests
Possible follow-up on https://github.com/Toblerity/Shapely/pull/979. With `pytest shapely --doctest-modules` we could actually run the doctests on CI (but that needs some fixes first, as there are some failures at the moment)

chadwhawkins:
Resolves #992.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,6 +100,7 @@ script:
   - export DYLD_LIBRARY_PATH=$HOME/geosinstall/geos-$GEOSVERSION/lib
   - python -c "from shapely.geos import geos_version; print(geos_version)"
   - python -m pytest --cov shapely --cov-report term-missing "${SPEEDUPS_FLAG}"
+  - python -m pytest shapely --doctest-modules
 
 after_success:
   - coveralls || echo "!! intermittent coveralls failure"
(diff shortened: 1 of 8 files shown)


chadwhawkins:
After passing locally, some issues have arisen because of the various permutations run under CI. There are issues with the following:

- Speedups doctest fails based on flag (could just ignore this file)
- Floating points
- Dictionary ordering

To get these to pass consistently the succinctness of the examples would be lost. Running the doctests did identify a couple of mistakes in the examples. Maybe those changes should be kept and the CI automation dropped?

jorisvandenbossche:
@chadwhawkins thanks for working on this!

If we run it on CI, we could maybe also limit it to a single build? (to not have to pass the tests in all possible setups)

chadwhawkins:
@jorisvandenbossche I think that's a great suggestion. I've gone ahead and done that and have it configured to only run for Python 3.8 when both NumPy and speedsups are enabled.

mwtoews (approved):
Thanks, looks great!

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks @chadwhawkins !

PR #459, opened 2017-01-22
Interpolate section

ewsterrenburg:
Added a section operation (https://github.com/Toblerity/Shapely/issues/456) for review.

PR #1008, opened 2020-10-25
Substring: negative values and direction issues + more robust testing

Jeremiah-England:
I set out to try some performance enhancements out on `ops.substring` and ran into some behavior issues. Here are two examples that capture the issues I found: 

```python
from shapely.ops import substring

ls = LineString([(i, i) for i in range(1000)])
print(substring(ls, -3, 3))
print(substring(ls, -1.1, 0.002, True))

# LINESTRING (996.8786796564403 996.8786796564403, 0 0, 1 1, 2 2, 2.121320343559642 2.121320343559642)
# LINESTRING (0 0, 0 0, 1 1, 1.998000000000014 1.998000000000014)
```
Just as a reminder from the docs of what behavior we're expecting here: 

> Negative distance values are taken as measured in the reverse direction from the end of the geometry. Out-of-range index values are handled by clamping them to the valid range of values.
>
> If the start distance equals the end distance, a point is being returned.
>
> If the normalized arg is True, the distance will be interpreted as a fraction of the geometry’s length.

Two things bother me about the first example. First and most importantly, it returned the wrong cut. It got the start and end correct, but the vertices in the middle are missing. Instead there are the vertices from the beginning of the LineString to the start of the cut. The second issue is the order. 

In the second example, the only issue is that the first vertex is duplicated. Probably not a big deal, but I didn't like it. So I fixed it. 

The test suit would have clued us in to the problem with the duplicate points if it wasn't using `.equals` to compare results. This returns true for reversed LineStrings and ones with more points than the other as long as those points are along the line. Since order matters in this case, I changed the tests to compare wkt instead. The failed tests are also nicer to read like this because the string diff is shown. 

I also added some more tests that cover the issues in the first example. 

Edits and comments are welcome! And take your time, I know things are busy around here these days. : )

Is this a good branch to point to?


mwtoews:
Thanks for the improvements! Yes, the `master` branch is the correct one. Did the manual section need improvement too?

xref #459; this feature was initially brought in three years ago

Jeremiah-England:
Thanks for looking at this. 

Yes, I think the manual should be updated, specifically to clarify that the reversed substring is returned when `start_dist` is *past* `end_dist` (though not necessarily *greater than* because the negatives wrap) and maybe provide an example with a negative argument. I also think it should also say 'LineString' instead of 'linear geometry' because it does not work on MultiLineStrings. 

I would also like to update the docstring [with this commit](https://github.com/Jeremiah-England/Shapely/commit/a5c43c8ab6a1d42d3f66c6f6b534cd3abac2b3d6), but I still need to add some examples to it. Originally I was thinking the docs should go in a separate pull request because this one already seemed busy to me. But now I can also see them being packaged together. Should I push the doc updates here?

One other improvement we may want to do is raise a `TypeError` instead of an `AssertionError` when `geom` is not a LineString. But maybe it would be better to not change the behavior any more?

PS: Sorry about the mess with the force pushes. I recently discovered interactive rebase and decided to try it here to clean up my commits. :)

Jeremiah-England:
I went ahead and added the documentation updates to this pull request. 

In the process I found another behavior issue that I would like to add and then I think this should be ready. In the docstring there is this example which fails the doctest. 

```python
from shapely.ops import substring

ls = LineString((i, 0) for i in range(6))
ls.wkt
# 'LINESTRING (0 0, 1 0, 2 0, 3 0, 4 0, 5 0)'
...
substring(ls, start_dist=2.5, end_dist=-2.5)
# 'POINT (2.5 0)'
```
What is *really* returned here is `'LINESTRING (2.5 0, 2.5 0)'`. That is inconistent with what the function does elsewhere when the substring consists of a single point.

Jeremiah-England:
Ok. Fixed the issue with returning a *LineString* instead of a *Point* when the distances meet but have different signs. 

I have one remaining question about whether to change the *AssertionError* to a *TypeError*. Other than that I think I'm pretty happy with this. But please let me know if you see anything that needs improved.

docs/manual.rst
```diff
@@ -2416,27 +2416,53 @@ The :func:`~shapely.ops.split` function in `shapely.ops` splits a geometry by an
...
+  If the normalized arg is ``True``, the distance will be interpreted as a
+  fraction of the geometry's length
 
-  >>> line = LineString(([0, 0], [2, 0]))
-  >>> result = substring(line, 0.5, 0.6)
-  >>> result.wkt
-  'LINESTRING (0.5 0, 0.6 0)'
+  `New in version 1.7.0`
+
+  Here are some examples that return `LineStrings`.
```
mwtoews, line 2442:
```suggestion
  Here are some examples that return `LineString` geometries.
```

docs/manual.rst
```diff
@@ -2416,27 +2416,53 @@ The :func:`~shapely.ops.split` function in `shapely.ops` splits a geometry by an
 Substring
 ---------
 
-The :func:`~shapely.ops.substring` function in `shapely.ops` returns a line segment between specified distances along a linear geometry.
+The :func:`~shapely.ops.substring` function in `shapely.ops` returns a line segment
```
mwtoews, line 2419:
```suggestion
The :func:`~shapely.ops.substring` function in :mod:`shapely.ops` returns a line segment
```

shapely/ops.py
```diff
@@ -553,14 +553,62 @@ def split(geom, splitter):
...
+    If the start distance is actually beyond the end distance, then the
+    reversed substring is returned such that the start distance is
+    at the first coordinate.
+
+    Parameters
+    ----------
+    geom : LineString
+        The geometry to get a substring of.
+    start_dist : float
+        The distance along `goem` of the start of the substring.
```
mwtoews, line 573:
```suggestion
        The distance along `geom` of the start of the substring.
```

shapely/ops.py
```diff
@@ -553,14 +553,62 @@ def split(geom, splitter):
...
+
+    Returns
+    -------
+    Union[Point, LineString]
+        The substring between `start_dist` and `end_dist` or a Point
+        if they are at the same location.
+
+    Raises
+    ------
+    AssertionError
```
mwtoews, line 588: probably change to `TypeError`; see below

mwtoews (commented):
Thanks for manual adjustments too! It's generally easier to put all related changes into on PR, so this is appreciated. Just a few minor changes to address, and I'll agree to change the type of error raised to `TypeError`.

shapely/ops.py
```diff
@@ -553,14 +553,62 @@ def split(geom, splitter):
...
+    'LINESTRING (1 0, 2 0)'
+
+    Returning a `Point` when `start_dist` and `end_dist` are at the
+    same location.
+
+    >>> substring(ls, 2.5, -2.5).wkt
+    'POINT (2.5 0)'
     """
 
     assert(isinstance(geom, LineString))
```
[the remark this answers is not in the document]
Jeremiah-England: Hmm... tabbing through `raise TypeError` and `raise ValueError` in the module, it looks like `transform`, `nearest_points`, and `split` all use `ValueError` for bad types while only `shared_paths` uses `TypeError`. Unless this misses a majority which are raised by sub-functions, it looks like the `ops` module tends to use `ValueError`. 

If I were starting from scratch, I think I would prefer `TypeError` because it seems to be what the language suggests. But I can also see a useful pattern where you just catch `ValueError` for mistakes that you make with your arguments. Anyway, I'm not sure a discussion about that preference would matter in light of the consistency issue. 

Does the information about which error is more common in `ops` change your suggestion?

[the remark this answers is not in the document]
mwtoews: given the inconsistency at the moment (see #1035), just pick one for now and we'll adjust it later.

mwtoews (approved):

PR #1010, opened 2020-10-28
CLN: remove test_suite runner from test files

jorisvandenbossche:
I don't think those are used any longer


mwtoews (approved):
Yup, these can be cleaned up, as this was for older support with unittest.

PR #1018, opened 2020-11-04
Deprecate setting of _geom attribute

jorisvandenbossche:
Related to https://github.com/Toblerity/Shapely/issues/999, but regardless the naming discussion / which properties we keep, the ability to *set* them from python is something that will go away in Shapely 2.0, so this PR is explicitly deprecating that.


mwtoews (approved):
Thanks, changes look good.

#1017 caused a conflict with `shapely/geometry/base.py` that will need to be fixed before merging.

PR #1019, opened 2020-11-04
Refactor affine_transform: use general function apply on coordinates

jorisvandenbossche:
In pygeos, we have a general `apply` method to apply a function on the coordinates and recreate geometries from those transformed geometries (we probably need to rename it to transform, but that's another issue). 

This speeds up the affine transforms again a bit (which saw a slowdown in https://github.com/Toblerity/Shapely/pull/983). The specific example of https://github.com/Toblerity/Shapely/pull/983#issuecomment-708276648 is now again at the same speed as with released Shapely:

```
In [1]: from shapely.geometry import Polygon

In [2]: poly = Polygon([(0,1), (1, 2), (2, 1), (1, 0), (0, 1)])

In [3]: %timeit poly.minimum_rotated_rectangle
275 µs ± 6.98 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

Now, for single geometries, I don't think we need to expect speed-ups, certainly not compared to the cython code (that I removed here).


jorisvandenbossche:
Note that the tests are still failing because `pygeos.apply` doesn't support 3D coords, yet (working on that)

mwtoews (approved):
Great to see similar performance to the "speedups" implementation and to preserve 3D capabilities, thanks!

PR #1020, opened 2020-11-05
Clean-up use of impl/lgeos/delegated/exceptNull in shapely/geometry/

jorisvandenbossche:
Follow-up on https://github.com/Toblerity/Shapely/pull/983, cleaning up a few more things in geometry/base.py

- The `delegated` decorator is no longer needed (no support for another "impl"), also removed the tests for this
- `project` is already implemented by PyGEOS, so using that one (same for `relate` and `relate_pattern` now)
- Remove the `exceptNull` decorator, as such error handling is now handled in pygeos C code. For interpolate, we only have a slight difference in behaviour, though (for which I updated the test), see inline comment.
- Removed `self.impl`, since this is now no longer rused


tests/test_operations.py
```diff
@@ -88,12 +88,18 @@ def test_interpolate(self):
...
+
+        # Issue #653; should nog segfault for empty geometries
         empty_line = loads('LINESTRING EMPTY')
-        assert(empty_line.is_empty)
-        with pytest.raises(ValueError):
-            empty_line.interpolate(.5, normalized=True)
-        
+        assert empty_line.is_empty
+        interpolated_point = empty_line.interpolate(.5, normalized=True)
+        assert interpolated_point.is_empty
```
jorisvandenbossche, line 96: This test/behaviour comes from https://github.com/Toblerity/Shapely/issues/653. GEOS segfaults when passing an empty linestring, and Shapely solved this with the `exceptNull` decorator checking if any of the arguments was empty. PyGEOS currently also checks for this explicitly (to avoid the segfault, see https://github.com/pygeos/pygeos/blob/32397d998d2fc7fe3afef673539dace9fd8efe7f/src/geos.c#L297), but we decided to return an empty point instead of raising an error (see https://github.com/pygeos/pygeos/blob/32397d998d2fc7fe3afef673539dace9fd8efe7f/src/ufuncs.c#L373)

Any strong opinions on the desired behaviour?

cc @snorfalorpagus

mwtoews: This change of behavior is welcome, particularly since this is what PostGIS does:
```
db=> SELECT ST_AsText(ST_LineInterpolatePoint('LINESTRING EMPTY'::geometry, 0.2));
  st_astext
-------------
 POINT EMPTY
(1 row)
```

mwtoews (approved):
Looks good. I see the remaining `delegated` functions are in `shapely.prepared`, which I suspect will be addressed later (as discussed in https://github.com/shapely/shapely-rfc/pull/1)

jorisvandenbossche:
> I see the remaining delegated functions are in shapely.prepared

Indeed, and the same for the `impl` / `DefaultImplementation`, which is now (with the latest updates to this PR) also only used in the shapely.prepared module. So fully cleaned usage of it in `geometry/base.py`, but we will only be able to fully remove all the code in `impl.py` if also the prepared module is updated to use pygeos (now, functionality related to "prepared" has been merged in pygeos, so that should be possible now).

jorisvandenbossche:
Also removed any remaining use of `lgeos` in shapely/geometry/ submodule, except for the usage in pickling of LinearRings

jorisvandenbossche:
Needed to do a force-push to this branch, because I rebased the `shapely-2.0` branch on latest master

jorisvandenbossche:
CI is failing because of a build failure with pygeos (opened https://github.com/pygeos/pygeos/issues/260 for that).

jorisvandenbossche:
Actually, GEOSVERSION changed to GEOS_VERSION, so that was the cause ;)

jorisvandenbossche:
OK, remaining failure is a doctest failure unrelated to this PR, so will fix that in a separate PR (we started checking the doctests on master, which was now brought into shapely-2.0 branch, and there are some failures now)

PR #1025, opened 2020-11-10
Deprecate ops.cascaded_union, as it was superseded by ops.unary_union

### LINKED ISSUE #1001: API: keep both cascaded_union and unary_union functions?
Originally, GEOS did have `cascased_union` operation, and `unary_union` was added later with better behaviour. Since GEOS 3.3, `GEOSUnionCascaded_r` is actually deprecated in favor of the `unary_union` version.  
In Shapely, similarly a `unary_union` function was added to `shapely.ops` in addition to `cascaded_union`. But as long as you have GEOS = 3.3, the `cascaded_union` in Shapely is actually a simple alias for `unary_union`, and no longer having different behaviour/implementation:

https://github.com/Toblerity/Shapely/blob/eae4dcbedb0756f658f9982825d5fa23b0061b2a/shapely/geos.py#L838

So I am mainly opening an issue to ask: do we want to keep `cascaded_union` as an alias forever, or do we actually want to deprecate it (raise a warning) and have people move to use `unary_union` ?

(note this is only about the `shapely.ops` function, on the Geometry classes, there is only a `unary_union` attribute, not a `cascaded_union` one)

mwtoews:
Resolves #1001 to deprecate `ops.cascaded_union` for Shapely 1.8.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -25,7 +25,7 @@ import shapely
 
 # For pyplots in code/, load functions here first, so they are visible
 from shapely import geometry, affinity, wkt, wkb
-from shapely.ops import cascaded_union
+from shapely.ops import unary_union
 
 # -- General configuration -----------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


jorisvandenbossche (approved):
Looks good to me, thanks!

Issue #1000, opened 2020-10-13
Minimal supported GEOS version ?


jorisvandenbossche:
What is the minimal GEOS version that Shapely still (aims to) supports?

At the moment, Shapely starts testing for GEOS 3.5 and up on CI, but actually has code in `shapely/geos.py` going back to GEOS 3.0. 

In PyGEOS,  we also test (and have workarounds) going back to GEOS 3.5, and have certain functionality that only works for more recent GEOS versions (raises an error for older versions, see eg the decorator used here for `get_z`: https://github.com/pygeos/pygeos/blob/32397d998d2fc7fe3afef673539dace9fd8efe7f/pygeos/geometry.py#L251). 
We did have some discussion related to this at https://github.com/pygeos/pygeos/pull/116, and @caspervdw gave the argument of supporting at minimum GEOS 3.5 since this is the GEOS version included in Ubuntu 16.04 (which reaches end of life in April 2021). Supporting GEOS versions back to 3.5 is not that big of a burden, because we need the machinery for version checks anyway to support functionalities added in the latest GEOS versions anyway. 

I don't think PyGEOS actively checks the used GEOS version and raises for older GEOS < 3.5 versions (@caspervdw is that correct?). 

**Are we fine to also have GEOS 3.5 as a minimum supported version in Shapely?** And thus not have workarounds for older versions, as is the case now (and next year, we can then discuss moving to GEOS 3.6 as a minimum)

(I would say that in practice, this is the only option if we don't even test older versions)

caspervdw:
I never actually tried compiling PyGEOS with GEOS 3.4, but there is a fat chance that there is at least one missing symbol in the CAPI. So I am :-1: on keeping GEOS <3.5 support.

mwtoews:
A minimum GEOS 3.5 (released 2015-08-15) for shapely 2.0 is fine with me.

Users with older GEOS versions (e.g. find some packaged distros [here](https://pkgs.org/search/?q=geos)) would need to stick to shapely<2.0 releases.

jorisvandenbossche:
Looking at the distros link, I already mentioned Ubuntu 16.04 that has GEOS 3.5 (16.04 is currently the oldest still supported version). 
Some others: CentOS 7 includes GEOS 3.4 and is officially still supported (for maintenance updates) until June 30, 2024. CentOS 8 includes GEOS 3.7, though. 
The oldest listed Debian version already includes GEOS 3.5. 
[...]

PR #1031, opened 2020-11-18
CLN: enforce a minimum GEOS version 3.3.0

mwtoews:
This PR cleans classes and logic from Shapely that supported GEOS versions before 3.3.0.

And yes, I did test this PR locally against GEOS 3.3.0, which passed most tests.

GEOS 3.3.0 was first released in May 2011, with GEOS 3.3.9 released Sept 2013.

See a range of GEOS versions in a few Linux distros here: https://pkgs.org/search/?q=geos

See also #1000 for Shapely-2.0, which will use a higher version.


tests/test_wkt.py
```diff
@@ -18,12 +18,12 @@ def pipi4():
...
 def test_wkt(pipi):
     """.wkt and wkt.dumps() both do not trim by default."""
-    assert pipi.wkt == "POINT ({0:.16f} {1:.16f})".format(pi, -pi)
+    assert pipi.wkt == "POINT ({0:.15f} {1:.15f})".format(pi, -pi)
 
 
 def test_wkt(pipi4):
     """.wkt and wkt.dumps() both do not trim by default."""
-    assert pipi4.wkt == "POINT ({0:.16f} {1:.16f})".format(pi*4, -pi*4)
+    assert pipi4.wkt == "POINT ({0:.14f} {1:.14f})".format(pi*4, -pi*4)
```
jorisvandenbossche, line 26: those changes seem consistent with what I needed to change on the shapely-2.0 branch, so good!

jorisvandenbossche (commented):
Skimmed through it, looks good!

jorisvandenbossche (approved):

tests/test_wkt.py
mwtoews: yes, parts of 3024adc06bdd6f110c87c3e08837790033a23eee where cherry-picked

sgillies (approved):

sgillies:
Thanks @mwtoews !

Issue #701, opened 2019-04-11
ops.split inconsistent behavior with LineStrings and MultiLineStrings


romainfontaine:
## Expected behavior and actual behavior & steps to reproduce the problem

Splitting a linestring A with an intersecting linestring B does not produce the expected output (returns A). Splitting A with a multilinestring containing B and C (C intersecting A as well) produces the desired output (3 segments). This seems inconsistent.

```
from shapely.ops import split
from shapely.geometry import LineString, MultiLineString

A = LineString([(0, 0), (10, 0)])
B = LineString([(5, 0), (5, 5)])
C = LineString([(1, -1), (1, 1)])
assert (split(A, B).wkt == "GEOMETRYCOLLECTION (LINESTRING (0 0, 10 0))")
# Does not split
# Expected: "GEOMETRYCOLLECTION (LINESTRING (0 0, 5 0), LINESTRING (5 0, 10 0))"

assert (A.intersection(B).wkt == "POINT (5 0)")
# Although A and B do intersect

assert (split(A, MultiLineString([B, C])).wkt ==
        "GEOMETRYCOLLECTION (LINESTRING (0 0, 1 0), LINESTRING (1 0, 5 0), LINESTRING (5 0, 10 0))")
# OK - A is split by both B and C - but A was not split by B in the previous example.

```

## Operating system

Ubuntu 18.04.2 LTS

## Shapely version and provenance
Python 3.6.7 - Shapely 1.6.4.post2 (installed from PyPI using pip)

Jeremiah-England:
Ok. I see the problem here.  

Visually, those linestrings look like this: 

![image](https://user-images.githubusercontent.com/34973839/96392898-7bcb9d00-118b-11eb-8862-cb1ee977bfb5.png)

When you split the horizontal (**A**) by the large vertical on top (**B**) which only *butts* into **A**, nothing happens. But splitting **A** by **C** does work because **C** *crosses* **A**. 

Now, when you split **A** by `MultiLineString([B, C])`, it splits into *three* linestrings, not *two* as you would expect. I agree that this seems inconsistent. 

The code causing this is actually pretty accessible ([permalink](https://github.com/Toblerity/Shapely/blob/eae4dcbedb0756f658f9982825d5fa23b0061b2a/shapely/ops.py#L414-L433)). 

```python
    @staticmethod
    def _split_line_with_line(line, splitter):
        """Split a LineString with another (Multi)LineString or (Multi)Polygon"""

[...]

PR #1034, opened 2020-11-19
ops.split: Split the line when the boundary of a splitter LineString touches the interior of the line

Jeremiah-England:
See #701 for background. 

I thought about using `splitter.relate_pattern(line, '...')` several times. But I'm not sure how much if any of that operation is cached, so I'm calculating one relation string and using it directly for the three checks. Fortunately the checks are all straightforward, each involving only one character.

Ping to @snorfalorpagus, just in case you would want to take a look since you have been heavily involved with the split function in the past.


sgillies (approved):
@Jeremiah-England thank you very much!

PR #1039, opened 2020-11-28
Refactor shapely.prepared to use prepared base geometries

jorisvandenbossche:
In PyGEOS, the GEOSPreparedGeometry is now stored on the same python extension type (C struct) as used for the base Geometry class (and there are functions to (un)prepare a geometry object "in place"). If an object is prepared, the predicate functions will automatically use it.

So in the future, we can make this interface in Shapely easier, but on the short term, I refactored the existing shapely PreparedGeometry to use the prepared functionality of PyGEOS. 
(since there is not yet an alternative for this in Shapely 1.8, so we can't deprecate this yet (if we would want that long term), and keep a working version initially in Shapely 2.0)

### DIFF
--- a/shapely/prepared.py
+++ b/shapely/prepared.py
@@ -1,9 +1,8 @@
 """
 Support for GEOS prepared geometry operations.
 """
+import pygeos
 
-from shapely.geos import lgeos
-from shapely.impl import DefaultImplementation, delegated
 from pickle import PicklingError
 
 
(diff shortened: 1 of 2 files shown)


shapely/prepared.py
```diff
@@ -20,75 +19,53 @@ class PreparedGeometry(object):
...
 
-    @delegated
     def contains_properly(self, other):
         """Returns True if the geometry properly contains the other, else False"""
-        return bool(self.impl['prepared_contains_properly'](self, other))
+        # TODO temporary hack until pygeos exposes contains properly as predicate function
+        from pygeos import STRtree
+        tree = STRtree([other])
+        idx = tree.query(self.context, predicate="contains_properly")
+        return bool(len(idx))
```
jorisvandenbossche, line 40: I opened https://github.com/pygeos/pygeos/pull/267 to add this to pygeos properly

mwtoews (approved):
Changes look good, and I appreciate `_geom` and `__geom__` have been cleaned up too, hopefully no downstream issues may rise from this.

I did a quick "timeit" comparison using:
```python
from shapely.geometry import Point, Polygon
triangle = Polygon(((0.0, 0.0), (1.0, 1.0), (1.0, -1.0)))
pt = Point(0.5, 0.5)
%timeit prep(triangle).intersects(pt)
```
and I see a minor slowdown from ~ 5 µs to 9 µs, which is acceptable. The first of these loops takes several times longer, which confirms that PyGEOS does some in-place caching of the prepared geometry.

This is also the last use of `delegated`, so it could be considered to be cleaned up soon. And I suppose the whole `impl` module could be cleaned up too.

sgillies (approved):

sgillies:
Yes, we're all in on GEOS now, so impl.py can be deleted.

jorisvandenbossche:
OK, next PR to remove `impl` !

PR #1041, opened 2020-11-30
Expand CI and tests to support GEOS 3.9.0beta1

mwtoews:
GEOS 3.9.0beta1 was recently [announced](https://lists.osgeo.org/pipermail/geos-devel/2020-November/009866.html), which should be paired with the Python 3.9 CI tests.

A few tests were refactored to accommodate the changes:

- There are a few inconsistencies within the `test_create_inconsistent_dimensionality.py` suite. I'm not sure if any of these are worth reporting upstream or not. The only exception is for `LineString`. Some additional tests and comments are added for completeness.
- The order of `union` changed, likely a result of the new OverlayNG
- WKB serialization of `POINT EMPTY` and `POINT Z EMPTY` are supported, so do an explicit test/xfail


mwtoews:
The AppVeyor failure is a [known issue](https://lists.osgeo.org/pipermail/geos-devel/2020-November/009856.html) with `ctest` only.

mwtoews:
The CI scripts are now closer aligned to what I [prepard for PyGEOS](https://github.com/pygeos/pygeos/pull/219), which may allow them to be used with Github Actions, or other CI offerings. This will probably help future CI migrations.

Another key difference is that I've disabled a failure from `ctest` for the Windows install script. This is intended to be temporary, until GEOS 3.9 is fixed for Windows. Ideally, only clean versions of GEOS should be tested and cached.

tests/test_create_inconsistent_dimensionality.py
```diff
@@ -29,9 +30,29 @@
...
+        # drop 3rd dimension
+        ('LINESTRING (1 1, 2 2 2)', 'LINESTRING (1 1, 2 2)'),
+    ]
+else:
+    wkt_cases += [
+        # drop 3rd dimension
+        ('LINESTRING (1 1 1, 2 2)', 'LINESTRING (1 1, 2 2)'),
+        # preserve 3rd dimension
+        ('LINESTRING (1 1, 2 2 2)', 'LINESTRING Z (1 1 0, 2 2 2)'),
+    ]
```
jorisvandenbossche, line 55: It seems to be quite inconsistent when an incomplete third dimension is preserved and when it is dropped, right? And also the change with GEOS 3.9 does not seem to make it more consistent

tests/test_svg.py
```diff
@@ -174,7 +174,8 @@ def test_collection(self):
         self.assertSVG(GeometryCollection(), '<g />')
         # Valid
         self.assertSVG(
-            Point(7, 3).union(LineString([(4, 2), (8, 4)])),
+            GeometryCollection(
+                [Point(7, 3), LineString([(4, 2), (8, 4)])]),
```
jorisvandenbossche, line 178: This is indeed related to OverlayNG (I once tested this, but forgot to do a PR for shapely to fix it). 
The order of the parts of the geometry collection changed

jorisvandenbossche (commented):
Thanks a lot for looking at this!

tests/test_create_inconsistent_dimensionality.py
mwtoews: I call this a minor annoyance; see [GEOS trac#1079](https://trac.osgeo.org/geos/ticket/1079)

mwtoews: Likely fixed here https://github.com/pramsey/geos/commit/9211fae903a22fc8fc68de47a1a0357fe1bfa4b8

PR #1047, opened 2020-12-14
auditwheel 2/3 compat

sgillies:
This is a "hotfix" with search path changes required by auditwheel versions > 3.0, which is required for Python 3.9.

We will not merge this into 1.7.x but we will need this for 1.8.

### DIFF
--- a/shapely/geos.py
+++ b/shapely/geos.py
@@ -68,22 +68,20 @@ if sys.platform.startswith('linux'):
     # Test to see if we have a wheel repaired by auditwheel which contains its
     # own libgeos_c. Note: auditwheel 3 changed the location of libs.
     geos_whl_so = glob.glob(
-        os.path.abspath(
-            os.path.join(
-                os.path.dirname(__file__), ".libs/libgeos_c-*.so.*"
-            )
-        )
+        os.path.abspath(os.path.join(os.path.dirname(__file__), ".libs/libgeos*.so*"))
     ) or glob.glob(
         os.path.abspath(
             os.path.join(
-                os.path.dirname(__file__), "..", "Shapely.libs", "libgeos_c-*.so.*"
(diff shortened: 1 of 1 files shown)


sgillies:
I've uploaded a `Shapely-1.7.1-1-cp39-cp39-manylinux1_x86_64.whl` to https://pypi.org/project/Shapely/1.7.1/#files with the hotfix made up of the commits in this PR: https://patch-diff.githubusercontent.com/raw/Toblerity/Shapely/pull/1047.diff. My changes in fc3e642 required the list of lib paths to be sorted and in 3a3f91b we do sort them. I've installed the wheel on my thinkpad with Python 3.9.1 and a basic test works.

```
Python 3.9.1 (default, Dec 15 2020, 11:13:51)
[GCC 7.5.0] on linux
>>> Point(0, 0).buffer(10).area
313.6548490545939
```

This resolves the new issue reported in https://github.com/Toblerity/Shapely/issues/1040#issuecomment-745420914.

daveisfera:
> If you've already downloaded and installed `Shapely-1.7.1-cp39-cp39-manylinux1_x86_64.whl` (the file from 2020-12-14 with no "-1" build tag), uninstall and remove it from your pip cache and reinstall, or do `pip install -U shapely -I --force-reinstall --no-cache`.

Can an updated release be made? We updated to Python 3.9 last night and trying to invalidate all of the docker builds that will now have this bad setup can be tricky and if I'm understanding this correctly, there is also a code change to go along with this, so wouldn't a new version make sense anyways?

sgillies:
No, I will not make a 1.7.2 release for this. 1.7.1-1 is like a build-specific patch applied by a downstream packager.

daveisfera:
Will 1.7.1-1 be published to PYPI or is there a way for me as a user to specify it in `requirement.txt` to be sure it's working, because we're having to play wack-a-mole with docker layer caches to resolve the problem currently and it keeps creeping in.

PR #1048, opened 2020-12-15
CI: test geos-3.9.0 release; show other debug information

mwtoews:
The final release of geos-3.9.0 was just released, so use this in the Travis CI test matrix. (We'll likely transition this to Github Actions at some point).

Other things:
- The `pytest>=3.8` command just redirect output to a `=3.8` file; the `--upgrade` flag should do the right thing.
- Testing with the GEOS master branch builds speedups, but they don't appear to work, so add a bit more debugging info. If it looks like an easy fix, then I'll add it here.
- No features from GEOS 3.9.0 are implemented yet, but are next up to investigate and discuss.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,17 +70,17 @@ matrix:
         NUMPY=0
     - python: "3.9"
       env:
-        GEOS_VERSION="3.9.0beta2"
+        GEOS_VERSION="3.9.0"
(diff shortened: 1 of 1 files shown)


jorisvandenbossche (approved):

mwtoews:
Superseded by #1076

PR #1052, opened 2020-12-16
Clean-up impl: remove impl.py + all helper classes

jorisvandenbossche:


jorisvandenbossche:
One question: the `shapely.iterops` module provided a set of functions for predicates that could filter an iterable of geometries. Which is something that has no direct replacement.  
As far as I know this was not documented, but it was tested (see removed test file), so: is this something we want to explicitly deprecate in Shapely 1.8? (I can still do that on the master branch, no problem) Or does somebody want to keep this functionality? (we could easily reimplement the function with plain shapely geometry methods)

sgillies:
@jorisvandenbossche the objects in the iterops module could be reimplemented by composing existing Python functions and shapely methods, it's true. I'm :+1: on removing it.

sgillies (approved):

jorisvandenbossche:
@sgillies so

jorisvandenbossche:
@sgillies so the `shapely.iterops` module will be removed in shapely 2.0, but I can still add explicit deprecation warnings for shapely 1.8 about it?

jorisvandenbossche:
It's easy to deprecate, so nice for users to just warn about it -> https://github.com/Toblerity/Shapely/pull/1063

Issue #1049, opened 2020-12-15
Availability of minimum_clearance


aharfoot:
## Expected behavior and actual behavior.

When I try to calculate the minimum clearance of a polygon an error is returned, I am expecting the result to match that printed in the documentation.

The Shapely documentation suggests that GEOS 3.6 or higher is needed, and as far as I can see Shapely is reporting that it is using GEOS 3.8.

## Steps to reproduce the problem.

Taking sample code from the user manual:

`sys.version`
'3.7.0 (v3.7.0:1bf9cc5093, Jun 27 2018, 04:59:51) [MSC v.1914 64 bit (AMD64)]'

`import shapely`
`shapely.__version__`
'1.7.0'

`import shapely.geos`
`shapely.geos.geos_version`
(3, 8, 0)

`shapely.geos.geos_version_string`
'3.8.0-CAPI-1.13.1 '

`from shapely.geometry import Polygon`
`Polygon([[0, 0], [1, 0], [1, 1], [0, 1], [0, 0]]).minimum_clearance`

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'Polygon' object has no attribute 'minimum_clearance'

## Operating system
MS Windows 10 64bit

## Shapely version and provenance
Version 1.7.0 installed via the OSGeo4W package installer

jorisvandenbossche:
The `minimum_clearance` property was only added in Shapely 1.7.1 (https://github.com/Toblerity/Shapely/pull/870), and you have 1.7.0. So you will need to upgrade the Shapely version to be able to use it

aharfoot:
Ok, thanks for clarifying! So [the manual](https://shapely.readthedocs.io/en/stable/manual.html) needs correcting as it references 1.7.0 and includes `minimum_clearance`

jorisvandenbossche:
>  So the manual needs correcting as it references 1.7.0 and includes minimum_clearance

Ah, good catch. Fully at the top, you can see that it are actually the docs for 1.7.1. But indeed the "Version: 1.7.0" should be updated as well. 
It seems that this version is hardcoded: https://github.com/Toblerity/Shapely/blame/master/docs/manual.rst#L8

PR #1057, opened 2021-01-05
DOC: Version of the minimum_clearance function

aharfoot:
Related to #1049

### DIFF
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -209,6 +209,7 @@ General Attributes and Methods
   such as a point, this will return `math.infinity`.
   
   `New in Shapely 1.7.1`
+  
   Requires GEOS 3.6 or higher.
 
 .. code-block:: pycon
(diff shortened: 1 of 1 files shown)


docs/manual.rst
```diff
@@ -207,7 +207,8 @@ General Attributes and Methods
   This can be thought of as a measure of the robustness of a geometry, where larger values of
   minimum clearance indicate a more robust geometry. If no minimum clearance exists for a geometry,
   such as a point, this will return `math.infinity`.
-
+  
+  `New in Shapely 1.7.1`
```
Jeremiah-England, line 211: This looks good to me. The one question I would have is whether we want the two requirement statements on one line:

image

Or on separate lines:

image
 
To get the later, we'll need an *extra* new between them like this:

```rst
  `New in Shapely 1.7.1`

  Requires GEOS 3.6 or higher.
```

jorisvandenbossche: On two lines might be nicer I think (in any case, otherwise there needs to be a punctuation (`.`) added at the end of this line)

jorisvandenbossche:
@aharfoot thanks for the PR! Would you like to update it for the comment above?

aharfoot:
Hi, I've added the additional newline.

sgillies (approved):
@aharfoot thank you very much!

Issue #991, opened 2020-09-25
API: custom classes for GEOS exceptions


jorisvandenbossche:
Moving an inline discussion to its separate dedicated issue: https://github.com/Toblerity/Shapely/pull/983#discussion_r489285146

Currently, Shapely has a few custom error classes, such as `TopologicalError`. Using an example from the tests:

```python
from shapely.geometry import Polygon
p1 = [(339, 346), (459,346), (399,311), (340, 277), (399, 173),
      (280, 242), (339, 415), (280, 381), (460, 207), (339, 346)]
p2 = [(339, 207), (280, 311), (460, 138), (399, 242), (459, 277),
      (459, 415), (399, 381), (519, 311), (520, 242), (519, 173),
      (399, 450), (339, 207)]

>>> Polygon(p1).within(Polygon(p2))
TopologyException: side location conflict at 459 346
Traceback (most recent call last):
  File "/home/joris/miniconda3/envs/DS-python-geospatial/lib/python3.8/site-packages/shapely/predicates.py", line 15, in __call__
    return self.fn(this._geom, other._geom, *args)
  File "/home/joris/miniconda3/envs/DS-python-geospatial/lib/python3.8/site-packages/shapely/geos.py", line 580, in errcheck_predicate
    raise PredicateError("Failed to evaluate %s" % repr(func))
shapely.errors.PredicateError: Failed to evaluate <_FuncPtr object at 0x7f0a641ce1c0>

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/joris/miniconda3/envs/DS-python-geospatial/lib/python3.8/site-packages/shapely/geometry/base.py", line 752, in within
    return bool(self.impl['within'](self, other))
  File "/home/joris/miniconda3/envs/DS-python-geospatial/lib/python3.8/site-packages/shapely/predicates.py", line 18, in __call__
    self._check_topology(err, this, other)
  File "/home/joris/miniconda3/envs/DS-python-geospatial/lib/python3.8/site-packages/shapely/topology.py", line 35, in _check_topology
    raise TopologicalError(
shapely.errors.TopologicalError: The operation 'GEOSWithin_r' could not be performed. Likely cause is invalidity of the geometry <shapely.geometry.polygon.Polygon object at 0x7f0a6488e490>
```

[...]

PR #1065, opened 2021-01-14
Refactor shapely.ops to not use lgeos

jorisvandenbossche:
Replacing direct `lgeos` calls with pygeos python functions in the `shapely.ops` module.

The remaining ones to do are `polygonize`, `polygonize_full` and `nearest_points`. For `polygonize` I already have a PR in pygeos, the others still need to be added there.

### DIFF
--- a/shapely/ops.py
+++ b/shapely/ops.py
@@ -46,16 +46,8 @@ class CollectionOperator:
             source = [source]
         finally:
             obs = [self.shapeup(l) for l in source]
-        geom_array_type = c_void_p * len(obs)
-        geom_array = geom_array_type()
-        for i, line in enumerate(obs):
-            geom_array[i] = line._geom
-        product = lgeos.GEOSPolygonize(byref(geom_array), len(obs))
-        collection = geom_factory(product)
-        for g in collection.geoms:
-            clone = lgeos.GEOSGeom_clone(g._geom)
-            g = geom_factory(clone)
-            yield g
+        collection = pygeos.polygonize(obs)
+        return collection.geoms
 
     def polygonize_full(self, lines):
         """Creates polygons from a source of lines, returning the polygons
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
This is ready now: it fully removed all remaining usage of ctypes/lgeos from ops.py (and then remaining usage is in strtree.py)

shapely/ops.py
```diff
@@ -310,24 +281,18 @@ def nearest_points(g1, g2):
 
     The points are returned in the same order as the input geometries.
     """
-    seq = lgeos.methods['nearest_points'](g1._geom, g2._geom)
+    seq = pygeos.shortest_line(g1, g2)
```
mwtoews, line 284: I was about to ask about why `nearest_points` wasn't in PyGEOS, but then I see https://github.com/pygeos/pygeos/pull/334

shapely/ops.py
```diff
@@ -210,26 +183,24 @@ def voronoi_diagram(geom, envelope=None, tolerance=0.0, edges=False):
...
+        result = pygeos.voronoi_polygons(
+            geom, tolerance=tolerance, extend_to=envelope, only_edges=edges
+        )
+    except pygeos.GEOSException as err:
+        errstr = "Could not create Voronoi Diagram with the specified inputs "
+        errstr += "({}).".format(str(err))
         if tolerance:
             errstr += " Try running again with default tolerance value."
-        raise ValueError(errstr)
+        raise pygeos.GEOSException(errstr) from err
```
mwtoews, line 195: I'm OK with exposing GEOSException, but I thinks @sgillies has preferred raising a Shapely error instead?

mwtoews (approved):
Changes look good! Just wait to see if to see if @sgillies wants to weigh-in on raising GEOSException vs a different exception.

shapely/ops.py
jorisvandenbossche: Yeah, we will need to decide on whether we want to expose both versions top-level in Shapely 2.0 (but that doesn't need to be done in this PR)

shapely/ops.py
jorisvandenbossche: For this specific case we have easy control over it (since we are catching the original GEOSException anyway to add some context to the error message), but the general issue of raising GEOSException to the user or not is a broader discussion, and some discussion is in https://github.com/Toblerity/Shapely/issues/991

sgillies: I very much prefer to raise an error that derives from `ShapelyError` (https://github.com/Toblerity/Shapely/blob/master/shapely/errors.py#L4). How about a `VoronoiDiagramError` ?

jorisvandenbossche: Note that `GEOSException` is "our own exception class", just like `ShapelyError` is (once pygeos is incorporated into shapely).

I would prefer if we can then revive the discussion in https://github.com/Toblerity/Shapely/issues/991, which exactly deals with this topic. 
I could certainly create a custom VoronoiDiagramError for this case (to get this PR merged), but it would then be adding a special case, since most other operations raise GEOSExceptions in the shapely-2.0 branch. So personally I would rather first get to conclusion for the general discussion.

jorisvandenbossche: Ping here. Is it OK to keep this discussion for #991, and move forward with this PR? 

I will keep the current ValueError instead of changing it to GEOSException, then this PR is just the status quo on that aspect.

sgillies: Thanks @jorisvandenbossche !

PR #1070, opened 2021-01-22
Initialize STRtree with capacity of 10 items per leaf node

dbaston:
Previous initialization was creating a tree with all geometries on a single leaf node.

See: https://gis.stackexchange.com/questions/353619/shapely-with-rtree-versus-strtree


jorisvandenbossche (approved):

andrerav:
Hi! We are running into some performance problems because of this, having high hopes that this will be merged soon :)

jorisvandenbossche:
Thanks @dbaston!

PR #1071, opened 2021-01-23
fix the multi libgeos for pyinstaller

sshuair:
If we have installed the `finoa`, `shapely`, `rasterio`, and package them all use Pyinstaller. 

There are multi `libgeos_c-*.so.*`, not only one. 

So, we should check the length of geos_pyinstaller_so greater and equal to 1.

image

image


sgillies:
@sshuair excellent, thank you! You'll see this fix in 1.8a1 (later today or tomorrow).

PR #1075, opened 2021-01-26
Deprecate len(..) of multi-part geometries

jorisvandenbossche:
xref https://github.com/Toblerity/Shapely/issues/932

PR #950 already deprecated `__getitem__` and `__iter__` of multi-part geometries, but not yet `__len__`. 
Similarly to that deprecation, users can do `len(obj.geoms)` instead of `len(obj)`. We could also add a `.ngeoms` property as a shortcut for this, if that is desired. 

Note that I think deprecating `__len__` is not strictly needed for compatibility with numpy (although I would need to check this to be sure), as long as the geometries are not iterable anymore (or have an array_interface). Personally I think it follows a bit from deprecating getitem/iter, where `__len__` typically is implemented for collection-like objects (and which multi-part geometries no longer are in Shapely 2.0). So I am fine with deprecating it (you also need to know if you have a multi-part geometry or not before calling it, since `len(point)` currently fails, so if you already know that, it shouldn't be any problem to do eg `len(multipoint.geoms)` instead).

But, we could also keep `__len__` if people prefer that. Speaking in terms of the python ABCs (https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes), multi-part geometries could be "Sized" without being "Iterable" / a "Collection".


sgillies:
@jorisvandenbossche `len(geom)` used to call a GEOS method to get the number of geoms without materializing them, and was fast. As long as `len(geom.geoms)` is equally fast and efficient, that works for me, no need for a .ngeoms.

jorisvandenbossche:
Yes, as long as we keep the `.geoms` attribute return a custom class and not a plain list or array, this is possible. That is also what is currently done in the shapely-2.0 branch: it kept the custom `GeometrySequence` class, and so checking the length of this directly calls the GEOS' method to check the number of parts of a Multi geometry:

https://github.com/Toblerity/Shapely/blob/34acbc1bd1fc12bcd07858f09133b2163fcb1e76/shapely/geometry/base.py#L705-L706

tomplex:
Personally I agree with your initial thought @jorisvandenbossche, `len` strongly implies iterability and geometries are not iterable anymore. I suspect it will be more confusing than helpful when you can ask for the `len` of a geometry but have to use `geometry.geoms` to iterate over it.

sgillies (approved):

tests/test_collection.py
```diff
@@ -95,3 +95,22 @@ def test_geometrycollection_adapter_deprecated(geometrycollection_geojson):
...
+    GeometryCollection(),
+    shape({"type": "GeometryCollection", "geometries": []}),
+    shape({"type": "GeometryCollection", "geometries": [
+        {"type": "Point", "coordinates": ()},
+        {"type": "LineString", "coordinates": (())}
+    ]}),
+    wkt.loads('GEOMETRYCOLLECTION EMPTY'),
+])
+def test_len_empty_deprecated(geom):
+    with pytest.warns(ShapelyDeprecationWarning):
```
mwtoews, line 109:
```suggestion
    with pytest.warns(ShapelyDeprecationWarning, match="__len__"):
```

tests/test_collection.py
```diff
@@ -95,3 +95,22 @@ def test_geometrycollection_adapter_deprecated(geometrycollection_geojson):
...
+        assert len(geom) == 0
+
+
+def test_len_deprecated(geometrycollection_geojson):
+    geom = shape(geometrycollection_geojson)
+    with pytest.warns(ShapelyDeprecationWarning):
```
mwtoews, line 115:
```suggestion
    with pytest.warns(ShapelyDeprecationWarning, match="__len__"):
```

mwtoews (changes requested):
Changes look good, and I agree there is no need for a `.ngeoms` property.

Other than suggested edits, the manual can be adjusted slightly too:
* https://github.com/Toblerity/Shapely/blob/master/docs/manual.rst#collections-of-polygons - remove the two lines around "len(polygons)"
* https://github.com/Toblerity/Shapely/blob/master/docs/manual.rst#constructive-methods adjust `len(clean)` -> `len(clean.geoms)`
* https://github.com/Toblerity/Shapely/blob/master/docs/manual.rst#merging-linear-features adjust `len(result)` -> `len(result.geoms)`

mwtoews (approved):
Perfect!

PR #1090, opened 2021-02-15
ENH: Add normalize() method

### LINKED ISSUE #994: Expose GEOSNormalize
`GEOSNormalize_r` is the C API for [`Geometry::normalize`](https://geos.osgeo.org/doxygen/classgeos_1_1geom_1_1Geometry.html#a2ea3b58d562bc183e118e2a68f86414d), which reorganizes geometries into 'canonical forms': points in a `LineString` may be deduplicated, members of a collection are sorted by type, points in a polygon's rings or a single `LinearRing` start at the top-left-most point and go clockwise, etc. This can be useful for unit testing or geometry deduplication. This was added [in libgeos 3.0.0](https://github.com/libgeos/geos/commit/9ecb499d71c279f4713f4aa8a2c0e9d4656b0e2c).

I tried to add the method myself but I never really played with `ctypes`, so this was mostly playing around, and I always got `-1` (exception). This method updates the geometry in place and I am not sure if this is a good thing; pygeos [implemented](https://github.com/pygeos/pygeos/blob/7a2798ab8ef43acf59e2023b8cfcb49e6d9384d3/src/ufuncs.c#L257) a wrapper that clones before normalizing.

Note that the `LineString::normalize` implementation has a bug in which the first point of a `LinearRing` *was not* the top-left-most point. It was fixed in [this commit](https://github.com/libgeos/geos/commit/a807833875976a29fbcf509297f4a7a5f837aa09#diff-a6950e44d48810137ee9f1460388168dR323), which is not included in any release yet (probably will be in libgeos 3.9.0). The workaround is to create a polygon using the LinearRing as an exterior ring, normalize the polygon, and get the ring back. Assuming normalize would be added to Shapely, we would get something like this:

```python3
>>> list(Polygon(((8, 8), (8, 5), (8, 5), (5, 5), (5, 8))).normalize().exterior.coords)
[(8, 8), (8, 5), (5, 5), (5, 8), (8, 8)]
>>> list(LinearRing(((8, 8), (8, 5), (8, 5), (5, 5), (5, 8))).normalize().coords)
[(5, 5), (5, 8), (8, 8), (8, 5), (5, 5)]
```

jorisvandenbossche:
Closes #994

This exposes `GEOSNormalize` as a method on the geometry classes. In GEOS, this works in place, but to keep the Shapely interface consistent (other operations like simplify also return a new geometry), I made the `normalize()` method to return a new geometry (and thus first clone the geometry before normalizing it).

This seems to work for me for some simple cases. cc @tomplex @Lucidiot  

Especially given the overlay changes in GEOS 3.9.0 (where you often first need to normalize the geometry if you want to be able to compare it to an expected result across different GEOS versions), I think it would be useful to still include this in Shapely 1.8.

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -649,7 +649,7 @@ class BaseGeometry(object):
         >>> from shapely.wkt import loads
         >>> p = loads("MULTILINESTRING((0 0, 1 1), (3 3, 2 2))")
         >>> p.normalize().wkt
-        "MULTILINESTRING ((2 2, 3 3), (0 0, 1 1))"
+        'MULTILINESTRING ((2 2, 3 3), (0 0, 1 1))'
         """
         # self.impl['normalize'](self)
         if self._geom is None:
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -637,6 +637,27 @@ class BaseGeometry(object):
             op = self.impl['simplify']
         return geom_factory(op(self, tolerance))
 
+    def normalize(self):
+        """Converts geometry to normal form (or canonical form).
+
+        This method orders the coordinates, rings of a polygon and parts of
+        multi geometries consistently. Typically useful for testing purposes
+        (for example in combination with `equals_exact`).
+
+        Examples
+        --------
+        >>> from shapely.wkt import loads
+        >>> p = loads("MULTILINESTRING((0 0, 1 1), (3 3, 2 2))")
+        >>> p.normalize().wkt
+        "MULTILINESTRING ((2 2, 3 3), (0 0, 1 1))"
+        """
+        # self.impl['normalize'](self)
+        if self._geom is None:
+            raise ValueError("Null geometry supports no operations")
+        geom_cloned = lgeos.GEOSGeom_clone(self._geom)
+        lgeos.GEOSNormalize(geom_cloned)
+        return geom_factory(geom_cloned)
+
     # Binary operations
     # -----------------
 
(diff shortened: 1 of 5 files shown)


shapely/geometry/base.py
```diff
@@ -637,6 +637,27 @@ def simplify(self, tolerance, preserve_topology=True):
...
+        This method orders the coordinates, rings of a polygon and parts of
+        multi geometries consistently. Typically useful for testing purposes
+        (for example in combination with `equals_exact`).
+
+        Examples
+        --------
+        >>> from shapely.wkt import loads
+        >>> p = loads("MULTILINESTRING((0 0, 1 1), (3 3, 2 2))")
+        >>> p.normalize().wkt
+        "MULTILINESTRING ((2 2, 3 3), (0 0, 1 1))"
```
mwtoews, line 652:
```suggestion
        'MULTILINESTRING ((2 2, 3 3), (0 0, 1 1))'
```

mwtoews (approved):
This feature should be welcome for 1.8.

mwtoews:
Should `normalize` should be a method property or regular method? E.g.:
* Method property: `line.normalize` returns a geometry
* Regular method: `line.normalize()` returns a geometry

I think a method property might be more fitting, since there are no arguments, and there are already many other method properties (e.g.) `centroid` returns a geometry.

tomplex:
I think `centroid` makes sense as a "property" because it's a geometric attribute of the geometry, whereas `normalize` is an operation on the geometry which is doing something to it, so a traditional method call feels right to me. Just my two cents though.

sgillies (approved):
Approve! Thank you @jorisvandenbossche !

PR #1094, opened 2021-03-02
A general purpose STRtree which can store any Python object

sgillies:
This STRtree can store almost anything and is backwards compatible with the one in 1.7. If initialized with a sequence of geometry objects, these are stored, and query and nearest return a list of geometries or a single geometry. If initialized with a sequence of `(value, geom)` pairs, the values are stored, and query and nearest return lists of these values or a single value. The values could be any Python object and don't need to be the same type.

In #1064 we have an STRtree that takes a sequence of geometry objects and stores their enumeration indices. Query returns a list of indices or a list of geometries (found by index from an internal list) depending on a flag.

The STRtree in #1064 and in this PR each have internal lists maintaining references to the initial data.

My understanding of the STRtree in #1064 is that it is designed to take a sequence of geometry objects (a DataSeries, yes?) from GeoPandas and store their enumeration indices. Syncronization of state between the enumeration indices in the tree and the DataFrame row indices would be required. The way GeoPandas would use the #1094 STRtree is to store DataFrame row indices directly. I believe this would make synchronization less fragile.

### DIFF
--- a/shapely/strtree.py
+++ b/shapely/strtree.py
@@ -35,9 +35,8 @@ def nearest_callback(func):
         value = ctypes.cast(arg1, ctypes.py_object).value
         geom = ctypes.cast(arg2, ctypes.py_object).value
         dist = ctypes.cast(arg3, ctypes.POINTER(ctypes.c_double))
-        userdata = ctypes.cast(arg4, ctypes.py_object).value
         try:
-            dist.contents.value = func(value, geom, userdata)
+            dist.contents.value = func(value, geom)
             return 1
         except Exception:
             log.exception()
@@ -49,8 +48,7 @@ def query_callback(func):
     @wraps(func)
     def wrapper(arg1, arg2):
         value = ctypes.cast(arg1, ctypes.py_object).value
-        userdata = ctypes.cast(arg2, ctypes.py_object).value
-        func(value, userdata)
+        func(value)
     return wrapper
 
 
@@ -105,62 +103,69 @@ class STRtree:
     """
 
     def __init__(self, initdata=None):
+        """Create a new STRtree.
+
+        Parameters
+        ----------
+        initdata : iterable
+            An iterable sequence of single geometry objects or (geom,
+            value) tuples.
+
+        Notes
+        -----
+        Items from initdata will be stored in two lists.
+
+        """
         warn(
             "STRtree will be completely changed in 2.0.0. The exact API is not yet decided, but will be documented before 1.8.0",
             ShapelyDeprecationWarning,
             stacklevel=2,
         )
         self._initdata = None
-        self._tree_handle = None
-        if initdata is not None:
-            self._initdata = []
-            for obj in initdata:
-                if not isinstance(obj, tuple):
-                    geom = obj
-                    value = obj
-                else:
-                    geom, value = obj
-                if not geom.is_empty:
-                    self._initdata.append(obj)
-
-        self._init_tree_handle(self._initdata)
-
-    def _init_tree_handle(self, initdata):
-        node_capacity = 10
-        self._tree_handle = lgeos.GEOSSTRtree_create(node_capacity)
+        self._tree = None
         if initdata:
-            for obj in initdata:
-                if not isinstance(obj, tuple):
-                    geom = obj
-                    value = obj
-                else:
-                    geom, value = obj
-                if not geom.is_empty:
-                    lgeos.GEOSSTRtree_insert(
-                        self._tree_handle, geom._geom, ctypes.py_object(value)
-                    )
+            self._initdata = list(self._iteritems(initdata))
+            self._init_tree(self._initdata)
+
+    def _iteritems(self, initdata):
+        for obj in initdata:
+            if not isinstance(obj, tuple):
+                geom = obj
+                value = obj
+            else:
+                geom, value = obj
+            if not geom.is_empty:
+                yield geom, value
+
+    def _init_tree(self, initdata):
+        if initdata:
+            node_capacity = 10
+            self._tree = lgeos.GEOSSTRtree_create(node_capacity)
+            for geom, value in self._iteritems(initdata):
+                lgeos.GEOSSTRtree_insert(
+                    self._tree, geom._geom, ctypes.py_object(value)
+                )
 
     def __getstate__(self):
         state = self.__dict__.copy()
-        del state["_tree_handle"]
+        del state["_tree"]
         return state
 
     def __setstate__(self, state):
         self.__dict__.update(state)
-        self._init_tree_handle(self._initdata)
+        self._init_tree(self._initdata)
 
     def __del__(self):
-        if self._tree_handle is not None:
+        if self._tree is not None:
             try:
-                lgeos.GEOSSTRtree_destroy(self._tree_handle)
+                lgeos.GEOSSTRtree_destroy(self._tree)
             except AttributeError:  # lgeos might be empty on shutdown.
                 pass
-            self._tree_handle = None
+            self._tree = None
 
     def query(self, geom):
         """
-        Search the index for geometry objects whose extents
-        intersect the extent of the given object.
+        Search the tree for nodes which intersect geom's envelope
 
         Parameters
         ----------
@@ -169,18 +174,17 @@ class STRtree:
 
         Returns
         -------
-        list of geometry objects
-            All the geometry objects in the index whose extents
-            intersect the extent of `geom`.
+        list
+            A list of the values stored at the found nodes. The list
+            will be empty if the tree is empty.
 
         Note
         ----
-        A geometry object's "extent" is its the minimum xy bounding
+        A geometry object's envelope is its the minimum xy bounding
         rectangle.
 
         Examples
         --------
-
         A buffer around a point can be used to control the extent
         of the query.
 
@@ -203,55 +207,42 @@ class STRtree:
         >>> [o.wkt for o in tree.query(query_geom) if o.intersects(query_geom)]
         ['POINT (2 2)']
 
-        To get the original indices of the returned objects, create an
-        auxiliary dictionary. But use the geometry *ids* as keys since
-        the shapely geometry objects themselves are not hashable.
-
-        >>> index_by_id = dict((id(pt), i) for i, pt in enumerate(points))
-        >>> [(index_by_id[id(pt)], pt.wkt) for pt in tree.query(Point(2,2).buffer(1.0))]
-        [(1, 'POINT (1 1)'), (2, 'POINT (2 2)'), (3, 'POINT (3 3)')]
-
         """
         result = []
 
         @query_callback
-        def callback(value, userdata):
+        def callback(value):
             result.append(value)
 
         self.query_cb(geom, callback)
-
         return result
 
     def query_cb(self, geom, callback):
-        """Search for tree nodes intersecting geom and execute the
-        callback function for each
+        """Searches the tree for nodes and applies a callback function.
 
         Parameters
         ----------
         geom : Geometry
             The query geometry
         callback : callable
-            This is a function that takes two arguments, value and
-            userdata, and which is decorated by
-            shapely.strtree.query_callback. See STRtree.query() for an
-            example. Value is the value stored with the tree node.
-            Support for userdata is not implemented.
+            This is a function which takes a node value as argument and
+            which is decorated by shapely.strtree.query_callback. See
+            STRtree.query() for an example.
 
         Returns
         -------
         None
 
         """
-        if not self._initdata:
+        if self._tree is None or not self._initdata:
             return
 
         lgeos.GEOSSTRtree_query(
-            self._tree_handle, geom._geom, lgeos.GEOSQueryCallback(callback), None
+            self._tree, geom._geom, lgeos.GEOSQueryCallback(callback), None
         )
 
     def nearest(self, geom):
-        """
-        Get the nearest object in the index to a geometry object.
+        """Finds the tree node nearest to a given geometry object.
 
         Parameters
         ----------
@@ -260,13 +251,9 @@ class STRtree:
 
         Returns
         -------
-        geometry object
-            The nearest geometry object in the index to `geom`.
-
-            Will always only return *one* object even if several
-            in the index are the minimum distance away.
-
-            `None` if the index is empty.
+        object or None
+            The value of the tree node nearest to geom or None if the
+            index is empty.
 
         Examples
         --------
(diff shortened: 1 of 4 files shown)


shapely/strtree.py
```diff
@@ -172,6 +173,11 @@ def callback(item, userdata):
 
         return result
 
+    def query_cb(self, geom, callback):
```
sgillies, line 176: Docstring is a TODO.

jorisvandenbossche:
@sgillies thanks for getting back to this!

Now, I think I don't yet fully understand how this `query_cb` would be used in practice by the user. 
First, it requires manual usage of `ctypes` (and knowing the types of the pointers that the callback function receives), and I am not sure if we want that users have to deal with ctypes? (certainly given that we are removing the use of ctypes in Shapely 2.0).
Further, I don't think this actually allows to get an integer index as the result? The callback function can only work with the data that was inserted into the tree, and at the moment that is still hardcoded to be the actual geometry object.

sgillies:
Indeed I need to think more about the callback. We can't ask users to navigate the Python/C issues. Perhaps requiring callbacks to be a method of an Extension type would help, they could use a helper method of that type to do conversions. I'll stop waving my hands now and come up with some running code.

sgillies:
Finally got the time to get back to this and overcome some of the issues. First, this PR will let us store any Python objects we want in the tree. Geometries, numbers, strings. Even a mix, which is not something one would normally do. The query_cb function takes an ordinary Python function as a callback and all the conversion between Python and C types is done by a decorator. A related decorator is used in STRtree.nearest, which also now supports trees that contain non-geometry objects.

shapely/strtree.py
```diff
@@ -19,11 +19,40 @@
...
+    return wrapper
+
+
+def query_callback(func):
+    @wraps(func)
+    def wrapper(arg1, arg2):
+        value = ctypes.cast(arg1, ctypes.py_object).value
+        userdata = ctypes.cast(arg2, ctypes.py_object).value
+        func(value, userdata)
+    return wrapper
```
sgillies, line 54: These need docs if we're going to go in this direction.

shapely/strtree.py
```diff
@@ -75,26 +104,41 @@ class STRtree:
     None
     """
 
-    def __init__(self, geoms):
+    def __init__(self, initdata=None):
```
sgillies, line 107: Not just geometries anymore. Docs needed.

shapely/strtree.py
```diff
@@ -75,26 +104,41 @@ class STRtree:
...
+        if initdata is not None:
+            self._initdata = []
+            for obj in initdata:
+                if not isinstance(obj, tuple):
+                    geom = obj
+                    value = obj
+                else:
+                    geom, value = obj
+                if not geom.is_empty:
+                    self._initdata.append(obj)
```
sgillies, line 124: Still saving a list of object references to keep things in the tree from getting GC'd.

shapely/strtree.py
```diff
@@ -167,20 +210,45 @@ def query(self, geom):
...
 
+        """
         result = []
 
-        def callback(item, userdata):
-            geom = ctypes.cast(item, ctypes.py_object).value
-            result.append(geom)
+        @query_callback
+        def callback(value, userdata):
+            result.append(value)
```
sgillies, line 219: Python callback function with our special decorator.

shapely/strtree.py
```diff
@@ -213,24 +281,29 @@ def nearest(self, geom):
...
-            lgeos.GEOSDistanceCallback(callback), None)
+        try:
+            geoms, values = zip(*self._initdata)
+        except TypeError:
+            geoms = self._initdata
+            values = self._initdata
+
+        @nearest_callback
+        def callback(value, geom, userdata):
+            value_geom = geoms[values.index(value)]
```
sgillies, line 298: When non-geometry objects are stored in the tree we need to look up the geometries they were inserted with.

tests/test_strtree.py
```diff
@@ -9,20 +10,79 @@
...
-    results = tree.query(Point(2, 2).buffer(1.0))
-    assert len(results) == 3
+@pytest.mark.parametrize("geoms", [[Point(i, i) for i in range(10)]])
+@pytest.mark.parametrize(
+    "query_geom,num_results", [(Point(2, 2).buffer(0.99), 1), (Point(2, 2).buffer(1.0), 3)]
+)
+def test_query(geoms, query_geom, num_results):
+    tree = STRtree(geoms)
+    results = tree.query(query_geom)
+    assert len(results) == num_results
```
sgillies, line 26: Made a bunch of test improvements that helped in debugging.

tests/test_strtree.py
```diff
@@ -27,6 +27,19 @@ def test_query(geoms, query_geom, num_results):
...
 
+@requires_geos_342
+@pytest.mark.parametrize("geoms", [[Point(i, i) for i in range(10)]])
+@pytest.mark.parametrize(
+    "query_geom,expected",
+    [(Point(2, 2).buffer(0.99), [2]), (Point(2, 2).buffer(1.0), [1, 2, 3])],
+)
+def test_query_idx_values(geoms, query_geom, expected):
+    with pytest.warns(ShapelyDeprecationWarning):
+        tree = STRtree(enumerate(geoms))
```
sgillies, line 38: Here's an example of storing the index of the geometry.

tests/test_strtree.py
```diff
@@ -33,13 +33,28 @@ def test_query(geoms, query_geom, num_results):
...
+@requires_geos_342
+@pytest.mark.parametrize("geoms", [[Point(i, i) for i in range(10)]])
+@pytest.mark.parametrize(
+    "query_geom,expected",
+    [(Point(2, 2).buffer(0.99), [2]), (Point(2, 2).buffer(1.0), [1, 2, 3])],
+)
+def test_query_enumeration(geoms, query_geom, expected):
+    """Store enumeration"""
+    with pytest.warns(ShapelyDeprecationWarning):
+        tree = STRtree(zip(enumerate(geoms), geoms))
```
sgillies, line 53: Also related to #1064. But instead of always only storing geometry indices in tree nodes and getting geometries through their index, we require the caller to explicitly store index and geometry.

shapely/strtree.py
```diff
@@ -35,9 +35,8 @@ def wrapper(arg1, arg2, arg3, arg4):
         value = ctypes.cast(arg1, ctypes.py_object).value
         geom = ctypes.cast(arg2, ctypes.py_object).value
         dist = ctypes.cast(arg3, ctypes.POINTER(ctypes.c_double))
-        userdata = ctypes.cast(arg4, ctypes.py_object).value
         try:
-            dist.contents.value = func(value, geom, userdata)
+            dist.contents.value = func(value, geom)
```
sgillies, line 39: I honestly can't think of a use for the userdata argument.

sgillies:
@jorisvandenbossche @mwtoews I've made some major improvements and have rewritten the introductory comment in this PR to better explain the intent and direction. I think this implementation covers existing features, the features in #1064, and more.

jorisvandenbossche:
Thanks, with those updates, the intent and how it would be used is much clearer now ;) (and avoids using ctypes as a user!)

A few follow-up questions / comments:

- Do you have use cases in mind where the user might want to get something else returned as either the geometry itself or either some ID? And a similar question for the ability to specify a custom callback function. 
  I am mainly looking at this from the point of view of what we need in geopandas (where we wouldn't need that), so this is of course a quite limited view! But at the same time, I think we should be cautious in implementing features "because it's possible", and think about practical use case if adding this. 
  (Because adding such flexible features also has a cost. We should probably need to try it out to know for sure, but my feeling says that implementing the same features in C for 2.0 would give quite some additional complexity.) 

- Regarding the ability to "store any Python objects we want in the tree": I think I mentioned this before in https://github.com/Toblerity/Shapely/pull/1064, but I want to mention again that to achieve this functionality, those objects don't necessarily have to be stored in the *GEOS C tree*, but could also be stored in the *Python tree object*. 
  If we want to enable this functionality for the end users, I think it is possible to keep track of the values on the python side, and still only store the actual geometries in the GEOS STRtree (as is currently done in pygeos). And based on the indices returned from this wrapped C tree, the *python* layer can still return any Python object we want.  


- I know that the GEOS STRtree explicitly allows storing generic values that get returned. But using Python objects for that instead of indices (or actually pointers, as currently done in pygeos), will still be more complicated, and for example won't allow to release the GIL for the query (I assume). Also a generic callback function that calls a user-provided python callback won't be able to release the GIL, so that might mean we need two different code paths to support both those cases.

jorisvandenbossche:
Only read the updated top post now, so a few comments/answers on that:

> In #1064 we have an STRtree that takes a sequence of geometry objects and stores their enumeration indices. Query returns a list of indices or a list of geometries (found by index from an internal list) depending on a flag.

A bit related to my comment above, but the actual API in that PR (different return value depending on a flag) is only "a" choice out of a set of options that *I* made when opening the PR, but is certainly something to be discussed. I think we are maybe discussing two (potentially) different aspects, which might be useful to separate. One is the actual implementation under the hood (i.e. how pygeos has implemented it in C / how #1064 does it in the current ctypes implementation), and the other is the user-facing API (eg the `query` method with a flag as in #1064, or a new `query_indices` method, or letting the user insert the value to be returned explicitly as in this PR, etc). 
My feeling is that we can separate those two aspects, and could implement any user facing API we want (maybe except the custom callback) on top of a low-level C STRtree that returns indices (which is probably the simplest to implement on that lower level).

> My understanding of the STRtree in #1064 is that it is designed to take a sequence of geometry objects (a DataSeries, yes?) from GeoPandas and store their enumeration indices. Syncronization of state between the enumeration indices in the tree and the DataFrame row indices would be required. The way GeoPandas would use the #1094 STRtree is to store DataFrame row indices directly. I believe this would make synchronization less fragile. 

The STRtree in #1064 just takes a sequence or iterable of geometries as input, which is exactly the same as the current version. So it's not specifically targetted at receiving a geopandas Series. 
Also, as a clarification, in geopandas we don't really want to store/get row indices (generic labels, which can be of any type), but rather integer positional indices (pandas allows you to index a dataframe in two ways: the actual row labels vs integer positions). This is because using integer positions is more performant (and also more robust, eg in theory you can have duplicate labels). 

I also don't think this is specific to geopandas, but more general when working with arrays of geometries. I know that eg @brendan-ward has mentioned he has already been using pygeos' strtree a lot in his work for fast spatial join like operations.

brendan-ward:
Thanks @sgillies for your work here and @jorisvandenbossche for keeping the conservation going!

I am likely biased, but I think keeping the underlying C STRtree limited to enumeration indices gives us a simpler API at the C level, avoids memory traps and GIL-related issues, and yet still provides a very flexible API at the Python level.  Those indices can always be used to return arbitrary values (e.g., original index for a DataFrame, the geometry at each index, some other value) at the Python level: just keep an indexable array (i.e., numpy array of dtype appropriate for values) of those values in the Python tier, and return those values using the enumeration indices depending on some parameter or a specific function layered on top of the C STRtree.

My hunch is similar to @jorisvandenbossche that storing other things directly in the C STRtree introduces a good bit more complexity and isn't necessary.  The slow parts are the underlying tree query and predicate operations against the results; indexing back into an arbitrary array of values afterward should be very fast.  I.e., we probably wouldn't see a performance improvement by storing the values you want back directly in the C tree (those may actually have a performance impact depending on how they are tracked there).

I have been using the enumeration indices heavily for a variety of different operations; even if the geometries originate from DataFrames or arrays, sometimes I just use these directly, sometimes I use them to get the original DataFrame index, and sometimes I use them to get the geometries or something else.  There is little overhead to using enumeration indices to get to any of those.  So I'm a huge fan of the enumeration indices approach at the core.

> we can separate those two aspects, and could implement any user facing API we want (maybe except the custom callback) on top of a low-level C STRtree that returns indices

I concur with this; the substantive part of the API discussion should be at the Python level, and there should be easy ways to return either the enumeration indices, the geometries, or arbitrary values based on the underlying enumeration indices.  This can also be a rich source of functionality in Shapely and be easier for contributors than the underlying implementation in C.  Both this and #1064 have elaborated the need for multiple types of results from the tree.


I'm unclear on the use cases around custom callbacks implemented in Python.  My first instinct is that I'd really want to keep these at the C level.  There are issues around memory management and releasing the GIL lurking here.  My preference would be to identify clear use cases for other callbacks that are needed and implement those at the C level and expose those through a well-defined API (also based on enumeration indices) back to the Python tier.

Now that predicates are implemented in `query` and `query_bulk` methods at the C level, I'm having a hard time thinking of other things I'd want to handle in a tree query callback.  The tree doesn't allow us to exit early on first result (which could be helpful for certain analyses), so other than tracking the results that you get back from the underlying `GEOSSTRtree_query_r` or some derivative of input geometries or values that you could do after getting enumeration indices from the tree, I don't know what else you'd do in a callback.

Likewise, the custom callback we had to work out in pygeos to support `nearest_all` (return all nearest equidistant results and not just the first, limited by an optional max distance) was a bit tricky and definitely needed to be done at the C level.  

What are some of the use cases you had in mind for custom callbacks?

sgillies:
@brendan-ward @jorisvandenbossche does storing only enumeration indices require the tree to be immutable? In #1104 I'm considering a tree that supports insertion and removal of items. It seems to me that in #1064 insertion must then be implemented as a strict append to the internal list so that indices already stored in the tree remain valid. I'm not sure if that breaks any pygeos requirements. Internally, item removal would be less constrained. If the item is left in the internal array/list all the indices remain valid.

GEOS could be modified to allow us to break out of the loop at https://github.com/libgeos/geos/blob/master/src/index/strtree/SimpleSTRtree.cpp#L226 but until then there's no great way to return early using a callback. My interest in the callbacks was coming from an interest in seeing if we could shift responsibility from mapping pointer addresses stored in the tree to Python objects onto the caller. But if we're going to release the GIL we cannot do that. Maybe it's best to shelve the callback idea for now. If we do change GEOS to allow an early return, we could add a callback keyword argument to query and switch to a branch that doesn't release the GIL. Yes?

brendan-ward:
@sgillies It is my understanding that STRtree trees are [immutable after first query](http://geos.refractions.net/ro/doxygen_docs/html/classgeos_1_1index_1_1strtree_1_1STRtree.html):
> once the tree has been built (explicitly or on the first call to query), items may not be added or removed

I know there are ongoing changes to the STRtree implementation, but I wasn't aware of any change to that constraint.

Pygeos expects the tree inputs to be immutable.

However, in my experience, building trees is very fast even for large (>1M) datasets, so I think it is entirely feasible to simply discard and rebuild the tree instead of trying to handle inserts.  Ideally, mutations should be batched to avoid rebuilding the tree on every change.

What is the use case for adding / removing items from the tree after building it?


If GEOS STRtree is modified, this would be through the addition of new functions against STRtree, so support would need to be built into the C wrapping in pygeos / shapely at that time - and that would be the place to provide an additional function to the Python tier to handle this case.  In Python, it could be exposed to users as a keyword parameter if appropriate, but use different underlying C functions.  This still allows us to leverage address-based indexing, release the GIL, and keep that complexity at the C level.

Issue #1035, opened 2020-11-19
Raise a consistent error for expected geometry type


mwtoews:
Throughout Shapely, there is a need to raise an error if a given geometry does not have the correct geometry type. Currently, this is done inconsistently, raising either [`TypeError`](https://docs.python.org/library/exceptions.html#TypeError) or [`ValueError`](https://docs.python.org/library/exceptions.html#ValueError).

There are a few options:
1. Do nothing
2. Convert these to `ValueError`, with the interpretation the `.geom_type` value is the issue
3. Convert these to `TypeError`, with the interpretation that the geometry type is the issue
4. Create a new `GeometryTypeError` or similar (in `shapely.errors`) as a subclass of either `ValueError` or `TypeError`

Thoughts?

jorisvandenbossche:
In PyGEOS, we currently have a specific PGERR_GEOMETRY_TYPE in the C code, which maps to a `TypeError` on the python side (https://github.com/pygeos/pygeos/blob/a524ae4eb590e736307c3054b30b9cd835eafb91/src/geos.h#L76-L79). This is used eg in `interpolate`. 
However, for other functions, I think we also let GEOS handle this, which then results in a GEOSException. 

@mwtoews do you have an idea of where in Shapely there currently already are such manual checks?

mwtoews:
Here is a partial list of cases that I see these inconsistencies.

AssertionError:
- [`shapely.ops.SplitOp`](https://github.com/Toblerity/Shapely/blob/6aeb8e7932076fc8700a3c36133255957989464c/shapely/ops.py#L406)
- [`shapely.ops.substring`](https://github.com/Toblerity/Shapely/blob/6aeb8e7932076fc8700a3c36133255957989464c/shapely/ops.py#L573) - to be addressed in #1008

TypeError:
- [`shapely.ops.shared_paths`](https://github.com/Toblerity/Shapely/blob/6aeb8e7932076fc8700a3c36133255957989464c/shapely/ops.py#L393)
- [`shapely.linref.LinearRefBase`](https://github.com/Toblerity/Shapely/blob/6aeb8e7932076fc8700a3c36133255957989464c/shapely/linref.py#L11)

ValueError:
- [`shapely.ops.transform`](https://github.com/Toblerity/Shapely/blob/6aeb8e7932076fc8700a3c36133255957989464c/shapely/ops.py#L325)
[...]

PR #1099, opened 2021-03-07
Raise consistent error for geometry type

rgangopadhya:
👋 thought it would be fun to contribute to this great library, having been an occasional consumer of it over the years and looking to learn more deeply about geospatial computing. 

Looking at https://github.com/Toblerity/Shapely/issues/1035, seemed to be consensus emerging on creating a `GeometryTypeError` and replacing that for the inconsistent uses of `AssertionError`, `TypeError`, and `ValueError`.

So, this attempts to do that -- please let me know if I am missing anything. I went through all the files. I _think_ this is exhaustive but make no guarantees as it was my first pass through.

Second commit is asking the question - do we consider passing a null geometry where it is not expected to be a `GeometryTypeError`? I believe it should be but not 100% sure and curious what people think. If we agree on that I'll make the change.


shapely/errors.py
```diff
@@ -41,3 +41,10 @@ class ShapelyDeprecationWarning(FutureWarning):
 
 class EmptyPartError(ShapelyError):
     """An error signifying an empty part was encountered when creating a multi-part."""
+
+
+class GeometryTypeError(ShapelyError):
+    """
+    An error raised when the type of the geometry in question is
+    unrecognized or inappropriate
```
rgangopadhya, line 49: Looking for feedback on this

sgillies: Excellent.

shapely/geometry/base.py
```diff
@@ -57,18 +57,20 @@ def dump_coords(geom):
...
     if g is None:
+        # Do we consider this a type error?
         raise ValueError("Null geometry has no type")
     return GEOMETRY_TYPES[lgeos.GEOSGeomTypeId(g)]
 
 
 def geom_factory(g, parent=None):
     # Abstract geometry factory for use with topological methods below
     if not g:
+        # Do we consider this a type error?
```
sgillies, line 73: This is not a type error. Also, geom_factory(), while not hidden from users, is for internal use and not part of the official public API. We can leave the ValueError here.

shapely/geometry/base.py
```diff
@@ -123,6 +125,7 @@ def exceptNull(func):
     @wraps(func)
     def wrapper(*args, **kwargs):
         if not args[0]._geom or args[0].is_empty:
+            # Do we consider this a type error?
```
sgillies, line 128: Same as above.

shapely/geometry/base.py
```diff
@@ -653,6 +656,7 @@ def normalize(self):
         """
         # self.impl['normalize'](self)
         if self._geom is None:
+            # Do we consider this a type error?
```
sgillies, line 659: This one we can leave as is. normalize() is part of the public API, yes. But this error will almost never be seen and exists to prevent a null C pointer from being passed to the GEOS function underneath. In Shapely 2.0 the implementation here will change quite a bit and null pointers will be less of a problem, or at least a different problem.

shapely/geos.py
```diff
@@ -514,6 +514,7 @@ def __del__(self):
     def write(self, geom):
         """Returns WKB byte string for geometry"""
         if geom is None or geom._geom is None:
+            # Do we consider this a type error?
             raise ValueError("Null geometry supports no operations")
```
sgillies, line 518: This suggests to me that we may benefit from an InvalidGeometryError .

shapely/geos.py
```diff
@@ -525,6 +526,7 @@ def write(self, geom):
     def write_hex(self, geom):
         """Returns WKB hex string for geometry"""
         if geom is None or geom._geom is None:
+            # Do we consider this a type error?
```
sgillies, line 529: Same as above.

shapely/linref.py
```diff
@@ -1,14 +1,15 @@
...
+from shapely.errors import GeometryTypeError
 from shapely.topology import Delegating
 
 
 class LinearRefBase(Delegating):
     def _validate_line(self, ob):
         super(LinearRefBase, self)._validate(ob)
         if not ob.geom_type in ['LinearRing', 'LineString', 'MultiLineString']:
-            raise TypeError("Only linear types support this operation")
+            raise GeometryTypeError("Only linear types support this operation")
```
sgillies, line 12: Yes.

shapely/topology.py
```diff
@@ -15,6 +15,7 @@ class Validating(object):
 
     def _validate(self, ob, stop_prepared=False):
         if ob is None or ob._geom is None:
+            # Do we consider this a type error?
```
sgillies, line 18: InvalidGeometryError again, I think.

sgillies (commented):
@rgangopadhya thank you for the contribution! This is completely in the spirit of #1035 . I have a few suggestions inline.

The master branch of Shapely will become version 1.8.0 and strictly speaking the changes here would break backwards compatibility. I have an idea, however. What would you think about making GeometryTypeError, for now, derive from TypeError and ValueError, and add a warning from GeometryTypeError.__init__()? I will make a suggestion inline about this.

shapely/errors.py
```diff
@@ -41,3 +41,10 @@ class ShapelyDeprecationWarning(FutureWarning):
...
+    """
```
sgillies, line 50:
```suggestion
class GeometryTypeError(TypeError, ValueError):
    """
    An error raised when the type of the geometry in question is
    unrecognized or inappropriate.
    """
    def __init__(self, msg):
        warnings.warn("", ShapelyDeprecationWarning, stacklevel=2)
        super()
```

rgangopadhya: Ah, good point! Wasn't thinking about backwards compatibility. Makes sense to me 👍

rgangopadhya:
Thanks for the review @sgillies ! I'll make suggested changes shortly

shapely/errors.py
```diff
@@ -43,8 +43,11 @@ class EmptyPartError(ShapelyError):
...
 
-class GeometryTypeError(ShapelyError):
+class GeometryTypeError(TypeError, ValueError):
     """
     An error raised when the type of the geometry in question is
-    unrecognized or inappropriate
+    unrecognized or inappropriate.
     """
+    def __init__(self, msg):
+        warnings.warn("GeometryTypeError will derive from ShapelyError and not TypeError or ValueError in Shapely 2.0.", ShapelyDeprecationWarning, stacklevel=2)
```
sgillies, line 52: @rgangopadhya looks like `import warnings` is needed at the top of errors.py.

rgangopadhya: Sorry, yeah. Spent like 10 mins on this last week and then got distracted by other things -- this week

rgangopadhya:
@sgillies sorry for the delay, ready for another pass

sgillies (approved):
@rgangopadhya great, thank you!

mwtoews (changes requested):
Thanks for getting the time to get to this, it looks great! Just a few requested changes to list here.

There are a few other exceptions with the message "Null geometry supports no operations" that should also raise `InvalidGeometryError`:

- shapely/geometry/base.py line 65 in normalize
- shapely/geos.py line 397 in WKTWriter.write

The only other specific recommendation is in shapely/speedups/_speedups.pyx line 576 should raise `GeometryTypeError`

jorisvandenbossche:
We could keep `GeometryTypeError` to be a `TypeError` subclass in the future? (only remove the ValueError in the future) 
Given that it's a kind of type error (as the name says itself), I think it would make sense to have it a TypeEror subclass.

shapely/geometry/geo.py
```diff
@@ -117,7 +118,7 @@ def shape(context):
         geoms = [shape(g) for g in ob.get("geometries", [])]
         return GeometryCollection(geoms)
     else:
-        raise ValueError("Unknown geometry type: %s" % geom_type)
+        raise GeometryTypeError("Unknown geometry type: %s" % geom_type)
```
jorisvandenbossche, line 121: I think this actually could make more sense as a ValueError, as this is about the *value* in the "type" key of the `__geo_interface__` dictionary (`context["type"]`)

shapely/linref.py
jorisvandenbossche: In PyGEOS, this actually returns a GEOSException, xref the discussion in https://github.com/Toblerity/Shapely/issues/991

shapely/ops.py
```diff
@@ -391,9 +391,9 @@ def shared_paths(g1, g2):
         The second geometry
     """
     if not isinstance(g1, LineString):
-        raise TypeError("First geometry must be a LineString")
+        raise GeometryTypeError("First geometry must be a LineString")
```
jorisvandenbossche, line 394: Same here about GEOSException

mwtoews: these are exceptions to the GIS methods (i.e. linear referencing is based on linear geometries, etc.), so I think they are an appropriate error type. (But yes, these are generally GEOS exceptions too).

rgangopadhya:
Thanks for the feedback @mwtoews and @jorisvandenbossche... I see now there is a lot of discussion on this topic which I hadn't seen before :) I'll take a look this week

shapely/ops.py
jorisvandenbossche: @mwtoews what I mean is that in PyGEOS, we let GEOS handle this case and bubble up the exception from GEOS as a GEOSException. If we are fine with doing that (xref #991), it doesn't necessarily make sense to change this error now to GeometryTypeError?

sgillies: @jorisvandenbossche I'm not fine with doing that in this case. If a call to a GEOS function is always going to fail due to an unusable type of geometry, the GEOS function shouldn't be called. An exception should be raised immediately instead. And I think an exception that's more specific than GEOSException should be raised. And I do see this pattern also used in PyGEOS, like at https://github.com/pygeos/pygeos/blob/63b474081352f4eab362a3b48e48c23328fcb3bd/pygeos/_geometry.pyx#L74, where a TypeError is raised instead of calling a GEOS geometry constructor that can't succeed.

rgangopadhya: Hmm, this seems to be a long-running discussion from #991 and other threads. To summarize my understanding of the discussion, a key question is whether we should let GEOS errors happen and bubble these back up to the user. However, it sounded like we weren't necessarily clear on whether different types of GEOS errors would be converted into different Python error classes, thus allowing for differential handling on the user side. Barring doing that, it makes more sense to me to handle errors in Shapely code before sending over to GEOS.

rgangopadhya:
@mwtoews thanks for the requests! I think I got 'em. Ready for another pass

sgillies:
Thank you @rgangopadhya !

mwtoews (approved):
Thanks for edits and patience @rgangopadhya. This looks good to go now. However, I'm cautious to see how much noise the new warnings will make. If it's too much, we could revise this aspect between merging and final release.

Issue #1102, opened 2021-03-12
Tests in test_wkb fail on big-endian platforms, such as s390x


musicinmybrain:
## Expected behavior and actual behavior.

Expect that all tests should pass; but most tests in `test_wkb.py` fail on s390x, which is big-endian.

## Steps to reproduce the problem.

Run the tests on a big-endian platform.

## Operating system

Fedora Rawhide (35) on s390x

## Shapely version and provenance

1.7.1 built from the GitHub release tarball

## Additional information

The problem is in the tests themselves, which contain little-endian expected outputs. The following downstream issue explains the problem clearly, with actual output from the failing tests: https://bugzilla.redhat.com/show_bug.cgi?id=1937719

I am one of the co-maintainers of the `python-shapely` package in Fedora Linux.

A PR to fix the tests will follow shortly.

PR #1103, opened 2021-03-12
Fix failures in test_wkb on big-endian hosts (#1102)

musicinmybrain:
Fixes https://github.com/Toblerity/Shapely/issues/1102.

Adds a `hostorder()` function in the test module to re-pack expected outputs in host endianness.

An alternative solution would be to manually compute the corresponding big-endian vectors, and branch on `sys.byteorder` in each test to decide which one to compare against. There are probably other reasonable alternative solutions, too.

This fix is verified on s390x, a big-endian architecture.

### DIFF
--- a/tests/test_wkb.py
+++ b/tests/test_wkb.py
@@ -1,4 +1,6 @@
 import binascii
+import struct
+import sys
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


sgillies:
Thank you very much @musicinmybrain !

Issue #1105, opened 2021-03-18
Bug when using Shapely in Windows conda


philippfischer-sd:
### Reproducing the Bug
The bug is very easy to reproduce by following these steps:

```bash
# Create a conda environment in Windows
conda create -n shapelybugpip python=3.8
conda activate shapelybugpip

# Inside the environment, use pip to install Shapely
pip install Shapely

# Then import some part of shapely that uses the geos library
python -c "import shapely.geometry"
```

### Resulting Error
```py
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\...\Miniconda3\envs\shapelybugpip\lib\site-packages\shapely\geometry\__init__.py", line 4, in <module>
    from .base import CAP_STYLE, JOIN_STYLE
  File "C:\...\Miniconda3\envs\shapelybugpip\lib\site-packages\shapely\geometry\base.py", line 19, in <module>
    from shapely.coords import CoordinateSequence
  File "C:\...\Miniconda3\envs\shapelybugpip\lib\site-packages\shapely\coords.py", line 8, in <module>
    from shapely.geos import lgeos
  File "C:\...\Miniconda3\envs\shapelybugpip\lib\site-packages\shapely\geos.py", line 154, in <module>
    _lgeos = CDLL(os.path.join(sys.prefix, 'Library', 'bin', 'geos_c.dll'))
  File "C:\...\Miniconda3\envs\shapelybugpip\lib\ctypes\__init__.py", line 381, in __init__
    self._handle = _dlopen(self._name, mode)
FileNotFoundError: Could not find module 'C:\...\Miniconda3\envs\shapelybugpip\Library\bin\geos_c.dll' (or one of its dependencies). Try using the full path with constructor syntax.
```

### How to Fix
The reason is an error in this function:
https://github.com/Toblerity/Shapely/blob/8ef5b092750e3499893f990ab6498dca17dc2c80/shapely/geos.py#L59

The function doesn't actually do what is says. It simply checks **if there is a conda environment**, but not **if shapely is installed as a conda package or as a pip package inside that environment**.

There are multiple ways to fix it:

1. Don't use a hard if-else in this line https://github.com/Toblerity/Shapely/blob/8ef5b092750e3499893f990ab6498dca17dc2c80/shapely/geos.py#L159 but instead use the conda path as an additional search path (should be first), or if loading the conda lib fails, fallback to the pip loader below.
[...]

Issue #924, opened 2020-06-05
pip install shapely does not work on Windows


FenryrMKIII:
Using an anaconda environment, doing `pip install shapely` does not work. It throws the error that `geos_c.dll` is not found.

Shapely version : 1.7.0
pip version : 20.1.1
Python version : 3.8.3

jorisvandenbossche:
If you are using ananconda, you can do `conda install shapely` ?

See https://github.com/Toblerity/Shapely/pull/843 for  some discussion on pip installing in a conda environment on Windows (this was broken, and is fixed now in master, but not yet released)

FenryrMKIII:
for some reasons, I would like to use pip

jorisvandenbossche:
Note this is supported in master (and will be in the next release), so that quoted comment is no longer up to date (in practice of course it's still the case with the releaed version of shapely, but not the intention of the comment)

PR #1108, opened 2021-03-19
Only load conda GEOS dll if it exists (on Windows)

jorisvandenbossche:
@sgillies I know you have said multiple times that you don't want to fix this .., but 1) it's a small change, 2) we basically do this for Linux/Mac as well (by first checking the wheel path) and 3) solves troubling import issues that can easily be avoided. 
So therefore I don't see much reason to *not* do this for Shapely 1.8 (it will still take a while until 2.0 is fully stabilized and released)

Related discussion in https://github.com/Toblerity/Shapely/pull/843#discussion_r386275247 (and also #1105, #924)

### DIFF
--- a/shapely/geos.py
+++ b/shapely/geos.py
@@ -156,9 +156,10 @@ elif sys.platform == 'darwin':
     free.restype = None
 
 elif sys.platform == 'win32':
-    if exists_conda_env():
+    _conda_dll_path = os.path.join(sys.prefix, 'Library', 'bin', 'geos_c.dll')
+    if exists_conda_env() and os.path.exists(_conda_dll_path):
         # conda package.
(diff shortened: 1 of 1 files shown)


sgillies:
Great, thanks @jorisvandenbossche ! I've got no technical objections, only no time and no easy access to Windows. So if you say this works, let's do it.

jorisvandenbossche:
Thanks!

philippfischer-sd:
> Great, thanks @jorisvandenbossche ! I've got no technical objections, only no time and no easy access to Windows. So if you say this works, let's do it.

As mentioned [here](https://github.com/Toblerity/Shapely/issues/1105#issuecomment-803786627), I quickly tested this fix. I didn't build the wheel from scratch, as that would involve installing GEOS etc. first, but I simply started from a 1.7.0 installation and replaced the code with the master branch (that contains the above fix) in-place. After that it worked for me.

PR #1109, opened 2021-03-21
MAINT: use super() as described by PEP 3135

mwtoews:
This PR refactors `super()` to be simpler, as described by [PEP 3135](https://www.python.org/dev/peps/pep-3135/) and approved in 2007 for Python 3.0.

The "old" convention (e.g. `super(C, self)`) was only required to support Python 2.


sgillies:
Yes! Thanks @mwtoews. This reminds me that I overlooked mentioning #884 in the change log. I'll fix that.

PR #1110, opened 2021-03-25
Doc buffer single side

ThomasG77:
Added doc in the manual about buffer function `single_sided` option merged in #806 with doc at API level but not in the manual


sgillies (changes requested):
Thank you @ThomasG77 ! Do you have the capability of making a screengrab of the resulting doc change?

ThomasG77:
Before

before-buffer-one-side

After

after-buffer-one-side

sgillies:
Thank you @ThomasG77 ! Eventually we may want to build docs for commits, but that isn't on the roadmap yet.

PR #1111, opened 2021-03-30
Wheels are only universal if they support Python 2 & 3

michael-k:
ref Toblerity/Shapely#743


sgillies (approved):
Thanks! We don't build wheels for PyPI from this repo anymore, which explains why this hasn't been changed.

Issue #1106, opened 2021-03-19
How to use strtree to find an element nearest to it but not its own?


liunx7594:
## Expected behavior and actual behavior.

How to use strtree to find an element nearest to it but not its own?My code is as follows, what I expect to find is (1,1), not (0,0).
If I want to start from (0,0), find the nearest point (1,1), and then find the second nearest point (2,2), how do I operate?

## Steps to reproduce the problem.

from shapely.strtree import STRtree
from shapely.geometry import Point

points=[Point(0,0),Point(1,1),Point(2,2)]

tree=STRtree(points)

pos=tree.nearest(Point(0,0))
print(pos)

#POINT (0 0)

## Operating system

windows 10    20H2
python3.7.0 64-bit (conda)

## Shapely version and provenance

Shapely-1.7.1 from pip install

sgillies:
I'm sorry to say @liunx7594 that the tree's nearest() method is behaving correctly here. However, I do have an idea for adding a new method to the STRtree API in version 1.8.0, one that would allow you to exclude the query geometry from the result.

liunx7594:
OK, I'm looking forward to the next features

liunx7594:
By the way, it's really a great library, and thank you so much for creating it

tomplex:
I [wrote a gist](https://gist.github.com/tomplex/0799e07c0a16381a32e85413be0102a7) solving a similar problem someone posted here, might help you as well if I'm understanding your issue

PR #1112, opened 2021-03-30
STRtree with a reverse mapping and query_items/query_geoms methods

sgillies:
This is very much like #1064. It stores only integer items. Exposes no callbacks. It's backwards compatible with the STRtree in shapely 1.7.0.

The differences from #1064:

- This tree can be initialized from a sequence of geometries (as #1064 and previous versions) but can also take a sequence of integer items to support applications that have their own integer identifiers. These sequences are separate, not zipped together.
- This tree maintains a reverse mapping of integer items to initializing geometries, instead of using a list or array, to support the user-provided item scenario.
- This tree has query_items/nearest_item and query_geoms/nearest_geom (aliased to geom and nearest) methods. The former return integers and the latter return geometries. I believe that this makes usage more clear and helps users who want to statically analyze their code using Python typing.
- Type hints have been added for the methods listed above.

Also tests and docstrings are much improved over the code in 1.8a1.


shapely/strtree.py
```diff
@@ -213,24 +262,71 @@ def nearest(self, geom):
         >>> tree = STRtree ([Point(0, 0), Point(0, 0)])
         >>> tree.nearest(Point(0, 0)).wkt
         'POINT (0 0)'
+
         """
-        if self._n_geoms == 0:
+        if self._tree is None or not self._rev:
             return None
```
sgillies, line 268: Raising an exception here could be less ambiguous than returning None, which is easy to confuse with an empty sequence (a perfectly normal query result).

shapely/strtree.py
```diff
@@ -213,24 +262,71 @@ def nearest(self, geom):
...
+            ctypes.py_object(geom),
+            envelope._geom,
+            lgeos.GEOSDistanceCallback(callback),
+            None,
+        )
         result = ctypes.cast(item, ctypes.py_object).value
-
         return result
+
+    def nearest_geom(self, geom):
```
sgillies, line 293: If we were to use type hints, this would be something like

`def nearest_geom(self, geom: Geometry) -> Geometry:`

I don't know how to write the type hints for 

`def nearest(self, geom: Geometry, return_geometry: bool = True) -> "Geometry if return_geometry else int":`

I'm not sure if Python supports conditional typing at all right now.

sgillies:
@jorisvandenbossche @brendan-ward what do you think of this modified version of #1064? Does it have any big problems from the C implementation perspective? It took me a while to remember how important it is to stick to storing C types when it comes to releasing the GIL. Thanks for being patient with me about that.

I'd still like to work on the feature requested in #1106. I wonder if a list of integer items to exclude or skip would be the way to go. After all, these are what we're storing in the tree and they are immediately accessible to the callback without returning to Python land. Meaning we can exclude/skip while the GIL is released.

jorisvandenbossche:
Thanks @sgillies for further looking at this!

Some quick remarks:

- The way you documented now is to require the custom ids (items) to be integers (although not sure the actual implementation also requires it). But if we allow custom ids, why limit it to only integers? 
- You mention explicitly to use a mapping (to keep track of item/geom combo's) instead of a list or array. Is there a specific reason you prefer a mapping? For the current shapely, it doesn't matter much whether we use a mapping (like this PR) or a list (like https://github.com/Toblerity/Shapely/pull/1064). But for Shapely 2.0 I think we rather want to use an array when we can use numpy (we can't yet use an array for 1.8, but a list is then closer in implementation).
- Implementation nitpick: currently it would be possible (I think) to have a mixture of (geom, id) tuples and plain geoms. I think we definitly want to disallow that.

> Does it have any big problems from the C implementation perspective? It took me a while to remember how important it is to stick to storing C types when it comes to releasing the GIL.

Ah, is this the reason you limited it to only integer items? 
This PR certainly does not impose any problems for a future version using the C implementation. Now, in practice, we would still *not* store the user-provided ids, but basically default 0,1,..n ids (using pointer arithmetic, this is how it's currently implemented in pygeos). But, it is easy to keep the user-provided ids on the Python tree object, and return those. So it would be easy to refactor the implementation in this PR on top of the pygeos C STRtree (although this basically would look like https://github.com/Toblerity/Shapely/pull/1064). 
The side-effect of not storing the user-provided ids in the C tree, is that those don't need to be limited to integers, but can be any kind of identifier / python object.

shapely/strtree.py
jorisvandenbossche: Python supports this with the `@overload` decorator (if we would want it): https://docs.python.org/3/library/typing.html#typing.overload

shapely/strtree.py
```diff
@@ -66,78 +68,94 @@ class STRtree:
...
                 pass  # lgeos might be empty on shutdown.
 
-            self._tree_handle = None
+            self._tree = None
 
-    def query(self, geom):
-        """
-        Search the index for geometry objects whose extents
-        intersect the extent of the given object.
+    def query_item(self, geom):
```
jorisvandenbossche, line 136: Naming/terminology question: would eg "ids" be clearer as "items"? 
If you see "(geom, item)" it's of course clear that "item" is the identifier. But if I read "item" by itself (as in this function name / docstring), for me it would not directly be clear whether it references the geometry item of the id item ..

shapely/strtree.py
```diff
@@ -66,78 +68,94 @@ class STRtree:
...
+            }
+            if self._rev:
+                self._init_tree(self._rev.items())
+
+    def _iterinitdata(self, initdata):
+        for enum_idx, item in enumerate(initdata):
+            if isinstance(item, tuple):  # a geom, idx pair
+                yield item[:2]
+            else:  # a single geometry
+                yield (item, enum_idx)
```
jorisvandenbossche, line 108: Here I think we would need to add some validation that the user doesn't mix both cases (otherwise a user-provided id could clash with a `enum_idx` of another geometry)

(it would of course be very bizarre for a user to do so, but I think still worth validating in some way)

sgillies: @jorisvandenbossche this is a "private" method called only from the constructor, so when our constructor's code is correct we'll never see a mixed mode.

shapely/strtree.py
sgillies: "item" is the term GEOS uses. An item is stored in a node of the tree. Whether the item is an "id" or not is up to the application.

sgillies:
@jorisvandenbossche we came to the conclusion that since items of a uniform C type are good for performance and since all Python objects can be mapped to ints via `id` and `hash`, the least common denominator was storing ints in the tree and asking the application to handle the mapping of these integer items to other objects.

sgillies:
Yay, builds are passing again. I think we had some cache related problems a few commits back.

shapely/strtree.py
jorisvandenbossche: Now it indeed is always called consistently from the constructor, but that was not the case when I commented :-) (then it was passing through the list of tuples from the user)

shapely/strtree.py
```diff
@@ -66,78 +78,107 @@ class STRtree:
...
-    def __init__(self, geoms):
+    def __init__(
+        self,
+        geoms: Iterable[BaseGeometry],
+        items: Iterable[int] = None,
+        node_capacity: int = 10,
+    ):
         warn(
-            "STRtree will be completely changed in 2.0.0. The exact API is not yet decided, but will be documented before 1.8.0",
+            "STRtree will be changed in 2.0.0. The exact API is not yet decided, but will be documented before 1.8.0",
```
sgillies, line 104: TODO: remove this warning.

shapely/strtree.py
```diff
@@ -160,45 +201,76 @@ def query(self, geom):
...
         -------
-        geometry object
-            The nearest geometry object in the index to `geom`.
+        sequence
+            A list or array of geometry objects.
+
+        """
+        return self.query_geoms(geom)
+
+    def nearest_item(self, geom: BaseGeometry) -> Union[int, None]:
```
sgillies, line 257: I think the return type of this function (and nearest_geom) are the only things that we'll change before 2.0. We may raise an error instead of returning None.

jorisvandenbossche:
> we came to the conclusion that since items of a uniform C type are good for performance and since all Python objects can be mapped to ints via id and hash, the least common denominator is storing ints in the tree and asking the application to handle the mapping of these integer items to other objects.

OK, just to be clear: the only additional "feature" this PR gives (compared to default indices 0...n mapping to the position in the sequence of geometries) is that you can use custom integer values instead of 0...n ?  (and no longer any custom python object)

I don't necessarily want to argue for allowing any python object as id, but note that when storing indices in the tree (as we will do in shapely 2.0, and as https://github.com/Toblerity/Shapely/pull/1064 does), there is not really a performance difference in using custom integers vs custom python objects.

jorisvandenbossche:
@sgillies and thanks for the updates! 

For a better comparison, I also updated https://github.com/Toblerity/Shapely/pull/1064 to follow the same API as you did in this PR (but didn't yet copy over all other changes, like better docstrings etc, just updated the actual implemenation). I think both achieve the same, but the version in https://github.com/Toblerity/Shapely/pull/1064 is closer to what we would do in the shapely-2.0 branch on top of the C-based STRtree.

sgillies:
@songololo @jorisvandenbossche I've relaxed the integer constraint on the constructor's `items` argument. In the future implementation of this class we will be converting items to a numpy array and numpy handles this very well. A sequence of strings is converted to an array of dtype something like `"U"` and a sequence of mixed objects is converted to an array of dtype `"object"`. In the future if you pass in a numpy array (like geopandas will do, yes?) it will be used as is, without any copying.

tests/test_strtree.py
```diff
@@ -8,21 +8,38 @@
...
+@requires_geos_342
+@pytest.mark.parametrize("geoms", [[Point(i, i) for i in range(10)]])
+@pytest.mark.parametrize(
+    "query_geom,expected",
+    [(Point(2, 2).buffer(0.99), [2]), (Point(2, 2).buffer(1.0), [1, 2, 3])],
+)
+def test_query_enumeration_idx(geoms, query_geom, expected):
+    """Store enumeration idx"""
+    with pytest.warns(ShapelyDeprecationWarning):
+        tree = STRtree((g, i) for i, g in enumerate(geoms))
```
jorisvandenbossche, line 40: @sgillies based on your last comments (and the updated top post which mentions *"These sequences are separate, not zipped together"* and your comment at https://github.com/Toblerity/Shapely/pull/1064#issuecomment-813720602) that the idea was the only support geometries and items as separate sequences (where the items is optional), and not in the combined (zipped) form.  
But this test above is testing that constructing the tree from a sequence of tuples actually works?

jorisvandenbossche: I opened https://github.com/Toblerity/Shapely/pull/1172 to potentially address this

sgillies: Oh my, you're right that this test shouldn't pass! :facepalm:

PR #1121, opened 2021-04-09
Drop Python 3.5 support

jorisvandenbossche:
See https://github.com/Toblerity/Shapely/issues/886, this is only for the shapely-2.0 branch


jorisvandenbossche:
Dropping 3.5 was already done in master for Shapely 1.8, so closing this.

PR #1122, opened 2021-04-09
DOC: update README to indicate Python 2.7 is not longer supported

jorisvandenbossche:
We dropped Python 2.7 support for Shapely 1.8

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -50,7 +50,7 @@ See the manual for more examples and guidance.
 Requirements
 ============
 
-Shapely 1.7 requires
+Shapely 1.8 requires
 
 * Python >=3.5
 * GEOS >=3.3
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -52,7 +52,7 @@ Requirements
 
 Shapely 1.7 requires
```
jamesmyatt, line 53: Shouldn't it be "Shapely 1.8 requires ..." then?

sgillies, line 53:
```suggestion
Shapely 1.8 requires
```

sgillies (approved):

README.rst
jorisvandenbossche: Ah, yes, indeed ;) 
Thanks @sgillies for the fixup

PR #1123, opened 2021-04-09
TST: make the WKB test robust for flavour of NaN

jorisvandenbossche:
This `test_wkb.py::test_point_empty` test is failing on the shapely-2.0 branch with the latest pygeos. I *think* this is because pygeos is using `Py_NAN` to write the WKB with NaN values for empty geometries, and `Py_NAN` might be platform-dependent (the error on appveyor indicates the WKB includes a negative quiet NaN (`-float(NaN)`) instead of a plain quiet NaN, by checking on my linux laptop). 

So this PR tries to make the test more robust, by using the default NaN representation from the system on which the test is running (which *should* match with the NaN value that pygeos is using, I think).

### DIFF
--- a/tests/test_wkb.py
+++ b/tests/test_wkb.py
@@ -99,8 +99,9 @@ requires_geos_380 = pytest.mark.xfail(
 @requires_geos_380
 def test_point_empty():
     g = wkt.loads("POINT EMPTY")
-    assert g.wkb_hex == hostorder(
-        "BIdd", "0101000000000000000000F87F000000000000F87F")
+    NAN = struct.pack("<d", float("nan")).hex()
+    POINT_EMPTY_WKB = "0101000000" + (NAN * 2)
+    assert g.wkb_hex == hostorder("BIdd", POINT_EMPTY_WKB)
 
 
 @pytest.mark.xfail(reason="Fails with latest pygeos")


jorisvandenbossche:
cc  @musicinmybrain does this look OK to ensure it still works on big-endian hosts as well?

jorisvandenbossche:
Hmm, that didn't actually help for windows. Trying the approach used in pygeos then to check the unpacked bytes with ``isnan``

musicinmybrain:
> cc @musicinmybrain does this look OK to ensure it still works on big-endian hosts as well?

The current code looks OK since it uses the `big_endian=False` option to `shapely.wkb.dumps`, so the resulting byte string should be little-endian regardless of the host endianness. The code also correctly uses `'>'` at the beginning of the `struct.unpack` format string to ensure the byte string is unpacked as little-endian.

I think that using `math.isnan` is a good approach. Except in the very rare cases where software actually uses the NaN “payload” bits to smuggle data around, there are too many ways that assuming a particular representation can go wrong.

jorisvandenbossche:
Thanks! Yes, the final version that I pushed shouldn't be sensitive to little/big endian host anymore, since I force little_endian in the output. Of course that then also doesn't test the big endian is correct ..

PR #1128, opened 2021-04-16
Fix shapely.vectorized install by adding a pyproject.toml

### LINKED ISSUE #1127: Missing module shapely.vectorized when installing shapely[vectorized] from sources

Two pre-requisites are needed to generate the [shapely.vectorized](../tree/master/shapely/vectorized) module from sources: `cython` and `numpy`

When a end-user module that uses shapely is installed, the install command should be:

```console 
$ pip install end-user-module --no-binary shapely
```
But issues are occurring on a fresh Python environment when doing so: as Cython and numpy are not pre-installed the vectorized module is not generated, even if the dependency in `end-user-module` is declared like `shapely[vectorized]`.

Main problem is that cython is still a build dependency that cannot be handled correctly by `setuptools`.

On a fresh Python environment:

```console
$ pip install shapely[vectorized] --no-binary shapely
$ python3 -c 'import shapely.vectorized'
  File "/home/aaaaa/Documents/Shapely/.venv/lib/python3.8/site-packages/shapely/vectorized/__init__.py", line 3, in <module>
    from ._vectorized import (contains, touches)
ModuleNotFoundError: No module named 'shapely.vectorized._vectorized'
```

## Operating system

Linux 64bits

1.7.1

remy-phelipot:
- Added a pyproject.toml to install cython and numpy at build time

Fix #1127


jorisvandenbossche:
Thanks @remy-phelipot. I suppose there is no way to let the build-system `requires` depend on the extras you want to install? (meaning, if you don't install [vectorized], you in principle don't need to install numpy in the build environment)

remy-phelipot:
> 
> 
> Thanks @remy-phelipot. I suppose there is no way to let the build-system `requires` depend on the extras you want to install? (meaning, if you don't install [vectorized], you in principle don't need to install numpy in the build environment)

It's an interesting question. Actually there may be a way if a custom build_backend (of pyproject.toml) is defined. There is a hook function `get_requires_for_build_wheel` that can filter the build dependencies but I'm not sure the `config_settings` input parameter contains enabled extras.

I'll try that and keep you informed!

PR #1134, opened 2021-05-14
Never pass empty geometries to GEOS_getCoordSeq

sgillies:
Fixes the issue reported in https://github.com/geopandas/geopandas/issues/1943.


jorisvandenbossche:
Thanks for the fix!

PR #1141, opened 2021-05-28
MAINT: aliases of builtin types is deprecated with numpy>=1.20

mwtoews:
With [numpy 1.20](https://numpy.org/doc/stable/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated), aliases of builtin types like `np.bool` is deprecated. This should clear up some warnings in the test suite.


jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

PR #1143, opened 2021-05-31
ENH: Added opacity option to polygon, point, linestring and their multiple variations

MejstrikRudolf:
Hello,
I've just edited your library a bit. I added to method svg option to change opacity for pylygons, points, linestrings and their multiple variations.
I hope you will accept it.

### DIFF
--- a/shapely/geometry/linestring.py
+++ b/shapely/geometry/linestring.py
@@ -56,7 +56,7 @@ class LineString(BaseGeometry):
             'coordinates': tuple(self.coords)
             }
 
-    def svg(self, scale_factor=1., stroke_color=None):
+    def svg(self, scale_factor=1., stroke_color=None, opacity=None):
         """Returns SVG polyline element for the LineString geometry.
 
         Parameters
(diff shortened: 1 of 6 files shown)


shapely/geometry/linestring.py
```diff
@@ -56,7 +56,7 @@ def __geo_interface__(self):
             'coordinates': tuple(self.coords)
             }
 
-    def svg(self, scale_factor=1., stroke_color=None):
+    def svg(self, scale_factor=1., stroke_color=None, opacity=None):
```
sgillies, line 59: @MejstrikRudolf what would you think about putting the default in the function declaration?

```suggestion
    def svg(self, scale_factor=1., stroke_color=None, opacity=0.8):
```

Currently the opacity is 1, but I think there are advantanges to 0.8, it reveals overlapping shapes. @om-henners (as the creator of this feature) do you see any downsides?

MejstrikRudolf: Sure, it's better to put the default value to the function declaration. I'm just beginner so I don't have best performence skills yet.
And I need this option for extracting shapes to svg and then putting the svg onto web, so I need for the shapes changeable opacity.

om-henners: @sgillies I think it's a good idea honestly - trying it locally the opacity of 0.8 looks pretty good - I certainly don't see any downsides!

sgillies:
@MejstrikRudolf thank you very much!

Issue #501, opened 2017-07-09
Vectorize all GEOS functions


wmay:
Vectorized functions would be a big help to GeoPandas and other projects trying to organize Shapely geometry in arrays, and there seems to be a consensus that it's best to implement the vectorized GEOS functions within Shapely. (See https://github.com/geopandas/geopandas/issues/430#issuecomment-291083993.)

After taking a look through the code I think we could implement this without too much trouble by adding a version of [topology.py](https://github.com/Toblerity/Shapely/blob/master/shapely/topology.py) that generates vectorized versions of the functions. At that point it would be easy to call the functions from the `vectorized` module, or add them to the `BaseGeometry` class, or call them in Geopandas (I think).

I just want to get some feedback to make sure I'm on the right track before I start this, and figure out where people should send the pull requests.

wmay:
OK, I've got a rudimentary version of this working in [my vectorize branch](https://github.com/wmay/Shapely/tree/vectorize). The relevant code is in [vectorfuncs.pyx](https://github.com/wmay/Shapely/blob/vectorize/shapely/vectorized/vectorfuncs.pyx), which duplicates the functionality of topology.py, predicates.py, and impl.py.

Even though I'm sure this code could be improved, it's already cutting running times in half or more for a lot of operations compared to unvectorized code. (I think that's mostly due to avoiding Python `for` loops right now.) [Here](http://nbviewer.jupyter.org/gist/wmay/a2b003f92db99f7c5a6e7f57db8b9ff5)'s a jupyter notebook demonstrating the functions, as the code stands now.

The tasks left for making this work really well are:
* Create appropriate validation functions for arrays.
* Add support for array/array binary functions. (The binary functions right now only take an array for the second argument.)
* Improve my awkward Cython code.
* Add tests.
* Decide how to make the functions available to regular users.

And my wish list of nice features:
* Add support for multidimensional arrays and broadcasting.
* Add an option to pass pointers instead of shapely objects (because that can be much faster).
* Add an option to return an array of shapely objects instead of pointers, for convenience.
* Vectorize `shapely.geometry.base.geom_factory()`, since that seems to slow things down.

sgillies:
@wmay this looks very promising! If I may suggest a few things: 

- test early and often, good tests help guide feature development and can be painful to write after the feature is finished (especially if they reveal that the feature isn't finished).
- generalize to N x M functions as soon as you can.
[...]

PR #1144, opened 2021-06-01
Add vectorized covers and intersects

aulemahal:
This PR goes in the direction of #501 and adds `covers` and `intersects` to the vectorized methods.

In `clisops`, we are constructing integer masks from xarray DataArrays and GeoDataframes. The code currently uses `vectorized.contains` and `vectorized.touches`. The latter is extremely slow. I'm not sure why, but see issue roocs/clisops#160.

What we are looking for is in fact `covers`, but as it wasn't available, we went with a logical OR of the two other results. This PR adds `covers` and `intersects`, the two remaining unimplemented predicates that make sense with points. 

I did try `PyGEOS`, but while I might be doing something wrong, the performance are orders of magnitude better with `vectorized.covers`.  Also, it seemed a low-effort high-reward PR to do for us.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -8,6 +8,10 @@ Python version support:
 
 Shapely 1.8 will support only Python versions >= 3.6.
 
+New features:
+
+- Add `covers` and `intersects` to the vectorized methods (#1144).
+
 Bug fixes:
 
 - GEOS error messages printed when GEOS_getCoordSeq() is passed an empty
--- a/CREDITS.txt
+++ b/CREDITS.txt
@@ -72,6 +72,7 @@ Shapely is written by:
 * Morris Tweed <tweed.morris@gmail.com>
 * Naveen Michaud-Agrawal <naveen.michaudagrawal@gmail.com>
 * Oliver Tonnhofer <olt@bogosoft.com>
+* Pascal Bourgault <pascal.bourgaulty@gmail.com
 * Paveł Tyślacki <tbicr@users.noreply.github.com>
 * Peter Sagerson <psagers.github@ignorare.net>
 * Phil Elson <pelson.pub@gmail.com>
--- a/shapely/vectorized/__init__.py
+++ b/shapely/vectorized/__init__.py
@@ -1,3 +1,3 @@
 """Provides multi-point element-wise operations such as ``contains``."""
 
-from ._vectorized import (contains, touches)
+from ._vectorized import (contains, covers, intersects, touches)
--- a/shapely/vectorized/_vectorized.pyx
+++ b/shapely/vectorized/_vectorized.pyx
@@ -41,6 +41,54 @@ def contains(geometry, x, y):
     return _predicated_elementwise(geometry, x, y, GEOSPreparedContains_r)
 
 
+def covers(geometry, x, y):
+    """
+    Vectorized (element-wise) version of `covers` which checks whether
+    multiple points are covered by the exterior of a single geometry.
+
+    Parameters
+    ----------
+    geometry : PreparedGeometry or subclass of BaseGeometry
+        The geometry which is to be checked to see whether each point is
+        covered. The geometry will be "prepared" if it is not already
+        a PreparedGeometry instance.
+    x : array
+        The x coordinates of the points to check. 
+    y : array
+        The y coordinates of the points to check.
+
+    Returns
+    -------
+    Mask of points covered by the exterior of the given `geometry`.
+
+    """
+    return _predicated_elementwise(geometry, x, y, GEOSPreparedCovers_r)
+
+
+def intersects(geometry, x, y):
+    """
+    Vectorized (element-wise) version of `intersects` which checks whether
+    multiple points are share any portion of space with a single geometry.
+
+    Parameters
+    ----------
+    geometry : PreparedGeometry or subclass of BaseGeometry
+        The geometry which is to be checked to see whether each point share any
+        portion of space with. The geometry will be "prepared" if it is not already
+        a PreparedGeometry instance.
+    x : array
+        The x coordinates of the points to check. 
+    y : array
+        The y coordinates of the points to check.
+
+    Returns
+    -------
+    Mask of points sharing space with the given `geometry`.
+
+    """
+    return _predicated_elementwise(geometry, x, y, GEOSPreparedIntersects_r)
+
+
 def touches(geometry, x, y):
     """
     Vectorized (element-wise) version of `touches` which checks whether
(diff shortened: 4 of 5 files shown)


jorisvandenbossche:
> I did try `PyGEOS`, but while I might be doing something wrong, the performance are orders of magnitude better with `vectorized.covers`. Also, it seemed a low-effort high-reward PR to do for us.

Do you have some code you can share that reproduces the performance issue? It would be good to investigate why pygeos is slower in this case / how that can be solved, because whatever code for vectorized operations is being added here will probably be replaced by the code of pygeos in Shapely 2.0 (for which reason I am also not that fond on adding more code to the current `shapely.vectorized` ..)

jorisvandenbossche:
I see there are some more details in https://github.com/roocs/clisops/issues/160#issuecomment-852456926. One thing that comes to mind: did you prepare the geometry when using pygeos? See eg here how the shapely-2.0 branch reimplements the shapely.vectorized operations on top of pygeos: 

https://github.com/Toblerity/Shapely/blob/98f4c8fba0f4754c5b121621934bd0fd888183b6/shapely/vectorized/__init__.py#L43-L47

Also be sure to compare the same predicate (the performance of one predicate to the other can vary a lot, which is determined by the underlying GEOS library). For pygeos you seem to mention to have used `covered_by`, while with shapely you used `covers` ?

aulemahal:
Your comments helped me better target what I should be testing.... and you were right! With proper coding, pygeos' performances are now close or even better than shapely.vectorized. I'm not sure what I missed in my last tests. ugh. I'll be closing this issue, but will put here my findings.

Here is the final code. It computes the timezones mask (value is the offset from UTC) on a generated global lat/lon grid. This is the kind of use case we are aiming for at `clisops`. I download a shapefile and subset it to reduce complexity, but I do not simplify the polygons.

```python
import warnings
import geopandas as gpd
import numpy as np
import pygeos
from shapely import vectorized
import xarray as xr
warnings.simplefilter('ignore', RuntimeWarning)

lon = xr.DataArray(np.arange(-180, 180, 1), dims=('lon',), name='lon')
lat = xr.DataArray(np.arange(-90, 90, 1), dims=('lat',), name='lat')
print(lon.size, lat.size)

gdf = gpd.read_file('https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/10m/cultural/ne_10m_time_zones.zip')

sub = gdf.iloc[:60]
polys = sub.geometry
index = np.arange(len(polys))
print(index.shape)

def create_mask(lon, lat, polys, index, method):
    """Create a 2D mask from a list of polygons. The values of the mask are taken from `index`.
       If a point falls in multiple polygons, the value with the lowest rank in `index` has priority.  
    """
    X, Y = np.meshgrid(lon.values, lat.values, indexing='ij')
    
    if method.startswith('pygeos'):
        pts = pygeos.points(X, Y)[np.newaxis, ...]
        pygeos.prepare(pts)

        geoms = pygeos.from_shapely(polys)[:, np.newaxis, np.newaxis]
        pygeos.prepare(geoms)
        
        vals = index[:, np.newaxis, np.newaxis]

        def get_intersection(geom):
            if method.endswith('covers'):
                return pygeos.covers(geom, pts)
            elif method.endswith('covered'):
                return pygeos.covered_by(pts, geom)
            elif method.endswith('contains'):
                return pygeos.contains(geom, pts)
            elif method.endswith('contains+touches'):
                return np.logical_or(pygeos.contains(geom, pts), pygeos.touches(geom, pts))
        
        if 'loop' in method:
            mask = np.full(X.shape, np.NaN)
            # Loop is reversed so that lower ranked value have priority
            for val, geom in zip(index[::-1], geoms[::-1]):
                intersection = get_intersection(geom).squeeze()
                mask[intersection] = val
        else:
             mask = np.nanmin(np.where(get_intersection(geoms), vals, np.NaN), axis=0)
    
    elif method.startswith('vectorized'):
        mask = np.full(X.shape, np.NaN)
        for val, geom in zip(index[::-1], polys[::-1]):
            if method.endswith('covers'):
                intersection = vectorized.covers(geom, X, Y)
            elif method.endswith('contains'):
                intersection = vectorized.contains(geom, X, Y)
            elif method.endswith('contains+touches'):
                intersection = np.logical_or(vectorized.contains(geom, X, Y), vectorized.touches(geom, X, Y))
            mask[intersection] = val

    return xr.DataArray(mask, dims=('lon', 'lat'), coords={'lon': lon, 'lat': lat})
```
And the benchmarking:

```
pygeos-covers  with timeit
  148 ms ± 9.53 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
pygeos-loop-covers  with timeit
  162 ms ± 13.3 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
vectorized-covers  with timeit
  305 ms ± 18.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
pygeos-contains  with timeit
  149 ms ± 4.89 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
vectorized-contains  with timeit
  318 ms ± 32.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
pygeos-covered with time
  CPU times: user 49.4 s, sys: 0 ns, total: 49.4 s
  Wall time: 49.4 s
pygeos-loop-covered with time
  CPU times: user 49.5 s, sys: 0 ns, total: 49.5 s
  Wall time: 49.5 s
pygeos-contains+touches with time
  CPU times: user 49.2 s, sys: 0 ns, total: 49.2 s
  Wall time: 49.3 s
pygeos-loop-contains+touches with time
  CPU times: user 49.2 s, sys: 6.39 ms, total: 49.2 s
  Wall time: 49.2 s
vectorized-contains+touches with time
  CPU times: user 51.1 s, sys: 9.79 ms, total: 51.1 s
  Wall time: 51.1 s
```
And with the full polygon list and a global grid at the 0.1° resolution. I use `memory_profiler` to estimate min/max memory usage during the computation. Here, `pygeos-covers` (with no loop) exploded my laptop RAM, which explains the long Wall time compared to the CPU time.
```
(lat, lon) = 3600 1800
len(polys) = (120,)
pygeos-covers  with time - full grid and geoms
  CPU times: user 24.2 s, sys: 27.1 s, total: 51.3 s
  Wall time: 1min 15s
  Min and peak mem usage : 62.296875 4803.6484375
pygeos-loop-covers  with time - full grid and geoms
  CPU times: user 18.8 s, sys: 914 ms, total: 19.7 s
  Wall time: 20.2 s
  Min and peak mem usage : 1126.25390625 1544.83984375
vectorized-covers  with time - full grid and geoms
  CPU times: user 55 s, sys: 130 ms, total: 55.1 s
  Wall time: 55.3 s
  Min and peak mem usage : 1298.66015625 1300.3203125
```

Conclusion : sorry again, I should have tested my code properly before opening this PR. And thanks for all the work done! I think we will use `pygeos` directly in the next version of `clisops`, returning to `shapely` when version 2.0 comes out!

jorisvandenbossche:
@aulemahal no need to feel sorry, I am happy that I could provide some pointers for you to figure it out.

Issue #962, opened 2020-08-11
Practical steps towards Shapely 2.0


jorisvandenbossche:
TODO before final release:

- [ ] update list of contributors again

TODO items for the Shapely / PyGEOS merge:

* [x] Merge packages into a single shapely package:
  * [x] Merge the python source code of both packages into a single, minimal working `shapely` package -> https://github.com/Toblerity/Shapely/pull/1237
  * [x] Consolidate both `strtree.py` files -> https://github.com/Toblerity/Shapely/pull/1251
* [x] Continuous Integration:
  * [x] Update and consolidate the github action workflow for testing -> https://github.com/Toblerity/Shapely/pull/1241
  * [x] Update the github action workflow for wheel building / releasing -> https://github.com/Toblerity/Shapely/pull/1253
  * [x] Ensure Travis CI is running (after moving to the shapely org?)
* [x] Packaging:
  * [x] Consolidate the `setup.py` parameters -> https://github.com/shapely/shapely/pull/1376
  * [x] Consolidate pyproject.toml and setup.cfg
  * [x] Consolidate MANIFEST.in, ensure sdist/wheel creation is working -> https://github.com/Toblerity/Shapely/pull/1249
  * [x] Ensure versioneer is working and has a correct shapely 2.0.0.dev+.. version 
* [ ] Tests
  * [ ] Move `/tests` to `/shapely/tests` -> [#1257](https://github.com/shapely/shapely/pull/1257), [#1275](https://github.com/shapely/shapely/pull/1275), ...
  * [ ] Combine / deduplicate tests where useful
* [ ] Documentation
  * [ ] Update docstrings
  * [x] Merge the sphinx documentation sources -> https://github.com/shapely/shapely/pull/1377, https://github.com/shapely/shapely/pull/1402

---

Opening an issue to discuss the more practical aspects of the work on Shapely 2.0 (https://github.com/shapely/shapely-rfc/pull/1/). Changes needed in Shapely 1.8 (deprecations, etc) related to this are tracked in https://github.com/Toblerity/Shapely/issues/932

The general route I am now trying out is to, iteratively, refactor the shapely code in-place. But @caspervdw mentioned at https://github.com/Toblerity/Shapely/issues/782/#issuecomment-592058242 also the option of starting "fresh" (for the code, not for the existing tests and things like that).

[...]

PR #1148, opened 2021-06-08
TST: rename head branch for GEOS from 'master' to 'main'

mwtoews:
GEOS renamed their git development head branch from 'master' to 'main' in March 2021 ([ref](https://lists.osgeo.org/pipermail/geos-devel/2021-March/010193.html)).


sgillies:
@mwtoews I'd be in favor of doing the same here.

mwtoews:
@sgillies I'd support that too. It would probably make sense to do two changes at once: move repo to https://github.com/shapely/shapely and rename the head branch to main.

jorisvandenbossche:
Also +1 on renaming to "main" (I only need to update my local scripts/aliases to handle both master and main for a while .. ;))

> It's a matter of timing and tasks related to #962

It could be done at the same time, but I don't think that's necessary. The fact that the content of the "shapely-2.0" branch will move the the master/main branch in the future, doesn't prevent already renaming the current master branch right now.

jorisvandenbossche:
And also +1 on moving to the github.com/shapely org. It will be good to have the different repo's (shapely-wheels, rfc) in a single organization.

PR #1150, opened 2021-06-10
arm64 wheels: RFC

tsahee:
This will create wheels usable by arm64 systems.

As arm64 system become more prevalent, many wheels are also published for arm64 systems. Would it be possible to also have shapely arm64 wheels?

The script could be used as-is on either arm64 systems (tested on Graviton2) or intel/amd systems with [qemu-docker](https://github.com/joshkunz/qemu-docker) (tested on intel MAC).

I would be happy to add / modify the script as needed to support your normal wheel-building process.

Differences between arm64 and x86 wheel generation:
* Using manylinux2014_aarch64
* Added a call to ldconfig after building libgeos
* Only generating wheels for cp3.7/8/9 - as matplotlib only publishes wheels for these versions (for both x86 and arm64).

### DIFF
--- a/Dockerfile.arm64.wheels
+++ b/Dockerfile.arm64.wheels
@@ -1,0 +1,24 @@
+FROM quay.io/pypa/manylinux2014_aarch64
+
+ENV GEOS_VERSION 3.5.0
+
+# Install geos
+RUN mkdir -p /src \
+    && cd /src \
+    && curl -f -L -O http://download.osgeo.org/geos/geos-$GEOS_VERSION.tar.bz2 \
+    && tar jxf geos-$GEOS_VERSION.tar.bz2 \
+    && cd /src/geos-$GEOS_VERSION \
+    && ./configure \
+    && make \
+    && make install \
+    && ldconfig \
+    && rm -rf /src
+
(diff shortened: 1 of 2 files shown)


tsahee:
Hello!
Is there any feedback?

Thank you!

jorisvandenbossche:
@tsahee sorry for the slow reply here. Personally, I don't have the bandwidth to help with following up on this wheel building, but some notes: I *think* the shapely wheel building currently happens at https://github.com/shapely/shapely-wheels (there also seems to be an issue/PR related to arm64), so to add arm64 wheels to the officially released wheels, it should probably be added there. 
In pygeos (to be integrated in shapely 2.0), we are at the moment experimenting with switching to cibuildwheel (https://github.com/pygeos/pygeos/pull/365, which also supports aarch64 for linux and arm64 for mac).

sgillies:
@tsahee thanks for the contribution, but we can't use it as it is. We've already got arm64 wheels for Pythons 3.8 and 3.9 in our build matrix at https://github.com/shapely/shapely-wheels/blob/master/.travis.yml#L66 and have published them for version 1.8a3: https://pypi.org/project/Shapely/1.8a3/#files.

PR #1155, opened 2021-06-19
Removed duplicate line in test_minimum_rotated_rectangle.py

jaehaaheaj:
removed duplicate instance type check


sgillies (approved):
Thanks @jaehaaheaj !

PR #1161, opened 2021-07-07
Refactor strtree to not use lgeos / ctypes

jorisvandenbossche:
This is for the shapely-2.0 branch, to do the equivalent of https://github.com/Toblerity/Shapely/pull/1112 or https://github.com/Toblerity/Shapely/pull/1064 (and ideally only gets done when one of those PRs is merged).

This shows how the `lib.STRtree` (the C object) will just contain primary keys, which are mapped to geometries (or other items) in the `query` method using the returned indices.

### DIFF
--- a/shapely/strtree.py
+++ b/shapely/strtree.py
@@ -17,14 +17,14 @@ References
      VLDB Conf. 497-506. 10.1109/ICDE.1997.582015.
      https://www.cs.odu.edu/~mln/ltrs-pdfs/icase-1997-14.pdf
 """
-
-import ctypes
 import logging
-from typing import Any, ItemsView, Iterable, Iterator, Optional, Sequence, Tuple, Union
-import sys
+from typing import Any, Iterable, Sequence, Union
 
 from shapely.geometry.base import BaseGeometry
-from shapely.geos import lgeos
+
+import numpy as np
+from pygeos import lib
+
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


shapely/strtree.py
```diff
@@ -168,18 +147,17 @@ def query(self, geom):
...
-            geom = ctypes.cast(item, ctypes.py_object).value
-            result.append(geom)
-
-        lgeos.GEOSSTRtree_query(self._tree_handle, geom._geom, lgeos.GEOSQueryCallback(callback), None)
+        indices = self._tree.query(geom, 0)
+        return self.geometries[indices]
 
-        return result
+    def query_items(self, geom):
+        """
```
snorfalorpagus, line 154: Missing docstring

shapely/strtree.py
```diff
@@ -168,18 +147,17 @@ def query(self, geom):
...
-        def callback(item, userdata):
-            geom = ctypes.cast(item, ctypes.py_object).value
-            result.append(geom)
-
-        lgeos.GEOSSTRtree_query(self._tree_handle, geom._geom, lgeos.GEOSQueryCallback(callback), None)
+        indices = self._tree.query(geom, 0)
+        return self.geometries[indices]
 
-        return result
+    def query_items(self, geom):
```
snorfalorpagus, line 153: This function needs a test

shapely/strtree.py
```diff
@@ -214,23 +192,26 @@ def nearest(self, geom):
...
+            geometry = np.expand_dims(geometry, 0)
+
+        indices = self._tree.nearest(geometry)
+        # nearest returns ndarray with shape (2, 1) -> index in input
+        # geometries and index into tree geometries
+        idx = indices[1, 0]
+        return self.geometries[idx]
+
+    def nearest_item(self, geom):
+        """
```
snorfalorpagus, line 206: Missing docstring

shapely/strtree.py
```diff
@@ -214,23 +192,26 @@ def nearest(self, geom):
...
+        if geometry.ndim == 0:
+            geometry = np.expand_dims(geometry, 0)
+
+        indices = self._tree.nearest(geometry)
+        # nearest returns ndarray with shape (2, 1) -> index in input
+        # geometries and index into tree geometries
+        idx = indices[1, 0]
+        return self.geometries[idx]
+
+    def nearest_item(self, geom):
```
snorfalorpagus, line 205: This function needs a test

shapely/strtree.py
```diff
@@ -75,44 +72,26 @@ class STRtree:
...
+    def __len__(self):
+        return self._tree.count
+
+    def __reduce__(self):
+        if self._has_custom_items:
+            return (STRtree, (self.geometries, self._items))
+        else:
+            return (STRtree, (self.geometries, ))
 
     def query(self, geom):
```
snorfalorpagus, line 96: This function now returns an `np.array` instead of a list which is causing the doctest to fail.

snorfalorpagus (changes requested):
Added some minor comments.

I like the new `query_items` function. This is a real pain in the current API.

jorisvandenbossche:
@snorfalorpagus 

> I like the new `query_items` function. This is a real pain in the current API.

That's actually already introduced in the main branch (and so will be in shapely 1.8), coming from #1112 / #1064

Related to the missing docstrings and tests, they were also introduced in #1112, so I need to rebase this branch to pull those in (I made this branch with just similar code changes before #1112 was merged).

jorisvandenbossche:
I updated the shapely-2.0 branch with the changes of latest master, and so also updated this branch (so docstrings and tests are now already included)

jorisvandenbossche:
Note that this PR disables the `exclusive` keyword again in the `nearest_item`/`nearest_geom` methods (which was introduced in https://github.com/Toblerity/Shapely/pull/1166), since pygeos does not yet support this. 
Of course ideally we fix this before an actual 2.0 release, but I would prefer to not keep up this PR until it's supported in the pygeos version.

shapely/strtree.py
mwtoews: docstring currently has "array or list of geometry objects.". Perhaps it should be updated to say just "array".

mwtoews (commented):
Admittedly, I don't know STRTree very well, but I've tested this locally and it seems to be in working order.

Perhaps the docstrings could be updated to say if an array is returned from (e.g.) `query`. Also, should `query` and `nearest` be at least deprecated? (These are still in consideration to be removed for 2.0)

Is there an issue in pygeos for the `exclusive` keyword? It hopefully shouldn't be too difficult to implement.

jorisvandenbossche:
> Is there an issue in pygeos for the exclusive keyword? It hopefully shouldn't be too difficult to implement.

Didn't do that yet (will do now). Somewhat related, I did just open an issue about some other API incompatibilities between the pygeos and Shapely STRtree: https://github.com/pygeos/pygeos/issues/398

jorisvandenbossche:
Opened https://github.com/pygeos/pygeos/issues/400 for the `exclusive` keyword.

shapely/strtree.py
jorisvandenbossche: Ah, yes, updated the docstrings.

jorisvandenbossche:
> Also, should `query` and `nearest` be at least deprecated? (These are still in consideration to be removed for 2.0)

Not fully sure, what I think that's a separate issue to discuss.

jorisvandenbossche:
Since this is only replacing the internals of the tree with pygeos (and not changing any API), I am going to merge this, so I can update https://github.com/Toblerity/Shapely/pull/1163

PR #1162, opened 2021-07-07
Refactor pickling of LinearRing to not use lgeos

jorisvandenbossche:
xref https://github.com/Toblerity/Shapely/issues/988

Together with https://github.com/Toblerity/Shapely/pull/1161, this is the last remaining usage of lgeos / ctypes in the shapely-2.0 branch.

### DIFF
--- a/shapely/geometry/polygon.py
+++ b/shapely/geometry/polygon.py
@@ -21,6 +21,11 @@ from shapely.errors import TopologicalError, ShapelyDeprecationWarning
 __all__ = ['Polygon', 'LinearRing']
 
 
+def _unpickle_linearring(wkb):
+    linestring = pygeos.from_wkb(wkb)
+    return pygeos.linearrings(pygeos.get_coordinates(linestring))
+
+
 class LinearRing(LineString):
     """
     A closed one-dimensional feature comprising one or more line segments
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
The windows failure is unrelated (it's also a pickling error, but already existing and related to encoding)

mwtoews (approved):
Changes look good, and great to see the relevant tests pass!

One curious question is why to change the method from `__setstate__` to `__reduce__`? The [`pickle` module docs](https://docs.python.org/3/library/pickle.html#object.__setstate__) say:
> Although powerful, implementing `__reduce__()` directly in your classes is error prone. For this reason, class designers should use the high-level interface (i.e. ... `__setstate__()`) whenever possible.

(And to reiterate, I'm fine with changing to `__reduce__()`)

jorisvandenbossche:
I am not too familiar with the internal details of pickling, but I know we use `__reduce__` in several other packages I contribute to as well. 
Typically, I think `__setstate__` can be used to update the "state" (`__dict__`) of a python object after it has been initialized. But specifically in this case, the pygeos Geometry's are now Python extension types that don't have a mutable state (like builtins, you also can't assign random attributes to it). So I think in this case we actually have to use `__reduce__`. For example, the underlying GEOSGeometry pointer (which would need to be changed from LineString to LinearRing) is stored in `_ptr` attribute, but that's a readonly attribute and can't be changed from Python (only in C).

jorisvandenbossche:
Going to merge this, so I can update https://github.com/Toblerity/Shapely/pull/1163

PR #1163, opened 2021-07-07
[2.0] Remove shapely.geos.lgeos ctypes interface to GEOS

jorisvandenbossche:
This is on top of https://github.com/Toblerity/Shapely/pull/1161 and https://github.com/Toblerity/Shapely/pull/1162 (the first 2 commits). 

This removes the content `shapely/geos.py` and `shapely/ctypes_declarations.py`. 
The `shapely.geos` module is kept for now just for the `geos_version(_string)` variables (we should discuss whether we want to keep them long term in a `geos` submodule or move them top-level as we have in pygeos, but I think that's something we can see later).


jorisvandenbossche:
I updated this PR now the last PRs that were removing usages of the ctypes interface are merged.

PR #1166, opened 2021-07-12
Optionally exclude input geom from results

### LINKED ISSUE #1106: How to use strtree to find an element nearest to it but not its own?

sgillies:
Resolves #1106


shapely/strtree.py
```diff
@@ -289,10 +295,14 @@ def nearest_item(self, geom: BaseGeometry) -> Union[Any, None]:
 
         def callback(item1, item2, distance, userdata):
             try:
+                callback_userdata = ctypes.cast(userdata, ctypes.py_object).value
                 idx = ctypes.cast(item1, ctypes.py_object).value
                 geom2 = ctypes.cast(item2, ctypes.py_object).value
                 dist = ctypes.cast(distance, ctypes.POINTER(ctypes.c_double))
-                lgeos.GEOSDistance(self._rev[idx]._geom, geom2._geom, dist)
+                if callback_userdata["exclusive"] and self._rev[idx].equals(geom2):
```
sgillies, line 302: There's a great insight about cost of an equality test in https://github.com/pygeos/pygeos/issues/324, but that could wait until a beta release. Slow is okay for 1.8a2, nailing down the interface is the important thing.

PR #1169, opened 2021-07-15
Get free from global symbol table on Linux as we do for macos.

sgillies:
I saw some errors like `OSError: Could not find lib c or load any of its variants` in production at work. I suspect that the instances I'm using were in a very bad state, but we don't need to load libc (or its equivalent) on a *nix platform: Python has already loaded it and `free` will be found in the global symbol table.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -23,6 +23,8 @@ New features:
 
 Bug fixes:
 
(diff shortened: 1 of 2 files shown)


sgillies:
Disregarding the coverage report.

PR #1170, opened 2021-07-17
Remove the DeprecationWarning from STRtree constructor

jorisvandenbossche:
As a follow-up on https://github.com/Toblerity/Shapely/pull/1112, if we don't have concrete plans to change the API, we shouldn't raising a warning in the constructor every time?


jorisvandenbossche:
@sgillies this might also be worth including for the next 1.8a?

sgillies (approved):
Thanks @jorisvandenbossche !

PR #1171, opened 2021-07-17
TST: add more tests for new query_items

jorisvandenbossche:
Another follow-up on https://github.com/Toblerity/Shapely/pull/1112. Some aspects of the new `query_items` were not yet covered in the tests:

- constructing the STRtree with explicit `items` sequence + using `query_items` _without_ providing explicit items (so using a default enumeration index then)
- testing that the returned item is still correct if some geometries were dropped before constructing the GEOS tree

### DIFF
--- a/tests/test_strtree.py
+++ b/tests/test_strtree.py
@@ -7,12 +7,17 @@ import sys
 import pytest
 
 from shapely.errors import ShapelyDeprecationWarning
-from shapely.geometry import Point, Polygon
+from shapely.geometry import Point, Polygon, box
 from shapely.geos import geos_version
 from shapely import strtree
 from shapely.strtree import STRtree
+from shapely import wkt
 
 from .conftest import requires_geos_342
+
+
+point = Point(2, 3)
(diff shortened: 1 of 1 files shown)


tests/test_strtree.py
```diff
@@ -42,6 +47,38 @@ def test_query_enumeration_idx(geoms, query_geom, expected):
...
+@pytest.mark.parametrize("items", [None, list(range(1, 6)), list("abcde")])
+@pytest.mark.parametrize(
+    "query_geom,expected",
+    [(Point(2, 2).buffer(0.99), [2]), (Point(2, 2).buffer(1.0), [1, 2, 3])],
+)
+def test_query_items(geoms, items, query_geom, expected):
+    """Store enumeration idx"""
+    with pytest.warns(ShapelyDeprecationWarning):
+        tree = STRtree(geoms, items)
+    results = tree.query_items(query_geom)
```
jorisvandenbossche, line 61: This is a version of the test above, but with different options for `items`

sgillies (approved):
Great!

PR #1172, opened 2021-07-17
Only support separate geoms, items sequences in STRtree constructor

jorisvandenbossche:
Another follow-up on #1112. See my comment at https://github.com/Toblerity/Shapely/pull/1112#discussion_r671652931 about supporting zipped geoms/items sequence vs only supporting separate sequences. 

Assuming the zipped sequences in #1112 were a left-over from the initial version of that PR, this PR updates the STRtree constructor (and tests) to:

- only support separate sequences as `STRtree(geoms, items)`, and thus no longer supports `STRtree(zip(geoms, items))`
- also removed support for `STRtree(None)`. This didn't work before in released versions, and IMO it's cleaner to keep requiring `geoms` to be a sequence (it can still be an empty sequence)

If the intention is to mainly support the separate sequences, I would personally not support the zipped sequences (i.e. sequence tuples) to keep the interface simple and explicit.

### DIFF
--- a/shapely/strtree.py
+++ b/shapely/strtree.py
@@ -20,7 +20,7 @@ References
 
 import ctypes
 import logging
-from typing import Any, ItemsView, Iterable, Iterator, Sequence, Tuple, Union
+from typing import Any, ItemsView, Iterable, Iterator, Optional, Sequence, Tuple, Union
 import sys
 from warnings import warn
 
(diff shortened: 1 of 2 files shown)


sgillies (approved):
Thank you @jorisvandenbossche !

PR #1174, opened 2021-07-20
Silence NumPy 1.21 DeprecationWarning by raising AttributeError

### LINKED ISSUE #1173: NumPy 1.21 __array_interface__ behavior
NumPy 1.21 has a new behavior described in [this release note](https://github.com/numpy/numpy/releases/tag/v1.21.0):
> Exceptions will be raised during array-like creation: When an object raised an exception during access of the special
attributes `__array__` or `__array_interface__`, this exception was usually ignored. A warning is now given when the exception is anything but AttributeError. To silence the warning, the type raising the exception has to be adapted to raise an AttributeError.

Shapely<2.0 versions still have a `__array_interface__` property for most (all?) geometry types, and some of them, e.g. Polygon raise `NotImplementedError("A polygon does not itself provide the array interface. Its rings do.")`. This is problematic with this example:
```
import numpy as np
import warnings
from random import random
warnings.filterwarnings('always')

geoms = [Polygon((random(), random()) for _ in range(3)) for _ in range(3)]
arr = np.empty(len(geoms), dtype=object)
arr[:] = geoms
```
which will show this message three times (one for each geometry)
> `<stdin>`:1: DeprecationWarning: An exception was ignored while fetching the attribute `__array_interface__` from an object of type 'Polygon'.  With the exception of `AttributeError` NumPy will always raise this exception in the future.  Raise this deprecation warning to see the original exception. (Warning added NumPy 1.21)

This is also an issue for [GeoPandas](https://github.com/geopandas/geopandas/issues/2013), which will flood a terminal with warnings for large GeoSeries.

One very quick fix is to change the exception raised by `Polygon.__array_interface__` to `AttributeError`. Other geometries may also need to be adjusted, but the BaseGeometry class should still raise NotImplementedError. If this seems reasonable, I can prepare a PR for 1.8.

mwtoews:
Possible fix for NumPy 1.21 DeprecationWarning for exceptions raised in `__array_interface__`.  Only `BaseGeometry.__array_interface__` still raise NotImplementedError, as it is an abstract class.

Tests for each geometry type (except GeometryCollection) are added, which creates/assigns an object array.

Comments/reviews are welcome!

Closes #1173

### DIFF
--- a/shapely/coords.py
+++ b/shapely/coords.py
@@ -116,7 +116,10 @@ class CoordinateSequence:
     def _ctypes(self):
         self._update()
         has_z = self._ndim == 3
-        n = self._ndim
+        n = self._ndim or 0
+        if n == 0:
+            # ignore with NumPy 1.21 __array_interface__
+            raise AttributeError("empty geometry sequence")
         m = self.__len__()
         array_type = c_double * (m * n)
         data = array_type()
--- a/tests/test_linestring.py
+++ b/tests/test_linestring.py
@@ -106,6 +106,7 @@ def test_numpy_object_array():
     geom = LineString([(0.0, 0.0), (0.0, 1.0)])
     ar = np.empty(1, object)
     ar[:] = [geom]
+    assert ar[0] == geom
 
 
 def test_from_invalid_dim():
--- a/tests/test_linestring.py
+++ b/tests/test_linestring.py
@@ -87,6 +87,7 @@ def test_from_numpy():
     assert line.coords[:] == [(1.0, 2.0), (3.0, 4.0)]
 
 
+@pytest.mark.filterwarnings("error:An exception was ignored")  # NumPy 1.21
 def test_numpy_empty_linestring_coords():
     np = pytest.importorskip("numpy")
 
@@ -95,6 +96,16 @@ def test_numpy_empty_linestring_coords():
     la = np.asarray(line.coords)
 
     assert la.shape == (0,)
+
+
+@shapely20_deprecated
+@pytest.mark.filterwarnings("error:An exception was ignored")  # NumPy 1.21
+def test_numpy_object_array():
+    np = pytest.importorskip("numpy")
+
+    geom = LineString([(0.0, 0.0), (0.0, 1.0)])
+    ar = np.empty(1, object)
+    ar[:] = [geom]
 
 
 def test_from_invalid_dim():
--- a/tests/test_polygon.py
+++ b/tests/test_polygon.py
@@ -124,6 +124,7 @@ def test_numpy_object_array():
     geom = Polygon([(0.0, 0.0), (0.0, 1.0), (1.0, 1.0)])
     ar = np.empty(1, object)
     ar[:] = [geom]
+    assert ar[0] == geom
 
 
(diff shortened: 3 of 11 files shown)


shapely/coords.py
```diff
@@ -116,7 +116,10 @@ def __getitem__(self, key):
     def _ctypes(self):
         self._update()
         has_z = self._ndim == 3
-        n = self._ndim
+        n = self._ndim or 0
+        if n == 0:
```
mwtoews, line 120: This is the same edit as with `shapely/speedups/_speedups.pyx`. It keeps and tests `n` as an integer, rather than testing `if n is None` (and yes, `self._ndim is None` for empty geometries).

tests/test_polygon.py
```diff
@@ -109,13 +109,23 @@ def test_numpy_linearring_coords():
...
 
+@pytest.mark.filterwarnings("error:An exception was ignored")  # NumPy 1.21
 def test_numpy_empty_linearring_coords():
     np = pytest.importorskip("numpy")
 
     ring = LinearRing()
     assert np.asarray(ring.coords).shape == (0,)
 
 
+@pytest.mark.filterwarnings("error:An exception was ignored")  # NumPy 1.21
```
mwtoews, line 120: This is the only test that didn't need `@shapely20_deprecated` to silence "ShapelyDeprecationWarning: The array interface is deprecated...", and I have no idea why. @jorisvandenbossche ?

mwtoews:
@sgillies I'm keen to see this in the next pre-release

jorisvandenbossche:
@mwtoews sorry for the slow response here. I'll try to take a look tomorrow

sgillies (approved):
Thanks @mwtoews !

tests/test_linestring.py
```diff
@@ -87,6 +87,7 @@ def test_from_numpy():
     assert line.coords[:] == [(1.0, 2.0), (3.0, 4.0)]
 
 
+@pytest.mark.filterwarnings("error:An exception was ignored")  # NumPy 1.21
 def test_numpy_empty_linestring_coords():
```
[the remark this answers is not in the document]
jorisvandenbossche: Whoops, sorry, missed the "error" :) (I thought it was ignoring those warnings)

tests/test_polygon.py
jorisvandenbossche: So I _think_ this is because Polygon already didn't have an array interface before (its rings have one), so for polygon the array interface didn't need to be deprecated. In addition, it also doesn't have deprecation of `__len__` or iteration like the multi-part geometries (that also didn't work before), where those warnings are also typically raises when interacting with numpy.

jorisvandenbossche (approved):
Thanks, looks good! Changing the error we raise is indeed the good solution for 1.8

I added a small change: I added a similar test for GeometryCollection (it should also work for that type, and the test passes), and also added an assert that the geometry stored in the array is correct (that it's eg not unpacked).

mwtoews:
Thanks for the additions @jorisvandenbossche ! And also nice to see you work on a general solution for geopandas too.

PR #1176, opened 2021-07-28
Added /opt/homebrew/lib for GEOS shared library

mcuntz:
I added the directory /opt/homebrew/lib to the list of directories to be searched for the GEOS
  shared library. Homebrew installs into /opt/homebrew rather than /usr/local on systems with the new Apple Silicon (M1) chip.

Actual behaviour of
`python -c "from shapely.geos import lgeos"`

is

```
  File "/Users/cuntz/.pyenv/versions/3.9.6/envs/test-cartopy/lib/python3.9/site-packages/shapely/geos.py", line 136, in <module>
    _lgeos = load_dll('geos_c', fallbacks=alt_paths)
  File "/Users/cuntz/.pyenv/versions/3.9.6/envs/test-cartopy/lib/python3.9/site-packages/shapely/geos.py", line 60, in load_dll
    raise OSError(
OSError: Could not find lib geos_c or load any of its variants ['/Library/Frameworks/GEOS.framework/Versions/Current/GEOS', '/opt/local/lib/libgeos_c.dylib', '/usr/local/lib/libgeos_c.dylib'].
```

## Operating system

macOS Big Sur (Version 11.5.1) on Mac mini (M1, 2020).
Python 3.9.6

Current master on Github.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -31,6 +31,8 @@ Bug fixes:
   geometry are avoided by never passing an empty geometry (#1134).
 - Python's builtin super() is now used only as described in PEP 3135 (#1109).
 - Only load conda GEOS dll if it exists (on Windows) (#1108).
+- Add /opt/homebrew/lib to the list of directories to be searched for the GEOS
+  shared library.
 
 1.8a1 (2021-03-03)
 ------------------
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -33,6 +33,7 @@ Bug fixes:
 - Only load conda GEOS dll if it exists (on Windows) (#1108).
 - Add /opt/homebrew/lib to the list of directories to be searched for the GEOS
   shared library.
+- Added new library search path to assist app creation with cx_Freeze.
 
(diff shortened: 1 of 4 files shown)


mayteio:
+1!

mcuntz:
I also added
`os.path.join(os.path.dirname(sys.executable), 'libgeos_c.1.dylib')`
to search path in case of _frozen_ applications (.exe, .app) to support cx_Freeze on top of py2app and pyinstaller.

b4stien:
I can confirm this would fix the current issue on ARM M1 Macs (and would be greatly appreciated). Thanks for your work!

mwtoews (approved):
These changes look fine to me. As there currently isn't any CI coverage on macOS, we'll have to trust mac users with these changes.

sgillies:
I don't love the idea of continuing to add more hard coded paths, but also don't see why we can't do it and continue to hold our noses until 2.0 allows us to stop using ctypes. Thanks @mcuntz and @mwtoews .

PR #1177, opened 2021-08-03
Refactor STRtree implementation to store geoms/idx sequences instead of reverse mapping

jorisvandenbossche:
This PR takes the implementation details from https://github.com/Toblerity/Shapely/pull/1064 (i.e. using  `geoms` and `idxs` sequences (lists) to store the data with which the GEOSSTRtree gets created, instead of using a mapping), but cleanly on top of the new interface (`query_geoms`/`query_items`) from https://github.com/Toblerity/Shapely/pull/1112, without all the discussion on the interface from #1064.

It doesn't change any user facing API now after #1112, but only refactors the implementation (how to store the items). As mentioned in https://github.com/Toblerity/Shapely/pull/1112#issuecomment-875108728, this version is closer to what we would do in the shapely-2.0 branch on top of the C-based STRtree (which I started in https://github.com/Toblerity/Shapely/pull/1161). IMO it is useful to already do this now (it will also make it a bit easier to have the master and shapely-2.0 branches next to each other for a while).

### DIFF
--- a/shapely/strtree.py
+++ b/shapely/strtree.py
@@ -104,44 +104,34 @@ class STRtree:
             ShapelyDeprecationWarning,
             stacklevel=2,
         )
-        self._tree = None
         self.node_capacity = node_capacity
-        self._rev = {
-            item: geom
-            for geom, item in self._iterinitdata(geoms, items)
-            if not geom.is_empty
-        }
-        if self._rev:
-            self._init_tree(self._rev.items())
 
-    def _iterinitdata(
-        self,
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
cc @sgillies what do you think about this PR?

sgillies (approved):
:+1: thanks @jorisvandenbossche !

sgillies:
I'll merge and then we could have another pre-release on Monday.

Issue #1096, opened 2021-03-03
Shapely 1.8 array interface deprecation for object arrays


mwtoews:
I'm looking at some of the shapely 1.8 deprecation impacts in a niche package, and I'm stumbling over one issue that can be represented in a small example:

```python
import numpy as np
from shapely.geometry import Point, MultiPoint
mp = MultiPoint([Point(1, 2), Point(3, 4)])
mp_list = [mp]
a = np.empty(len(mp_list), dtype=object)
a[:] = mp_list
print(a[0].wkt)
```
This example works for all current and development versions of shapely, printing `MULTIPOINT (1 2, 3 4)` from the object array.
 
Shapely 1.7 does not show any warnings.

Shapely 1.8dev shows two warnings:
> /usr/bin/ipython3:1: ShapelyDeprecationWarning: `__len__` for multi-part geometries is deprecated and will be removed in Shapely 2.0. Check the length of the `geoms` property instead to get the  number of parts of a multi-part geometry.
  #! /bin/sh
/usr/lib/python3/dist-packages/IPython/core/interactiveshell.py:2882: ShapelyDeprecationWarning: The array interface is deprecated and will no longer work in Shapely 2.0. Convert the '.coords' to a numpy array instead.
  exec(code_obj, self.user_global_ns, self.user_ns)

Shapely 2.0dev does not show any warnings, and works the same as any other version.

---

It seems that the warnings from shapely 1.8dev are crying wolf. I understand these shenanigans are permitted with object arrays. Is there any way for shapely 1.8 to omit these warnings? Or should I just refactor to ether of these?
```python
a = np.empty(len(mp_list), dtype=object)
for idx, obj in enumerate(mp_list):
    a[idx] = obj

# or

import warnings
with warnings.catch_warnings():
    warnings.simplefilter("ignore")
    a = np.empty(len(mp_list), dtype=object)
    a[:] = mp_list
```

jorisvandenbossche:
Yeah, this is an annoying aspect when dealing with numpy arrays. I encountered the same issue in GeoPandas, and wrote a helper for the "ignore warnings" approach, see https://github.com/geopandas/geopandas/pull/1659 (we do this a lot in geopandas ..)

I am not fully sure why it happens (because in the end the geometries are not converted to its parts or coordinates), but apparently, the warnings are still triggered somewhere in the numpy internals, outside our control. 

[...]

PR #1180, opened 2021-08-21
Add migration guide with explanation how to update your code for Shapely 1.8 / 2.0

jorisvandenbossche:
In preparation of a next (final?) Shapely 1.8 release, I wrote up a more detailed page on the warnings included in Shapely 1.8 (and upcoming changes in 2.0), and how to update your code to get rid of the warnings / prepare it for Shapely 2.0.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -37,7 +37,9 @@ Bug fixes:
 
 Shapely 1.8.0 will be a transitional version. There are a few bug fixes and new
 features, but it is mainly about warning of the upcoming changes in 2.0.0.
-Several more pre-releases before 1.8.0 are expected.
+Several more pre-releases before 1.8.0 are expected. See the migration guide
+to Shapely 1.8 / 2.0 for more details on how to update your code
+(https://shapely.readthedocs.io/en/latest/migration.html).
 
 Python version support:
 
@@ -50,7 +52,6 @@ version 2.0.0.
 
 - ops.cascaded_union
 - geometry .empty()
-- geometry .to_wkb() and .to_wkt()
 - geometry .ctypes and .__array_interface__
 - multi-part geometry .__len__
 
@@ -61,6 +62,9 @@ not yet decided, but will be decided before 1.8.0 is released.
 
 Deprecation warnings will be emitted in 1.8a1 when any of these features are
 used.
+
+The deprecated .to_wkb() and .to_wkt() methods on the geometry objects have
+been removed.
 
 New features:
 
--- a/docs/migration.rst
+++ b/docs/migration.rst
@@ -194,9 +194,8 @@ convert the ``.coords`` attribute instead::
 The ``array_interface()`` method and ``ctypes`` attribute will be removed in
 Shapely 2.0, but since Shapely will start requiring NumPy as a dependency,
 you can use NumPy or its array interface directly. Check the NumPy docs on
-the `ctypes attribute <https://numpy.org/doc/stable/reference/generated/numpy.ndarray.ctypes.html>`__
-or the `array interface <https://numpy.org/doc/stable/reference/arrays.interface.html>`__
-for more details.
+the :py:attr:`ctypes <numpy:numpy.ndarray.ctypes>` attribute
+or the :ref:`array interface <numpy:arrays.interface>` for more details.
 
 Creating NumPy arrays of geometry objects
 -----------------------------------------
--- a/docs/migration.rst
+++ b/docs/migration.rst
@@ -273,8 +273,8 @@ manager that can be copied into your project::
         def ignore_shapely2_warnings():
             yield
 
-This can then be used when creating NumPy arrays (be careful to *only* for
-this specific purpose, and not generally suppress those warnings)::
+This can then be used when creating NumPy arrays (be careful to *only* use it
+for this specific purpose, and not generally suppress those warnings)::
 
     geoms = [...]
     arr = np.empty(len(geoms), dtype="object")
--- a/docs/migration.rst
+++ b/docs/migration.rst
@@ -46,7 +46,7 @@ In Shapely 1.8, this will start raising a warning::
     in place is deprecated, and will not be possible any more in Shapely 2.0
 
 and starting with version 2.0.0, all geometry objects will become immutable
-(as a consequence, they will also become hashable and thefore usable as, for
+(as a consequence, they will also become hashable and therefore usable as, for
 example, dictionary keys).
 
 **How do I update my code?** There is no direct alternative for mutating the
@@ -137,11 +137,11 @@ support those features, and for those classes there is no change in behaviour
 for this aspect.
 
 
-The array interface and conversion to numpy
+The array interface and conversion to NumPy
 ===========================================
 
 Shapely provides an array interface to have easy access to the coordinates as,
-for example, numpy arrays (:ref:`manual section <array-interface>`).
+for example, NumPy arrays (:ref:`manual section <array-interface>`).
 
 A small example::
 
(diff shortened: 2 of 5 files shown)


CHANGES.txt
```diff
@@ -50,7 +52,6 @@ version 2.0.0.
 
 - ops.cascaded_union
 - geometry .empty()
-- geometry .to_wkb() and .to_wkt()
```
jorisvandenbossche, line 53: Those were actually removed and not deprecated, so updated the note (they were deprecated long time ago, see https://github.com/Toblerity/Shapely/pull/974)

docs/migration.rst
```diff
@@ -0,0 +1,249 @@
...
+assigning custom attributes, which currently works, will no longer be possible.
+
+Currently you can do::
+
+    >>> line.name = "my_geometry"
+    >>> line.name
+    'my_geometry'
+
+This doesn't raise a warning in Shapely 1.8, but will start raising an
+AttributeError in Shapely 2.0.
```
jorisvandenbossche, line 70: I realized this while writing this guide, but we could actually still add a warning for this. That might be useful?

jorisvandenbossche:
I should probably still add some more content about putting geometries in numpy arrays, cfr https://github.com/Toblerity/Shapely/issues/1096

docs/migration.rst
```diff
@@ -0,0 +1,249 @@
...
+    >>> np.array(line.coords)
+    array([[0., 0.],
+           [1., 1.],
+           [2., 2.]])
+
+The ``array_interface()`` method and ``ctypes`` attribute will be removed in
+Shapely 2.0, but since Shapely will start requiring NumPy as a dependency,
+you can use NumPy or its array interface directly.
+
+
```
mwtoews, line 195: Another related issue is creating an object array of shapely geometries, described in #1096. Is it worth discussing? It only impacts version 1.8.

mwtoews (approved):
Documentation looks well written, and covers the main user-cases that might be encountered.

There's only one potentially missing item, concerning object arrays.

docs/migration.rst
jorisvandenbossche: Yes, I certainly think that's worth describing (see also my comment https://github.com/Toblerity/Shapely/pull/1180#issuecomment-903147126 above). Will add a section about that.

jorisvandenbossche:
@mwtoews thanks for the fixes and review! 
I added a section about numpy array creation.

docs/migration.rst
```diff
@@ -0,0 +1,338 @@
...
+those deprecation warnings will be removed (and thus the ``filterwarnings``
+will no longer be necessary), and creation of NumPy arrays will generally be
+more robust.
+
+If you maintain code that depends on Shapely, and you want to have it work
+with multiple versions of Shapely, the above code snippet provides a context
+manager that can be copied into your project::
+
+    import contextlib
+    import shapely
```
jorisvandenbossche, line 254: If including this full code snippet might make the section a bit long to read, we could alternatively also link to the github issue's comment from where I copied this: https://github.com/Toblerity/Shapely/issues/1096#issuecomment-789998514 (which might also make it easier to comment / ask questions about it on the issue)

docs/migration.rst
jorisvandenbossche: Looking into it, what I wrote here is actually not completely correct. So PyGEOS objects are indeed fully immutable (a C extension type):

```python
In [1]: import pygeos

In [2]: p = pygeos.Geometry("POINT(1 1)")

In [3]: p
Out[3]: <pygeos.Geometry POINT (1 1)>

In [4]: p.name = "test"
AttributeError                            Traceback (most recent call last)
<ipython-input-4-3cf5ac6b0436> in <module>
----> 1 p.name = "test"

AttributeError: 'pygeos.lib.Geometry' object has no attribute 'name'
```

But when using the `shapely-2.0` branch, we actually create *python subclasses* for each geometry type (and register those to pygeos), and those subclasses are still mutable:

```python
In [1]: import pygeos

In [2]: p = pygeos.Geometry("POINT(1 1)")

In [3]: p
Out[3]: <shapely.geometry.Point POINT (1 1)>

In [4]: p.name = "test"

In [5]: p.name
Out[5]: 'test'
```

So in the end, it's not an inevitable consequence of the refactor, but we still have the _choice_ whether to allow or disallow this. 
I made a PR https://github.com/Toblerity/Shapely/pull/1181 to illustrate how we can disallow this in Shapely 2.0, but to discuss of course whether we want this or not.

jorisvandenbossche:
cc also @caspervdw @brendan-ward @martinfleis  feedback on those docs is very welcome!

sgillies (approved):
Beautiful writing @jorisvandenbossche. Is it okay if we go ahead with 1.8a3 before this is done and introduce this in 1.8b1?

jorisvandenbossche:
It's a doc only change, so it will be directly available at https://shapely.readthedocs.io/en/latest/ after merging (without a new release). So yes, you certainly don't have to wait on this one to cut a new alpha.

martinfleis (approved):
It looks really nice!

docs/migration.rst
```diff
@@ -0,0 +1,338 @@
...
+    LINESTRING (0 0, 10 0, 10 10)
+
+In Shapely 1.8, this will start raising a warning::
+
+    >>> line.coords = [(0, 0), (10, 0), (10, 10)]
+    ShapelyDeprecationWarning: Setting the 'coords' to mutate a Geometry
+    in place is deprecated, and will not be possible any more in Shapely 2.0
+
+and starting with version 2.0.0, all geometry objects will become immutable
+(as a consequence, they will also become hashable and therefore usable as, for
```
brendan-ward, line 49:
```suggestion
.  As a consequence, they will also become hashable and therefore usable as, for
```
Suggest removing from parenthetical, this is a nice benefit.

docs/migration.rst
```diff
@@ -0,0 +1,338 @@
...
+    >>> line.coords
+    <shapely.coords.CoordinateSequence at 0x7f2e09e88d60>
+    >>> np.array(line.coords)
+    array([[0., 0.],
+           [1., 1.],
+           [2., 2.]])
+
+The ``array_interface()`` method and ``ctypes`` attribute will be removed in
+Shapely 2.0, but since Shapely will start requiring NumPy as a dependency,
+you can use NumPy or its array interface directly.
```
brendan-ward, line 196: This might benefit from an example showing how to do this directly instead of `line.array_interface()`.

docs/migration.rst
```diff
@@ -0,0 +1,338 @@
...
+  outside Shapely geometries are deprecated and will be removed in Shapely
+  2.0 (e.g. created using ``asShape()``). They have little to no benefit
+  compared to the normal geometry classes, as thus you can convert to your
+  data to a normal geometry object instead. Use the ``shape()`` function
+  instead to convert a GeoJSON-like dict to a Shapely geometry.
+
+- The ``empty()`` method on a geometry object is deprecated.
+
+- The ``shapely.ops.cascaded_union`` function is deprecated. Use
+  ``shapely.ops.unary_union`` instead.
```
brendan-ward, line 338:
```suggestion
  ``shapely.ops.unary_union`` instead, which internally already uses a cascaded union operation for better performance.
```

(just to indicate that the functionality / perf. benefit is not being taken away, it is just absorbed into other functions)

brendan-ward (commented):
This looks great, thanks for working on this @jorisvandenbossche !

A few minor comments to consider.

docs/migration.rst
jorisvandenbossche: I am a bit hesitant to add an explicit example of accessing `np.array(line.coords).__array_interface__`, because I can't think of many good reasons you would need to do that directly yourself (also because it is somewhat legacy, with now the more general python buffer interface: https://numpy.org/doc/stable/reference/arrays.interface.html), but it would basically be:

```
In [1]: from shapely.geometry import LineString

In [2]: line = LineString([(0, 0), (1, 1), (2, 2)])

In [4]: coords = np.array(line.coords)

In [5]: coords.__array_interface__
Out[5]: 
{'data': (94533168670016, False),
 'strides': None,
 'descr': [('', '<f8')],
 'typestr': '<f8',
 'shape': (3, 2),
 'version': 3}
```

And numpy arrays also have a `ctypes` attribute, although it is a bit different:
```
In [6]: coords.ctypes
Out[6]: <numpy.core._internal._ctypes at 0x7f7ca9abdd00>

In [7]: coords.ctypes.data
Out[7]: 94533168670016
```

For this we could point to https://numpy.org/doc/stable/reference/generated/numpy.ndarray.ctypes.html

I could maybe at least point to those urls.

brendan-ward: Pointing to those URLs is fine.

docs/migration.rst
jorisvandenbossche: And also opened a PR to deprecate it for Shapely 1.8: https://github.com/Toblerity/Shapely/pull/1183

docs/migration.rst
mwtoews: Hope you don't mind using `sphinx.ext.intersphinx` for the numpy links. Outputs were checked locally to make sure they are the same.

jorisvandenbossche: Certainly not, thanks!

caspervdw (approved):
The migration guide looks solid @jorisvandenbossche , I have no comments or suggestions.

jorisvandenbossche:
OK, merging this. We can always further update the guide based on feedback / issues we notice.

jorisvandenbossche:
Rendered version can now be seen here: https://shapely.readthedocs.io/en/latest/migration.html

PR #1181, opened 2021-08-23
[2.0] Disallow setting custom attributes on geometry objects

jorisvandenbossche:
xref https://github.com/Toblerity/Shapely/pull/1180/files#r693378882

PyGEOS geometries are "immutable", not only regarding the coordinates of the geometry, but also the python object: you can't assign random attributes as you can do with a normal class defined in python (this is because they are C extension types):

```python
>>> import pygeos
>>> p = pygeos.Geometry("POINT(1 1)")
>>> p
<pygeos.Geometry POINT (1 1)>
>>> p.name = "test"
...
AttributeError: 'pygeos.lib.Geometry' object has no attribute 'name'
```

However, because in Shapely we still create a Python-defined subclass for each of the geometry types (and register this, so the C code will actually instantiate those subclasses), we get the typical behaviour of standard python classes still being "mutable" objects:

```python
# using shapely-2.0 branch
>>> p = pygeos.Geometry("POINT(1 1)")
>>> p
<shapely.geometry.Point POINT (1 1)>
>>> p.name = "test"
>>> p.name
'test'
```

It's however relatively easy to keep the "can't set custom attribute" behaviour of the base C class (*if* we want this), by adding an empty `__slots__` to each of the python subclasses. (which is what this PR is doing)

---

Personally, I liked the fact that pygeos Geometries were completely immutable, so I would be in favor of preserving this characteristic in Shapely as well. 
(for example, we will generally re-use objects (eg when broadcasting) because they are immutable, and for such cases, setting attributes might give surprising behaviour)

_If_ we would want to disallow setting attributes in Shapely 2.0, following PyGEOS behaviour, that's of course a breaking change compared to Shapely 1.x, but I think it should be relatively easy to add a deprecation warning for this change as well (adding a warning in `__setattr__`).

cc @sgillies @caspervdw

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -99,7 +99,7 @@ class BaseGeometry(pygeos.Geometry):
     # _ndim : int
     #     Number of dimensions (2 or 3, generally)
 
-    _coords = None
+    __slots__ = []
 
     def __new__(self):
         # TODO create empty geometry - should we deprecate this constructor?
@@ -182,10 +182,8 @@ class BaseGeometry(pygeos.Geometry):
     @property
     def coords(self):
         """Access to geometry's coordinates (CoordinateSequence)"""
-        if self._coords is None:
-            coords_array = pygeos.get_coordinates(self, include_z=self.has_z)
-            self._coords = CoordinateSequence(coords_array)
-        return self._coords
+        coords_array = pygeos.get_coordinates(self, include_z=self.has_z)
+        return CoordinateSequence(coords_array)
 
     @property
     def xy(self):
@@ -663,6 +661,8 @@ class BaseGeometry(pygeos.Geometry):
 
 
 class BaseMultipartGeometry(BaseGeometry):
+
+    __slots__ = []
 
     def shape_factory(self, *args):
         # Factory for part instances, usually a geometry class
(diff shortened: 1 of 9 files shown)


jorisvandenbossche:
Additional benefit of defining `__slots__` is that this reduces the memory overhead of the Python class (you of course still have the underlying GEOS object that doesn't change and can generally be a lot larger (for large geometries), but certainly for eg Points, reducing the Python object overhead can probably be beneficial for large arrays of Points).

(from a quick test with `sys.getsizeof`, the (python) size of a Point object seems to reduce from 64 to 32 bytes with this PR)

jorisvandenbossche:
cc @caspervdw (correct name this time ;))

jorisvandenbossche:
I also opened a PR for the shapely 1.8 version with a deprecation: https://github.com/Toblerity/Shapely/pull/1183

shapely/geometry/base.py
```diff
@@ -182,10 +182,8 @@ def __ne__(self, other):
     @property
     def coords(self):
         """Access to geometry's coordinates (CoordinateSequence)"""
-        if self._coords is None:
-            coords_array = pygeos.get_coordinates(self, include_z=self.has_z)
-            self._coords = CoordinateSequence(coords_array)
-        return self._coords
+        coords_array = pygeos.get_coordinates(self, include_z=self.has_z)
+        return CoordinateSequence(coords_array)
```
jorisvandenbossche, line 186: We can see later if we want to "cache" this again (there might be other attributes you could cache as well)

PR #1183, opened 2021-08-24
Deprecate setting of custom attributes on geometry objects

jorisvandenbossche:
See https://github.com/Toblerity/Shapely/pull/1181 (and https://github.com/Toblerity/Shapely/pull/1180/files#r693378882). _If_ we want to disallow it for Shapely 2.0, this is adding a deprecation warning for it for 1.8

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -53,6 +53,7 @@ version 2.0.0.
 - geometry .to_wkb() and .to_wkt()
 - geometry .ctypes and .__array_interface__
 - multi-part geometry .__len__
+- setting custom attributes on geometry objects
 
 Geometry objects will become immutable in version 2.0.0.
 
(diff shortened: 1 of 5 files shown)


jorisvandenbossche:
@sgillies @mwtoews you're fine with this additional deprecation?

mwtoews (approved):
Changes look good for shapely 1.8.

mwtoews:
The only thing we can't warn against is users that extend geometry types:
```python

class MyPoint(Point):
    extra = 123

mp = MyPoint(2, 4)
print(mp.extra)  # 123
mp.extra = 456
print(mp.extra)  # 456
```

jorisvandenbossche:
I suppose that in _theory_ we could also add a warning when subclassing the geometry classes (with a metaclass, see eg https://stackoverflow.com/questions/16056574/how-does-python-prevent-a-class-from-being-subclassed), but not directly sure how that impacts other things.

jorisvandenbossche:
I am going to interpret the thumbs up of @sgillies at https://github.com/Toblerity/Shapely/pull/1181 as an approval of the general idea, and merge this.

PR #1191, opened 2021-09-08
Add test for importing shapely.vectorized on Appveyor

jorisvandenbossche:
Starting with a test for https://github.com/Toblerity/Shapely/issues/1184


jorisvandenbossche:
So this at least reproduces the error from the issue (for Python 3.8 and Python 3.9)

PR #1192, opened 2021-09-08
Support LinearRing objects in the shape() function

### LINKED ISSUE #1179: shapely.geometry.shape does not work with LinearRing
Shapely 1.7.1, Python 3.9.6 on mac

The `shape` function knows how to process Polygons but not LinearRings. So I need to convert my LinearRing into a Polygon to use it with certain libraries that attempt to work with various geometry types by using `shape`.

To remedy, the `shape` function should be taught about LinearRing.

```python

>>> from shapely.geometry import LinearRing, shape, Polygon
>>> l = LinearRing([(0,0), (0,1), (1,0)])
>>> p = Polygon([(0,0), (0,1), (1,0)])
>>> shape(p)
<shapely.geometry.polygon.Polygon object at 0x1032e2b80>
>>> shape(l)
  File ".venv/lib/python3.9/site-packages/shapely/geometry/geo.py", line 116, in shape
    raise ValueError("Unknown geometry type: %s" % geom_type)
ValueError: Unknown geometry type: linearring
```

jorisvandenbossche:
Closes #1179

Note that LinearRing type geometry is not actually supported in the GeoJSON spec, and the `__geo_interface__` spec doesn't explicitly mention it (https://gist.github.com/sgillies/2217756). 

But since we create python dicts with "LinearRing" type in `mapping(linearring)` or `linearring.__geo_interface__`, it seems most logical to indeed support the roundtrip in `shape(..)` as well.

### DIFF
--- a/shapely/geometry/geo.py
+++ b/shapely/geometry/geo.py
@@ -8,7 +8,7 @@ from shapely.errors import ShapelyDeprecationWarning
 
 from .point import Point, asPoint
 from .linestring import LineString, asLineString
-from .polygon import Polygon, asPolygon
+from .polygon import LinearRing, Polygon, asPolygon
 from .multipoint import MultiPoint, asMultiPoint
 from .multilinestring import MultiLineString, asMultiLineString
 from .multipolygon import MultiPolygon, MultiPolygonAdapter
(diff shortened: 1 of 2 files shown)


sgillies (approved):
Yes, let's do this. Thanks @jorisvandenbossche !

PR #1194, opened 2021-09-30
TST: creating linearring with too few coordinates gives different error with latest GEOS

jorisvandenbossche:
GEOS changed to actually allow this, so PyGEOS now checks for this itself (giving a ValueError instead of GEOSException). Leaving the GEOSException for now in the pytest.raises as well, for if you are using this with released pygeos.

See https://github.com/pygeos/pygeos/pull/378


jorisvandenbossche:
I am going to merge this, since it makes CI green again for the other PRs (the failure in appveyor is the known WKB error I still need to figure out).

PR #1214, opened 2021-10-29
Fix unintentional calling of property

martinxsliu:
`_array_interface_base` is a decorated property, but is called like a function in the public `array_interface_base` method.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -223,7 +223,7 @@ Bug fixes:
 - A missing special case for rectangular polygons has been added to the
   polylabel algorithm (#644).
 - LinearRing can be created from a LineString (#638).
-- The prepared geoemtry validation condition has been tightened in #632 to fix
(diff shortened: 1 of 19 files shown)


jorisvandenbossche:
Thanks for the catch! 
Small question: were you actually using this? (or just noticed the mistake in the code?) We clearly don't seem to have tests for it, and internally it is actually only the private `_array_interface_base` that is being used in eg the public `array_interface()` of the Point class, and there it is called correctly. And we are only testing those `array_interface` calls, not the `array_interface_base` ..

jorisvandenbossche:
One more thing: could you rebase this against the maint-1.8 branch? (I changed the base here in the PR interface, but now that gives conflicts ..)

jorisvandenbossche:
This was merged in https://github.com/shapely/shapely/pull/1235

Thanks again @martinxsliu

PR #1217, opened 2021-11-03
Allow usage in a venv from a conda environment

TD22057:
We use conda to install newer versions of python for users than we can get on our basic linux systems.  We also install shapely (and cartopy , etc) into that conda environment for people.  The problem is that people want to be able to make small virtual environments with those conda env's and use them.  But the shared library loading path is hard coded in geos.py and fails when doing this.  There is a fairly simple fix using the site package which shows the correct search paths when in a venv.

Here is the basic error.  Starting by installing python 3.9 and shapely in a new conda package.  Then create a venv and try to import shapely.
```
conda create -y -p ./tools_v1 python=3.9 shapely
conda activate /home/trd/tools_v1
python -m venv --system-site-packages tools_v1_venv
source tools_v1_venv/bin/activate
python
Python 3.9.7 (default, Sep 16 2021, 13:09:58) 
[GCC 7.5.0] :: Anaconda, Inc. on linux
>>> import shapely.geometry
  File "/home/trd/tools_v1/lib/python3.9/site-packages/shapely/geometry/__init__.py", line 4, in <module>
  File "/home/trd/tools_v1/lib/python3.9/site-packages/shapely/geometry/base.py", line 19, in <module>
    from shapely.coords import CoordinateSequence
  File "/home/trd/tools_v1/lib/python3.9/site-packages/shapely/coords.py", line 8, in <module>
    from shapely.geos import lgeos
  File "/home/trd/tools_v1/lib/python3.9/site-packages/shapely/geos.py", line 81, in <module>
    _lgeos = CDLL(os.path.join(sys.prefix, 'lib', 'libgeos_c.so'))
  File "/home/trd/tools_v1/lib/python3.9/ctypes/__init__.py", line 382, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: /home/trd/tools_v1_venv/lib/libgeos_c.so: cannot open shared object file: No such file or directory
>>> 
```

The site package contains the list PREFIXES which stores the locations of the both the venv directory and base python directory.  So in this example, you can see:
```
>>> import site
>>> site.PREFIXES
['/home/trd/tools_v1_venv', '/home/trd/tools_v1', '/home/trd/tools_v1']
```

This patch changes geos.py in the CONDA_PREFIX branch to loop over these directories instead of the hard coded sys.prefix (which in this case is the venv).    The same fix might work in the darwin code branch but I have no way to test that so I didn't make that change.

### DIFF
--- a/shapely/geos.py
+++ b/shapely/geos.py
@@ -11,6 +11,7 @@ import logging
 import os
 import re
 import sys
+import site
 import threading
 from functools import partial
 
@@ -89,8 +90,14 @@ if sys.platform.startswith('linux'):
             _lgeos = CDLL(geos_pyinstaller_so[0])
             LOG.debug("Found GEOS DLL: %r, using it.", _lgeos)
     elif exists_conda_env():
-        # conda package.
-        _lgeos = CDLL(os.path.join(sys.prefix, 'lib', 'libgeos_c.so'))
+        # Using site.PREFIXES allows finding the lib from a venv as well.
+        for prefix in site.PREFIXES:
+            libpath = os.path.join(prefix, 'lib', 'libgeos_c.so' )
+            if os.path.exists(libpath):
+                _lgeos = CDLL(libpath)
+                break
+        else:
+           raise OSError("Could not find libgeos_c.so in {}".format(site.PREFIXES))
     else:
         alt_paths = [
             'libgeos_c.so.1',


TD22057:
It looks like you added my patch to the new geos.py - so thanks.

But - I think it might be still broken.  Instead of checking for $CONDA_PREFIX, the new geos.py says this:
```
def exists_conda_env():
    """Does this module exist in a conda environment?"""
    return os.path.exists(os.path.join(sys.prefix, 'conda-meta'))
```

But that fails for exactly the same reason that this PR was created.  conda-meta only exists in the base conda env.  Running python from the venv will show it's not a conda env.

PR #1218, opened 2021-11-03
Doc-migration : add alternative for custom attributes

LauLauThom:
I just came across this use-case of setting custom attributes for geometry object, and I though the solution of creating a custom object holding a reference to a geometry object and the custom attributes could be proposed.
Kind of simple but maybe not obvious to anyone.

### DIFF
--- a/docs/migration.rst
+++ b/docs/migration.rst
@@ -74,7 +74,17 @@ AttributeError in Shapely 2.0.
 **How do I update my code?** There is no direct alternative for adding custom
 attributes to geometry objects. You can use other Python data structures such as
 (GeoJSON-like) dictionaries or GeoPandas' GeoDataFrames to store attributes
-alongside geometry features. 
+alongside geometry features.  
+You can also create a costum object holding a reference to a shapely geometry object, in addition to the custom attribute(s)::
+
+    from shapely.geometry import LineString
+    
+    class CustomObject():
+      
+      def __init__(self, attribute1, coords):
+        self.attribute1 = attribute1
+        self.line = LineString(coords)
+
 
 Multi-part geometries will no longer be "sequences" (length, iterable, indexable)
 =================================================================================


docs/migration.rst
```diff
@@ -74,7 +74,17 @@ AttributeError in Shapely 2.0.
 **How do I update my code?** There is no direct alternative for adding custom
 attributes to geometry objects. You can use other Python data structures such as
 (GeoJSON-like) dictionaries or GeoPandas' GeoDataFrames to store attributes
-alongside geometry features. 
+alongside geometry features.  
+You can also create a costum object holding a reference to a shapely geometry object, in addition to the custom attribute(s)::
+
+    from shapely.geometry import LineString
+    
+    class CustomObject():
```
jorisvandenbossche, line 82: An alternative could also be to use a dataclass like:

```
from shapely.geometry.base import BaseGeometry
from dataclasses import dataclass

@dataclass
class CustomFeature:
    attribute1: int
    geometry: BaseGeometry
```

(and because there are many ways to do this, I am not fully sure to what extent we should give a very explicit code example, versus explaining in words)

jorisvandenbossche (commented):
@LauLauThom thanks for looking to improve the migration guide!

docs/migration.rst
LauLauThom: Hi Joris, well the data-class is special case of the CustomObject I propose in the end, right ? it still illustrates the alternative of holding a reference to a geometry object.

I agree they are many ways to do it, but for having been confronted with the issue, a code example can save some time and is indeed more explicit if one is not familiar with the actual terms (like "using an attribute holding a reference to a geometry object" might not be obvious to everyone).

We typed the example code already actually (thanks for the tip about dataclass decorator btw, did not know about it) so why not put both and let the user decide what they want to use. This might not be the best solution but it's short, functional and does not rely on any other package.

PR #1219, opened 2021-11-05
Clean-up old/unused files and scripts

jorisvandenbossche:
Cleaning up some (AFAIK) unused files in the repo. 

All of those files were not touched for years, empty, or unused wheel build scripts (which are not used, since we have shapely-wheels). So I _think_ all of them can be removed, but of course would be good if someone with more history in the project can check this.


mwtoews (commented):
The `DLLs_*` directories are cross-referenced in `setup.py` and `MANIFEST.in` (via found `git grep DLLs_`). It might be easier to address changes to `setup.py` when it gets rewritten for the 2.0 merge. `MANIFEST.in` could be changed in this PR.

Otherwise I agree that all of these files should be cleaned up.

jorisvandenbossche:
Good find, thanks! Will clean up those references.

PR #1222, opened 2021-11-06
Remove now unused shapely._buildcfg module

jorisvandenbossche:
The `shapely/_buildcfg.py` is not actually used at the moment (AFAIK), so we can remove that.


setup.py
```diff
@@ -69,10 +69,6 @@
 
 from _vendor.packaging.version import Version
 
-# Get geos_version from GEOS dynamic library, which depends on
-# GEOS_LIBRARY_PATH and/or GEOS_CONFIG environment variables
-from shapely._buildcfg import geos_version_string, geos_version, \
-        geos_config, get_geos_config
```
jorisvandenbossche, line 75: All those imported variables are either not used in this file or either redefined a bit lower

mwtoews (approved):
I'm certain `setup.py` is due for a rewrite soon, so this private module can be dropped.

Some background on this with PR #327

PR #1226, opened 2021-11-08
CI: Upgrade GEOS versions

mwtoews:
Another maintenance update for shapely-1.8, to use the latest versions of GEOS.

- GEOS 3.9.2 [announcement](https://lists.osgeo.org/pipermail/geos-devel/2021-November/010526.html)
- GEOS 3.10.1 [announcement](https://lists.osgeo.org/pipermail/geos-devel/2021-November/010556.html)

If we get around to shapely-1.8.1, the binary wheels should probably be packaged with GEOS 3.9.2.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -146,7 +146,7 @@ jobs:
           python -m pytest --cov shapely --cov-report term-missing "${SPEEDUPS_FLAG}"
 
       - name: Run doctests
-        if: ${{ matrix.python == '3.8' && matrix.numpy && matrix.speedups == '1' }}
+        if: ${{ matrix.python == '3.9' && matrix.numpy && matrix.speedups == '1' }}
         shell: bash
         run: |
           python -m pytest shapely --doctest-modules
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -48,27 +48,27 @@ jobs:
             speedups: 0
           # 2020
           - python: 3.9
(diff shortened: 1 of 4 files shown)


mwtoews:
Might as well also upgrade:

- GEOS 3.6.5 [announcement](https://lists.osgeo.org/pipermail/geos-devel/2020-December/009994.html)
- GEOS 3.8.2 [announcement](https://lists.osgeo.org/pipermail/geos-devel/2021-April/010201.html)

mwtoews:
Doctests fail with GEOS 3.8.2, as it seems there are small differences with coordinate values in WKT and in area calculations. I can't tell if the results from that version are better, but all other releases don't have these issues. Considering rolling GEOS back to 3.8.1

mwtoews:
Doctests now run on Python 3.9, and were made to be a bit more robust, and should run the same on any combination of Python and GEOS.

ecf0d6cbb692e398c028228b2b95c273c3554368 could be cherry-picked into `main`

mwtoews:
GEOS 3.8.1 is kept for Windows, as 3.8.2 does not compile without being patched with https://github.com/libgeos/geos/commit/da1b246f28df4c2b7863920fbc913b316ca77062

jorisvandenbossche (approved):

jorisvandenbossche:
> ecf0d6c could be cherry-picked into main

I am doing that in https://github.com/Toblerity/Shapely/pull/1228

PR #1228, opened 2021-11-08
Fix doctests to be more robust with different versions of GEOS

jorisvandenbossche:


jorisvandenbossche:
I directly moved it to python 3.10 / GEOS 3.10 (to match with pygeos). Seems no additional fixes are needed.

mwtoews (approved):

caspervdw (approved):

PR #1231, opened 2021-11-15
ENH: add normalize keyword to equals_exact() to normalize input geometries

### LINKED ISSUE #889: equals is True but almost_equals is False

`almost_equals` should return True, if `equals` is True.

```python
import numpy as np
from shapely.geometry import Polygon, box

coords = np.array(
       [[ 577., 1250.],
       [ 593., 1325.],
       [2447., 1002.],
       [2447.,  945.]], dtype=np.float32
)

p = Polygon(coords)
rect = Polygon.from_bounds(*p.bounds)
rect2 = box(*p.bounds)

print(rect.equals(rect2))
print(rect.almost_equals(rect2))
```

## Operating system

Ubuntu Linux 16

1.7.0

tfardet:
This PR fixes #889 by making sure that the coordinates are properly normalized before being compared.
Thus, we no longer get cases where ``equals`` is True but ``equals_exact`` is False.


tfardet:
mhh... tests are failing but I don't think that's on me, as far as I can tell this is probably due to main being a WIP for other functions: ``test_equality.py`` passes fine

jorisvandenbossche (commented):
Thanks for the contribution! 

I think we will probably want to have a keyword for doing this. So you can do `geom1.equals_exact(geom2, normalize=True)` or something like that to opt in to this, and not do it by default.

tfardet:
Right, keyword sounds reasonable.
Are you sure that we don't want to make this the default behavior, though?
I understand that this would be a behavior change but it does seem like the normally expected behavior to me...

tfardet:
OK, moved to keyword (default to True until discussion converges) and adapted the docstring accordingly

snowman2:
I just got the deprecation notice for `almost_equals` and wound my way through the issues and ended up here. What are your thoughts on calling `normalize` in `almost_equals` and keeping the method instead of adding `normalize` kwarg to `equals_exact`?

tfardet:
I pulled main to solve the conflicts.
I was wondering, @jorisvandenbossche : given that 2.0 will break a lot of things, would it not be the occasion to change the default behavior of ``equals_exact``, or would you still rather have ``nomalize`` be False by default?

jorisvandenbossche:
@tfardet sorry for letting this drag on for so long! And thanks for the update.

One code suggestion: we have a similar `equals_exact` function in `shapely/predicates.py`. I think we should add the keyword there as well (and actually can pass through the keyword from `BaseGeometry/equals_exact`). This would be something like:

```
if normalize:
    a = lib.normalize(a)
    b = lib.normalize(b)
```

jorisvandenbossche:
> I just got the deprecation notice for `almost_equals` and wound my way through the issues and ended up here. What are your thoughts on calling `normalize` in `almost_equals` and keeping the method instead of adding `normalize` kwarg to `equals_exact`?

That could also have been a option, indeed. But given that we already have deprecated it a while ago (and people might already have updated their code), and that this name is also used in other places (eg R sf or JTS also have an "equal exact", PostGIS doesn't seem to have this), I would personally not change it anymore. 
Also the "exact" makes sense for the aspect that you need to have the exact same vertices (eg LineString(0 0, 22) and LineString(0 0, 1 1, 2 2) are spatially equal, but not "exact" equal). Of course for the fact that you an specify a tolerance for the coordinates itself, the "almost" is more appropriate.

> given that 2.0 will break a lot of things, would it not be the occasion to change the default behavior of equals_exact, or would you still rather have nomalize be False by default?

It's indeed the moment to change this, if we want to do so (personally, I think I would maybe still lean towards keeping `normalize=False` as the default, to keep it more in line with other implementations and the "exact" name).

@caspervdw @brendan-ward @sgillies @mwtoews any preferences for whether we should `normalize` by default or not? (currently we do not)

tfardet:
> sorry for letting this drag on for so long! And thanks for the update.

@jorisvandenbossche no worries, I know how easy it is to forget about a PR ;)


> One code suggestion: we have a similar `equals_exact` function in `shapely/predicates.py`. I think we should add the keyword there as well (and actually can pass through the keyword from `BaseGeometry/equals_exact`). This would be something like:

I did not know about this one!


> It's indeed the moment to change this, if we want to do so (personally, I think I would maybe still lean towards keeping `normalize=False` as the default, to keep it more in line with other implementations and the "exact" name).

I understand. I think the problem for me is that, with the deprecation of ``almost_equals``, ``equals_exact`` becomes the de facto replacement for it (unless I missed another function) and I therefore considered it from that perspective.
But I guess if it has historically been tied to coordinates comparison in a specific order, and you want to keep it that way, then it changes things.

**NB:** the problem is, the docstrings for the two implementations of ``equals_exact`` are completely different:

* the one in `base.py` reads simply "True if geometries are equal to within a specified tolerance." (hence my expectations and why I considered that `normalize` should be True by default)
* the one in `predicates.py` has a more explicit "This method uses exact coordinate equality".

So I guess we also should decide a) which of the two docstrings interpretation we want to go with and b) change them accordingly so that both are the same.

brendan-ward:
> whether we should normalize by default or not

I don't have a strong opinion on this, but in agreement with the reasons stated above, I think this should be opt-in (not normalize by default).

jorisvandenbossche:
> * the docstring of the [main class implementation](https://github.com/shapely/shapely/blob/9a7200dc46f0d0ec22935cb5819a5ed31aa70215/shapely/geometry/base.py#L655) of `equals_exact` was unclear about its reliance on precise coordinate order

I think it actually includes a note about this? 

> This method considers coordinate equality, which requires
> coordinates to be equal and in the same order for all components
> of a geometry.

The same is mentioned in the docstring of `almost_equals`, so it has always been intentional that this method relied on the coordinate order.

To be clear, the only difference between `equals_exact` and `almost_equals` in the past has always been the way how to specify the precision (an absolute tolerance in equals_exact, a number of decimals in almost_equals, where the second is under the hood translated to the first).

tfardet:
@jorisvandenbossche 

> I think it actually includes a note about this?

True, I did not expect something like that to be below the parameters.

Do you have an opinion about how we should proceed, then? I still feel that given the name of the functions, restoring ``almost_equals`` and having it call normalize would make more sense, but if you still want to remove it and add the normalize option to ``equals_exact`` that's fine with me.

jorisvandenbossche:
> Do you have an opinion about how we should proceed, then?

I think for this PR we can certainly move forward with adding the `normalize` keyword to `equals_exact`, with a default of False to preserve the current behaviour. I think having this keyword will be useful in any case, regardless of whether we want to reinstate `almost_equals` or not (which is based on `equals_exact`, and thus could then pass a different default value for this keyword).

---

For adding back `almost_equals`, I would like to see some more opinions of other contributors (and we can consider this for a separate PR, to move forward with this one). 
I am personally not super convinced of the value of doing adding it back. While it could be nice to have a method that normalizes by default, IMO it is still confusing to have those two slightly different methods (eg that it has a different way to specify the tolerance)., and even when it normalizes by default, I still find "almost" misleading, as it still requires exactly the same coordinates (only the order is relaxed by first normalizing).

snowman2:
> I still find "almost" misleading

I think that the same could be said for "exact" with the tolerance and normalization kwargs. This makes me wonder if we should just call the method `equals` for clarity?

snowman2:
> > I still find "almost" misleading
> 
> I think that the same could be said for "exact" with the tolerance and normalization kwargs. This makes me wonder if we should just call the method `equals` for clarity?

I spoke too soon, looks like it is taken. I wonder if it would make sense to give it a different name? "coordinates_equals"?

snowman2:
Moved almost equals to discussion to #1555.

jorisvandenbossche:
Thanks @snowman2 !

tfardet:
Should I also deprecate [``almost_equals``](https://shapely.readthedocs.io/en/stable/manual.html#object.almost_equals) in the manual and add ``equals_exact`` there?

Also, in ``predicates.py``, tolerance is optional (default: 0) whereas it is mandatory in the method... should I make it default to zero in the method too?

jorisvandenbossche:
> should I make it default to zero in the method too?

Yes, that's a good idea I think

tfardet:
@jorisvandenbossche I also updated the docstrings and manual, please check that I did not make any mistakes in the content

docs/manual.rst
```diff
@@ -1097,9 +1097,35 @@ differently.
   >>> b == c
   False
 
-.. method:: object.equals_exact(other, tolerance)
-
-  Returns ``True`` if the object is within a specified `tolerance`.
+.. method:: object.equals_exact(other, tolerance=0.0, normalize=False)
```
idanmiara, line 1100: I think that no additions to `manual.rst` are required since the docs will be generated from the docstrings of `shapely/predicates.py`

shapely/geometry/base.py
```diff
@@ -721,23 +721,22 @@ def dwithin(self, other, distance):
...
+        tolerance : float, optional (default: 0.)
             Absolute tolerance in the same units as coordinates.
-
-        This method considers coordinate equality, which requires
-        coordinates to be equal and in the same order for all components
-        of a geometry.
-
-        Because of this it is possible for "equals()" to be True for two
-        geometries and "equals_exact()" to be False.
+        normalize : bool, optional (default: False)
```
idanmiara, line 737: should be `normalize : bool default False`. optional means that it accepts `None` as an input, i.e. `Optional[bool]`.

shapely/predicates.py
```diff
@@ -947,25 +947,36 @@ def within(a, b, **kwargs):
...
+    and MultiPolygons, two topologically equal geometries can still be
+    structurally different even with `normalize` set to True if the order of the
+    rings or Polygons differs.
 
     Parameters
     ----------
     a, b : Geometry or array_like
-    tolerance : float or array_like
+    tolerance : float or array_like (default: 0.)
+    normalize : bool, optional (default: False)
```
idanmiara, line 970: same comment about `optional` as above

idanmiara:
Looks good to me 👍 Just a few minor remarks above.

shapely/geometry/base.py
snowman2: Related: https://github.com/shapely/shapely/issues/1701

idanmiara:
@tfardet have you seen the comments in the review?

tfardet:
@idanmiara yes, I was waiting for feedback from someone in the project so I don't do things that I need to undo afterwards, are you now a member of the Shapely project?

idanmiara:
> @idanmiara yes, I was waiting for feedback from someone in the project so I don't do things that I need to undo afterwards, are you now a member of the Shapely project?

Nope 🙄

idanmiara:
@jorisvandenbossche I've reviewed this PR, it looks good to me and almost ready. 
I just had a few minor suggestions, @tfardet is waiting for a review from a maintainer before addressing them.

idanmiara:
@jorisvandenbossche @brendan-ward 
Could you review please?

docs/manual.rst
```diff
@@ -1097,9 +1097,35 @@ differently.
...
+
+    Returns True if the geometries are structurally equal within a given
```
brendan-ward, line 1102:
```suggestion
    Returns True if the geometries are structurally equivalent within a given
```

docs/manual.rst
```diff
@@ -1097,9 +1097,35 @@ differently.
...
+    tolerance.
+
+    This method uses exact coordinate equality, which requires coordinates
+    to be equal (within specified tolerance) and in the same order for
+    all components of a geometry. This is in contrast with the
+    :meth:`~object.equals` function which uses spatial (topological) equality.
```
brendan-ward, line 1108:
```suggestion
    :meth:`~object.equals` function which uses spatial (topological) equality and does not require all components to be in the same order.
```

docs/manual.rst
```diff
@@ -1097,9 +1097,35 @@ differently.
...
+    For two given geometries, it is thus possible for :meth:`~object.equals` to
```
brendan-ward, line 1109:
```suggestion
    Because of this, it is thus possible for :meth:`~object.equals` to
```

docs/manual.rst
```diff
@@ -1097,9 +1097,35 @@ differently.
...
-
-  Returns ``True`` if the object is within a specified `tolerance`.
+.. method:: object.equals_exact(other, tolerance=0.0, normalize=False)
+
+    Returns True if the geometries are structurally equal within a given
+    tolerance.
+
+    This method uses exact coordinate equality, which requires coordinates
+    to be equal (within specified tolerance) and in the same order for
+    all components of a geometry. This is in contrast with the
```
brendan-ward, line 1107:
```suggestion
    all components (vertices, rings, or parts) of a geometry. This is in contrast with the
```

shapely/geometry/base.py
```diff
@@ -721,23 +721,22 @@ def dwithin(self, other, distance):
         """
         return _maybe_unpack(shapely.dwithin(self, other, distance))
 
-    def equals_exact(self, other, tolerance):
-        """True if geometries are equal to within a specified
-        tolerance.
+    def equals_exact(self, other, tolerance=0.0, normalize=False):
+        """Returns True if the objects have the same sequence of coordinates
```
brendan-ward, line 725: Should match the first part of the docstring for `equals_exact()` function.

brendan-ward (commented):
Thanks for this @tfardet ; added a few text suggestions.

I like that `normalize` is `False` here; I still think that should be opt-in.

Longer term, I'm wondering if for GEOS >= 3.12, using a tolerance of 0 here should use `GEOSEqualsIdentical` instead of `GEOSEqualsExact`, but that is separate from this PR.

shapely/geometry/base.py
tfardet: sorry, I did not get that, could you be more explicit?

idanmiara: I think that he meant that the docstring for `equals_exact` in `base.py` should be the same as the first part (i.e. first few lines) of `equals_exact` in `predicates.py` (which holds the full doc of this function)

tfardet: Ah, right!
If it's fine with you an until @idanmiara deals with the duplication issue in the doc, I'd propose to copy the text from manual.rst to both ``base.py`` and ``predicates.py`` if that's fine with both of you.

shapely/geometry/base.py
idanmiara: > > optional means that it accepts None as an input, i.e. Optional[bool].
> 
> I'm not sure this is true? According to [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#parameters), `optional` in the docstring like this means that there are default values, which can be of the same type.

I see your reference in numpydoc, but I find it very confusing. Why would you use the keyword `optional` which has a specific meaning in python when you actually mean `default`.  I also can't find such use of the keyword `optional` in shapely docs.

shapely/geometry/base.py
idanmiara: @tfardet there is no need to duplicate the docstring in `manual.rst`.
The manual will also contain the docs for this function as it pulls is from `predicates.py`, see https://shapely.readthedocs.io/en/stable/predicates.html for some examples.

A single copy of the full docstring should be where the function is implemented, which is in `predicates.py`.
the convenience wrapper in `base.py` should only contain the first few lines (i.e. the gist) of the full docstring (from `predicates.py`)

tfardet: In my experience, the manual is the most consulted page, so until it pulls automatically from predicates, I'm not willing to reduce the docstring there ;)

idanmiara: > In my experience, the manual is the most consulted page, so until it pulls automatically from predicates, I'm not willing to reduce the docstring there ;)

Please note that that both `manual` and `predicates` are pages of the main shapely docs (see the links on the left):
https://shapely.readthedocs.io/
I'm thinking that since `equals_exact` already is mentioned in the `manual` page as a gist, maybe update the it in the same way as in `base`. 
I think in general we should avoid duplicating docs also because it makes it harder for keeping them in sync when editing, so we only maintain a single copy of the full doc and copy short gists where appropriate.
@brendan-ward what do you think ?

tfardet: IMO the right solution for this is to use

```rst
.. autosummary::

    shapely.predicates.equals_exact
```

or that of ``BaseGeometry.equals_exact`` in ``manual.rst``.

Once that is decided, we can put the main docstring in only one of the two and reuse it directly in manual.rst.
I thought that was what you meant when you said you would work on cleaning up the doc.

shapely/geometry/base.py
tfardet: the use of ``optional`` in docstrings predates annotations and historically was used to indicate that the argument is optional in the sense that you do not need to provide it because it has a default value; this is the meaning that users understand by default; only developers may link it to the ``Optional`` annotation.

shapely/geometry/base.py
brendan-ward: Sorry for the confusion!  Yes, what I meant was that the authoritative and complete docstring should be on the `equals_exact` function, and just a snippet of it in the `equals_exact` method.  Unfortunately I added a lot of my text suggestions to the entry in the manual not the docstring! (wasn't paying close enough attention to location, only content)  Sorry, my intent was that it would apply to the docstring instead.

For now, having that duplicated into the manual is OK; it would be fine to resolve that in a follow-up PR.  But if using 

```
.. autosummary::

    shapely.predicates.equals_exact
```

is a drop-in change, that is also fine to do here (using function not method).

tfardet: No pb!
Regarding the autosummary, I think that needs to be done properly an consistently throughout the manual, once and for all, with a proper discussion, so I'll leave that for a later PR. For now let's stick with the duplicated entry in the manual ;)

shapely/geometry/base.py
brendan-ward: `optional` is used in many places in shapely docstrings; e.g., [points constructor](https://shapely.readthedocs.io/en/stable/reference/shapely.points.html#shapely.points) for exactly the reasons provided by @tfardet ; at least some of these came via PyGEOS.  `default: <value>` is rarely used in the docstrings; I believe when this came up in PyGEOS the preference there was that the default `kwarg` parameter values was sufficient documentation because that was easily visible in the function definition shown in the docs (i.e., using `default: <value>` was viewed as redundant).

tfardet:
OK, hopefully I integrated all changes.
If it fine, please let me know what's the procedure for a first merge (do you want me to squash the commits, sign off the commit, whatever else?)

brendan-ward:
> please let me know what's the procedure for a first merge

No need for you to squash (in fact, it makes review easier if you don't), we do that automatically when merging in Github.  Once reviewers have had a chance to review, they mark it as approved, and one of the maintainers will merge it.

tfardet:
> No need for you to squash (in fact, it makes review easier if you don't), we do that automatically when merging in Github. Once reviewers have had a chance to review, they mark it as approved, and one of the maintainers will merge it.

OK, great!
I think everything's good then: failing tests seem unrelated.

docs/manual.rst
```diff
@@ -1098,9 +1098,37 @@ differently.
...
+    Returns True if the geometries are structurally equivalent within a given
+    tolerance.
+
+    This method uses exact coordinate equality, which requires coordinates
+    to be equal (within specified tolerance) and in the same order for
+    all components (vertices, rings, or parts) of a geometry. This is in
+    contrast with the :meth:`~object.equals` function which uses spatial
+    (topological) equality and does not require all components to be in the
+    same order.
+    Because of this, it is thus possible for :meth:`~object.equals` to be True
```
brendan-ward, line 1112:
```suggestion
    Because of this, it is possible for :meth:`~object.equals` to be ``True``
```

Sorry I didn't catch this sooner, but inline code like True needs to have double-backticks in RST.  This isn't consistent throughout Shapely yet, but it looks like these should be double-backticks in the docstrings as well since they get pulled into the docs (or will in the future, for this one).

Also, this can be part of the prior paragraph ("This method uses...") instead of a new paragraph.

docs/manual.rst
```diff
@@ -1098,9 +1098,37 @@ differently.
...
+    while :meth:`~object.equals_exact` is False.
+    The order of the coordinates can be normalized (by setting the
```
brendan-ward, line 1114: Please add new line separating paragraphs

shapely/predicates.py
```diff
@@ -947,25 +947,39 @@ def within(a, b, **kwargs):
...
+        If True, normalize the two geometries so that the coordinates are
+        in the same order.
     **kwargs
         For other keyword-only arguments, see the
         `NumPy ufunc docs <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs>`_.
 
     See Also
     --------
-    equals : Check if A and B are spatially equal.
+    equals : Check if A and B are topologically equal.
```
brendan-ward, line 982: Not sure this change is strictly necessary, and possibly confusing.  The description used by GEOS is
`True if geometries cover the same space on the plane`, which to me leans more heavily on the spatial rather than topological side.  `spatially (topologically)`, which is similar to the usage in the docstring above, is probably OK.

brendan-ward (commented):
Thanks for the updates @tfardet !  Just caught a couple minor text edits, but otherwise this is looking good to me.

docs/manual.rst
```diff
@@ -1098,9 +1098,37 @@ differently.
...
+
+    This method uses exact coordinate equality, which requires coordinates
+    to be equal (within specified tolerance) and in the same order for
+    all components (vertices, rings, or parts) of a geometry. This is in
+    contrast with the :meth:`~object.equals` function which uses spatial
+    (topological) equality and does not require all components to be in the
+    same order. Because of this, it is possible for :meth:`~object.equals` to
+    be ``True`` while :meth:`~object.equals_exact` is ``False``.
+
+    The order of the coordinates can be normalized (by setting the `normalize`
```
brendan-ward, line 1114:
```suggestion
    The order of the coordinates can be normalized (by setting the ``normalize``
```

docs/manual.rst
```diff
@@ -1098,9 +1098,37 @@ differently.
...
+    keyword to True) so that this function will return `True` when geometries
```
brendan-ward, line 1115:
```suggestion
    keyword to ``True``) so that this function will return ``True`` when geometries
```

docs/manual.rst
```diff
@@ -1098,9 +1098,37 @@ differently.
...
+    are structurally equivalent but differ only in the ordering of vertices.
+    However, this function will still return False if the order of interior
```
brendan-ward, line 1117:
```suggestion
    However, this function will still return ``False`` if the order of interior
```

shapely/predicates.py
```diff
@@ -947,25 +947,39 @@ def within(a, b, **kwargs):
...
-    components of a geometry. This is in contrast with the ``equals`` function
-    which uses spatial (topological) equality.
+    to be equal (within specified tolerance) and in the same order for
+    all components (vertices, rings, or parts) of a geometry. This is in
+    contrast with the :meth:`~object.equals` function which uses spatial
+    (topological) equality and does not require all components to be in the
+    same order. Because of this, it is possible for :meth:`~object.equals` to
+    be ``True`` while :meth:`~object.equals_exact` is ``False``.
+
+    The order of the coordinates can be normalized (by setting the `normalize`
```
brendan-ward, line 962:
```suggestion
    The order of the coordinates can be normalized (by setting the ``normalize``
```

shapely/predicates.py
```diff
@@ -947,25 +947,39 @@ def within(a, b, **kwargs):
...
+    keyword to True) so that this function will return `True` when geometries
```
brendan-ward, line 963:
```suggestion
    keyword to ``True``) so that this function will return ``True`` when geometries
```

shapely/predicates.py
```diff
@@ -947,25 +947,39 @@ def within(a, b, **kwargs):
...
+    are structurally equivalent but differ only in the ordering of vertices.
+    However, this function will still return False if the order of interior
```
brendan-ward, line 965:
```suggestion
    However, this function will still return ``False`` if the order of interior
```

brendan-ward (commented):
Very sorry to drag this out @tfardet , just found a few more instances where double-ticks are needed.  Unfortunately, it looks like we aren't building docs automatically as part of PRs, so there isn't an easy way to proof this via GH actions.

After installing the required dependencies, it is possible to build docs locally by running `make html` in the `docs` directory, then opening `_build/html/index.html` in a browser (this is what I did to check these).

docs/manual.rst
tfardet: If shapely follows numpy doc recommendations (which I think it is supposed to), then function arguments should be referenced with a single `

brendan-ward: @jorisvandenbossche - perhaps you can answer this one?

brendan-ward: Either way, I still wanted review from @jorisvandenbossche before this gets merged.  Sorry for the delay in proceeding with this.

idanmiara: @brendan-ward maybe someone else could approve this ? It's a bit sad that this PR stalls for a such a minor issue ...

jorisvandenbossche (approved):
Looks good to me!

I just added two tiny changes:

- in the docstring of the function, change `` :meth:`~object.equals` `` to `` :func:`equals` `` so it points to the reference page, and not to the manual
- added a small test for the top-level function as well (just verifying that the normalize keyword is used here as well), the other tests only did that for the method (which indirectly uses the function, but it's good to have an explicit test as well)

I checked the doc build locally, and it looks good

jorisvandenbossche:
The failures on the builds using GEOS main are unrelated (but something we have to investigate ..), so merging.

@tfardet thanks again, for the contribution and the perseverance (apologies for the far from ideal response time ..) !

jorisvandenbossche:
To follow-up on two of the documentation style related discussions above:

- How do specify optional keywords with a default? (https://github.com/shapely/shapely/pull/1231/files#r1128291423)
  - The numpydoc guide indeed indicates "optional" can be used if you don't have to specify anything, but in practice for keywords that do have a default value, I think we always use the "type, default .." style (so writing this now, we should have done that here as well). Related issue: https://github.com/shapely/shapely/issues/1701
- How to reference keyword names: single or double backticks? (https://github.com/shapely/shapely/pull/1231/files#r1129821468)
  -  The numpydoc guide indeed uses single backticks. However, how this gets rendered in the html version actually depends on your sphinx settings (single backticks can be given a default meaning). I don't think we have been strictly following this one way or the other (although maybe rather tend to use double backticks)

PR #1235, opened 2021-11-23
Fix the array_interface_base property

maximlt:
Same as https://github.com/Toblerity/Shapely/pull/1214 but based on the right branch.

Just for your information `geoviews` was using this property in a couple of internal utils and in its test suite.


maximlt:
Ah, someone will need to approve running the CI since its my first contribution!

sgillies (approved):
Thanks @maximlt !

PR #1237, opened 2021-12-01
Minimal working shapely package with pygeos code included

jorisvandenbossche:
After merging in the https://github.com/pygeos/pygeos repo into Shapely, this are some minimal changes to get a the shapely package in a "working state":

- First commit: move `/pygeos/*` directory into `/shapely/*`. The only conflict is `strtree.py`, so the pygeos version is temporarily named `strtree_pygeos.py` (and in a follow-up we need to merge both strtree files)
- Second commit: replacing "pygeos" with "shapely" in the package source and tests (not yet in docs, CI, etc, also not yet in docstring examples (for doctests))
- Third commit: minimal changes to get a working package:
  - minimal setup.py (I kept the old one as `setup_shapely.py` for now, we can in a follow-up consolidate the setup files (the exact setup arguments, author, maintainer, description, etc)
  - renamed `geometry.py` to `_geometry.py`, because it otherwise conflicts with `geometry/__init__.py` submodule (and therefore also renamed `_geometry.pyx` to `_geometry_helpers.pyx`). If there are other suggestions for names, that's certainly welcome.

With those changes, building and installing the new shapely package works (`python setup.py build_ext -i` / `python setup.py develop` / `pip install -e . --no-build-isolation`), and the package is in a working state: it imports, most functionality works, the tests run (but there are still a bunch of failures)

(it's probably easiest to check the commits separately)

cc @sgillies @caspervdw


caspervdw:
Looks good @jorisvandenbossche !

jorisvandenbossche:
Since this blocks other PRs, and is mostly some moving things around and renaming, going to merge this.

PR #1239, opened 2021-12-01
BUG: fixup the hash-ability of geometries

jorisvandenbossche:
On top of #1237, so only the last commit is relevant.

The tests for hash-ability were still failing, for two reasons:

- There was still a `__hash__ = None` in the code, effectively making a class unhashable
- We defined a `__eq__` in the python-level BaseGeometry class, and when you subclass, you need to implement neither or both of `__eq__` and `__hash__` (see https://docs.python.org/3/reference/datamodel.html#object.__hash__)
  - I now removed the `__eq__` in the python layer, keeping the `GeometryObject_richcompare` in the C extension type in `pygeos.c` (but in principle we could also do it the other way around I think)

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -165,17 +165,6 @@ class BaseGeometry(shapely.Geometry):
     def __xor__(self, other):
         return self.symmetric_difference(other)
 
-    def __eq__(self, other):
-        return (
-            type(other) == type(self) and
-            bool(shapely.equals_exact(self, other))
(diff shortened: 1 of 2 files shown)


caspervdw (approved):

jorisvandenbossche:
Merging this, so I can try to get the tests passing in https://github.com/Toblerity/Shapely/pull/1241 (we can later still revisit the exact approach of doing it in C vs python)

PR #1240, opened 2021-12-01
Minimal pygeos->shapely renaming in C code to get cython API working

jorisvandenbossche:
I missed one aspect in https://github.com/Toblerity/Shapely/pull/1237 to get a basic working package (due to testing it in an environment in which I still had pygeos installed, so the cython code was picking up the pygeos package instead ..).

I didn't yet do a full rename of all PyGEOS_.. functions.

cc @caspervdw

### DIFF
--- a/src/c_api.h
+++ b/src/c_api.h
@@ -42,7 +42,7 @@
 #define PyGEOS_API_num_pointers 3
 
 #ifdef PyGEOS_API_Module
-/* This section is used when compiling pygeos.lib C extension.
+/* This section is used when compiling shapely.lib C extension.
  * Each API function needs to provide a corresponding *_PROTO here.
  */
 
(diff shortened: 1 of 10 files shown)


src/c_api.h
```diff
@@ -75,9 +75,9 @@ static void **PyGEOS_API;
  * PyCapsule_Import will set an exception on error.
  */
 static int
-import_pygeos_c_api(void)
+import_shapely_c_api(void)
 {
-    PyGEOS_API = (void **)PyCapsule_Import("pygeos.lib._C_API", 0);
+    PyGEOS_API = (void **)PyCapsule_Import("shapely.lib._C_API", 0);
```
jorisvandenbossche, line 80: This was the crucial change to have our cython code actually pick up shapely instead of pygeos if it happens to be co-installed

caspervdw (approved):

PR #1241, opened 2021-12-01
CI: consolidate the build scripts and github actions tests workflow

jorisvandenbossche:
xref https://github.com/Toblerity/Shapely/issues/962

This PR consolidates the github actions workflow for tests + the build scripts required for that, and gets the tests "passing":

- I basically used the pygeos workflow and build scripts (with some minor adjustments for shapely), but kept the name of the shapely files. So as a result the diff actually shows the difference with the original shapely files. 
- Also removed appveyor file because this is now covered by github actions
- I still added some skips, and temporarily disabled warning checking and doctests (all are marked with a `TODO(shapely-2.0)` comment), to get the tests CI green here (we can fix those in targeted follow-ups, but for those follow-ups a green CI will be useful)


.github/workflows/tests.yml
```diff
@@ -24,91 +30,134 @@ jobs:
...
 
       - name: Install python dependencies
-        shell: bash
         run: |
-          pip install --disable-pip-version-check --upgrade pip
-          pip install --upgrade wheel
-          pip install git+https://github.com/pygeos/pygeos.git -v
+          python -m pip install --disable-pip-version-check --upgrade pip
+          pip install --upgrade wheel setuptools
           if [ "$GEOS_VERSION" = "main" ]; then
```
caspervdw, line 114: The condition `[ "$GEOS_VERSION" = "main" ]` coincidentally aligns with whether or not matrix.numpy is specified, but it might not be like that in the future.

```suggestion
          if [ -n "${{ matrix.numpy }}" ]; then
```

.github/workflows/tests.yml
```diff
@@ -24,91 +30,134 @@ jobs:
...
-          pip list
+        run: pip list
 
       - name: Run tests
-        shell: bash
         continue-on-error: ${{ matrix.geos == 'main' }}
         run: |
-          python -c "from shapely import geos; print(geos.geos_version_string)"
-          python -m pytest --cov shapely --cov-report term-missing
+          pytest shapely/tests -cov shapely --cov-report term-missing ${{ matrix.extra_pytest_args }}
```
caspervdw, line 154:
```suggestion
          pytest tests shapely/tests -cov shapely --cov-report term-missing ${{ matrix.extra_pytest_args }}
```
And drop the next line.

tests/test_linestring.py
```diff
@@ -111,6 +112,8 @@ def test_numpy_object_array():
     assert ar[0] == geom
 
 
+# TODO(shapely-2.0)
+@pytest.mark.xfail(strict=False, reason="Not yet implemented for Shapely 2.0")
```
caspervdw, line 116: Maybe use `shapely20_todo` here?

caspervdw (approved):
Some minor nitpicks, for the rest there is one test failing on python 3.10, something with a deprecation warning on setting attributes. OK if you want to postpone solving that. Maybe good to decorate those with `shapely20_todo` so that we can start working with a green CI.

tests/test_linestring.py
jorisvandenbossche: Yeah, I initially did, but that one is currently a "strict" xfail, and this test also seemed to pass in some builds (it might depend on the GEOS version, in any case something to verify and clean-up in a follow-up). 
I could also make the general marker less strict, but the strict version is also nice because then you get notified if something starts passing.

shapely/geometry/polygon.py
```diff
@@ -263,27 +263,6 @@ def interiors(self):
             return []
         return InteriorRingSequence(self)
 
-    def __eq__(self, other):
```
jorisvandenbossche, line 266: This was an oversight in https://github.com/Toblerity/Shapely/pull/1239 where I removed the other `__eq__` methods

.github/workflows/tests.yml
```diff
@@ -24,91 +30,136 @@ jobs:
...
-          pip list
+        run: pip list
 
       - name: Run tests
-        shell: bash
         continue-on-error: ${{ matrix.geos == 'main' }}
         run: |
-          python -c "from shapely import geos; print(geos.geos_version_string)"
-          python -m pytest --cov shapely --cov-report term-missing
+          pytest shapely/tests tests -cov shapely --cov-report term-missing ${{ matrix.extra_pytest_args }}
```
caspervdw, line 155:
```suggestion
          pytest shapely/tests tests --cov shapely --cov-report term-missing ${{ matrix.extra_pytest_args }}
```

maybe you can also include the coverage stuff in the “extra_pytest_args” so that the coverage report is only generated on a specific run?

jorisvandenbossche: Good catch, thanks :)

> maybe you can also include the coverage stuff in the “extra_pytest_args” so that the coverage report is only generated on a specific run?

I think it can actually be useful to run coverage on multiple builds (and the coverage report should then get combined), to run cover code that is version specific as well (although most of the GEOS-version specific code lives in C, and that's not covered by this anyway)

caspervdw:
There is one issue with coverage also reporting test coverage on the tests in shapely/tests themselves.

According to the [docs](https://coverage.readthedocs.io/en/latest/source.html#specifying-source-files) , pytest-cov needs this in the `.coveragerc`:

`[run]
source = shapely
omit = shapely/tests/* `

caspervdw:
Btw I am mostly not using coverage in my projects, because I think it gives a false sense of confidence. So I am actually 👍 on dropping it all together

jorisvandenbossche:
Yeah, I personally also don't use coverage that much (it's mostly because it was present, that I kept it). I think as long as you are aware of the limitations and don't use it as indication that "all is good", it can have some use (if something is _not_ covered, that's still useful information)

jorisvandenbossche:
Test workflows are all passing here, so going to merge this so have proper CI running again.

caspervdw:
Good! Are you planning to look into the wheelbuild workflows? I could do that maybe in the coming days.

jorisvandenbossche:
> Are you planning to look into the wheelbuild workflows?

Didn't plan yet, so feel free to do so! (I think it should be only a minor edit, for example at least updating the test invocation to run shapely tests, but maybe not much more, since the GEOS build scripts are already updated in this PR)

PR #1242, opened 2021-12-01
Linting: blacken shapely code

jorisvandenbossche:
From https://github.com/Toblerity/Shapely/issues/962#issuecomment-961071354, one of the discussion items is the use of linting tools (black, flake8, isort), which pygeos was already using (and enforcing on CI using pre-commit). 
If we want to use them for shapely as well, we of course need to make the existing code apply, and I think black will have the biggest impact. So this PR shows what it would do for the existing shapely code (I see the largest part of the diff is actually string normalization (`'` -> `"`), which could be turned off if people prefer).


sgillies:
Running black on the code would be fine with me. I've been using darker to incrementally blacken lines of code that I've changed for a couple years. About the quote changes: at least they are easy to see and pass over in a visual diff!

mwtoews (approved):
Black as a pre-commit check is fine with me. The default configuration is also fine too, including line length and quote options.

There are a few missing subdirectories, including tests. Was that intentional, or to be done later?

jorisvandenbossche:
Whether to use black or darker (https://github.com/akaihola/darker) doesn't matter that much to me (as long as we automate the fixing up and checking). The original author of black has a small blogpost explaining why _he_ prefers to do apply black once instead of incrementally: https://lukasz.langa.pl/36380f86-6d28-4a55-962e-91c2c959db7a/

Now, it seems people are generally fine with it. So will go forward with black here.

> There are a few missing subdirectories, including tests. Was that intentional, or to be done later?

Indeed, the tests I also need to do (I left out the `examples` subdir with the idea we should maybe move them outside of the package anyway?)

PR #1245, opened 2021-12-02
Pickling: fix SRID handling + update linearring test

### LINKED ISSUE #988: Pickling of LinearRings
While implementing pickling for the extension type in PyGEOS (https://github.com/pygeos/pygeos/pull/190), we are having a discussion about how faithful the roundtrip of pickle should be. And specifically for LinearRings, because that geometry type is not supported in WKB, which is the serialization format being used for pickling.

Currently, Shapely preserves LinearRings itself, but LinearRings that are part of GeometryCollections become LineStrings after pickle roundtrip:

```
In [1]: from shapely.geometry import LinearRing, GeometryCollection

In [2]: import pickle   

In [3]: ring = LinearRing([(0,0), (1,1), (0,1), (0,0)])  

In [4]: col = GeometryCollection([ring])  

In [5]: print(ring)
LINEARRING (0 0, 1 1, 0 1, 0 0)

In [6]: print(pickle.loads(pickle.dumps(ring)))
LINEARRING (0 0, 1 1, 0 1, 0 0)

In [7]: print(col)  
GEOMETRYCOLLECTION (LINEARRING (0 0, 1 1, 0 1, 0 0))

In [8]: print(pickle.loads(pickle.dumps(col)))
GEOMETRYCOLLECTION (LINESTRING (0 0, 1 1, 0 1, 0 0))
```

The reason for this is that a WKB roundtrip cannot preserve LinearRings, as they get represented as linestrings:

```
In [21]: from shapely import wkb  

In [22]: print(wkb.loads(wkb.dumps(ring)))  
LINESTRING (0 0, 1 1, 0 1, 0 0)
```

For actual LinearRings objects, Shapely overrides this by having a custom pickling function specifically for this class (which is not possible this way for GeometryCollection):

https://github.com/Toblerity/Shapely/blob/9a54c7eb4946b8eb420091bd53d8d3930b61ea5d/shapely/geometry/polygon.py#L80-L87

So the questions we are having:

- Would we be fine with documenting this as a "known limitation" of pickling that it doesn't preserve LinearRings? Or does Shapely want to keep this roundtrip faithful?
- In any case, I don't think we want to fix the LinearRing-in-GeometryCollection case?

The main motivation for this issue is that "just using WKB" without having to keep track of the original geometry type in some way is easier to implement the reducer on the extension type in C. 

(Now I am writing this, I realize we could also implement the `__reduce__` on the C level for the base class using just WKB, and if we want faithful roundtrip for LinearRings, we can still *override* this base `__reduce__` in Python *only* for LinearRings (as is actually done now as well). That will mean that pickling an array of LinearRings will be slower than if only implemented in C, but since that is a special corner case anyway, that might be worth it for preserving the correct roundtrip)

cc @caspervdw @brendan-ward

jorisvandenbossche:
The goal here is to consolidate the pickling behaviour between pygeos and shapely. The differences:

- Shapely preserved LinearRing in the roundtrip, pygeos not (some discussion about this in https://github.com/Toblerity/Shapely/issues/988)
- Pygeos preserved the SRID in roundtrip, shapely not

This PR combines both features, preserving both LinearRing type and SRID.

I currently did this in the Python layer, while we also have a base ``__reduce__`` implementation in C (which is probably unused now in practice): 

https://github.com/Toblerity/Shapely/blob/2e513b143c1f0a84b9cd23cbe409f72bfa78f40f/src/pygeom.c#L408

I suppose we could also keep the logic in C. The main reason that I currently need to override `__reduce__` in the python subclasses is because the C version ends up passing the WKB to the class `__new__`. The base `lib.Geometry` class accepts WKB for construction, but the subclasses currently do not (eg `shapely.geometry.Point(..)` only accepts actual coordinate values).

Closes #988


caspervdw (approved):

caspervdw:
In pygeos, we didn't have the `Geometry` subclasses, so things like pickling were implemented in C. I am really :+1: on keeping that kind of per-object logic on the Python side.

jorisvandenbossche:
I _suppose_ I can then actually remove the C version (`GeometryObject_reduce`)

sgillies:
Looks good. I still don't like the idea of SRID support. Not supporting SRID up to this point has eliminated a source of complexity and thus has been good for Shapely. And the GEOS project doesn't consider SRID to be a useful feature, either.

jorisvandenbossche:
As long as we limit the SRID support strictly to IO (i.e. pickling and read/write WKB), I don't think it will really add a source of complexitly / issues. If someone opens an issue related to SRID for anything else, we can simply close as out-of-scope / defer to GEOS. 

I personally don't care that much about preserving SRID for pickling, as long as we support it for WKB (which is needed for example for PostGIS interaction)

caspervdw:
Personally I also think that srids are out of scope. However if users want to carry this information (any information really) along with the geometries, then I think it is fine to support that. As long as we never base any logic on it.

sgillies:
I looked in geopandas and it seems that it is only depending on a geometry's SRID due to a design choice and not because there is no other way. Instead of checking to see if an SRID is set on the first geometry at https://github.com/geopandas/geopandas/blob/d006fad416a9f33fd88c1a00744ae9f98cb9c6d0/geopandas/io/sql.py#L88 (if there was a bug here it would be super confusing), this function could instead call https://postgis.net/docs/Find_SRID.html via `pd.read_sql()` or some lower level function.

jorisvandenbossche:
We still use SRID for WKB export to PostGIS as well (that doesn't need to be _on the geometry_ though, but as mentioned above still need it in the WKB IO functions) 
(this relates to https://github.com/pygeos/pygeos/issues/137)

jorisvandenbossche:
For the reading side, it is true that we don't need the `get_srid(geom)` for PostGIS, as we could indeed use one of the PostGIS function as well. However, I personally think we should still support reading the SRID from a WKB object. You can have other sources of WKB than PostGIS, and I think generally being able to extract all information contained in the WKB is a good thing to support.

jorisvandenbossche:
Since this PR is only fixing the LinearRing pickling and making it consistent with the other geometry classes, I am planning to merge this. 
We can continue the discussion about to what extent we want to expose SRID related functionality in general in another issue.

PR #1249, opened 2021-12-04
Update MANIFEST.in to enable 'python -m build` to work

mwtoews:
A few edits to `MANIFEST.in` were necessary to get `python -m build` to work as expected.

This should probably be integrated into the CI at some point.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,9 @@ on:
     paths:
       - ".github/workflows/release.yml"
       - "ci/*"
+      - "MANIFEST.in"
+      - "pyproject.toml"
+      - "setup.py"
 
 jobs:
   build_sdist:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,9 +30,8 @@ jobs:
 
       - name: Build a source tarball
         run: |
-          python -m pip install --upgrade pip
-          pip install setuptools
-          python setup.py sdist
+          python -m pip install --upgrade pip build
+          python -m build --sdist
 
       - uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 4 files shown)


sgillies (approved):

jorisvandenbossche:
@mwtoews could you remove `MANIFEST_pygeos.in` then? (or check if that this has other things to add)

mwtoews:
thanks @jorisvandenbossche, this merge has been reviewed. The significant include was for pyproject.toml. The Cython-generated C files don't need to be excluded from `shapely/`, as they already seem to be ignored.

So at some point, a good CI check would be to:
```
python setup.py sdist
# figure out what dist/*.tar.gz to use
tar xf $SDIST_TGZ
cd $SDIST
python -m build
```

jorisvandenbossche:
>  The Cython-generated C files don't need to be excluded from shapely/, as they already seem to be ignored.

From a quick test locally, for me they are included. 
Maybe the difference is whether you started from a cleaned repo or not? It's always good to first do a `git clean -xdf` when running packaging stuff, but explicitly ignoring the shapely/*.c files might still be useful?

mwtoews:
Strange, as I have several `shapely/*.c` files, but these don't get included. There must be some implicit magic happening somewhere. I've added further commit to explicitly exclude these, and `python setup.py sdist` has a harmless warning:
> warning: no previously-included files matching '*.c' found under directory 'shapely'

jorisvandenbossche:
> This should probably be integrated into the CI at some point.

With the copied release.yml workflow from pygeos, we actually do have that now. I will push a small update so that this gets triggered here.

jorisvandenbossche:
The sdist artifact can be downloaded from https://github.com/Toblerity/Shapely/actions/runs/1540852171

mwtoews:
Thanks @jorisvandenbossche, the additions look good.

PR #1250, opened 2021-12-04
BUG/TST: fix hashing for Polygon + update tests

mwtoews:
Supersedes #985

The hash implementation is better than the previous one that existed in shapely several versions ago, as it is now based on WKB.

Now geometries can be used as dict keys to do things like this:
```
>>> d = {}
>>> d[Point(1, 2)] = "X"
>>> d[Point(1.0, 2.0)]
'X'
```

### DIFF
--- a/shapely/geometry/polygon.py
+++ b/shapely/geometry/polygon.py
@@ -263,27 +263,6 @@ class Polygon(BaseGeometry):
             return []
         return InteriorRingSequence(self)
 
-    def __eq__(self, other):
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
Yes, so I actually already did this in https://github.com/Toblerity/Shapely/pull/1239, but missed the one `__eq__` for Polygon. Thanks!

jorisvandenbossche (approved):

PR #1253, opened 2021-12-09
CI: Fix the release workflows (GHA and travis CI)

caspervdw:
The goal is here to make the wheel build and travis CIs pass.

Some open ends:
- The tests are still outside of the package, which makes running them after building a wheel more complex. I prefer moving them in a separate folder. Maybe we should have `tests_shapely` and `tests_pygeos` submodules for now?
- Travis seems to be out of credits. @sgillies  Do you know if we are entitled to OSS credits?
- We need a PyPI token to be generated, [encrypted](https://docs.travis-ci.com/user/encryption-keys#usage) and configured in the `.travis.yml`
- We need another PyPI token set up as `pypi_password` "secret" in Github Actions (see [docs](https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository))
- Appveyor needs to be switched off.

BTW, I don't have manage privileges on this repo, is that correct @sgillies? I'd like to edit the ticket in #962 to reference this PR.


jorisvandenbossche:
Regarding the secrets and manage access, I would maybe directly / first move to the shapely org, and fix it there (otherwise we will need to do it twice), see https://github.com/Toblerity/Shapely/issues/1204

jorisvandenbossche:
> The tests are still outside of the package, which makes running them after building a wheel more complex. I prefer moving them in a separate folder. Maybe we should have tests_shapely and tests_pygeos submodules for now?

My plan was to progressively move the tests from /tests into /shapely/tests (eg https://github.com/shapely/shapely/pull/1257 as a start). Personally I wouldn't care too much about the /tests here for now and just run /shapely/tests in the packaging CI (I suppose we won't make an actual release before the tests are actually moved. Or otherwise we can already move them in bulk into a subdirectory of /shapely/tests (I would avoid creating tests_pygeos / tests_shapely, as that would unnecessarily move once more the tests that are already in /shapely/tests)

jorisvandenbossche:
@caspervdw I think you should have the privileges now to continue this PR?

caspervdw:
Yes that’s ok, only Travis is still showing the message “ Builds have been temporarily disabled for public repositories due to a negative credit balance”

I am not sure how to proceed here. Did the negative credit balance get carried along with the organisation change of tobleriy to here? @sgillies do you have an idea on how to get Travis running again? Maybe we can reach out to Travis?

jorisvandenbossche:
> Did the negative credit balance get carried along with the organisation change of tobleriy to here?

This organization was already used for the shapely-wheels repo, so I suppose the credits have been used there

jorisvandenbossche:
I emailed Travis' support for requesting OSS credits.

In the meantime, I think we can already get this PR merged?

sgillies (approved):

caspervdw:
Sure we can get this merged right away, however the PyPI tokens still need to be configured.

@sgillies  Could you generate two tokens for the shapely repo? Or grant me rights (username: caspervdw) at PyPI to do so myself?

jorisvandenbossche:
BTW, I got reply from Travis and we should now have free credits (at least for some time ..) 
I see already some things running at https://app.travis-ci.com/github/shapely/shapely. But in order to limit our use of credits, we should maybe disable Travis by default, and only have it run on tags or on specific demand? (to test before a release) But not at every commit on master, or certainly not at every commit on PRs

jorisvandenbossche:
But so it was running of this PR, so that is useful: https://app.travis-ci.com/github/shapely/shapely/jobs/561140861

jorisvandenbossche:
There seems to be one test failure on Python 3.8 for the aarch64 builds:

```
=================================== FAILURES ===================================
____________________________ test_from_invalid_dim _____________________________
    def test_from_invalid_dim():
        # TODO(shapely-2.0) better error message?
        # pytest.raises(ValueError, match="at least 2 coordinate tuples|at least 2 coordinates"):
        with pytest.raises(shapely.GEOSException):
            LineString([(1, 2)])
    
        with pytest.raises(
            ValueError, match="Input operand 0 does not have enough dimensions"
        ):
>           LineString([(1, 2, 3), (4, 5)])
/tmp/tmp.HukXYlwDK9/venv/lib/python3.8/site-packages/shapely/tests/geometry/test_linestring.py:112: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/tmp/tmp.HukXYlwDK9/venv/lib/python3.8/site-packages/shapely/geometry/linestring.py:65: in __new__
    geom = shapely.linestrings(coordinates)
/tmp/tmp.HukXYlwDK9/venv/lib/python3.8/site-packages/shapely/decorators.py:80: in wrapped
    return func(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
coords = [(1, 2, 3), (4, 5)], y = None, z = None, indices = None, out = None
kwargs = {}
    @multithreading_enabled
    def linestrings(coords, y=None, z=None, indices=None, out=None, **kwargs):
        """Create an array of linestrings.
    
        This function will raise an exception if a linestring contains less than
        two points.
    
        Parameters
        ----------
        coords : array_like
            An array of lists of coordinate tuples (2- or 3-dimensional) or, if ``y``
            is provided, an array of lists of x coordinates
        y : array_like, optional
        z : array_like, optional
        indices : array_like, optional
            Indices into the target array where input coordinates belong. If
            provided, the coords should be 2D with shape (N, 2) or (N, 3) and
            indices should be an array of shape (N,) with integers in increasing
            order. Missing indices result in a ValueError unless ``out`` is
            provided, in which case the original value in ``out`` is kept.
        out : ndarray, optional
            An array (with dtype object) to output the geometries into.
        **kwargs
            Ignored if ``indices`` is provided.
    
        Examples
        --------
        >>> linestrings([[[0, 1], [4, 5]], [[2, 3], [5, 6]]]).tolist()
        [<pygeos.Geometry LINESTRING (0 1, 4 5)>, <pygeos.Geometry LINESTRING (2 3, 5 6)>]
        >>> linestrings([[0, 1], [4, 5], [2, 3], [5, 6], [7, 8]], indices=[0, 0, 1, 1, 1]).tolist()
        [<pygeos.Geometry LINESTRING (0 1, 4 5)>, <pygeos.Geometry LINESTRING (2 3, 5 6, 7 8)>]
    
        Notes
        -----
        - Usage of the ``y`` and ``z`` arguments will prevents lazy evaluation in ``dask``.
          Instead provide the coordinates as a ``(..., 2)`` or ``(..., 3)`` array using only ``coords``.
        """
        coords = _xyz_to_coords(coords, y, z)
        if indices is None:
>           return lib.linestrings(coords, out=out, **kwargs)
E           TypeError: ufunc 'linestrings' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

Since it was passing on older python versions, it might be dependent on the numpy version?

.travis.yml
```diff
@@ -23,16 +23,17 @@ jobs:
...
-    - CIBW_TEST_COMMAND="pytest --pyargs pygeos.tests"
+    - CIBW_TEST_COMMAND="pytest --pyargs shapely.tests"
+
 
 branches:
   only:
-  - master
+  - main
   - travis
+  - /^\d+\.\d+(\.\d+)?(-\S*)?$/
```
jorisvandenbossche, line 36: According to https://docs.travis-ci.com/user/conditions-v1, we should be able to add something like `if: tag IS present` for having it only run on tags (on main)

jorisvandenbossche:
For Travis, the wheel build job is passing, but the other ones are still failing (see https://app.travis-ci.com/github/shapely/shapely/builds/249895523). Since this PR is focusing on fixing the release workflow, I am going to leave those other builds for a follow-up.

.github/workflows/release.yml
```diff
@@ -132,7 +130,7 @@ jobs:
           CIBW_BEFORE_BUILD_WINDOWS: pip install delvewheel
           CIBW_REPAIR_WHEEL_COMMAND_WINDOWS: delvewheel repair --add-path ${{ runner.temp }}\geos-${{ env.GEOS_VERSION }}\bin -w {dest_dir} {wheel}
           CIBW_TEST_REQUIRES: pytest
-          CIBW_TEST_COMMAND: pytest --pyargs pygeos.tests
+          CIBW_TEST_COMMAND: pytest --pyargs shapely.tests
```
mwtoews, line 133: couldn't this simply be `pytest` without any other arguments? (similar for `.travis.yml`)

jorisvandenbossche: Maybe? But we struggled quite a bit with that in the original PR https://github.com/pygeos/pygeos/pull/365, and this was eventually the result of different tries (this should ensure that the installed package is tested, and not the source. And we included the tests in the installed package)

So if we would want to revisit this, I prefer doing that in a separate PR :)

jorisvandenbossche:
The proper tokens are still missing here for PyPI uploads, but let's also do that in a follow-up PR. I am going to merge this, to have the release workflow working again (and https://github.com/shapely/shapely/pull/1373 can be updated)

PR #1255, opened 2021-12-10
Replace Toblerity/Shapely -> shapely/shapely

jorisvandenbossche:


CITATION.txt
```diff
@@ -6,5 +6,5 @@ below:
   organization = {toblerity.org},
```
mwtoews, line 6: What's a better organization name?

jorisvandenbossche: We can maybe also leave out this line completely? (no idea if an "organization" is important for the citation if there is an author already)

mwtoews: yes, leave it out.

CITATION.txt
```diff
@@ -6,5 +6,5 @@ below:
...
   title =     {Shapely: manipulation and analysis of geometric objects},
   year =      {2007--},
-  url = "https://github.com/Toblerity/Shapely"
+  url = "https://github.com/shapely/shapely"
```
mwtoews, line 9:
```suggestion
  url =       {https://github.com/shapely/shapely},
```

PR #1264, opened 2021-12-17
Update MANIFEST.in to exclude *.c files

### LINKED ISSUE #1263: Speedups not available in shapely 1.8.0 with Python 3.10
In Python 3.9.9, `speedups.available` is `True`, but it is `False` with Python 3.10.1.

I'm using this Dockerfile to test:

```Dockerfile
FROM python:3.10.1-slim

RUN apt-get update && \
    apt-get -y install libgeos-dev
RUN pip install shapely==1.8.0
RUN python -c "import sys; from shapely import speedups; sys.exit(not speedups.available)"
```

``docker build .`` fails on 3.10.1 on my two systems:

* macOS Big Sur (11.6), Intel x86_64
* macOS Monterey (12.1), Apple M1 arm64

If I change to ``python:3.9.9-slim``, both systems pass.

mwtoews:
This is similar to #1249 and somewhat fixes #1263 for shapely-1.8

It seems that cythonized C files were packaged in the source distribution, which is not what we want to have.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,7 +5,8 @@ prune DLLs_AMD64
 prune DLLs_x86
 exclude *.txt
 exclude MANIFEST.in
-include CHANGES.txt CREDITS.txt LICENSE.txt README.rst VERSION.txt
+include CHANGES.txt CREDITS.txt LICENSE.txt README.rst
+include pyproject.toml
 recursive-include _vendor *.py
 recursive-include tests *.py *.txt
 recursive-include shapely/examples *.py
@@ -13,4 +14,5 @@ recursive-include shapely/speedups *.pyx
 recursive-include shapely/vectorized *.pyx
 recursive-include shapely *.pxi
 recursive-include shapely/DLLs *.dll *.rst *.txt
+recursive-exclude shapely *.c
 include docs/*.rst


jorisvandenbossche (approved):
Looks good to me. We should probably do a 1.8.1 release?

mwtoews:
I'd like to see a 1.8.1 release soon, particularly for Python 3.10.

mwtoews:
partial rollback to re-add VERSION.txt in 5c006173f16cdab6f5eaaae3fe7bb5403defe0cd, as this is written by setup.py

sgillies:
@mwtoews since we've overcome the issue between Cython and Python 3.10, I'm considering adding the .c files back to the sdist for 1.8.1.

jorisvandenbossche:
I think it might still be good in general to leave out the generated .c files. For example, when python 3.11 comes out, there could again be similar issues. Not having the .c files in there would avoid that, AFAIU. 
Also, nowadays with pyproject.toml which lists cython, installing from source should automatically install cython in the build environment (and cython itself is not a difficult to install built-time dependency)

jorisvandenbossche:
At least, that's what we are also doing at the moment for the main branch

sgillies:
There's a recent (long) python-dev thread about the Cython-CPython relationship: https://mail.python.org/archives/list/python-dev@python.org/thread/RS2C53LDZPXHRR2VCY2G2YSPDVA4LNQU/#K7LZAJTGDBFDM5TEQE7EALZMXQTCMQUS. Everybody wants to avoid breakage like at 3.10.0.

I think it's a good idea to remove the .c files for 2.0. But including them was never much of a problem until 3.10 and since that's resolved, I think I will include them in case someone is looking for them. And I'll be more proactive about 3.11.

jorisvandenbossche:
On that thread, Thomas Caswell (from matplotlib) also argued to not include cython output in the sdists: https://mail.python.org/archives/list/python-dev@python.org/thread/LJX5X7NH6F6TD2XN2VV5PSEFR52R2MU7/#LJX5X7NH6F6TD2XN2VV5PSEFR52R2MU7, and the cython maintainer seems to agree with it.

Now, certainly fine to keep it for 1.8 to be clear. It just seems a confirmation that it is indeed fine to do this for 2.0

sgillies:
Yes, Alan Snow and I saw this kind of recommendation before. Rasterio 1.3 will strictly require Cython to install from source (https://github.com/rasterio/rasterio/pull/2340). I only think it's too much of a change for shapely 1.8.1.

mwtoews:
I don't have any super strong preference, but I feel it's more future-proof to exclude the generated `.c` files. E.g. re-add `recursive-exclude shapely *.c` to MANIFEST.in for maint-1.8, then:
```
$ pip install build
$ python -m build --sdist
...
Successfully built Shapely-1.8.1.dev0.tar.gz
$ pip install dist/Shapely-1.8.1.dev0.tar.gz --verbose
...
  Installing collected packages: wheel, setuptools, numpy, cython, oldest-supported-numpy
...
  running build_ext
  cythoning shapely/vectorized/_vectorized.pyx to shapely/vectorized/_vectorized.c
...
Successfully installed Shapely-1.8.1.dev0
$ cd /tmp
$ python -c "from shapely import speedups; print(speedups.enabled)"
True
```
I'd expect this to work with future versions of Python/Cython too.

PR #1265, opened 2021-12-17
Linting: update pre-commit and setup.cfg configuration + pass black / flake8 / isort

jorisvandenbossche:
This PR:

- Updates the pre-commit configuration file from pygeos to shapely
- Consolidates the setup.cfg files from shapely and pygeos (eg adding some flake8 ignore rules that we were using in pygeos + isort config)
- Run pre-commit hooks for black and isort (in separate commits) + some manual fixes for passing the flake8 checks (mostly unused imports)


jorisvandenbossche:
With those changes, our CI is not fully green!

mwtoews (approved):
Changes all look like great improvements and clean-ups of unused thinks like `_longest`. And nice to see all checks pass!

setup.cfg
```diff
@@ -5,6 +5,22 @@ versionfile_source = shapely/_version.py
...
 tag_prefix =
 
+[flake8]
+max-line-length = 88
+ignore =
+    # space before : (needed for how black formats slicing)
+    E203,
+    # too many leading '#' for block comment
+    E266,
+    # TODO line length too long
```
caspervdw, line 15: Why TODO?

caspervdw (approved):
Looks good!

setup.cfg
jorisvandenbossche: Because I think we should, long term, actually remove this ignore rule and fix the lines that are too long. Currently that are mostly lines in docstrings (which black doesn't reformat), but also for docstrings it's useful to keep the line length limited.

Issue #1204, opened 2021-10-25
Post 1.8 tasks


mwtoews:
Now that #1066 is closed, here are a few follow-up tasks:

- [x] Create `maint-1.8` maintenance branch, i.e. this is the last release series with ctypes support, which sometimes require patching
- [x] Review items for the [1.8 milestone](https://github.com/Toblerity/Shapely/milestone/37) and either close or move forward, depending on what makes sense
- [x] Move `master` -> `main` -- this is as good enough of a time as any; see #1148 for a bit of discussion
- [x] Consider moving repo from https://github.com/Toblerity/Shapely to https://github.com/shapely/shapely as vaguely discussed in #819. I'm neutral on this task, but it should be discussed more thoroughly before.
- [x] Add `maint-1.8` version to https://readthedocs.org/projects/shapely/

After these have been cleared, there is plenty of work ahead with #962 and beyond.

Comments before action on any of these tasks is welcome here.

sgillies:
Done! Thank you @jorisvandenbossche .

jorisvandenbossche:
> yes, the repo transfer is just a click, it's all the follow up that is time consuming

@sgillies for the repo transfer: I am now an owner of the github.com/shapely org so I can help with the follow-up tasks after the transfer, but you will still have to click the button for doing the actual transfer (I am not an owner of Toblerity/Shapely, so can't do that)

jorisvandenbossche:
@sgillies small reminder for this (moving the repo)

sgillies:
@jorisvandenbossche thanks for the reminder! Here we go: https://github.com/shapely/shapely.

jorisvandenbossche:
Cool, thanks a lot!

hobu:
What was the reason to move away from the Toblerity org? If it is broken, confusing, or just not functional, we should disband it generally I guess.

sgillies:
@hobu yeah, we're moving away from what are largely mature or defunct projects. It should increase visibility a bit and will be useful in the context of PyPI's upcoming organization accounts.

hobu:
Ok. Rtree and Fiona are not defunct, so we'll need to move them.

hobu:
Actually, can we put Rtree under the Shapely org then? github.com/rtree is taken and I am already the owner of 13+ github orgs.

jorisvandenbossche:
[...]

PR #1268, opened 2021-12-17
Replace Toblerity/Shapely -> shapely/shapely

mwtoews:
This is very similar to #1255 but for the shapely-1.8 maintenance branch.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -28,7 +28,7 @@ libraries. Shapely is not concerned with data formats or coordinate systems,
 but can be readily integrated with packages that are. For more details, see:
 
 * `Shapely GitHub repository <https://github.com/shapely/shapely>`__
-* `Shapely documentation and manual <https://shapely.readthedocs.io/en/latest/>`__
+* `Shapely documentation and manual <https://shapely.readthedocs.io/en/maint-1.8/>`__
 
 Usage
 =====
--- a/README.rst
+++ b/README.rst
@@ -4,14 +4,14 @@ Shapely
 
 |github-actions| |appveyor| |coveralls|
 
-.. |github-actions| image:: https://github.com/Toblerity/Shapely/workflows/Tests%20(Linux)/badge.svg?branch=maint-1.8
-   :target: https://github.com/Toblerity/Shapely/actions?query=branch%3Amaint-1.8
+.. |github-actions| image:: https://github.com/shapely/shapely/workflows/Tests%20(Linux)/badge.svg?branch=maint-1.8
+   :target: https://github.com/shapely/shapely/actions?query=branch%3Amaint-1.8
 
-.. |appveyor| image:: https://ci.appveyor.com/api/projects/status/github/Toblerity/Shapely?branch=maint-1.8&svg=true
+.. |appveyor| image:: https://ci.appveyor.com/api/projects/status/github/shapely/shapely?branch=maint-1.8&svg=true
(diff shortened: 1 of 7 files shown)


README.rst
```diff
@@ -27,7 +27,7 @@ planar geometric objects. It is based on the widely deployed `GEOS
 libraries. Shapely is not concerned with data formats or coordinate systems,
 but can be readily integrated with packages that are. For more details, see:
 
-* `Shapely GitHub repository <https://github.com/Toblerity/Shapely>`__
+* `Shapely GitHub repository <https://github.com/shapely/shapely>`__
 * `Shapely documentation and manual <https://shapely.readthedocs.io/en/latest/>`__
```
mwtoews, line 31: There's one outstanding task in #1204 which would enable this link to be updated to the correct version of documentation.

jorisvandenbossche: https://shapely.readthedocs.io/en/stable/ url seems to work?

jorisvandenbossche (approved):
Thanks!

README.rst
mwtoews: I was looking for https://shapely.readthedocs.io/en/maint-1.8/ and just realised that it works! But wasn't a link to click in RTD for some reason.

jorisvandenbossche: It works because I just activated both maint-1.8.0 branch and 1.8.0 tag ;) 
The link should appear in the RTD versions box, but you might need a browser hard page refresh to get those (I suppose the available versions are cached)

mwtoews:
@sgillies just realized there is one follow-up task that only you are able to complete, which is to re-create an AppVeyor installation for shapely/shapely done here: https://ci.appveyor.com/account/frsci/projects/new . Not sure if existing project can be re-assigned to a new repo, or if the former needs to be rename, and a new one created with the same project name. This will hopefully fix the "build unknown" badge now shown in README.rst for `maint-1.8`

jorisvandenbossche:
I activated the AppVeyor app (github "Integration") again, but it might need a new commit on maint-1.8 to see if that is working again?

But quite likely it also needs some updates in the settings of https://ci.appveyor.com/project/frsci/shapely, which I don't have access too

mwtoews:
I've update it to only run for `maint-1.8` branch. I'll test a new commit directly...

mwtoews:
Hmm, seems like it didn't work. But also not a big deal. Not sure if it's worth investing too much time in getting some of the shapely-2.0 GitHub Actions to work for maint-1.8

jorisvandenbossche:
> Not sure if it's worth investing too much time in getting some of the shapely-2.0 GitHub Actions to work for maint-1.8

I think that should actually be pretty straightforward. 
The main change was the optional numpy or not, speedups or not, etc that made it more complicated. But could also just add windows to the current github actions workflow in maint-1.8, and copy over the windows specific bits of the workflow file in main:

https://github.com/shapely/shapely/blob/907c7ee19219b74c45e032f085d3dd3020e0544d/.github/workflows/tests.yml#L64-L68

https://github.com/shapely/shapely/blob/907c7ee19219b74c45e032f085d3dd3020e0544d/.github/workflows/tests.yml#L101-L105

https://github.com/shapely/shapely/blob/907c7ee19219b74c45e032f085d3dd3020e0544d/.github/workflows/tests.yml#L134-L142

(they should all have a `if: ${{ matrix.os == 'windows-2019' }}`)

PR #1272, opened 2021-12-24
CI: remove AppVeyor, add Windows and macOS to CI testing, remove 'dev'

mwtoews:
This is for the `maint-1.8` branch. I'll leave a few inline comments...


.github/workflows/tests.yml
```diff
@@ -36,15 +42,15 @@ jobs:
             speedups: 0
           # 2019
           - python: 3.8
-            geos: 3.8.2
+            geos: 3.8.1
```
mwtoews, line 45: downgrade is necessary to compile GEOS for Windows

.github/workflows/tests.yml
```diff
@@ -120,42 +145,62 @@ jobs:
...
       - name: Overview of the Python environment (pip list)
-        shell: bash
         run: |
           pip list
 
       - name: Run tests
-        shell: bash
-        continue-on-error: ${{ matrix.geos == 'main' }}
         run: |
+          export DYLD_LIBRARY_PATH=${{ env.DYLD_LIBRARY_PATH }}
```
mwtoews, line 189: this appears to be necessary! Despite being part of `$GITHUB_ENV`, it cannot be seen i.e. `echo "DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH"` shows a blank entry.

ci/install_geos.sh
```diff
@@ -1,10 +1,11 @@
-#!/bin/sh
+#!/bin/bash
```
mwtoews, line 1: pushd / popd are bash-specific

.github/workflows/tests.yml
```diff
@@ -70,46 +76,65 @@ jobs:
           - python: "3.10"
             geos: 3.10.1
             speedups: 0
-          # dev
-          - python: "3.10"
-            geos: main
-            speedups: 1
```
mwtoews, line 76: given that this is a maintenance branch, it does not need to test against the latest features.

jorisvandenbossche (approved):
Thanks!

I suppose the changes to install script are just a copy from the main branch?

mwtoews:
yes, except for one difference
```diff
1c1
< #!/bin/sh
---
> #!/bin/bash
```
because pushd and popd are bash specific.

PR #1274, opened 2021-12-28
BUG: fix linestring/linearring creation (copyFromBuffer usage) to check for dimension

### LINKED ISSUE #1267: BUG: creating LineString from array with invalid dimensions uses wrong coordinates
This changed in the last pygeos release (and thus also happens on shapely main):

```python
In [1]: import pygeos

# and using GEOS 3.9.1
In [2]: pygeos.__version__
Out[2]: '0.10.2'

In [3]: pygeos.linestrings(np.array([(1, 2, 3, 4), (4, 5, 6, 7)]))
GEOSException                             Traceback (most recent call last)
<ipython-input-3-9a89aa951713> in <module>
----> 1 pygeos.linestrings(np.array([(1, 2, 3, 4), (4, 5, 6, 7)]))

~/miniconda3/envs/geo/lib/python3.9/site-packages/pygeos/decorators.py in wrapped(*args, **kwargs)
     78             for arr in array_args:
     79                 arr.flags.writeable = False
---> 80             return func(*args, **kwargs)
     81         finally:
     82             for arr, old_flag in zip(array_args, old_flags):

~/miniconda3/envs/geo/lib/python3.9/site-packages/pygeos/creation.py in linestrings(coords, y, z, indices, **kwargs)
    101     coords = _xyz_to_coords(coords, y, z)
    102     if indices is None:
--> 103         return lib.linestrings(coords, **kwargs)
    104     else:
    105         return simple_geometries_1d(coords, indices, GeometryType.LINESTRING)

GEOSException: IllegalArgumentException: Unknown ordinate index 3
```

vs

```python
In [1]: import pygeos

# and uses GEOS 3.10.1
In [2]: pygeos.__version__
Out[2]: '0.12.0'

In [3]: pygeos.linestrings(np.array([(1, 2, 3, 4), (4, 5, 6, 7)]))
Out[3]: <pygeos.Geometry LINESTRING (1 2, 3 4)>
```

I suppose this is related to https://github.com/pygeos/pygeos/pull/436

jorisvandenbossche:
Closes #1267

### DIFF
--- a/src/geos.c
+++ b/src/geos.c
@@ -883,8 +883,9 @@ GEOSGeometry* PyGEOSForce3D(GEOSContextHandle_t ctx, GEOSGeometry* geom, double 
 }
 
 GEOSCoordSequence* coordseq_from_buffer(GEOSContextHandle_t ctx, const double* buf,
-                                        unsigned int size, unsigned int dims, char ring_closure,
-                                        npy_intp cs1, npy_intp cs2) {
+                                        unsigned int size, unsigned int dims,
+                                        char ring_closure, npy_intp cs1, npy_intp cs2,
+                                        char* last_error) {
   GEOSCoordSequence* coord_seq;
   char *cp1, *cp2;
   unsigned int i, j;
@@ -896,12 +897,22 @@ GEOSCoordSequence* coordseq_from_buffer(GEOSContextHandle_t ctx, const double* b
     if ((cs1 == dims * 8) && (cs2 == 8)) {
       /* C-contiguous memory */
       int hasZ = dims == 3;
+      if (dims > 3) {
+        snprintf(last_error, 1024,
+                 "IllegalArgumentException: Ordinate dimension cannot be larger than 3");
+        return NULL;
+      }
       coord_seq = GEOSCoordSeq_copyFromBuffer_r(ctx, buf, size, hasZ, 0);
       return coord_seq;
     }
     else if ((cs1 == 8) && (cs2 == size * 8)) {
       /* F-contiguous memory (note: this for the subset, so we don't necessarily
       end up here if the full array is F-contiguous) */
+      if (dims > 3) {
+        snprintf(last_error, 1024,
+                 "IllegalArgumentException: Ordinate dimension cannot be larger than 3");
+        return NULL;
+      }
       const double* x = buf;
       const double* y =  (double*)((char*)buf + cs2);
       const double* z = (dims == 3) ? (double*)((char*)buf + 2 * cs2) : NULL;
--- a/src/geos.c
+++ b/src/geos.c
@@ -882,10 +882,15 @@ GEOSGeometry* PyGEOSForce3D(GEOSContextHandle_t ctx, GEOSGeometry* geom, double 
   return force_dims(ctx, geom, 3, z);
 }
 
+/* Create a GEOSCoordSequence from an array
+ *
+ * Note: this function assumes that the dimension of the buffer is already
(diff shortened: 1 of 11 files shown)


src/geos.c
```diff
@@ -896,12 +897,22 @@ GEOSCoordSequence* coordseq_from_buffer(GEOSContextHandle_t ctx, const double* b
     if ((cs1 == dims * 8) && (cs2 == 8)) {
       /* C-contiguous memory */
       int hasZ = dims == 3;
+      if (dims > 3) {
+        snprintf(last_error, 1024,
+                 "IllegalArgumentException: Ordinate dimension cannot be larger than 3");
+        return NULL;
```
jorisvandenbossche, line 903: This is a bit an ugly solution (the usage of `last_error` to set a GEOS exception message on our side, and the need to pass through `last_error` in the function). 

But so the problem here is that this function needs to return NULL to signal an error (the way it is currently written), and inside the ufunc any NULL (it can also come from GEOS itself) is considered as a "GEOSException". So if I don't set any `last_error`, you get a "blank" GEOSException error.

I first tried to raise a Python error, but, then we would need to pass through the `errstate` instead, and setting a python error string (eg `PyErr_SetString`) also requires the GIL, which is not the case within this function. 

One alternative that is maybe a bit cleaner is to check for the dimension one level up in ufuncs.c (then we don't need to passthrough `last_error`, as that will be available there anyway). Or that would maybe even allow to make it a python error.

caspervdw: I see the issue, I am fine with this solution actually, but maybe you could consider an alternative: let this function return an `errstate` and put the `CoordinateSequence` pointer as an additional argument. That would of course involve a minor change in the places where this function is called.

This specific error case should be registered as an additional enum value with corresponding error raising in the `GEOS_FINISH` macro.

This pattern is used in some GEOS functions that return doubles. Also I tended to use it in some custom PyGEOS functions.

jorisvandenbossche: Changing it to return the errstate and registering the case in the enum and handling it in GEOS_FINISH sounds good (I suppose with setting it as a Python error then?). 

Now, thinking about it a bit more, you get a bit an inconsistent state where depending on the memory layout of the input array you would get a GEOSException or a ValueError if you have >3 ordinate dimension. 
It might be nice to always get a consistent error for that case. And so if we are handling this with a new errstate enum, I could also set this inside the ufunc code, _before_ calling `coordseq_from_buffer` (instead of checking it inside `coordseq_from_buffer`).

caspervdw: Sounds better! That would be an easy fix, if you do it before releasing the GIL then you can just directly set the PyErr.

In the cython code this is already checked.

jorisvandenbossche: OK, updated with this much simpler solution. 

I also added a check that the dimension is at least 2, because right now you could actually create a LineString/LinearRing with a 1D ordinate dimension, which then gets filled with zeros for the y dimension:

```python
In [2]: shapely.linestrings(np.random.randn(3, 2, 1))
Out[2]: 
array([<shapely.geometry.LineString LINESTRING (0.9497887286261655 0, -0.3278272663692793 0)>,
       <shapely.geometry.LineString LINESTRING (1.9726398481813794 0, 0.8734770826550462 0)>,
       <shapely.geometry.LineString LINESTRING (1.054188743128901 0, -1.0106490603521021 0)>],
      dtype=object)
```

I didn't use the exact same error message as the cython one, because here it needs to be more general (in the cython code the input is always a 2D array).

caspervdw (approved):

PR #1284, opened 2022-01-03
Update manual version from 1.7.0 to 1.8.0

akadouri:
Found a typo in the docs that listed the version as 1.7.0.


mwtoews (approved):
Thanks for the catch @akadouri !

PR #1295, opened 2022-01-23
RFC: add svg export support to basegeometry

jreiberkyle:
Discussion piece for implementing svg export support for shapes.

Adds support for SVG export to BaseGeometry through three new methods:
* svg_path_element(): returns the svg path as an ElementTree element
* svg_doc_element(): returns the svg document as an ElementTree element
* svg_doc(): returns the svg document as a string

_repr_svg_ now uses svg_doc() to generate the svg document string

Tests are passing, but new methods probably warrant new tests


shapely/geometry/base.py
```diff
@@ -181,14 +204,23 @@ def svg(self, scale_factor=1.0, **kwargs):
...
+
     def _repr_svg_(self):
         """SVG representation for iPython notebook"""
-        svg_top = (
-            '<svg xmlns="http://www.w3.org/2000/svg" '
-            'xmlns:xlink="http://www.w3.org/1999/xlink" '
-        )
         if self.is_empty:
-            return svg_top + "/>"
+            return ET.tostring(empty_svg_doc_element(), encoding="unicode")
```
sgillies, line 223: Great! Seeing `+ "/>"` in code makes me nervous. Using a tested serializer is way better.

shapely/geometry/base.py
```diff
@@ -67,6 +73,23 @@ class JOIN_STYLE:
     bevel = 3
 
 
+def empty_svg_doc_element():
+    doc = ET.Element('svg', xmlns=SVG_NAMESPACE)
+    doc.set('xmlns:xlink', SVG_XLINK_NAMESPACE)
+    return doc
```
sgillies, line 79: @jreiberkyle In my opinion this function isn't called in enough places to be justified. I subscribe to the rule of 3: duplication two times is fine, make a new function the third time.

My suggestion is to copy this into svg_doc_element and then for the empty case below, return `ET.Element('svg', xmlns=SVG_NAMESPACE)`.

shapely/geometry/base.py
```diff
@@ -181,14 +204,23 @@ def svg(self, scale_factor=1.0, **kwargs):
...
+
+        # svg is y-down, shapely is y-up
+        _, ymin, _, ymax = self.bounds
+        path.set('transform', f'matrix(1,0,0,-1,0,{ymax + ymin})')
+        return path
+
+    def svg_doc(self, width, height, xmin, ymin, xmax, ymax, stroke_scale_factor=1):
+        path = self.svg_path_element(stroke_scale_factor)
+        doc_element = svg_doc_element([path], width, height, xmin, ymin, xmax, ymax)
+        return ET.tostring(doc_element, encoding="unicode")
```
sgillies, line 218: How about eliminating this function and where it is called, substitute `svg_doc_element([self.svg_path_element(stroke_scale_factor)], width, height, xmin, ymin, xmax, ymax)` ?

sgillies:
@jreiberkyle we're using black and isort as pre-commit hooks for this project. If you run `pre-commit run --all-files` in your working directory, I expect it to make the changes that will satisfy the Lint GitHub workflow.

This is good stuff, thank you!

jreiberkyle:
Thanks for the feedback @sgillies . These are great suggestions and I'll get right on them. As for linting, I am having a hard time finding where the `pre-commit` command comes from. It doesn't appear to be available after I installed shapely and requirements-dev.txt. Additionally, I can't see it mentioned anywhere in the repo other than the YAML file. Should I install the [pre-commit](https://pypi.org/project/pre-commit/) pypi package manually?

shapely/geometry/base.py
jreiberkyle: makes sense, will do!

shapely/geometry/base.py
jreiberkyle: With the elimination of `svg_doc()`, the user needs to figure out that they need to set the encoding to "unicode" or the SVG won't be valid. That wasn't immediately obvious to me, but I was able to figure it out after brushing up on ElementTrees. It seems reasonable to make this compromise, as the answer is already in the Python docs.

sgillies:
@jreiberkyle it looks like the project is using a pre-commit GHA action in its CI, but I don't think it will hurt to add it to the dev requirements.

jreiberkyle:
Closing this in favor of #1610. I took a slightly different approach, streamlining to one new function and also streamlining the tests and accounting for possible float mismatching.

PR #1299, opened 2022-01-31
PERF: speed-up bounds function with GEOSGeom_getXMin et al (GEOS >= 3.7.0)

jorisvandenbossche:
Instead of going through `GEOSEnvelope` and getting the coordinates of that rectangular geometry object. This is simpler code, and quite a bit faster (it avoids the construction/destruction of the polygon object).

See https://github.com/libgeos/geos/pull/554 for context


caspervdw (approved):
Looks good, thanks!

PR #1302, opened 2022-02-04
Update repr for the Geometry classes

### LINKED ISSUE #1248: Decide on the repr of the geometry objects
I currently skipped the repr tests in https://github.com/Toblerity/Shapely/pull/1241 because they are still failing. 

As an overview, what we have been doing in the past / in the different packages (using a simple Point as example):

- Shapely 1.8 and before: `<shapely.geometry.point.Point at 0x7fae395b7e50>` 
  - This is the default repr inherited from python
- Shapely shapely-2.0 branch: `<shapely.geometry.Point POINT (1 1)>`
  - I changed this in https://github.com/Toblerity/Shapely/pull/983, but wasn't really discussed at the time (and a simple python `__repr__()` using `self.wkt`)
- PyGEOS: `<pygeos.Geometry POINT (1 1)>`
  - This is implemented in C (`GeometryObject_repr`). This also deals with 1) failure in WKT writer, and 2) limiting total length to 80 characters (both could also be done in python of course)

I think the questions to answer are (with the second and third question depending on answering yes on the first:

1) Do we want to include a WKT (preview) or coordinates in the repr?
2) Which class name to show?
3) Implement in C or Python?

----

**1) Do we want to include a WKT (preview) or coordinates in the repr?**

There have been several issues/PRs in the past asking about adding a `repr` (https://github.com/Toblerity/Shapely/issues/718, https://github.com/Toblerity/Shapely/pull/478), but until now there was never added one in Shapely. Some arguments back then were about how to deal with many coordinates (which can easily fill the screen), about ambiguity (if only showing something like `"Point(1, 1)"`), about the "repr-eval" use case / round-trippability (eg doing `"shapely.geometry.Point(1, 1)"` works, but this easily gets complicated  for other geometry types / large geometries)

I personally like the choice me made in PyGEOS for having a *truncated WKT* within a `<pygeos. ...>`:

- It is not "just" a WKT representation (in which case `repr` would be equivalent to `str`, which we don't want I think), by still including the WKT in a `<shapely... >` context.
- Especially when working with arrays of geometries, it is IMO really important to limit the length of the repr (it's annoying to flood your console output when inspecting data), but at the same time it is also very valuable to see some preview of the data (a single object you can also print to get str instead of repr, but for an array this is also less convenient)
- We could also make a more Python-code like repr instead of using WKT (so like `shapely.geometry.LineString([(1, 1), (2, 3)])`), but 1) this is more complex to write (and for more more complex geometry will also get more harder to interpret I think), and 2) it will never be guaranteed to be "eval"-able if we limit the length of the repr. So for those reasons, I would personally not put the effort in writing custom code for this, but just go with truncated WKT (but it would certainly be possible if preferred).

**2) Which class name to show?**

One drawback of using WKT with the full subclass name as I now did in the main branch (eg  `<shapely.geometry.Point POINT (1 1)>`) is that you get a repetition in the class name and geometry type in the WKT. But on the other hand, I think it would be better to use the full subclass name (`shapely.geometry.Point`) instead of the base class (`shapely.Geometry`), since it's the subclass that users actually interact with.

**3) Implement in C or Python?**

A technical point, but I suppose it's actually easier to just do this in Python instead of the current C function (in PyGEOS it was only possible to do this in C, but now with the Python subclasses we can also do it there) 

@sgillies @snorfalorpagus @snorfalorpagus @caspervdw @brendan-ward

jorisvandenbossche:
Closes #1248


shapely/tests/test_io.py
```diff
@@ -385,38 +385,34 @@ def test_to_wkt_multipoint_with_point_empty_errors():
...
+    assert repr(point) == "<shapely.geometry.Point POINT (2 3)>"
 
 
-@shapely20_todo
 def test_repr_max_length():
     # the repr is limited to 80 characters
     geom = shapely.linestrings(np.arange(1000), np.arange(1000))
     representation = repr(geom)
-    assert len(representation) == 80
+    assert len(representation) == 92
```
jorisvandenbossche, line 396: Because the "fixed" part became longer ("pygeos.Geometry" vs "shapely.geometry.LineString"), this number increased. Another option if we want to keep the length limited here at 80 is to reduce the width of the truncated WKT, but that also shows even a smaller part of the WKT ..

jorisvandenbossche: @caspervdw do you have a strong opinion here? We might want to keep the full truncated version at 80 chars for avoiding wrapping on terminals?

jorisvandenbossche: (sidenote: another option is to move the geometry subclasses to the main namespace. That's something I have been considering for other reasons as well ("what to do with the Geometry class?"), will try to open an issue for that)

src/pygeom.c
```diff
@@ -199,16 +199,26 @@ static PyObject* GeometryObject_ToWKB(GeometryObject* obj) {
...
     return PyUnicode_FromString("<shapely.Geometry Exception in WKT writer>");
   }
+
+  GEOS_INIT;
+  GEOSGeometry* geom = self->ptr;
+  geom_type = GEOSGeomTypeId_r(ctx, geom);
+  GEOS_FINISH;
+
+  int type_lengths[8] = {5, 10, 10, 7, 10, 15, 12, 18};
+  int max_length = 80 - (11 + type_lengths[geom_type]);
```
jorisvandenbossche, line 217: @caspervdw I was thinking that it would be nice that all reprs have the exact same max length (eg when you have an array with a mixture of geometry types, that the repr truncates at the same place). 
Now, while the above does that, we can probably just as well do this in Python (which would make this simpler)? I don't think there is a specific benefit of keeping the repr definition in C? (the full (non-truncated) WKT string python object is constructed anyway)

(and to be clear, if we keep it in C, the above needs some cleanup to properly handle possible errors)

caspervdw: I agree on moving the  repr logic to Python space. Am I right that the Geometry (base)class is now internal (not user-facing)? If not maybe good to keep the original repr (in C) around, else, I am in favour of dropping that altogether.

shapely/tests/test_io.py
caspervdw: I don’t have a strong opinion about the truncation length.
I prefer moving the geometry type to the root namespace though, to shorten that part of the repr and to ease imports.

jorisvandenbossche: > I prefer moving the geometry type to the root namespace though, to shorten that part of the repr and to ease imports.

Yes, that's already done! -> https://github.com/shapely/shapely/pull/1339

src/pygeom.c
jorisvandenbossche: > Am I right that the Geometry (base)class is now internal (not user-facing)? If not maybe good to keep the original repr (in C) around, else, I am in favour of dropping that altogether.

At the moment I didn't yet remove the public exposure, but that's indeed the idea. Although I might actually prefer to keep the class exposed as the "base class" (eg to use for isinstance check), something to further discuss in https://github.com/shapely/shapely/issues/1330

PR #1303, opened 2022-02-04
Deprecate calling the BaseGeometry constructor + EmptyGeometry class

### LINKED ISSUE #1022: Shapely 2.0: deprecate BaseGeometry() and EmptyGeometry() ?
In current Shapely, the base class `BasecGeometry` constructor can be called, creating an empty GeometryCollection:

```
>>> from shapely.geometry.base import BaseGeometry
>>> geom = BaseGeometry()
>>> print(geom)
GEOMETRYCOLLECTION EMPTY
```

However, if you want an empty geometry collection, you can also explicitly create one with eg `GeometryCollection()`. So I think it is not needed to have a working constructor for the BaseGeometry. 

I know that at some point in geopandas we did use this to create an empty geometry, so it might be worth actually deprecating this in Shapely 1.8.

---

In addition, a more explicit `EmptyGeometry` was added (actually to address the issue that geopandas was using `BaseGeometry()` to get an empty geometry -> https://github.com/Toblerity/Shapely/pull/514). 
But since there is no "generic" empty type in GEOS, but only empty geometries of one of the specific geometry types (`EmptyGeometry()` also creates an empty GeometryCollection), in hindsight, this `EmptyGeometry` class doesn't seem that useful, and even somewhat misleading. 

So I would propose to also deprecate this class (in favor of using the explicit constructors, like `Point()`, `Polygon()`, `GeometryCollection()`, ...)

jorisvandenbossche:
Closes #1022

All the geometry subclasses can now properly create an empty geometry when called with no input (eg `Point()`, `Polygon()`). So I personally don't see a good reason why people should do `BaseGeometry()`, while this is creating an empty geometry collection, so `GeometryCollection()` is more explicit. 
See https://github.com/shapely/shapely/issues/1022 for some more argumentation.


mwtoews:
Shouldn't this raise `ShapelyDeprecationWarning` (which is a subclass of `FutureWarning`)?

jorisvandenbossche:
That's fine for me as well (I wasn't sure we were going to use ShapelyDeprecationWarning just for the 2.0 release, or whether we would keep using that class for other deprecations in the future as well, but either way is fine).

caspervdw (approved):

jorisvandenbossche:
The black installation issue is getting fixed in https://github.com/shapely/shapely/pull/1355

PR #1304, opened 2022-02-04
TST: remove filterwarnings for numpy 1.21

jorisvandenbossche:
Since the geometry objects are now no longer sequence-like objects / objects with an array interface, those objects should now interoperate fine with numpy arrays, and the warning should be no longer be raised.


mwtoews (approved):
Indeed this is no longer required. Xref #1173 that noted the initial changes.

PR #1305, opened 2022-02-04
Consolidate error class for unsupported GEOS version functionality

jorisvandenbossche:
We had two classes as a left-over from the merger. I removed the top-level one (from pygeos) and used the one in the `shapely.errors` submodule.


caspervdw (approved):
Looks good!

PR #1306, opened 2022-02-04
CLN: remove no longer used custom error classes

jorisvandenbossche:
Related to the discussion in https://github.com/shapely/shapely/issues/991

This removes some custom exception subclasses that are no longer used after the merge with pygeos (eg reading invalid WKB or WKT now raises a GEOSException)

### DIFF
--- a/shapely/errors.py
+++ b/shapely/errors.py
@@ -10,28 +10,12 @@ class UnsupportedGEOSVersionError(ShapelyError):
     """Raised when the system's GEOS library version is unsupported."""
 
 
-class ReadingError(ShapelyError):
-    """A WKT or WKB reading error."""
-
-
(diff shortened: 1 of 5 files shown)


sgillies (approved):
@jorisvandenbossche and GEOSException subclasses ShapelyError?

jorisvandenbossche:
> and GEOSException subclasses ShapelyError?

No, not yet (so this PR itself doesn't close https://github.com/shapely/shapely/issues/991, just some initial clean-up).

We will have to move ShapelyError to C to have GEOSException to subclass it. Which shouldn't that hard, the current definition of GEOSException is two lines of code:

https://github.com/shapely/shapely/blob/b47dab627f796e3a89e340a6e533504037535e88/src/geos.c#L15-L16

PR #1307, opened 2022-02-05
ENH: add maximum_inscribed_circle (polylabel)

jorisvandenbossche:
MaximumInscribedCircle and LargestEmptyCircle are new functions from GEOS 3.9 (https://github.com/libgeos/geos/commit/0909042df5f8d4b9ae1832b3fd67713e206057fb, port of https://github.com/locationtech/jts/pull/530).

This PR is exposing them in Shapely. For now, I just added bindings for both functions separately. However, MaximumInscribedCircle only accepts polygonal input, and LargestEmptyCircle also works for other geometries. And for example in PostGIS, they combined both in a single user-facing function: [`ST_MaximumInscribedCircle`](https://postgis.net/docs/ST_MaximumInscribedCircle.html). 

Further, the PostGIS wrapper also has some additional logic: checking for empty input (and returning empty point and zero radius), automatically calculating a `tolerance`, and thus choosing between both functions depending on the input geometry type (see https://github.com/postgis/postgis/blob/6f335e8117bcbc43daa9e0536c0703e3926f8acd/postgis/lwgeom_geos.c#L376-L440)

I would be inclined to follow the PostGIS logic (in which case I need to update the PR, need to add more tests and clean up in any case, but first want to decide on this before doing that).

This can also replace a custom python `polylabel` implementation that we currently have (there is one test case that gives different results, but that might be something to report to GEOS).

Update: this PR now only adds `maximum_inscribed_circle`

Also closes https://github.com/shapely/shapely/issues/1836


jorisvandenbossche:
Ah, I will of course have to keep the Python implementation for `polylabel` at least for older GEOS versions.

shapely/constructive.py
```diff
@@ -888,6 +890,90 @@ def minimum_bounding_circle(geometry, **kwargs):
...
+    The class supports polygons with holes and multipolygons.
+
+    Returns a two-point linestring, with the first point at the center of the
+    inscribed circle and the second on the boundary of the inscribed circle.
+
+    Parameters
+    ----------
+    geometry : Geometry or array_like
+    tolerance : float or array_like
+        Stop the algorithm when the search area is smaller than this tolerance
```
jorisvandenbossche, line 922: In PostGIS, this is not exposed as an argument to the user, but always calculated based on the size of the geometry (https://github.com/postgis/postgis/blob/6f335e8117bcbc43daa9e0536c0703e3926f8acd/postgis/lwgeom_geos.c#L395-L398):

```
		if (!gserialized_get_gbox_p(geom, &gbox))
			PG_RETURN_NULL();

		width = gbox.xmax - gbox.xmin;
		height = gbox.ymax - gbox.ymin;
		size = width > height ? width : height;
		tolerance = size / 1000.0;
```

So even if we still expose the keyword (if we think that is useful), we could use similar logic for the default value if it was not specified by the user.

tests/test_polylabel.py
```diff
@@ -68,7 +37,7 @@ def test_rectangle_special_case(self):
         polygon = Polygon([(32.71997,-117.19310), (32.71997,-117.21065),
                            (32.72408,-117.21065), (32.72408,-117.19310)])
         label = polylabel(polygon)
-        self.assertEqual(label.coords[:], [(32.722025, -117.201875)])
+        self.assertEqual(label.coords[:], [(32.722025, -117.208595)])
```
jorisvandenbossche, line 40: This test was added after this report (and fix): https://github.com/mapbox/polylabel/issues/3 
So it seems that the GEOS implementation has the same issue with rectangular polygons (will bring it up with GEOS to see if they are interested in the fix, though it is not strictly a bug I think)

shapely/constructive.py
caspervdw: I think that is a very sensible thing to do.

caspervdw (commented):
Looks good, some things:

- tests needs finishing (especially corner cases / weird geom types / nan-containing geoms - we want to prevent segfaults at all times)
- maybe add the functions to the docs?
-

jorisvandenbossche:
> For now, I just added bindings for both functions separately. However, MaximumInscribedCircle only accepts polygonal input, and LargestEmptyCircle also works for other geometries. And for example in PostGIS, they combined both in a single user-facing function: [`ST_MaximumInscribedCircle`](https://postgis.net/docs/ST_MaximumInscribedCircle.html).  ...
> I would be inclined to follow the PostGIS logic

I was chatting with @martinfleis a few weeks ago about this new feature, and there might actually be a reason for keeping those as two separate functions, as they can be different conceptually for polygons. 
The "maximum inscribed circle" gives you the largest circle that is inside the interior of the polygon. But a "largest empty circle" could be the largest circle that is within the _envelope_ of the polygon, but outside of its interior. I think a typical example are building footprints, where you might want to get the circle in an open area within a building block (polygon hole or in the middle of MultiPolygon). 

Now, while that distinction might be nice to have, that's not the actual behaviour currently of GEOS. In LargestEmptyCircle, it just considers the linework of polygonal input, and then doesn't care whether the largest circle it finds is in the interior of the polygon or not (it's just lines). 

Using the example of the PostGIS docs of [`ST_MaximumInscribedCircle`](https://postgis.net/docs/ST_MaximumInscribedCircle.html), we get:


<details>

```python
import shapely
from shapely.plotting import plot_polygon, plot_line

polygon = shapely.from_wkt("POLYGON ((40 180, 110 160, 180 180, 180 120, 140 90, 160 40, 80 10, 70 40, 20 50, 40 180), (60 140, 50 90, 90 140, 60 140))")

line_radius = shapely.maximum_inscribed_circle(polygon)
circle = shapely.get_point(line_radius, 0).buffer(shapely.length(line_radius))

plot_polygon(polygon)
plot_line(line_radius, color="C1")
plot_polygon(circle, color="C1", add_points=False)
```

</details>

image

and you get exactly the same result whether we are using `shapely.maximum_inscribed_circle` or `shapely.largest_empty_circle` (while the largest "empty" circle could be in the hole)

If I manually construct a different polygon, we can see that the two functions can potentially return something different (using `'POLYGON ((10 10, 10 60, 70 60, 70 10, 40 10, 40 20, 60 20, 60 50, 20 50, 20 10, 10 10))'`):

<details>

```python
polygon = shapely.Polygon([(10, 10), (10, 60), (70, 60), (70, 10), (40, 10), (40, 20), (60, 20), (60, 50), (20, 50), (20, 10), (10, 10)])

# maximum inscribed circle
line_radius = shapely.maximum_inscribed_circle(polygon)
circle = shapely.get_point(line_radius, 0).buffer(shapely.length(line_radius))
plot_polygon(polygon)
plot_line(line_radius, color="C1")
plot_polygon(circle, color="C1", add_points=False)

# largest empty circle
line_radius = shapely.largest_empty_circle(polygon, polygon.convex_hull, tolerance=0.1)
circle = shapely.get_point(line_radius, 0).buffer(shapely.length(line_radius))
plot_polygon(polygon)
plot_line(line_radius, color="C1")
plot_polygon(circle, color="C1", add_points=False)
```

</details>

Maximum inscribed circle:

image

Largest empty circle:

image

However, since GEOS just regards the polygon as general linework, it depends on your exact polygon whether you get a result that is within or outside the input for `largest_empty_circle` (which doesn't sound ideal if you want to use this function with polygonal input).

The `largest_emtpy_circle` has a boundary keyword, where you can limit the area where the center can be located. So we / a user could actually manually set that area to be something that does _not_ include the original polygon. Instead of using the default envelope, we could subtract the input area:

```
# full envelope without polygon interior
shapely.largest_empty_circle(
    polygon.boundary, tolerance=0.1,
    boundary=polygon.envelope.difference(polygon).union(polygon.boundary)
)
```

image

```
# only considering holes
shapely.largest_empty_circle(
    polygon.boundary, tolerance=0.1, 
    boundary=shapely.Polygon(polygon.exterior).difference(polygon).union(polygon.boundary).buffer(0.001)
)
```

image


But that seems a bit brittle with current GEOS (I need to union it with the boundary to ensure that it fully covers the input, but in the second case I therefore also needed to buffer it because the added line gave problems)

And I suppose it will also be very use case dependent whether you want to use the envelope / convex hull / concave hull / actual polygon exterior as the boundary. 

But, assuming the user can do this step of providing the boundary manually to fit their use case, that might still be a reason to expose `largest_empty_circle` separately from `maximum_inscribed_circle`.

jorisvandenbossche:
Some relevant discussion at https://github.com/libgeos/geos/issues/855 as well

dr-jts:
> But, assuming the user can do this step of providing the boundary manually to fit their use case, that might still be a reason to expose `largest_empty_circle` separately from `maximum_inscribed_circle`.

I agree with this.  Once the Largest Empty Circle algorithm can support polygonal obstacles, there is then a clear difference between it's functionality and that of Maximum Inscribed Circle.   This is doubly so with the upcoming enhancement for Largest Empty Circle to accept _any_ polygonal geometry as a boundary (see https://github.com/locationtech/jts/pull/973).

I think PostGIS should add an `ST_LargestEmptyCircle` function as well.  The overloading of `ST_MaximumInscribedCircle` was convenient at the time, but makes less sense as functionality is extended. I see the signature being `ST_LargestEmptyCircle(obstacles, tolerance [, boundary])`.

shapely/algorithms/polylabel.py
```diff
@@ -108,56 +34,9 @@ def polylabel(polygon, tolerance=1.0):
     >>> from shapely import LineString
     >>> polygon = LineString([(0, 0), (50, 200), (100, 100), (20, 50),
     ... (-100, -20), (-150, -200)]).buffer(100)
-    >>> polylabel(polygon, tolerance=10).wkt
-    'POINT (59.35615556364569 121.83919629746435)'
+    >>> polylabel(polygon, tolerance=0.001)
```
sgillies, line 37: @jorisvandenbossche has something about tolerance changed in the implementation so that we'd need to tighten the value? Like, does `10` no longer succeed?

sgillies (approved):

shapely/algorithms/polylabel.py
jorisvandenbossche: If I recall correctly, I changed this because 10 was just not really a realistic example. 

The docstring explains `tolerance` as "`tolerance` represents the highest resolution in units of the input geometry that will be considered for a solution."  
So it stops when the search area is smaller than 10 (wide or high, not sure exactly). But so for coordinates in the range of 0-200, that's seems a bit stupid to already stop with a precision of 10

jorisvandenbossche: (of course, it is is meant for just labeling on a plot, you might not care about a good precision, but I think for the example it is better to set a bit lower tolerance)

jorisvandenbossche: And if you have lat/lon coordinates, the default of 1.0 is also typically too large, see eg https://github.com/shapely/shapely/issues/1383 where the default caused confusion. For projected coordinates that default is of course typically fine.

Issue #1024, opened 2020-11-06
BUG: failure when importing shapely/pygeos in subprocess (outside pytest)


jorisvandenbossche:
On the shapely-2.0 branch, we have a failure on Appveyor (skipped at the moment), see https://github.com/Toblerity/Shapely/pull/983#issuecomment-707706532, and some more analysis of @mwtoews at https://github.com/Toblerity/Shapely/pull/983#issuecomment-718557666

It *might* automatically go away when the source code of pygeos and shapely gets actually merged / usage of ctypes is completely removed, but in any case something to keep track off to check before final Shapely 2.0 release.

jorisvandenbossche:
Going to close this as no longer relevant

PR #1309, opened 2022-02-05
TST: enable test_pickle_persistence again on Windows

jorisvandenbossche:
See https://github.com/shapely/shapely/issues/1024

Trying to remove to see if has fixed itself in the meantime (now we don't have a mixture of pygeos/shapely anymore)


jorisvandenbossche:
So it's still failing to find the GEOS library (*ImportError: DLL load failed: The specified module could not be found.*), but now consistently for all Python versions on Windows (so I am not fully sure it is related to the change in finding libraries since 3.8, xref https://github.com/shapely/shapely/pull/983#issuecomment-718557666 cc @mwtoews)

caspervdw:
Maybe it is wise to try an approach  different from executing a script with `sys.executable`.

I have a gist containing a `may_segfault` function to protect against segfaults/aborts over here. It uses `multiprocessing.Process` to fork/spawn a process.

https://gist.github.com/caspervdw/10d9fd373a31642498cdea494d2f1bc4

jorisvandenbossche:
Testing this three years later again, now it seems to be passing :)

PR #1310, opened 2022-02-08
Backport releasing workflow from main

sgillies:


.github/workflows/release.yml
```diff
@@ -0,0 +1,188 @@
...
+          CIBW_ENVIRONMENT_WINDOWS:
+            GEOS_INSTALL='${{ runner.temp }}\geos-${{ env.GEOS_VERSION }}'
+            GEOS_LIBRARY_PATH='${{ runner.temp }}\geos-${{ env.GEOS_VERSION }}\lib'
+            GEOS_INCLUDE_PATH='${{ runner.temp }}\geos-${{ env.GEOS_VERSION }}\include'
+          CIBW_BEFORE_ALL: ./ci/install_geos.sh
+          CIBW_BEFORE_ALL_WINDOWS: ci\install_geos.cmd
+          CIBW_BEFORE_BUILD_WINDOWS: pip install delvewheel
+          CIBW_REPAIR_WHEEL_COMMAND_WINDOWS: delvewheel repair --add-path ${{ runner.temp }}\geos-${{ env.GEOS_VERSION }}\bin -w {dest_dir} {wheel}
+          CIBW_TEST_REQUIRES: pytest
+          CIBW_TEST_COMMAND: python -m pytest --pyargs shapely.tests
```
jorisvandenbossche, line 135: I don't think the `--pyargs shapely.tests` is possible for the maint-1.8 branch, since the tests are there top-level and not installed with the package. 

It should be possible to run the tests from the project checkout though (without changing the repo structure for the 1.8 branch), but I seem to remember that @caspervdw also struggled a long time to get this test invocation working ...

jorisvandenbossche: The PR is at https://github.com/pygeos/pygeos/pull/365#issuecomment-878531326, after some struggling we decided to include the tests in the installed package, I think

sgillies: I ended up falling back to `{project}/tests`.

sgillies:
Couldn't have done this without the examples from pygeos and shapely-wheels (appveyor config)!

jorisvandenbossche:
Out of curiosity: the changes you needed to make compared the the version in main (from pygeos), were those all related to how shapely in the 1.8 branch loads the library at runtime with ctypes (and where it expected to find the libs in case they are included in the wheels)? 

Or to phrase it differently: is there anything that you figured out here that would be useful for the main branch?

sgillies:
@jorisvandenbossche everything was related to ctypes and especially dlopening the shared libs at build time. I can't think of anything the main branch needs.

PR #1311, opened 2022-02-08
add tests for BaseGeometry._repr_svg_

jreiberkyle:
This PR adds tests for `BaseGeometry._repr_svg_()` in preparation for changes to the function in #1295.

### DIFF
--- a/tests/test__repr_svg_.py
+++ b/tests/test__repr_svg_.py
@@ -1,0 +1,128 @@
+from xml.dom.minidom import parseString as parse_xml_string
+
+from . import unittest
+from shapely.geometry import Point, MultiPoint, LineString, MultiLineString,\
+    Polygon, MultiPolygon
+from shapely.geometry.collection import GeometryCollection
+
+
+class _Repr_Svg_TestCase(unittest.TestCase):
+
+    def assert_repr_svg_(self, geom, expected):
+        """Helper function to check SVG representation for iPython notebook"""
+        self.assertEqual(geom._repr_svg_(), expected)
+
+    def test_empty(self):
+        self.assert_repr_svg_(
+            Point(), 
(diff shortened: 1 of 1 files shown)


jreiberkyle:
@sgillies I prepared this PR so we can get some tests in before changing the SVG functionality.

tests/test__repr_svg_.py
```diff
@@ -0,0 +1,131 @@
+import xml.etree.ElementTree as ET
+
+from . import unittest
+from shapely.geometry import Point, MultiPoint, LineString, MultiLineString,\
```
sgillies, line 4: Nit: black would eliminate the forward slash and turn these into a list.

tests/test__repr_svg_.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def test_multipoint(self):
+        self.assert_repr_svg_(
+            MultiPoint([(6, 7), (3, 4)]),
+            '<svg xmlns="http://www.w3.org/2000/svg" '
+            'xmlns:xlink="http://www.w3.org/1999/xlink" '
+            'width="100.0" height="100.0" '
+            'viewBox="2.88 3.88 3.24 3.24" preserveAspectRatio="xMinYMin meet">'
+            '<g transform="matrix(1,0,0,-1,0,11.0)"><g>'
+            '<circle cx="6.0" cy="7.0" r="0.09720000000000001" '
```
sgillies, line 44: When I see something like that `r` value, I worry that it will be platform dependent. On a different system it might be `"0.09720000000000002"`.

The most important thing might be to check that we have two circle elements and that their radii are roughly correct?

sgillies (changes requested):
@jreiberkyle these tests will help a lot! My only concern is that asserting on string representations of floats could be a bit flaky.

Issue #1315, opened 2022-02-12
Improving performance of `shapely.algorithms.cga.signed_area`


whophil:
I have a program whose performance is limited by the method `shapely.algorithms.cga.signed_area`, which is not vectorized.

https://github.com/shapely/shapely/blob/7396b0823ca7fdb8b4115fb5b458bbb1057de59d/shapely/algorithms/cga.py#L4-L13

Benchmarks and a proposal follow.

## Expected behavior and actual behavior.

`shapely.algorithms.cga.signed_area` should be faster. This affects the performance of:
- `shapely.geometry.polygon.orient`
- `shapely.geometry.LinearRing.is_ccw`

## Benchmarking

```python
import numpy as np
import shapely.geometry as sg
import shapely.speedups
from shapely.algorithms.cga import signed_area
shapely.speedups.enable()

from time import time


def fast_signed_area(ring):
    """Vectorized version of shapely.algorithms.cga.signed_area"""
    coords = np.array(ring.coords)[:, :2]
    xs, ys = np.vstack([coords, coords[1]]).T
    return np.sum(xs[1:-1] * (ys[2:] - ys[:-2])) / 2.0


n_repeats = 200


cir = sg.Point((0, 0)).buffer(1.0, resolution=1024)
ring = cir.exterior

t0 = time()
[cir.area for _ in range(n_repeats)]
print('cir.area', time() - t0)

t0 = time()
[signed_area(ring) for _ in range(n_repeats)]
print('signed_area(ring)', time() - t0)

t0 = time()
[ring.is_ccw for _ in range(n_repeats)]
print('ring.is_ccw', time() - t0)

t0 = time()
[fast_signed_area(ring) for _ in range(n_repeats)]
print('fast_signed_area(ring)', time() - t0)

t0 = time()
[sg.polygon.orient(cir) for _ in range(n_repeats)]
print('orient(ring)', time() - t0)
```

Which yields
```
cir.area               0.009488105773925781
signed_area(ring)     11.10140085220337
ring.is_ccw           10.961623668670654
fast_signed_area(ring) 0.07963943481445312
orient(ring)          11.738571882247925
```
## Proposal
The vectorized `fast_signed_area` method in the benchmarking code should be a drop-in replacement. Is there any reason not to use it?

## Operating system

Linux

## Shapely version and provenance

(For example, 1.6b4 installed from PyPI using pip)
```
shapely                   1.8.0           py310h1b8dd88_5    conda-forge
```

[...]

PR #1323, opened 2022-02-18
PERF: vectorized implementation for signed area

whophil:
As discussed in #1315 

1. Vectorizes `shapely.algorithms.cga.signed_area`
2. Modifies `shapely.polygon.orient` to use `LinearRing.is_ccw` instead of `shapely.algorithms.cga.signed_area` to detect the handedness of rings. This defers to a fast GEOS call when possible.


whophil:
@jorisvandenbossche this is ready for review. Thanks!

jorisvandenbossche (commented):
Thanks for the PR, looks good to me!

There is one potential downside in using `is_ccw` for orient, and that is for invalid geometries. I don't know to what extent we want to ensure those are properly handled. But from reading the GEOS comments about this (https://github.com/shapely/shapely/issues/1315#issuecomment-1037108201), that is a case where the signed area check is more robust. 
I am wondering if it would be worth doing a `is_valid` check, and only then using `is_ccw`, and otherwise the `signed_area` version. It might well be that the combined `is_valid`/`is_ccw` is still faster.

whophil:
As an alternative, geos offers `Algorithm::isCCWArea` (public)
https://github.com/libgeos/geos/blob/13a3bbb015f82c4458fb7768fd24fc6231a11367/src/algorithm/Orientation.cpp#L137-L142

It also seems that geos offers `Area::ofRingSigned` (public), so the Shapely implementation in `shapely.algorithm.cga` could probably just go away entirely if you wanted to use that instead.

jorisvandenbossche:
> As an alternative, geos offers Algorithm::isCCWArea (public)

That's in the C++ API, while we use the (more stable) C API. So if we want to use that, it first requires a change in GEOS to expose this. 

I actually have been thinking it might be nice if GEOS would provide the whole "orient" functionality (but that's a longer term thought, not relevant for this PR now)

jorisvandenbossche:
@whophil sorry for the slow follow-up! I updated it with latest main branch, and limited it to just the signed_area improvement, which we can certainly merge now.

whophil:
Thanks @jorisvandenbossche

PR #1346, opened 2022-03-15
CI: upgrade GEOS 3.10.1 (2021-11-02) to 3.10.2 (2022-01-15)

mwtoews:
Upgrade GEOS used in the CI to [3.10.2 released 2022-01-15](https://lists.osgeo.org/pipermail/geos-devel/2022-January/010638.html)


jorisvandenbossche:
Thanks!

The wheel builds are failing here, because that first needs https://github.com/shapely/shapely/pull/1253 to get merged. So if everything else passes, I think it good to go.

PR #1356, opened 2022-03-31
TST: update tests to pass with numpy 1.22 and pytest 8

jorisvandenbossche:
Numpy 1.22 starts to check the dtype of an array before the dimensionality, and so when passing a 1D object array into a ufunc that expects a 2+D array, we now get a TypeError instead of a ValueError (and this 1D object array comes from passing an inconsistent dimensioned list of coordinates).

We have numpy 1.22 on the build with GEOS main, but the failures were not visible because that build allows failures .. (there are more failures related to GEOS that would need to be fixed first)


jorisvandenbossche:
I opened a follow-up PR to fix some more tests on the GEOS main build: https://github.com/shapely/shapely/pull/1357

Issue #1348, opened 2022-03-18
Wheel distribution for Python 3.10 on macosx x86_64 is missing


kvashchuka:
## Expected behavior and actual behavior.

After installing Shapely using pip, I import `lgeos` from` shapely.geos` on Mac with Intel processor and python version 3.10 (in GitHub Actions and on my machine). This results in the following error:
```
... Python/3.10.2/x64/lib/python3.10/site-packages/shapely/geos.py:54: in load_dll
    raise OSError(
E   OSError: Could not find lib geos_c or load any of its variants ['/Library/Frameworks/GEOS.framework/Versions/Current/GEOS', '/opt/local/lib/libgeos_c.dylib', '/usr/local/lib/libgeos_c.dylib', '/opt/homebrew/lib/libgeos_c.dylib']
```
I believe it happens due to the fact that there is no wheel for Python 3.10 Intel Mac OS on PyPi, and therefore, Shapely is build from source. 

Could you please provide a wheel distribution?

## Steps to reproduce the problem.

```
pip install Shapely
python
import shapely.geometry
```
or
```
pip install Shapely
python
from shapely.geos import lgeos
```

## Operating system

Mac OS 11.6.5

## Shapely version and provenance

1.8.1.post1 installed from PyPI using pip

mwtoews:
Looking at the [files for 1.8.1.post1](https://pypi.org/project/Shapely/1.8.1.post1/#files), there are indeed `*macosx_10_9_x86_64.whl` files for Python 3.6 to 3.9, but not 3.10. Unfortunately I'm not familiar enough with the [build recipe](https://github.com/shapely/shapely/blob/maint-1.8/.github/workflows/release.yml) to understand why Python 3.10 was not built. Perhasp @sgillies knows?

adamjstewart:
I'm guessing https://github.com/shapely/shapely/blob/maint-1.8/.github/workflows/release.yml#L120 is the culprit and explains why cp310-win32 and cp310-manylinux_i686 are also missing. @sgillies any reason why these were excluded?

adamjstewart:
@sgillies? I can submit a PR to remove those skips, but I'm not sure which branch(es) to submit it against. I'm also sure they were there for a reason, so maybe it will cause the tests to fail. Will this require a post2 release or can they be applied to the existing release?

jorisvandenbossche:
> but I'm not sure which branch(es) to submit it against

[...]

PR #1367, opened 2022-04-19
Add Python 3.10 wheels for additional platforms

adamjstewart:
See #1348 for discussion.

If this works on maint-1.8 I'll submit the same PR to main.

@kvashchuka @mwtoews @sgillies @jorisvandenbossche

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -117,7 +117,7 @@ jobs:
         uses: pypa/cibuildwheel@v2.3.1
         env:
           CIBW_ARCHS: ${{ matrix.arch }}
-          CIBW_SKIP: pp* *musllinux*
+          CIBW_SKIP: pp* *musllinux* cp310-manylinux_i686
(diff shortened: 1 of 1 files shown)


adamjstewart:
Yay, everything passes! I'm guessing these skips were simply inadvertently copied from pygeos as @jorisvandenbossche surmised.

mwtoews (approved):
Tests on the new wheels look good!

caspervdw (commented):
Thanks for the PR! I noticed that the `cp310-manylinux_i686` build is actually building numpy from source. I tend to avoid that situation and leave in the skip for that specific platform.

https://github.com/shapely/shapely/runs/6084981342?check_suite_focus=true#step:8:5601

adamjstewart:
Works for me, re-added.

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks @adamjstewart ! 
(do you want to open a PR with the same change applied to the `main` branch? Otherwise I can also get to it)

Issue #1349, opened 2022-03-21
Crash when allocating too much memory


caspervdw:
## Expected behavior and actual behavior.

I expect a MemoryError when attempting to create too many geometries in memory. However, I get a crash:

```
>>> import pygeos
>>> import numpy as np
>>> pygeos.points(*np.random.random((2, 10**9)))
Killed
```
## Operating system

Ubuntu 18.04

## Shapely version and provenance

Pygeos 0.11.1, Python 3.8.10

caspervdw:
I don’t think there is a generic way of checking this beforehand- the Geometry object is an opaque object. 

I would expect some “bad_alloc” being raised by the C++ code that does the geometry construction, and so a NULL returned by the CAPI which leads to a Python exception. Apparently the bad_alloc is not raised before the OS decides to kill the process.

PR #1370, opened 2022-04-24
ENH: Check signals every 10000 ufunc iteration

caspervdw:
This mitigates #1349 by breaking the ufunc inner loop on a SIGINT.

There is really no good way of adjusting the memory allocation of GEOS. One could try to link against a custom `malloc` function, but that will also affect the Python interpreter. This seems a bad idea to me. If we want to control the memory allocation itself, GEOS will need to add custom memory allocators in its CAPI. trac ticket: https://trac.osgeo.org/geos/ticket/540.

Anyway, this PR offers the user to stop the function gracefully by sending a SIGINT (Ctrl+C). Now the numpy loop is stopped whenever a SIGINT is received. Before, the signals were only handled after the function resurfaced into the Python world. 

Test by calling `points` with a big array and then hit (Ctrl+C).

For this to work, we need to acquire the GIL every 10000 iterations. I tried to hit a balance here between the performance hit and usability. 

By the way, there are GEOS CAPI functions for an interrupt callback too: https://github.com/libgeos/geos/blob/5f02e3c4b70d10225c54602f69bb2106d94ae731/capi/geos_c.h.in#L301 . I experimented with these, but in the end they appeared to be not called at all in the points constructor. I guess they are mainly for the more heavy GEOS functions. Breaking the numpy loop instead seemed to be a better (more generic) way for now.


brendan-ward (commented):
Thanks for working on this @caspervdw - I've often wanted something like this.

Possibly out of scope for this PR (which is immediately useful as-is), is the idea of adding a number of iterations to check that varies for different GEOS functions depending on their respective "weight": i.e., use a smaller number like 1,000 for `make_valid` but stick to the default of `10,000` for fast operations like `intersects`.  Could this be a second parameter to this function, with a global default of 10,000?

(untested code)

```C
const int MAX_INTERRUPT_ITER=10000;

#define CHECK_INTERRUPT(I, count)            \
  if (((I + 1) % count) == 0) {        \
    if (PyErr_CheckSignals() == -1) { \
      errstate = PGERR_INTERRUPT;     \
    };                                \
  }



// in ufunc loop for expensive GEOS functions:
CHECK_INTERRUPT(i, 1000);

// or this for most ufuncs
CHECK_INTERRUPT(i, MAX_INTERRUPT_ITER);
```

jorisvandenbossche:
> By the way, there are GEOS CAPI functions for an interrupt callback too: https://github.com/libgeos/geos/blob/5f02e3c4b70d10225c54602f69bb2106d94ae731/capi/geos_c.h.in#L301 . I experimented with these, but in the end they appeared to be not called at all in the points constructor. I guess they are mainly for the more heavy GEOS functions. Breaking the numpy loop instead seemed to be a better (more generic) way for now.

Looking at the GEOS code base, they seem to have a `GEOS_CHECK_FOR_INTERRUPTS` macro, which is indeed only used in certain places (such as overlay ops)

jorisvandenbossche:
This looks like a nice solution. 
I am a little bit worried about the impact of regularly acquiring the GIL again, if you would use this in a multithreaded context (something to check, I can also try to do that)

caspervdw:
Some timing tests with 4 threads:

```
interrupt interval = 10000: 187 ms ± 906 µs
interrupt interval = 10000: 187 ms ± 906 µs
interrupt interval = 10000: 187 ms ± 906 µs

**Comment by caspervdw** [comment]
Some further investigation (see also the commit):

> Python signal handlers are always executed in the main Python thread of the main interpreter, even if the signal was received in another thread. This means that signals can’t be used as a means of inter-thread communication. You can use the synchronization primitives from the [threading](https://docs.python.org/3/library/threading.html#module-threading) module instead.
source: https://docs.python.org/3/library/signal.html#signals-and-threads

Indeed the function I used (`PyErr_CheckSignals`) first checks if the current thread is the main one (and if the current interpreter is the main one) over here: https://github.com/python/cpython/blob/804f2529d8e545a8d59eaf260ee032d014e681ba/Modules/signalmodule.c#L1798

So it only makes sense to check for signals from the main thread. There are however no public C API functions to check for that... I added a draft for that in the latest commit that checks the current thread `ident` against a one that I set when importing `pygeos`, but I am not very enthusiastic about this solution.

Next question is: how to terminate a multithreading script like the one in https://github.com/pygeos/pygeos/issues/164?
Tou have to set this up manually like for instance done using `threading.local` in the dask threadpool scheduler (https://distributed.dask.org/en/stable/_modules/distributed/threadpoolexecutor.html)

**Comment by caspervdw** [comment]
@brendan-ward  @jorisvandenbossche   I have finished this PR, now every ufunc does the signal check and (if it is running on the main thread) can be terminated with Ctrl+C.

The benchmark did not show any noticable performance deterioration. This branch is even a bit faster than the main branch, which is weird. Benchmarks are hard.

this branch:
```
interval = 10     182 ms ± 612 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
interval = 100    182 ms ± 703 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
interval = 1000   182 ms ± 507 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
interval = 10000  181 ms ± 434 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
interval = 100000 182 ms ± 260 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

main branch:
```
190 ms ± 603 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```


Script used:

(see https://github.com/pygeos/pygeos/issues/164 for source for test_parallel)

```
import shapely
import numpy as np

coords = np.random.rand(100_000, 2)

# Running this 4 times but in parallel:
@test_parallel(4)
def g():
    shapely.points(coords)

%timeit g()
```

File: src/geos.h [hunk 1]
```diff
@@ -111,6 +112,8 @@ enum {
       PyErr_WarnFormat(PyExc_Warning, 0,                                                 \
                        "Invalid GeoJSON: geometry is returned as None. %s", last_error); \
       break;                                                                             \
+    case PGERR_PYSIGNAL:                                                                \
```
brendan-ward, line 115: Suggest adding a comment that this error is raised on an interrupt signal does not raise its own exception or warning.

brendan-ward (commented):
This looks great @caspervdw ; I tried it out and it definitely was nice to be able to CTRL-C out of long running loops.

What do you think about adding this to the loops within STRtree (could be a follow-up PR)?  That would probably mean moving the implementation of the signal handler to `geos.c` so it could be used in more places.

Also, what do you think about similar breaking logic within the Cython loops?

shapely/errors.py
```diff
@@ -1,5 +1,35 @@
 """Shapely errors."""
-from shapely.lib import GEOSException, ShapelyError  # NOQA
+import threading
+
+from shapely.lib import _setup_signal_checks, GEOSException, ShapelyError  # NOQA
+
+
+def setup_signal_checks(interval=10000, main_thread_id=None):
```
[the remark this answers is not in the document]
caspervdw: It's not really an error, true, but a signal does lead to an error. I don't see any advantage of moving it to geos.py.

caspervdw:
Thanks @brendan-ward  for your review. I included your suggestions and by dropping Py3.6 the tests now pass. Not sure if this can be included in 2.0 still @jorisvandenbossche  ?

brendan-ward:
@caspervdw raising this again, just because it impacts location of some of the code in this PR:

> What do you think about adding this to the loops within STRtree (could be a follow-up PR)? That would probably mean moving the implementation of the signal handler to geos.c so it could be used in more places.

It's also fine for that to be refactored in a follow-up PR.

caspervdw:
Yes sorry I forgot to reply to that. Absolutely in favour of doing this also in the STRtree. But maybe as a next PR indeed.

shapely/errors.py
```diff
@@ -1,5 +1,35 @@
...
+    Parameters
+    ----------
+    interval : int, default 10000
+        Check for interrupts every x iterations. The higher the number, the slower
+        shapely will respond to a signal. However, at low values there will be a negative effect
+        on performance. The default of 10000 does not have any measureable effects on performance.
+    main_thread_id : int, defaults to ``threading.get_ident``
+        Python signal handlers are always executed in the main Python thread of the main
+        interpreter. Shapely needs to know the main thread id to prevent needless signal
+        checks which would deteriorate performance in multithreading contexts.
```
jorisvandenbossche, line 21: Are there use cases for specifying this manually?

brendan-ward (approved):
Thanks for working on this!

shapely/errors.py
jorisvandenbossche: @caspervdw one question above

caspervdw: I guess things go wrong if someone would `import shapely` for the first time in a subthread AND then try to break a shapely operation in a main thread. I just exposed it here because the C-level function exposes it (and that is because it is easier to get the current thread id from python).

So what do you say, should I remove this parameter?

jorisvandenbossche: > I guess things go wrong if someone would `import shapely` for the first time in a subthread AND then try to break a shapely operation in a main thread.

Do you think it is reasonable in that case to set the id of the main thread manually to avoid that situation? (an easier alternative might be to ensure you import shapely in the main thread?)

> So what do you say, should I remove this parameter?

I was just wondering about a use case. If there are cases were it could be useful, it's certainly fine to keep it.

caspervdw: I replaced the thread-id fetch with `threading.main_thread().ident` and removed the parameter. I think it is good to merge now.

caspervdw:
Test failures with GEOS main branch are unrelated:

```
FAILED tests/test_create_inconsistent_dimensionality.py::test_create_from_wkt[MULTIPOINT (1 1 1, 2 2)-MULTIPOINT Z (1 1 1, 2 2 0)]
FAILED tests/test_create_inconsistent_dimensionality.py::test_create_from_wkt[MULTIPOINT (1 1, 2 2 2)-MULTIPOINT Z (1 1 0, 2 2 2)]
FAILED tests/test_create_inconsistent_dimensionality.py::test_create_from_wkt[LINESTRING (1 1 1, 2 2)-LINESTRING Z (1 1 1, 2 2 0)]
FAILED tests/test_create_inconsistent_dimensionality.py::test_create_from_wkt[POLYGON ((0 0 0, 1 0 0, 1 1, 0 1 0, 0 0 0))-POLYGON Z ((0 0 0, 1 0 0, 1 1 0, 0 1 0, 0 0 0))]
FAILED tests/test_create_inconsistent_dimensionality.py::test_create_from_wkt[LINESTRING (1 1, 2 2 2)-LINESTRING (1 1, 2 2)]
FAILED tests/test_create_inconsistent_dimensionality.py::test_create_from_wkt[POLYGON ((0 0, 1 0 1, 1 1, 0 1, 0 0))-POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0))]
```

jorisvandenbossche (approved):

PR #1373, opened 2022-04-25
Add Python 3.10 wheels for additional platforms

adamjstewart:
Forward-port of #1367 

@jorisvandenbossche

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -109,7 +109,7 @@ jobs:
         uses: pypa/cibuildwheel@v2.3.0
         env:
           CIBW_ARCHS: ${{ matrix.arch }}
-          CIBW_SKIP: pp* *musllinux* cp310-win32 cp310-macosx_x86_64 cp310-manylinux_i686
+          CIBW_SKIP: pp* *musllinux* cp310-manylinux_i686
           CIBW_ENVIRONMENT_LINUX:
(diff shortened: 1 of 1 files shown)


adamjstewart:
Not sure why tests are failing...

mwtoews:
This is the line that causes the failure: https://github.com/shapely/shapely/blob/fc55006ecac8ad7835fe1618d6d391bfe076403c/.github/workflows/release.yml#L135

(arm64 macos-10.15 does not fail due to message "Warning: While arm64 wheels can be built on x86_64, they cannot be tested. The ability to test the arm64 wheels will be added in a future release of cibuildwheel, once Apple Silicon CI runners are widely available.")

This line should probably just ` CIBW_TEST_COMMAND: pytest`

jorisvandenbossche:
Yeah, there is actually a PR that is fixing up the wheel building the main branch (and eg is fixing that pytest invocation): https://github.com/shapely/shapely/pull/1253. I will try to get that merged first.

jorisvandenbossche:
@adamjstewart I merged https://github.com/shapely/shapely/pull/1253, so you should be able to update this.

adamjstewart:
Alright, GitHub Actions now passes! Travis CI is still failing, but it's also failing on main.

jorisvandenbossche (approved):
Thanks!

Issue #975, opened 2020-08-25
API: geom_type vs type vs GeometryType()


jorisvandenbossche:
To get the type of a geometry object, there are currently 3 method/properties that are aliases of each other:

https://github.com/shapely/shapely/blob/d1ad37cab7c58ffd0a9f96f1f2f57e4a522f6356/shapely/geometry/base.py#L364-L369

https://github.com/shapely/shapely/blob/d1ad37cab7c58ffd0a9f96f1f2f57e4a522f6356/shapely/geometry/base.py#L438-L441

Is there appetite for reducing the number of options, and deprecate some of those? (given the deprecations in 1.8 / changes in 2.0, it might be a good time to clean this up as well)

If we choose one, I would go with `geom_type` (that is the only one that is documented in manual.rst). It seems that `type` is used quite a bit in our own codebase, though. But at least `GeometryType` is never used/documented, and also not following PEP8.

mwtoews:
I did a bit of code archaeology:
* `geom_type` was part of the first commit at 1fda2de1df1cef55b6c24bf009c81030643dde68 and is used extensively internally and publicly, so should be the primary property for the geometry type.
* `geometryType()` was also part of the first commit (1fda2de1df1cef55b6c24bf009c81030643dde68) but I agree that this function is an outlier with PEP 8 and should be deprecated with 1.8, and I'd support removing for 2.0. This function is rarely used internally, and I doubt has much public use.
* `type` was added at 84a06cef3e8a3e5343f1ca1c5a55b2774b1d8036 for point and e9aaf049a6b847aae5685f31bbb5a7b8e8c1110b for polygon (back when only two geometry types existed; this property was later moved to base.py). The comments suggest that `type` and `coordinates` were to support "Python feature protocol" or I assume [PEP 544](https://www.python.org/dev/peps/pep-0544/). I don't know anything about this PEP, but I'm inclined to support that this property should be deprecated for version 1.8.

jorisvandenbossche:
Thanks for the research!

> The comments suggest that type and coordinates were to support "Python feature protocol" or I assume PEP 544.

Since that PEP is 10 years older than the commit, I don't think it is related ;) Maybe the comment pointed to a precursor of the protocol for geospatial data (https://gist.github.com/sgillies/2217756)

mwtoews:
Now that `.geometryType()` is deprecated, is there any further appetite to deprecate `.type`? Or is it harmless to have two properties that serve the same purpose?

[...]

PR #1375, opened 2022-04-30
DEPR: deprecate the GeometryType attribute

jorisvandenbossche:
Part of https://github.com/shapely/shapely/issues/975. This PR already deprecates the `GeometryType()` method, which isn't used internally or used in the documentation.


mwtoews (approved):
Perfect, thanks! I don't recall seeing this anywhere else either, so it probably won't be missed.

PR #1377, opened 2022-05-01
DOC: Move pygeos doc pages into shapely docs

jorisvandenbossche:
This PR starts merging the docs. For now, it "just" moves the actual content into the `/docs` directory, with some minimal additions to conf.py (things that were done differently in pygeos). The docs don't yet build correctly (eg the content has still many references to pygeos etc), but the current doc build is failing on readthedocs anyway, so that doesn't change.


jorisvandenbossche:
Since this is just moving some content (and not yet actually merging / rewriting anything), going to merge this

PR #1378, opened 2022-05-01
Consolidate README files

jorisvandenbossche:
This PR tries to merge the two README files, basically moving some content from the old PyGEOS README file into the existing Shapely one (and removing the old pygeos one).


README.rst
```diff
@@ -2,55 +2,133 @@
...
+
 .. |coveralls| image:: https://coveralls.io/repos/github/shapely/shapely/badge.svg?branch=main
    :target: https://coveralls.io/github/shapely/shapely?branch=main
 
+.. Zenodo
+
+.. .. image:: https://zenodo.org/badge/191151963.svg
+..   :alt: Zenodo
+..   :target: https://zenodo.org/badge/latestdoi/191151963
+
```
mwtoews, line 45: This DOI is for pygeos. I've never set these up, but we'd probably need to set one up for Shapely.

README.rst
```diff
@@ -2,55 +2,132 @@
...
-* `Shapely GitHub repository <https://github.com/shapely/shapely>`__
-* `Shapely documentation and manual <https://shapely.readthedocs.io/en/latest/>`__
+planar geometric objects. It is using the widely deployed open-source
+geometry library `GEOS <https://libgeos.org/>`__ (the engine of `PostGIS
+<https://postgis.net/>`__, and a port of `JTS <https://locationtech.github.io/jts/>`__).
+Shapely wraps the GEOS operations in NumPy ufuncs providing a performance
+improvement when operating on arrays of geometries, as well as provides a
+rich scalar `Geometry` interface.
+Shapely is not concerned with data formats or coordinate systems, but can be
+readily integrated with packages that are.
```
jorisvandenbossche, line 60: This is the main part that is rewritten a little

README.rst
```diff
@@ -2,55 +2,132 @@
...
+diminishing the overhead of the Python interpreter.
+
+Multithreading
+--------------
+
+PyGEOS functions support multithreading. More specifically, the Global
+Interpreter Lock (GIL) is released during function execution. Normally in Python, the
+GIL prevents multiple threads from computing at the same time. PyGEOS functions
+internally releases this constraint so that the heavy lifting done by GEOS can be
+done in parallel, from a single Python process.
```
jorisvandenbossche, line 77: For Shapely, those two paragraphs can probably be moved to the actual documentation instead of being in the README (but that can be done later when reworking the docs, currently the README gets inserted in the docs anyway)

README.rst
jorisvandenbossche: Yeah, indeed (that's the reason that I left it commented out for now). Thanks for opening the issue about it.

README.rst
```diff
@@ -2,55 +2,132 @@
...
+           [ 80.,  72.,  64.,  56.,  48.],
+           [ 70.,  63.,  56.,  49.,  42.],
+           [ 60.,  54.,  48.,  42.,  36.]])
+
+See the documentation for more examples and guidance: https://shapely.readthedocs.io
 
 Requirements
 ============
 
 Shapely 1.8 requires
```
mwtoews, line 127:
```suggestion
Shapely 2.0 requires
```

README.rst
```diff
@@ -2,55 +2,132 @@
...
-    >>> from shapely.geometry import Point
-    >>> patch = Point(0.0, 0.0).buffer(10.0)
+    >>> import shapely
+    >>> patch = shapely.Point(0.0, 0.0).buffer(10.0)
     >>> patch
-    <shapely.geometry.polygon.Polygon object at 0x...>
+    <shapely.Polygon POLYGON ((10 0, 9.952 -0.98, 9.808 -1.951, 9.569 -2.903, 9....>
     >>> patch.area
-    313.65484905459385
+    313.6548490545941
```
mwtoews, line 92: Just to note, this is a change with geos, not shapely. Probably related to [geos-trac-743](https://trac.osgeo.org/geos/ticket/743).

README.rst
```diff
@@ -2,55 +2,132 @@
...
+improvement when operating on arrays of geometries, as well as provides a
+rich scalar `Geometry` interface.
+Shapely is not concerned with data formats or coordinate systems, but can be
+readily integrated with packages that are.
+
+What is a ufunc?
+----------------
+
+A universal function (or ufunc for short) is a function that operates on
+n-dimensional arrays in an element-by-element fashion, supporting array
```
mwtoews, line 66:
```suggestion
*n*-dimensional arrays in an element-by-element fashion, supporting array
```

mwtoews (commented):
Looks good so far. A few other details that could seek improvement:

- Replace "OS X" with "macOS"
- Replace "Numpy" with "NumPy" for consistency
- The "Development and Testing" section replace [`virtualenv`](https://virtualenv.pypa.io/en/latest/) (primarily a Python 2 tool) with the [`venv` module](https://docs.python.org/3/library/venv.html)

README.rst
jorisvandenbossche: Yep, thanks for the link, I just reran the example with shapely main and latest released GEOS (to update the repr)

jorisvandenbossche:
@mwtoews Thanks for the review!

> The "Development and Testing" section replace [`virtualenv`](https://virtualenv.pypa.io/en/latest/) (primarily a Python 2 tool) with the [`venv` module](https://docs.python.org/3/library/venv.html)

Is that a 1:1 replacement in that code snippet? (I don't use those myself) 
I am planning to update the "development" section anyway later on, to combine it with the instructions from pygeos (and I would also move this out of the README to the docs), so can also update this later when doing that.

mwtoews:
the 1:1 replacement is:
```diff
- $ virtualenv .
+ $ python3 -m venv .
```
(although I tend to used burner directories like `python3 -m venv /tmp/py38` then `source /tmp/py38/bin/activate`).

See further info at the [`venv` docs](https://docs.python.org/3/library/venv.html).

jorisvandenbossche:
@mwtoews I updated this virtualenv -> venv in https://github.com/shapely/shapely/pull/1396

README.rst
```diff
@@ -2,55 +2,132 @@
...
+*n*-dimensional arrays in an element-by-element fashion, supporting array
+broadcasting. The for-loops that are involved are fully implemented in C
+diminishing the overhead of the Python interpreter.
+
+Multithreading
+--------------
+
+PyGEOS functions support multithreading. More specifically, the Global
+Interpreter Lock (GIL) is released during function execution. Normally in Python, the
+GIL prevents multiple threads from computing at the same time. PyGEOS functions
```
brendan-ward, line 75:
```suggestion
GIL prevents multiple threads from computing at the same time. Shapely functions
```

README.rst
```diff
@@ -2,55 +2,132 @@
...
+internally releases this constraint so that the heavy lifting done by GEOS can be
```
brendan-ward, line 76:
```suggestion
internally release this constraint so that the heavy lifting done by GEOS can be
```

README.rst
```diff
@@ -2,55 +2,132 @@
...
 
 Here is the canonical example of building an approximately circular patch by
-buffering a point.
+buffering a point, using the scalar Geometry interface:
 
 .. code-block:: pycon
 
-    >>> from shapely.geometry import Point
-    >>> patch = Point(0.0, 0.0).buffer(10.0)
+    >>> import shapely
```
brendan-ward, line 87:
```suggestion
    >>> from shapely import Point
```

(better matches examples in #1391)

README.rst
```diff
@@ -2,55 +2,132 @@
...
+    <shapely.Polygon POLYGON ((10 0, 9.952 -0.98, 9.808 -1.951, 9.569 -2.903, 9....>
     >>> patch.area
-    313.65484905459385
+    313.6548490545941
+
+Using the vectorized ufunc interface, compare a grid of points with a polygon:
+
+.. code:: python
+
+    >>> geoms = shapely.points(*np.indices((4, 4)))
```
brendan-ward, line 98: This is a little hard to visualize; even though it is a little long, it might be helpful to print out `geoms`

README.rst
```diff
@@ -2,55 +2,132 @@
...
+           [False,  True, False, False],
+           [False, False, False, False],
+           [False, False, False, False]])
+
+Compute the area of all possible intersections of two lists of polygons:
 
-See the manual for more examples and guidance.
+.. code:: python
+
+    >>> polygons_x = shapely.box(range(5), 0, range(10, 15), 10)
```
brendan-ward, line 111: Likewise, these are a bit hard to visualize from the function calls.  Suggest using a smaller set of combinations, and printing them out.

brendan-ward (commented):
A few text edits to consider.  The only bits I'm not yet sure about are the vectorized examples; I'm finding them hard to visualize from the function calls, so I think we may want to simplify those a bit and print out the inputs to make it easier to understand what is going on, esp. for folks that are new to the vectorized APIs.

jorisvandenbossche:
@brendan-ward thanks for the review! Fully agreed that the vectorized examples are hard to follow if you don't already have good knowledge numpy and broadcasting. I think in general we should also add a page in the documentation going into more detail on those topics, but will already do a pass to try to improve those examples here.

jorisvandenbossche:
I drastically simplified the example now. I think the example using reshaping trick and broadcasting (to compute intersection of all combinations) is something we move to the user guide.

README.rst
```diff
@@ -91,33 +91,18 @@ buffering a point, using the scalar Geometry interface:
...
-Using the vectorized ufunc interface, compare a grid of points with a polygon:
+Using the vectorized ufunc interface (instead of using a manual for loop),
+compare an array of points with a polygon:
 
 .. code:: python
 
-    >>> geoms = shapely.points(*np.indices((4, 4)))
+    >>> import shapely
+    >>> import numpy as np
+    >>> geoms = np.array([Point(0, 0), Point(1, 1), Point(2, 2)])
```
jorisvandenbossche, line 101: I changed the array construction to a manual array instead of the `shapely.points` constructor, as I think this example should focus on showing the vectorized contains (and that way it is also more visual already what the array looks like)

README.rst
mwtoews: xref #1455 as a starting point

PR #1382, opened 2022-05-04
CI: only build Travis on the main branch (not PRs)

jorisvandenbossche:
Given that we have only limited "credits" for Travis, we can best limit the number of builds, and for example not build on _every_ commit in a PR.

Using the conditions from https://docs.travis-ci.com/user/conditions-v1

Will additionally check if we can still run it when we edit eg the build configuration (as we detect with github actions)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ dist: focal
 language: python
 python: '3.8'
 
+if: branch = main OR tag IS present
+
 env:
   global:
   - GEOS_VERSION=3.10.1
@@ -27,13 +29,6 @@ jobs:
     - CIBW_BEFORE_ALL="./ci/install_geos.sh"
     - CIBW_TEST_REQUIRES="pytest"
     - CIBW_TEST_COMMAND="pytest --pyargs shapely.tests"
-
-
-branches:
-  only:
-  - main
-  - travis
-  - /^\d+\.\d+(\.\d+)?(-\S*)?$/
 
 install:
 - |
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ dist: focal
 language: python
 python: '3.8'
 
-if: branch = main OR tag IS present
+if: (branch = main OR tag IS present) AND (type = push)
 
 env:
   global:


.travis.yml
```diff
@@ -3,6 +3,8 @@ dist: focal
 language: python
 python: '3.8'
 
+if: (branch = main OR tag IS present) AND (type = push)
```
jorisvandenbossche, line 6: This should also allow to easily change this to `type = pull_request` when working on a PR that affects Travis

mwtoews: What does `tag IS present` mean?

mwtoews:
Is there a way to trigger this with a commit message, e.g. "CI: fix something [travis]"? (I'm thinking of the days when there was "[ci skip]" to do the opposite.)

.travis.yml
jorisvandenbossche: My understanding is that it will then also run on tagged commits. 

It's not fully clear to me whether it is important to include this, as the commit we tag for a release will typically also be on the main branch (and if we at some point have a release branch, we should include that branch here). Although based on https://docs.travis-ci.com/user/customizing-the-build/#safelisting-or-blocklisting-branches, it might be needed.

jorisvandenbossche:
> Is there a way to trigger this with PRs using a commit message, e.g. "CI: fix something [travis]"? (I'm thinking of the days when there was "[ci skip]" to do the opposite.)

Not that I found (only the skipping when it is enabled, but not the other way around). 

But so as far as I know, there are two ways to trigger Travis:

- In the PR where you want to check this, change travis.yml to inlude "pull_request" (see my comment above)
- In the web interface (https://app.travis-ci.com/github/shapely/shapely), builds can be triggered for any branch. Although for a PR that means that you would need to do the PR using a branch in the main repo and not from your fork (but it means we can trigger a build on the main branch eg to ensure it is passing before tagging)

No idea if you can still enable travis on your fork, and have some initial credits

mwtoews (approved):
This a sensible strategy to conserve Travis CI credits. Thanks!

PR #1388, opened 2022-05-13
CI: fix tests with latest setuptools

jorisvandenbossche:
Tests are failing on the main branch. It might be that setuptools broke something (that's the only thing that I notice that is different with the passing build of 4 days ago, a version update of setuptools of 62.1 -> 62.2), although I don't directly see anything mentioned in their changelog that would explain this (https://setuptools.pypa.io/en/latest/history.html#v62-2-0). It seems that it no longer ignores the packages listed in pyproject.toml, even when directly calling `python setup.py build_ext`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -143,7 +143,7 @@ jobs:
 
       - name: Build and install Shapely
         run: |
-          pip install .
+          pip install -e .
 
       - name: Overview of the Python environment (pip list)
         run: pip list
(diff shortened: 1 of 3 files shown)


mwtoews (approved):
Good fix!

On a related note, see blog "[Why you shouldn't invoke setup.py directly](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html)" (although it does not specifically discuss `build_ext`).

PR #1389, opened 2022-05-13
CLN: remove unused vendored packaging, consolidate gitignore, remove old pygeos pyproject.toml

jorisvandenbossche:
Various clean ups after the pygeos merge (the vendored packaging was only used inside the old shapely setup.py file, but that is no longer used and is being removed in https://github.com/shapely/shapely/pull/1376)


mwtoews (approved):
Thanks, looks good.

PR #1391, opened 2022-05-16
DOC: update examples in manual.rst

jorisvandenbossche:
This updates the examples in manual.rst: it both updates it for some behavioural changes, as well as the changed output repr.

With those changes, the doctests pass locally for this file (still need to set it up on CI)


docs/manual.rst
```diff
@@ -604,32 +604,12 @@ Figure 5. a) a green and a yellow line that intersect along a line and at a
...
 .. code-block:: pycon
 
-  >>> pprint(list(x.geoms))
-  [<shapely.geometry.point.Point object at 0x...>,
-   <shapely.geometry.linestring.LineString object at 0x...>]
-  >>> pprint(list(x))
-  [<shapely.geometry.point.Point object at 0x...>,
-   <shapely.geometry.linestring.LineString object at 0x...>]
-
-Collections can also be sliced.
```
jorisvandenbossche, line 619: This was deprecated in 1.8 and no longer works, so removed that example

docs/manual.rst
```diff
@@ -1290,21 +1250,18 @@ Shapely's binary :meth:`within` predicate.
...
 
 As the howto says, the `less than` comparison is guaranteed to be used in
-sorting. That's what we'll rely on to spatially sort, and the reason why we use
-:meth:`within` in reverse instead of :meth:`contains`. Trying it out on features
+sorting. That's what we'll rely on to spatially sort. Trying it out on features
 `d` and `c`, we see that it works.
 
 .. code-block:: pycon
 
-  >>> d < c
```
jorisvandenbossche, line 1306: Such comparison operator as shortcut for contains also didn't work in Shapely 1.7 or 1.8, not sure if that ever worked in the past

docs/manual.rst
```diff
@@ -1530,9 +1483,9 @@ boundaries.
 .. code-block:: pycon
 
   >>> a.union(b).boundary
-  <shapely.geometry.polygon.LinearRing object at 0x...>
+  <shapely.LineString LINESTRING (1.435 -0.435, 1.293 -0.471, 1.147 -0.493, 1 ...>
```
jorisvandenbossche, line 1486: Also in Shapely 1.7/1.8, the `boundary` already returned a LineString instead of LinearRing

brendan-ward (approved):
The new `__repr__` certainly looks nicer here.

There is a bit of inconsistency between the raw `__repr__` version used for some of the earlier examples (as before), whereas the later examples use <geom>.wkt to print WKT (e.g., STRtree examples).  This isn't necessarily a problem, but might be good to be more consistent where appropriate.



I wonder, though, if now the `__repr__` is a bit rendundant, given that it include the specific geometry type class as well as the WKT type, e.g.,

```Python
<shapely.MultiPoint MULTIPOINT (0 0, 1 1)>
```

Compared to in pygeos it was 

```
<pygeos.Geometry MULTIPOINT (0 0, 1 1)>
```

Which seemed less redundant.

I wonder if 

```
<shapely.MultiPoint (0 0, 1 1)>
```

Would be more terse, since the class name is the same as the WKT type except for casing (though this doesn't hold up for Z types, so bears more thinking through).

jorisvandenbossche:
> There is a bit of inconsistency between the raw `__repr__` version used for some of the earlier examples (as before), whereas the later examples use .wkt to print WKT

I cleaned this up a bit, now preferring to use the repr directly in most cases, instead of using .wkt to show the result (basically .wkt is only useful now if the WKT is longer and you want to show the full output)

mwtoews:
Should "Migrating to Shapely 1.8 / 2.0" (`docs/migration.rst`) be part of the 2.0 manual, or be removed? It shows example object outputs for 1.8, which is a bit awkward for a manual for later releases. Also it uses deprecated `from distutils.version import LooseVersion`, so it's already outdated. My sense is to remove the document from main.

Note that this document will remain at the RTD version for `maint-1.8` (https://shapely.readthedocs.io/en/maint-1.8/migration.html), where it could be improved to use `from packaging.version import Version`, as done with geopandas.

mwtoews:
Other examples that could be updated in this PR are in README.rst (search for "object at").

jorisvandenbossche:
> Should "Migrating to Shapely 1.8 / 2.0" (`docs/migration.rst`) be part of the 2.0 manual, or be removed?

Good question. I personally think it is useful to keep (it still contains relevant details on differences between 1.7 vs 1.8 vs 2.0, and how to update your code; many people reading the docs in the coming months will still be using an older version). But of course it is also true that they still exist in the older version docs, and we can link to that (although that makes them less discoverable IMO).

Now, I am only updating manual.rst examples here in this PR, and not touching anything else from the docs, so maybe we can discuss this further in a separate issue?

> Also it uses deprecated `from distutils.version import LooseVersion`, so it's already outdated.

Good catch, that's indeed something we can update in the docs (whether we keep it or not on main, we can still do that for 1.8.x)

jorisvandenbossche:
> Other examples that could be updated in this PR are in README.rst (search for "object at").

I indeed missed one example there in a previous PR, thanks! -> https://github.com/shapely/shapely/pull/1452

Issue #1276, opened 2021-12-28
Shapely 2.0 naming inconsistencies to resolve


jorisvandenbossche:
What follows is an overview of some (naming) inconsistencies we currently have between the vectorized functions (moved from pygeos) and existing shapely geometry methods or functions. 
We can use this as a general issue, but can also open separate issue to discuss some of them in more detail if needed. 

(the naming inconsistency is listed as "original shapely name" *vs* "vectorized function name (moved from pygeos)")

**Unary union:**

* `unary_union` vs `union_all`

Unary union is the typical way this is called (also in PostGIS), and this is an often used attribute. But, we also have `intersection_all` and `symmetric_difference_all`, for which the `_all` naming scheme probably makes more sense than adding `unary_`. For this reason, we could keep both `unary_union` and `union_all` as aliases?

This is both an attribute on the geometry objects, as a function in the ops.py submodule.

PR -> https://github.com/shapely/shapely/pull/1536

**Linear referencing related methods**

* `interpolate` vs `line_interpolate_point`
* `project` vs `line_locate_point`

PostGIS also uses the [`ST_LineInterpolatePoint`](https://postgis.net/docs/ST_LineInterpolatePoint.html) and [`ST_LineLocatePoint`](https://postgis.net/docs/ST_LineLocatePoint.html) terminology, while the `interpolate` and `project` naming follows GEOS.

Those are methods on the geometry objects

**Other methods on the geometry class:**

* `parallel_offset` vs `offset_curve` (method only on LineString)
* `representative_point` vs `point_on_surface`

PostGIS has [`ST_OffsetCurve`](https://postgis.net/docs/ST_OffsetCurve.html), and also GEOS uses this naming scheme (`GEOSOffsetCurve_r`).\
Similar for [`ST_PointOnSurface`](https://postgis.net/docs/ST_PointOnSurface.html) / `GEOSPointOnSurface_r`

**Buffer resolution keyword**

* `resolution` vs `quadsegs`

In Shapely, the `quadsegs` keyword was deprecated in favor of `resolution` at some point (https://github.com/shapely/shapely/commit/5689abfa81dc38d882e9e8140fc81bbc23d3b10a), but the vectorized function (moved from pygeos) is using `quadsegs` (which maps to the name used in GEOS and more or less in PostGIS). @sgillies do you remember the reason for changing it to `resolution`? (to have a more readable name?)

**Simplify preserve_topology default**

[...]

PR #1392, opened 2022-05-22
Change simplify() preserve_topology default to True (match the Geometry method)

jorisvandenbossche:
xref https://github.com/shapely/shapely/issues/1276

The Geometry class' `simplify()` method has a default of `preserve_topology=True`, while the vectorized function from pygeos was using `True`. But I think we have been thinking in the past we should probably move to `True` anyway (to return valid geometries by default, eg https://github.com/pygeos/pygeos/issues/233#issuecomment-943856854), and that also ensures that the function and geometry class method use a consistent default.

### DIFF
--- a/shapely/constructive.py
+++ b/shapely/constructive.py
@@ -709,7 +709,7 @@ def segmentize(geometry, tolerance, **kwargs):
 
 
 @multithreading_enabled
-def simplify(geometry, tolerance, preserve_topology=False, **kwargs):
+def simplify(geometry, tolerance, preserve_topology=True, **kwargs):
     """Returns a simplified version of an input geometry using the
     Douglas-Peucker algorithm.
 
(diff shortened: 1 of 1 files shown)


brendan-ward (approved):

caspervdw (approved):

mwtoews:
I think in the future, there is scope to allow more than two simplify methods, as there are multiple "simplify" algorithms out there (published or coded in some form under a liberal licence). Currently, the `preserve_topology` bool keyword just enables two methods, but a future-proof way is to have a `method` keyword instead (base on an enum or str).

jorisvandenbossche:
@mwtoews that's a good point (eg GEOS 3.11 will have a PolygonHullSimplify method, so we can (in another issue) start discussing whether to expose this under `simplify` with a `method` keyword, or as a separate function/method).

Regardless of that, I think we still want to switch the default anyway (and we need to keep the existing keyword for some time for backwards compatibility), and so this PR should still be fine as is for now?

mwtoews (approved):
Looks good. We can discuss changes to support more than two algorithms later.

PR #1393, opened 2022-05-22
Rename shapely.apply to shapely.transform

jorisvandenbossche:
xref https://github.com/shapely/shapely/issues/1276 about naming inconsistencies

Shapely uses "transform" for what is called `apply` for the vectorized functions (merged from pygeos). This PR renames `apply()` to `transform()`.

I personally think both names are fine, so this change keeps it a bit more consistent in shapely. 

(I might actually even like "transform" a bit more, as it signals more that the geometry coordinates are only transformed, keeping the exact structure. While applying a function in general could return something else as geometries. For example, in pandas, there is a more generic "apply" (that can basically return anything) vs more specific "transform" (which is guaranteed to keep the same shaped output))


mwtoews (approved):
Yes to "transform" for this type of operation.

A new GEOS 3.11 feature is [`GEOSGeom_transformXY`](https://github.com/libgeos/geos/blob/2a82afc1ac2327966f56bdf35a6d88aaadb81330/capi/geos_c.h.in#L3846-L3861) (and [`GEOSTransformXYCallback`](https://github.com/libgeos/geos/blob/2a82afc1ac2327966f56bdf35a6d88aaadb81330/capi/geos_c.h.in#L278-L293)). This functionality has not yet made it into Shapely yet, although I can see that it would be useful optimization for "(N, 2)->(N, 2)" cases within `shapely.coordinates.transform()`.

brendan-ward (approved):
+1 on the name change

sgillies (approved):

shapely/coordinates.py
```diff
@@ -3,10 +3,10 @@
 from . import Geometry  # NOQA
 from . import lib
 
-__all__ = ["apply", "count_coordinates", "get_coordinates", "set_coordinates"]
+__all__ = ["transform", "count_coordinates", "get_coordinates", "set_coordinates"]
 
 
-def apply(geometry, transformation, include_z=False):
+def transform(geometry, transformation, include_z=False):
```
caspervdw, line 9: The docstring should also be adjusted

caspervdw (changes requested):

jorisvandenbossche:
> And maybe also change the names all the way down to C?

The method currently only lives on the python side (the actual C funcs being called here are `get_coordinates` and `set_coordinates`, `apply`/`transform` being a python wrapper around those with calling the transform func in the middle)

shapely/coordinates.py
jorisvandenbossche: Updated the docstring

PR #1394, opened 2022-05-22
DOC: replace pygeos -> shapely in the reference docs

jorisvandenbossche:
This already allows to build the docs and all those autodoc reference pages (I still get a bunch of warnings to resolve though).


jorisvandenbossche:
Going to merge this one, since it is a trivial search/replace.

PR #1395, opened 2022-05-22
DOC: update class (init/new) docstrings of Geometry subclasses

jorisvandenbossche:
This makes a few changes to the main docstrings of the Geometry subclass objects:

- "Example" -> "Examples" to follow numpydoc convention
- Combined the class docstring with the `__new__` docstring. I think this is more useful for users (interactively, you will typically only check the class docstring (`Polygon?` in IPython)).
- Some small formatting changes, and ensuring everything builds fine with sphinx.

### DIFF
--- a/shapely/geometry/linestring.py
+++ b/shapely/geometry/linestring.py
@@ -10,17 +10,18 @@ __all__ = ["LineString"]
 
 class LineString(BaseGeometry):
     """
-    A one-dimensional geometry type composed of one or more line segments.
+    A geometry type composed of one or more line segments.
 
-    A LineString has non-zero length and zero area. It may approximate a curve
-    and need not be straight. Unlike a LinearRing, a LineString is not closed.
+    A LineString is a one-dimensional feature and has a non-zero length but
+    zero area. It may approximate a curve and need not be straight. Unlike a
+    LinearRing, a LineString is not closed.
 
     Parameters
     ----------
     coordinates : sequence
-        A sequence of (x, y [,z]) numeric coordinate pairs or triples, or
-        an object that provides the numpy array interface, including
-        another instance of LineString.
+        A sequence of (x, y, [,z]) numeric coordinate pairs or triples, or
+        an array-like with shape (N, 2) or (N, 3).
+        Also can be a sequence of Point objects.
 
     Examples
     --------
--- a/shapely/geometry/linestring.py
+++ b/shapely/geometry/linestring.py
@@ -10,31 +10,30 @@ __all__ = ["LineString"]
 
 class LineString(BaseGeometry):
     """
-    A one-dimensional figure comprising one or more line segments
+    A one-dimensional figure comprising one or more line segments.
 
     A LineString has non-zero length and zero area. It may approximate a curve
     and need not be straight. Unlike a LinearRing, a LineString is not closed.
+
+    Parameters
+    ----------
+    coordinates : sequence
+        A sequence of (x, y [,z]) numeric coordinate pairs or triples or
+        an object that provides the numpy array interface, including
+        another instance of LineString.
+
+    Examples
+    --------
+    Create a line with two segments
+
+    >>> a = LineString([[0, 0], [1, 0], [1, 1]])
+    >>> a.length
+    2.0
     """
 
     __slots__ = []
 
     def __new__(self, coordinates=None):
-        """
-        Parameters
-        ----------
-        coordinates : sequence
-            A sequence of (x, y [,z]) numeric coordinate pairs or triples or
-            an object that provides the numpy array interface, including
-            another instance of LineString.
-
-        Example
-        -------
-        Create a line with two segments
-
-          >>> a = LineString([[0, 0], [1, 0], [1, 1]])
-          >>> a.length
-          2.0
-        """
         if coordinates is None:
             # empty geometry
             # TODO better constructor
--- a/shapely/geometry/linestring.py
+++ b/shapely/geometry/linestring.py
@@ -10,7 +10,7 @@ __all__ = ["LineString"]
 
 class LineString(BaseGeometry):
     """
-    A one-dimensional figure comprising one or more line segments.
+    A one-dimensional geometry type composed of one or more line segments.
 
     A LineString has non-zero length and zero area. It may approximate a curve
     and need not be straight. Unlike a LinearRing, a LineString is not closed.
@@ -18,7 +18,7 @@ class LineString(BaseGeometry):
     Parameters
     ----------
     coordinates : sequence
-        A sequence of (x, y [,z]) numeric coordinate pairs or triples or
+        A sequence of (x, y [,z]) numeric coordinate pairs or triples, or
         an object that provides the numpy array interface, including
         another instance of LineString.
 
--- a/shapely/geometry/linestring.py
+++ b/shapely/geometry/linestring.py
@@ -25,7 +25,7 @@ class LineString(BaseGeometry):
 
     Examples
     --------
-    Create a line with two segments
+    Create a LineString with two segments
 
     >>> a = LineString([[0, 0], [1, 0], [1, 1]])
     >>> a.length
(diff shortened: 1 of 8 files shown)


shapely/geometry/polygon.py
```diff
@@ -24,35 +24,39 @@ def _unpickle_linearring(wkb):
 
 class LinearRing(LineString):
     """
-    A closed one-dimensional feature comprising one or more line segments
+    A closed one-dimensional feature comprising one or more line segments.
```
brendan-ward, line 27:
```suggestion
    A geometry type composed of one or more line segments that forms a closed loop (a line segment exists between the last and first coordinate).
```

shapely/geometry/polygon.py
```diff
@@ -24,35 +24,39 @@ def _unpickle_linearring(wkb):
...
 
     A LinearRing that crosses itself or touches itself at a single point is
     invalid and operations on it may fail.
+
+    Parameters
+    ----------
+    coordinates : sequence
+        A sequence of (x, y [,z]) numeric coordinate pairs or triples.
+        Also can be a sequence of Point objects.
```
brendan-ward, line 36: Consistency check: I'm assuming this can also come from something that provides the numpy array interface with appropriate shape.

shapely/geometry/polygon.py
```diff
@@ -24,35 +24,39 @@ def _unpickle_linearring(wkb):
...
+
+    Notes
+    -----
+    Rings are implicitly closed. There is no need to specify a final
```
brendan-ward, line 40:
```suggestion
    Rings are automatically closed. There is no need to specify a final
```

shapely/geometry/polygon.py
```diff
@@ -24,35 +24,39 @@ def _unpickle_linearring(wkb):
...
+    --------
+    Construct a square ring.
+
+    >>> ring = LinearRing( ((0, 0), (0, 1), (1 ,1 ), (1 , 0)) )
+    >>> ring.is_closed
+    True
+    >>> list(ring.coords)
+    [(0.0, 0.0), (0.0, 1.0), (1.0, 1.0), (1.0, 0.0), (0.0, 0.0)]
+    >>> ring.length
+    4.0
```
brendan-ward, line 53: shouldn't lengths always be whole integers?

shapely/geometry/polygon.py
```diff
@@ -172,42 +176,41 @@ def _get_ring(self, i):
 
 class Polygon(BaseGeometry):
     """
-    A two-dimensional figure bounded by a linear ring
+    A two-dimensional figure bounded by a linear ring.
```
brendan-ward, line 179:
```suggestion
    A geometry type representing an area that is enclosed by a linear ring.
```

brendan-ward (commented):
Thanks for working on this @jorisvandenbossche , the expanded docstrings are helpful.

A few suggestions to consider that attempt to make the terms more accessible to newcomers (i.e., using dimensionality to define geometry types is maybe a little less clear if you aren't familiar with the mathmatical interpretation), while hopefully not compromising too much on precision of terms.

Note: comments made on the first few geometry types are implicit for subsequent ones to avoid repeating the same type of comment...

shapely/geometry/linestring.py
```diff
@@ -10,31 +10,30 @@
...
 
     A LineString has non-zero length and zero area. It may approximate a curve
     and need not be straight. Unlike a LinearRing, a LineString is not closed.
+
+    Parameters
+    ----------
+    coordinates : sequence
+        A sequence of (x, y [,z]) numeric coordinate pairs or triples or
+        an object that provides the numpy array interface, including
+        another instance of LineString.
```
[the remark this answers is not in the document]
jorisvandenbossche: Ah, that's actually something to update. With 2.0, LineStrings don't directly implement the array interface (only its `.coords` do), so the "including" is no longer correctly phrased.

In general, all those subclass constructors accept an instance of itself (I don't know how important that is to mention every time). So it's not a "sequence of LineStrings" (that would be for the MultiLineString constructor).

shapely/geometry/polygon.py
jorisvandenbossche: I left out the "(a line segment exists between the last and first coordinate).", because a LinearRing actually has a start and end coordinate pair that is identical (potentially automatically added, but once constructed, the closing coordinate pair is included in the representation), so it's not an implicit line segment between first and last coordinate pair.

shapely/geometry/polygon.py
jorisvandenbossche: Indeed, the same as for a LineString. I made their docstring entries for this parameter match.

shapely/geometry/polygon.py
jorisvandenbossche: This is a bit confusing, but so it's the number of line segments, but the actual length of the line, and that is correctly a float.

jorisvandenbossche: I don't think there is actually a direct property on the class that gives this. There is the new top-level `shapely.get_num_coordinates` function, or one could do `len(ring.coords)`

jorisvandenbossche:
Thanks a lot for the review @brendan-ward !

> A few suggestions to consider that attempt to make the terms more accessible to newcomers (i.e., using dimensionality to define geometry types is maybe a little less clear if you aren't familiar with the mathmatical interpretation), while hopefully not compromising too much on precision of terms.

Thanks, that's useful. I didn't really touch the existing explanations, but it's good to have done a pass over that now as well.

shapely/geometry/linestring.py
```diff
@@ -10,31 +10,31 @@
...
+    Parameters
+    ----------
+    coordinates : sequence
+        A sequence of (x, y, [,z]) numeric coordinate pairs or triples, or
+        an array-like with shape (N, 2) or (N, 3).
+        Also can be a sequence of Point objects.
+
+    Examples
+    --------
+    Create a line with two segments
```
brendan-ward, line 28:
```suggestion
    Create a LineString with two segments
```

shapely/geometry/multilinestring.py
```diff
@@ -11,33 +11,31 @@
...
+        sequences or array-likes (see accepted input for LineString).
+
     Attributes
     ----------
     geoms : sequence
         A sequence of LineStrings
+
+    Examples
+    --------
+    Construct a collection containing two LineStrings.
```
brendan-ward, line 31:
```suggestion
    Construct a MultiLineString containing two LineStrings.
```

(using concrete geometry types for consistency with the GeometryCollection example above)

shapely/geometry/multipoint.py
```diff
@@ -10,39 +10,37 @@
...
-    """A collection of one or more points
+    """
+    A collection of one or more Points.
 
     A MultiPoint has zero area and zero length.
 
+    Parameters
+    ----------
+    points : sequence
+        A sequence Points, or a sequence of (x, y [,z]) numeric coordinate
```
brendan-ward, line 21:
```suggestion
        A sequence of Points, or a sequence of (x, y [,z]) numeric coordinate
```

shapely/geometry/multipoint.py
```diff
@@ -10,39 +10,37 @@
...
+        pairs or triples, or an array-like of shape (N, 2) or (N, 3).
+
     Attributes
     ----------
     geoms : sequence
         A sequence of Points
+
+    Examples
+    --------
+    Construct a 2 point collection
```
brendan-ward, line 31:
```suggestion
    Construct a MultiPoint containing two Points
```

shapely/geometry/multipolygon.py
```diff
@@ -9,45 +9,44 @@
...
+        (see LinearRing) and holes is a sequence of such linear rings.
+
     Attributes
     ----------
     geoms : sequence
         A sequence of `Polygon` instances
+
+    Examples
+    --------
+    Construct a collection from a sequence of coordinate tuples
```
brendan-ward, line 32:
```suggestion
    Construct a MultiPolygon from a sequence of coordinate tuples
```

shapely/geometry/polygon.py
brendan-ward: 🤦 `ring.length` != `len(ring)`; sorry - I wasn't reading this carefully enough!

brendan-ward (approved):
Just a few more suggestions re: consistency, but otherwise this looks good to me.  Thanks!

PR #1396, opened 2022-05-22
DOC: update the installation documentation

jorisvandenbossche:
This updates / consolidates the documentation on installing shapely. Apart from some pygeos->shapely renaming, I did:

* I moved most content about installation in the README to the specific documentation page (and linking that page), and in the README kept it short. My goal here is to avoid duplication (and going into details in the README). 
* I tried to combine the installation doc page from pygeos with some of the content from Shapely's README. But I didn't keep everything (some aspects were a bit outdated), so it would be good to give this a check.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -11,7 +11,7 @@ available via system package management tools like apt, yum, and Homebrew.
 Installation from PyPI
 ^^^^^^^^^^^^^^^^^^^^^^
 
-Shapely is available as a binary distribution (wheel) for Linux, OSX and Windows platforms.
+Shapely is available as a binary distribution (wheel) for Linux, macOS and Windows platforms.
 The distribution includes a GEOS version that was most recent at the time of the Shapely release.
 Install the binary wheel with pip as follows::
 
@@ -39,7 +39,7 @@ On Linux::
     $ sudo apt install libgeos-dev  # skip this if you already have GEOS
     $ pip install shapely --no-binary shapely
 
-On OSX::
+On macOS::
 
     $ brew install geos  # skip this if you already have GEOS
     $ pip install shapely --no-binary shapely
@@ -101,13 +101,13 @@ Shapely can be tested using ``pytest``::
 GEOS discovery (compile time)
 -----------------------------
 
-If GEOS is installed on Linux or OSX, normally the ``geos-config`` command line utility
+If GEOS is installed on Linux or macOS, normally the ``geos-config`` command line utility
 will be available and ``pip`` will find GEOS automatically.
-If the correct ``geos-config`` is not on the PATH, you can add it as follows (on Linux/OSX)::
+If the correct ``geos-config`` is not on the PATH, you can add it as follows (on Linux/macOS)::
 
     $ export PATH=/path/to/geos/bin:$PATH
 
-Alternatively, you can specify where Shapely should look for GEOS (on Linux/OSX)::
+Alternatively, you can specify where Shapely should look for GEOS (on Linux/macOS)::
 
     $ export GEOS_INCLUDE_PATH=/path/to/geos/include
     $ export GEOS_LIBRARY_PATH=/path/to/geos/lib
@@ -120,7 +120,7 @@ specified manually in any case::
 
 Common locations of GEOS (to be suffixed by ``lib``, ``include`` or ``bin``):
 
-* Anaconda (Linux/OSX): ``$CONDA_PREFIX/Library``
+* Anaconda (Linux/macOS): ``$CONDA_PREFIX/Library``
 * Anaconda (Windows): ``%CONDA_PREFIX%\Library``
 * OSGeo4W (Windows): ``C:\OSGeo4W64``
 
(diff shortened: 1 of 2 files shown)


docs/installation.rst
```diff
@@ -1,70 +1,101 @@
 Installation
 ============
 
+Built distributions
+-------------------
+
+Built distributions don't require to compile Shapely and its Dependencies,
```
mwtoews, line 7:
```suggestion
Built distributions don't require to compile Shapely and its dependencies,
```

mwtoews (commented):
Thanks, this is a good revision of the installation documents. It's nice to see several details move from README.rst to the manual. The `venv` section also looks good. A few other remarks:

- Note that `pip install -e .[test]` depends on `extras_require` in #1376
- Consider renaming "OSX" to the modern "macOS" ([ref](https://en.wikipedia.org/wiki/MacOS#macOS))
- Let's stick with title case "Shapely" for now, except when dealing with the repo or package name (with pip or conda), or other obvious code bits. There are several place I see the name switch between "shapely" and "Shapely", so it would be nice to see this more consistent.

jorisvandenbossche:
Thanks for the feedback!

mwtoews (approved):

docs/installation.rst
```diff
@@ -1,82 +1,113 @@
...
+    $ pip install shapely --no-binary shapely
 
-On OSX::
+On macOS::
 
     $ brew install geos  # skip this if you already have GEOS
-    $ pip install pygeos --no-binary
+    $ pip install shapely --no-binary shapely
+
+If you've installed GEOS to a standard location, the installation will automatically
```
brendan-ward, line 47:
```suggestion
If you've installed GEOS to a standard location on Linux or macOS, the installation will automatically
```

docs/installation.rst
```diff
@@ -1,82 +1,113 @@
...
 - Download and install a GEOS binary (https://trac.osgeo.org/osgeo4w/)
 - Set GEOS_INCLUDE_PATH and GEOS_LIBRARY_PATH environment variables (see below for notes on GEOS discovery)
-- Run ``pip install pygeos --no-binary``
+- Run ``pip install shapely --no-binary``
 - Make sure the GEOS .dll files are available on the PATH
 
 
-Installation from source
-------------------------
+Installation of development version
```
brendan-ward, line 61:
```suggestion
Installation for local development
```

docs/installation.rst
```diff
@@ -1,82 +1,113 @@
...
+-----------------------------------
 
-The same as installation with a custom GEOS binary, but then instead of installing pygeos
-with pip, you clone the package from Github::
+The same as installation with a custom GEOS binary, but then instead of installing
```
brendan-ward, line 64:
```suggestion
This is similar to installing with a custom GEOS binary, but then instead of installing
```

docs/installation.rst
```diff
@@ -1,82 +1,113 @@
...
-    $ pytest --pyargs pygeos.tests
+    $ pip install pytest  # or shapely[test]
+    $ pytest --pyargs shapely.tests
 
 
 GEOS discovery (compile time)
 -----------------------------
 
-If GEOS is installed on Linux or OSX, normally the ``geos-config`` command line utility
+If GEOS is installed on Linux or macOS, normally the ``geos-config`` command line utility
```
brendan-ward, line 104:
```suggestion
If GEOS is installed on Linux or macOS the ``geos-config`` command line utility should be available
```

docs/installation.rst
```diff
@@ -1,82 +1,113 @@
...
 will be available and ``pip`` will find GEOS automatically.
-If the correct ``geos-config`` is not on the PATH, you can add it as follows (on Linux/OSX)::
+If the correct ``geos-config`` is not on the PATH, you can add it as follows (on Linux/macOS)::
 
     $ export PATH=/path/to/geos/bin:$PATH
 
-Alternatively, you can specify where PyGEOS should look for GEOS (on Linux/OSX)::
+Alternatively, you can specify where Shapely should look for GEOS (on Linux/macOS)::
```
brendan-ward, line 110:
```suggestion
Alternatively, you can specify where Shapely should look for GEOS library and header files using environment variables (on Linux/macOS)::
```

brendan-ward (commented):
Thanks for working on this @jorisvandenbossche.

A few text edits to consider.

docs/installation.rst
```diff
@@ -1,82 +1,113 @@
 Installation
 ============
 
+Built distributions
+-------------------
+
+Built distributions don't require to compile Shapely and its dependencies,
+and can be installed using ``pip`` or ``conda``. In addition, Shapely is also
+available via system package management tools like apt, yum, and Homebrew.
```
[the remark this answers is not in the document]
jorisvandenbossche: Yeah, I copied this from the text in the README, but agreed that we should maybe not mention it (certainly if it is not even possible, like in the case of Homebrew).

> (installing GEOS via package manager then compiling Shapely using that may make sense, but as an advanced case)

I think that is indeed a use case. I certainly have seen issues of people install GEOS with brew, and then installing shapely on top of that. But then that doesn't belong in this section.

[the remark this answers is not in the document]
jorisvandenbossche: Shapely 1.8 seems to be available on debian unstable (https://packages.debian.org/search?keywords=python3-shapely), so that seems relatively up to date. 
So maybe I will generalize the sentence a bit, and remove the reference of Homebrew.

PR #1397, opened 2022-05-22
DOC: fix / consolidate the readthedocs configuration

jorisvandenbossche:
Trying to get the readthedocs build working again. I activated this branch, and so the build output can be seen here (it is now "passing"): https://readthedocs.org/projects/shapely/builds/16978294/

The actual doc build is still failing, because the plot directives are still erroring.

I also switched to use mamba instead of conda, which speeds up the env install step from 4 to 2 min.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,6 +5,11 @@
 # Required
 version: 2
 
+build:
+  os: "ubuntu-20.04"
+  tools:
+    python: "mambaforge-4.10"
+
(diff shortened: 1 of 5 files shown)


jorisvandenbossche:
With the last commit (to not fail the build on warnings), we actually get a result: https://shapely.readthedocs.io/en/docs-readthedocs/

Now, given that there are some more PRs needed to further clean up the docs, it might be better to not yet have this as  https://shapely.readthedocs.io/en/latest/ (so revert the last commit before merging)

PR #1402, opened 2022-05-23
DOC: restructure the combined docs + new theme

jorisvandenbossche:
This PR does:

- Uses a more modern sphinx theme (instead of "sphinxdoc"). There are plenty of possibilities, but I went with sphinx-book-theme, because it's a theme where I am also involved with somewhat (we use a variant of it for pandas/geopandas)
- Restructures the `index.rst`: moved the "project" page to index.rst, split "changes" to its own page, divided the docs in user guide vs reference docs (in the future we could further split the "User Manual" page into multiple pages if we would want)

You can see the result live at https://shapely.readthedocs.io/en/docs-restructure/index.html


jorisvandenbossche:
What do people think of the theme and general layout? -> https://shapely.readthedocs.io/en/docs-restructure/index.html

docs/index.rst
```diff
@@ -1,27 +1,24 @@
-=======
-Shapely
-=======
+.. include:: ../README.rst
+
+.. include:: ../CREDITS.txt
```
brendan-ward, line 3: This is long enough that it probably would do better on its own top level page.  Alternatively, we could instead adopt the approach of numpy and pandas and list contributors for each release, but not maintain a single long list (per-release seems easier from a maintenance perspective).  This list could then go into a dedicated "contributors prior to 2.0" page (not quite sure where).

brendan-ward (commented):
Thanks for the changes here @jorisvandenbossche !

I like the new layout.  

It's a bit hard to judge the content of the docs cohesively since there are concurrent changes in other PRs, so I think my only real comment here is around trying to move the long list of contributors off the main page onto a dedicated page, and to think about how we want to maintain that post 2.0 (e.g., alongside release notes).

jorisvandenbossche:
Yes, agreed that we should reconsider how to handle the credits/contributors (keep the overall listing? Move to a separate page? List them per release?). For now this keeps it as is (include the file on the main doc page), so let's have that as a follow-up discussion / PR.

jorisvandenbossche:
Going to merge this as people generally seemed to like the new theme, and then we can better judge other PRs that update the docs based on the new layout/theme.

PR #1403, opened 2022-05-23
Update LICENSE file and copyright holder

jorisvandenbossche:
Updating / consolidating the LICENSE files:

* Removes the second LICENSE file originally from pygeos
* Reformats the existing LICENSE a bit (copied plain text from https://choosealicense.com/licenses/bsd-3-clause/, this matches also better with https://opensource.org/licenses/BSD-3-Clause, eg using 1, 2, 3). Note this is purely some formatting changes, not actual text changes, but this should help with github (and other tools) to better recognize the license (eg right now github's UI doesn't recognize it as "BSD-3-Clause license")
* Updated the copyright holder to include "Shapely Contributors" for now, but similar question as in https://github.com/shapely/shapely/pull/1376#discussion_r862481338 about what exactly we want to put here (see comment below).


jorisvandenbossche:
So for the copyright holder (currently "Copyright (c) 2007, Sean C. Gillies"), some notes/questions:

- Can we include the "Copyright (c) 2019, Casper van der Wel" somewhere? (from pygeos, in the removed LICENSE file)
  - Looking at other projects, I think it is a general practice that if a package includes some vendored code, it also includes a copy of the license or some note about that. Although in this case it is not really vendoring. In Apache software projects this is called a "code donation", and checking in the Arrow repo, in such a case we don't include the original license.
  - Could also add this to the current license file as a separate copyright line (instead of a separate license file somewhere in the code base).
- Comparison with some other projects: NumPy uses "2005-2022, NumPy Developers" (geopandas uses the same pattern). [Pandas](https://github.com/pandas-dev/pandas/blob/d2a7eff0024f074b742e5d69667555bbd8d0c9d6/LICENSE#L3-L6) kept an original copyright for the first years, and then added a similar general author for later years. [Scipy](https://github.com/scipy/scipy/blob/50a29c4109e4ec2b6f61db95aa4e49297db869f9/LICENSE.txt#L1) also does the same. So that is basically what I did here right now with "Copyright (c) 2007, Sean C. Gillies. 2007-2022, Shapely Contributors."

Personally I don't care too much about the exact copyright notice, but we need to pick _something_ (and then update docs/conf.py to match it). But so would like to hear from @sgillies (and @caspervdw) what your preference is.

caspervdw (approved):

sgillies (approved):

PR #1408, opened 2022-06-07
CI: upgrade GEOS patch versions for tests and release

mwtoews:
See https://lists.osgeo.org/pipermail/geos-devel/2022-June/010685.html

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-2019]
         architecture: [x64]
-        geos: [3.6.5, 3.7.2, 3.8.3, 3.9.3, 3.10.3, main]
+        geos: [3.6.5, 3.7.4, 3.8.3, 3.9.3, 3.10.3, main]
         include:
           # 2017
           - python: 3.6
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -14,27 +14,27 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-2019]
         architecture: [x64]
-        geos: [3.6.5, 3.7.3, 3.8.1, 3.9.2, 3.10.1, main]
+        geos: [3.6.5, 3.7.2, 3.8.3, 3.9.3, 3.10.3, main]
```
jorisvandenbossche, line 17:
```suggestion
        geos: [3.6.5, 3.7.4, 3.8.3, 3.9.3, 3.10.3, main]
```

jorisvandenbossche (approved):
Thanks!

We might also want to backport this to 1.8.x?

mwtoews:
It seems that GEOS 3.7.4 does not work with CMake; see https://github.com/libgeos/geos/issues/629

There may be a GEOS 3.7.5 patch release soon, so I'll hold out a bit longer.

mwtoews:
> We might also want to backport this to 1.8.x?

xref #1413

PR #1412, opened 2022-06-08
ENH: Remove "shapely" name and duplicate WKT type from __repr__

### LINKED ISSUE #1411: Refine __repr__ to avoid duplication of geometry type label
Originally noted on #1391 

The current `__repr__` produces something like `<shapely.Point POINT (0 0)>`, which has duplication between the class name and the WKT type (except for Z types). 

One way to solve this would be to omit the first part of the WKT type from `__repr__`:
* `Point(0,0)` => `<shapely.Point (0 0)>`
* `Point(0,0,0)` => `<shapely.Point Z (0 0 0)>`
* `Point()` => `<shapely.Point EMPTY>`
* `GeometryCollection(Point(0,0,0))` => `<shapely.GeometryCollection Z (POINT Z (0 0 0))>` 

The only case where this gets a little awkard is for goemetry collections, which use the WKT all-caps type names for members, but likely not that big a deal.

brendan-ward:
Resolves #1411 , now `Point(0, 0)` produces ~~`<shapely.Point (0 0)>`~~ `<POINT (0 0)>`, etc.

This updates the two tests that used the __repr__ output, but does not touch any of the docstring / doctests as those are in progress in #1378, #1391

Based on `__repr__` decision [here](https://github.com/shapely/shapely/issues/1411#issuecomment-1160728980)


jorisvandenbossche:
@brendan-ward do you have time to update this PR based on https://github.com/shapely/shapely/issues/1411#issuecomment-1160728980? (`<WKT>`)

brendan-ward:
Yes, I'll work on this.  I was waiting until the STRtree changes were in, since those will need to be updated with the new __repr__ outputs, but can work on this for non-STRtree stuff so this is more ready.

shapely/geometry/base.py
```diff
@@ -92,19 +92,18 @@ def __nonzero__(self):
         return self.__bool__()
 
     def __repr__(self):
-        class_name = str(self.__class__.__name__)
         try:
             wkt = super().__str__()
         except (GEOSException, ValueError):
             # we never want a repr() to fail; that can be very confusing
-            return "<shapely.{} Exception in WKT writer>".format(class_name)
+            return "<Exception in WKT writer>"
```
jorisvandenbossche, line 99: Maybe in this specific case we could still include the class? (so you know whether it is a MultiPoint, or ...)

jorisvandenbossche (approved):

jorisvandenbossche:
>  I was waiting until the STRtree changes were in, since those will need to be updated with the new repr outputs, but can work on this for non-STRtree stuff so this is more ready.

I think I will merge both: also the other docstring need to be updated (not only of STRtree), and merging this one further allows to update the outputs in the docs etc.

Issue #1386, opened 2022-05-13
Finalizing the STRtree interface


jorisvandenbossche:
For 1.8.0, we had quite some discussion in https://github.com/shapely/shapely/pull/1064 (and in https://github.com/shapely/shapely/pull/1094 and https://github.com/shapely/shapely/pull/1112), about the exact API we wanted for STRtree. In the end we settled on adding `query_items()` and `query_geoms` methods (in addition to `query`) instead of adding a keyword (like `return_geometries=True/False`) in `query()`. https://github.com/shapely/shapely/pull/1112 also added the functionality to pass custom items to the the `STRtree(.., items=..)` constructor.

But in https://github.com/shapely/shapely/pull/1251, where I merged the pygeos functionality with this Shapely interface, we had again some discussion about this, indicating that it might not yet be fully resolved. 

One of the discussion points is the usage of the term "items" (https://github.com/shapely/shapely/pull/1251#issuecomment-1051407542). While the method is called `query_items()`, I used "index" in the update of the docstring. 

> Return the index (or stored item) of all geometries in the tree
> with extents that intersect the envelope of the input geometry.
>
> The returned items may be identified by an integer index (default) or
> arbitrary item values (optional) if those are provided when
> constructing the tree.

But this triggered some comments about it (eg https://github.com/shapely/shapely/pull/1251#discussion_r795242461 or https://github.com/shapely/shapely/pull/1251#discussion_r795242032). While I think we can certainly get agreement on how to understand the default behaviour, it might point to the fact that the API is not yet really ideal (also given the fact that we have a hard time to clearly / succinctly explain this in the documentation).

Some other issues with the current naming in the API:

- The default is to return integer indices (if the user only provided geometries to the STRtree constructor), but the method is actually called `query_items()` using "items".  
 (I personally _think_ that most people will never pass explicit items, but will rather use the indices) 
- The `query_bulk()` method _only_ returns integer indices, and never items (even if the user passed those in the constructor), creating some more inconsistency. And the same applies to `nearest_all`. 
  (The reason for this is that this methods takes an array of query geometries as input, and so returns indices into both the tree and input geometries. Alternatively, we could require that you pass `items` as well to the `query_bulk` method, but this would further complicate this method)

---

So possible outcomes:

- Decide we are fine with how it is now, and maybe only tweak the docstring a bit further.
- Reconsider the ability to pass custom items, to avoid the dichotomy between "items" and "indices" (and it easy to mimic this as a user). 
[...]

PR #1414, opened 2022-06-09
ENH: Remove Geometry() constructor usage and use concrete geometry type constructors or from_wkt

brendan-ward:
As part of the PyGEOS to Shapely migration, this removes pygeos geometry types from docstrings, and replaces the `Geometry("...wkt...")` constructor used in many places with either concrete geometry type constructors like `Point(0, 0)` or `from_wkt("...wkt...")`.

Includes some lightweight cleanup of some of the doc pages as I'm touching them in the process, mostly to make sure they refer to Shapely rather than PyGEOS, but fully expecting we will want to modify those more heavily in follow-on PRs.

Includes a few minor updates to doctests where I saw a benefit to reordering them or adding another example; in general tried to keep this to a minimum to make this easier to review.

Where possible, this standardizes on using the un-prefixed geometry constructors `shapely.Point()`  => `Point()` but only where I was already modifying those files.  `shapely/tests/common.py` could use this treatment as well but was outside the scope of this PR.

This deliberately avoids updating STRtree docstrings / tests, as that is subject to change under #1386 

This does not actually remove the `Geometry()` constructor, but should make it possible to do so in a later PR.


docs/geometry.rst
```diff
@@ -1,78 +1,86 @@
...
   >>> point_1 != point_2
   True
 
 
-Properties
-~~~~~~~~~~
-
-Geometry objects have neither properties nor methods.
-Instead, use the functions listed below to obtain information about geometry objects.
+Geometry types
```
brendan-ward, line 78: This is intended to be a placeholder for now.  We should probably refactor so that each geometry type gets an individual page

shapely/constructive.py
```diff
@@ -128,33 +130,34 @@ def buffer(
...
     >>> buffer(line2, 2, cap_style="flat", join_style="bevel")
-    <pygeos.Geometry POLYGON ((18 12, 18 20, 22 20, 22 10, 20 8, 10 8, 10 12, 18...>
+    <shapely.Polygon POLYGON ((18 12, 18 20, 22 20, 22 10, 20 8, 10 8, 10 12, 18...>
     >>> buffer(line2, 2, cap_style="flat", join_style="mitre")
-    <pygeos.Geometry POLYGON ((18 12, 18 20, 22 20, 22 8, 10 8, 10 12, 18 12))>
+    <shapely.Polygon POLYGON ((18 12, 18 20, 22 20, 22 8, 10 8, 10 12, 18 12))>
     >>> buffer(line2, 2, cap_style="flat", join_style="mitre", mitre_limit=1)
-    <pygeos.Geometry POLYGON ((18 12, 18 20, 22 20, 21.828 9, 21 8.172, 10 8, 10...>
-    >>> square = Geometry("POLYGON((0 0, 10 0, 10 10, 0 10, 0 0))")
+    <shapely.Polygon POLYGON ((18 12, 18 20, 22 20, 22 9.172, 20.828 8, 10 8, 10...>
```
brendan-ward, line 153: Not sure why, but I got slightly different values (macOS 10.15 / > GEOS 3.10)

shapely/_geometry.py
```diff
@@ -104,13 +104,16 @@ def get_dimensions(geometry, **kwargs):
...
-    >>> get_dimensions(Geometry("POLYGON((0 0, 0 10, 10 10, 10 0, 0 0))"))
+    >>> polygon = Polygon(((0, 0), (0, 10), (10, 10), (10, 0), (0, 0)))
+    >>> get_dimensions(polygon)
     2
-    >>> get_dimensions(Geometry("GEOMETRYCOLLECTION (POINT(0 0), LINESTRING(0 0, 1 1))"))
-    1
-    >>> get_dimensions(Geometry("GEOMETRYCOLLECTION EMPTY"))
+    >>> get_dimensions(GeometryCollection([point, polygon]))
+    2
+    >>> get_dimensions(GeometryCollection())
```
jorisvandenbossche, line 116: For this specific case, I find `"GEOMETRTCOLLECTION EMPTY` a bit more explicit / clear. 
Or maybe using `GeometryCollection([])` (with the empty list) also helps?

shapely/_geometry.py
```diff
@@ -337,18 +351,19 @@ def get_point(geometry, index, **kwargs):
...
+    <shapely.Point POINT (2 2)>
     >>> get_point(line, [0, 3]).tolist()
-    [<pygeos.Geometry POINT (0 0)>, <pygeos.Geometry POINT (3 3)>]
-    >>> get_point(Geometry("LINEARRING (0 0, 1 1, 2 2, 0 0)"), 1)
-    <pygeos.Geometry POINT (1 1)>
-    >>> get_point(Geometry("MULTIPOINT (0 0, 1 1, 2 2, 3 3)"), 1) is None
+    [<shapely.Point POINT (0 0)>, <shapely.Point POINT (3 3)>]
+    >>> get_point(LinearRing(((0, 0), (1, 1), (2, 2), (0, 0))), 1)
+    <shapely.Point POINT (1 1)>
+    >>> get_point(MultiPoint(((0, 0), (1, 1), (2, 2), (3, 3))), 1) is None
```
jorisvandenbossche, line 364:
```suggestion
    [<shapely.Point POINT (0 0)>, <shapely.Point POINT (3 3)>]

    The function works the same for LinearRing input:

    >>> get_point(LinearRing(((0, 0), (1, 1), (2, 2), (0, 0))), 1)
    <shapely.Point POINT (1 1)

    For non-linear input geometries, it returns None:

    >>> get_point(MultiPoint(((0, 0), (1, 1), (2, 2), (3, 3))), 1) is None
```

In general, this kind of divisions of the example blob would make them more readable, I think. But that's probably a bit out of scope for this PR, so feel free to defer.

shapely/_geometry.py
```diff
@@ -435,10 +451,17 @@ def get_interior_ring(geometry, index, **kwargs):
 
     Examples
     --------
-    >>> polygon_with_hole = Geometry("POLYGON((0 0, 0 10, 10 10, 10 0, 0 0), (2 2, 2 4, 4 4, 4 2, 2 2))")
+    >>> from shapely import Point, Polygon
+    >>> polygon_with_hole = Polygon(((0, 0), (0, 10), (10, 10), (10, 0), (0, 0)), \
+holes=[((2, 2), (2, 4), (4, 4), (4, 2), (2, 2))])
```
jorisvandenbossche, line 456:
```suggestion
    >>> polygon_with_hole = Polygon(
    ...     [(0, 0), (0, 10), (10, 10), (10, 0), (0, 0)],
    ...     holes=[[(2, 2), (2, 4), (4, 4), (4, 2), (2, 2)]]
    ... )
```

Possible alternative, that might be more readable? (although it takes quite a bit more vertical space)

shapely/_geometry.py
```diff
@@ -104,13 +104,16 @@ def get_dimensions(geometry, **kwargs):
 
     Examples
     --------
-    >>> get_dimensions(Geometry("POINT (0 0)"))
+    >>> from shapely import GeometryCollection, Point, Polygon
+    >>> point = Point(0, 0)
+    >>> get_dimensions(point)
     0
-    >>> get_dimensions(Geometry("POLYGON((0 0, 0 10, 10 10, 10 0, 0 0))"))
+    >>> polygon = Polygon(((0, 0), (0, 10), (10, 10), (10, 0), (0, 0)))
```
jorisvandenbossche, line 111: The linestring above (line 82) uses a _list_ of coordinate pair tuples, while here it is a _tuple_ of tuples. Both work fine, but we should maybe try to be consistent within the docs? 
I personally find the list of tuples more readable (otherwise you have all those same type of brackets):

```suggestion
    >>> polygon = Polygon([(0, 0), (0, 10), (10, 10), (10, 0), (0, 0)])
```

(this also doesn't necessarily need to happen in this PR, as it is quite some work I can imagine)

shapely/_geometry.py
```diff
@@ -677,11 +709,14 @@ def get_precision(geometry, **kwargs):
...
+    >>> from shapely import Point
+    >>> point = Point(1, 1)
+    >>> get_precision(point)
     0.0
-    >>> geometry = set_precision(Geometry("POINT (1 1)"), 1.0)
+    >>> geometry = set_precision(point, 1.0)
     >>> get_precision(geometry)
     1.0
+    >>> import numpy as np
     >>> np.isnan(get_precision(None))
```
jorisvandenbossche, line 720: In this case we could maybe actually also do something more simple like:

```
>>> get_precision(None)
nan
```

(since np.nan actually prints a value in contrast to `None`, we don't necessarily have to do the workaround)

jorisvandenbossche (commented):
Thanks for working on this! Only a partial review, looked at the first few files.

shapely/constructive.py
jorisvandenbossche: It's annoying when writing those examples on a mac, but you will need to keep the output as it was. We currently test the docstrings on one CI build (ubuntu python 3.10 build), and so it has to pass on that build. And currently this example is failing: https://github.com/shapely/shapely/runs/6824676810?check_suite_focus=true#step:17:54

shapely/_geometry.py
brendan-ward: I'll add this one here but we should certainly do a more comprehensive pass at this in another PR.

shapely/_geometry.py
brendan-ward: I'd prefer to leave that for a follow-on PR to do just that, in order to keep that both changesets a bit smaller.

shapely/_geometry.py
brendan-ward: Scratch that; I'm updating these now.  All these lines were touched in this PR anyway, best to have them clean before this gets merged.

brendan-ward:
I ended up standardizing coord lists from tuples of tuples (or tuples of lists) to lists of tuples everywhere I saw them; this ended up reaching a bit beyond the areas originally touched by this PR.  It also unexpectedly led to some reformatting in files that apparently were out of sync with the pinned version of `black` (sorry!  But probably good to get these in sync).

In nearly all cases, the only changes to those files are a few isolated changes of tuples to lists or splitting a test case across multiple lines; there are no other substantive logic changes introduced in 1da1414974762d192072e3cd43ef49b0aac9c781

docs/geometry.rst
```diff
@@ -1,78 +1,96 @@
...
+be changed in place. Every Shapely operation will result in a new object being
+returned.
+
 
 Construction
 ~~~~~~~~~~~~
 
-For convenience, the ``Geometry`` class can be constructed with a WKT (Well-Known Text)
-or WKB (Well-Known Binary) representation of a geometry:
+Geometries can be constructed directly using Shapely geometry classes:
```
brendan-ward, line 19: seemed like we needed a canonical example here of using the classes, rather than starting with WKT / WKB examples.

shapely/examples/intersect.py
```diff
@@ -24,7 +24,7 @@
 # speed: 50*sqrt(2)
 #
 # as a line
-vector = LineString(((-25.0, -25.0), (25.0, 25.0)))
+vector = LineString([(-25.0, -25.0), (25.0, 25.0)])
```
brendan-ward, line 27: Only intentional change in this file; everything else is auto formatting

tests/test_linear_referencing.py
```diff
@@ -6,14 +6,13 @@
...
 class LinearReferencingTestCase(unittest.TestCase):
-
     def setUp(self):
         self.point = Point(1, 1)
-        self.line1 = LineString(([0, 0], [2, 0]))
-        self.line2 = LineString(([3, 0], [3, 6]))
-        self.multiline = MultiLineString([
-            list(self.line1.coords), list(self.line2.coords)
-        ])
+        self.line1 = LineString([(0, 0), (2, 0)])
```
brendan-ward, line 11: Lines 11-12 are the only intentional changes in this file; everything else is autoformatting

tests/test_operations.py
```diff
@@ -87,21 +90,21 @@ def test_hausdorff_distance(self):
 
     def test_interpolate(self):
         # successful interpolation
-        test_line = LineString(((1,1),(1,2)))
-        known_point = Point(1,1.5)
-        interpolated_point = test_line.interpolate(.5, normalized=True)
+        test_line = LineString([(1, 1), (1, 2)])
```
brendan-ward, line 93: Only intentional change in this file; everything else is autoformatting (or fix on line 33 to keep as one line since autoformatting did bad things there otherwise)

jorisvandenbossche (approved):
I did an update for the new (shortened) repr (https://github.com/shapely/shapely/pull/1412).

I also did a full pass over the changes, looks all good to me!

I am personally not always fully a fan of replacing all WKT strings in test parametrizations with subclass constructors (I think using the WKT strings in the parametrize values and then using from_wkt in the test can be a bit cleaner / shorter for large parametrizations). But on the other hand simply always using the subclass constructors makes it easier to know how to write tests as contributor.

Issue #999, opened 2020-10-13
API: read-only access to GEOS pointer (__geom__, _geom, _ptr)


jorisvandenbossche:
Moving discussion from https://github.com/Toblerity/Shapely/pull/983#discussion_r496481732

In current Shapely, there are two attributes that both return the GEOS pointer:

```
In [7]: from shapely.geometry import Point

In [8]: p = Point(1, 1)

In [9]: p
Out[9]: <shapely.geometry.Point POINT (1 1) >

In [10]: p.__geom__
Out[10]: 94617737200752

In [11]: p._geom
Out[11]: 94617737200752
```

In the branch subclassing from the C extension type (https://github.com/Toblerity/Shapely/pull/983), there is now a third attribute inherited from PyGEOS that does the same:

```
In [12]: p._ptr
Out[12]: 94617737200752
```

I think it is fine to keep *some* way to get access to the GEOS pointer (eg if you want to use numba or cython yourself in some application where you know you are using the same GEOS version as Shapely, and thus passing along pointers can be safe). So we should maybe keep one of those attributes (although for cython, a function to get the GEOS pointer from a python object could also be used for this instead of an attribute). 
But we don't necessarily need to keep all of them, and we can deprecate others?

As @mwtoews mentions at https://github.com/Toblerity/Shapely/pull/983#discussion_r496481732, `__geom__` is not an official "dunder" method of python (and here also not used in a context for which typically dunder methods are defined), we could maybe deprecate `__geom__` in Shapely 1.8, keeping `_geom` in Shapely 2.0 (since that is backwards compatible with older Shapely versions)?

cc @sgillies @caspervdw

jorisvandenbossche:
Note that in the shapely-2.0 branch, the attributes are *actually* read-only, unlike on current Shapely where you can set those `_geom` / `__geom__` attributes yourself from python, and thus run into ugly errors.

jorisvandenbossche:
Concrete proposal for Shapely 1.8:

* Deprecate accessing `__geom__`
* Deprecate *setting* both `__geom__` and `_geom`

Thoughts?

mwtoews:
The proposed plan for `__geom__` and `_geom` are reasonable.

Access to both attributes `_geom` and `_ptr` (as the same object) should be fine for Shapely 2+.

PR #1417, opened 2022-06-13
API: remove __geom__, keep _geom read-only access to GEOS pointer

jorisvandenbossche:
Closes https://github.com/shapely/shapely/issues/999. This removes the `__geom__` dunder attribute and keeps `_geom` (renaming the `_ptr` inherited from pygeos to `_geom`).

### DIFF
--- a/shapely/tests/test_io.py
+++ b/shapely/tests/test_io.py
@@ -88,51 +88,6 @@ GEOJSON_COLLECTION_EXPECTED = [
 ]
 
 
-class ShapelyGeometryMock:
-    def __init__(self, g):
-        self.g = g
-        self.__geom__ = g._ptr if hasattr(g, "_ptr") else g
-
-    @property
-    def __array_interface__(self):
-        # this should not be called
-        # (starting with numpy 1.20 it is called, but not used)
-        return np.array([1.0, 2.0]).__array_interface__
-
-    @property
-    def wkb(self):
-        return shapely.to_wkb(self.g)
-
-    @property
-    def geom_type(self):
-        idx = shapely.get_type_id(self.g)
-        return [
-            "None",
-            "Point",
-            "LineString",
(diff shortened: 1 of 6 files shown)


mwtoews:
Looks great, two quick questions.

1. Is this `_geom` correct?
https://github.com/shapely/shapely/blob/8d1911e4684f53261419abfb015e53e64b7acfb4/shapely/_geos.pxi#L59

2. Slightly related to this PR, should this `_ptr` also be renamed? https://github.com/shapely/shapely/blob/8d1911e4684f53261419abfb015e53e64b7acfb4/src/strtree.c#L1283

shapely/tests/test_io.py
```diff
@@ -88,51 +88,6 @@
 ]
 
 
-class ShapelyGeometryMock:
-    def __init__(self, g):
-        self.g = g
-        self.__geom__ = g._ptr if hasattr(g, "_ptr") else g
```
jorisvandenbossche, line 94: This was some leftover code from pygeos (testing it could convert shapely geometries), that could be cleaned-up (while removing all references of `__geom__`)

jorisvandenbossche:
> 1. Is this `_geom` correct?

Ah, that's actually in a file that is no longer used. Cleaning that up separately -> https://github.com/shapely/shapely/pull/1419

> 2\. Slightly related to this PR, should this `_ptr` also be renamed?

Yeah, I wondered the same. But the problem is that renaming it to `_geom` would not be correct since it is not a geometry pointer (but STRtree). For consistency we _could_ rename it to something else than `_ptr`, but leaving it as `_ptr` also seems fine for me (it won't be exactly the same as `_geom` for the geometry class anyway)

brendan-ward (approved):
Thanks @jorisvandenbossche , looks good to me.

PR #1418, opened 2022-06-13
PERF: use numpy matmul for affine transformation

mwtoews:
This is an idea borrowed from @ThomasHilger in https://github.com/shapely/shapely/discussions/1336

### DIFF
--- a/shapely/affinity.py
+++ b/shapely/affinity.py
@@ -69,7 +69,7 @@ def affine_transform(geom, matrix):
         off = np.array([xoff, yoff, zoff], dtype=float)
 
     def _affine_coords(coords):
-        return np.dot(A, coords.T).T + off
+        return np.matmul(A, coords.T).T + off
 
     return shapely.apply(geom, _affine_coords, include_z=ndim == 3)
 
--- a/shapely/affinity.py
+++ b/shapely/affinity.py
@@ -10,7 +10,7 @@ __all__ = ["affine_transform", "rotate", "scale", "skew", "translate"]
 
 
 def affine_transform(geom, matrix):
-    r"""Returns a transformed geometry using an affine transformation matrix.
+    r"""Return a transformed geometry using an affine transformation matrix.
 
     The coefficient matrix is provided as a list or tuple with 6 or 12 items
     for 2D or 3D transformations, respectively.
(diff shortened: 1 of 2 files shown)


shapely/affinity.py
```diff
@@ -54,33 +54,22 @@ def affine_transform(geom, matrix):
...
-            return np.hstack((np.atleast_2d(xp).T, np.atleast_2d(yp).T))
-        elif ndim == 3:
-            z = coords[:, 2]
-            xp = a * x + b * y + c * z + xoff
-            yp = d * x + e * y + f * z + yoff
-            zp = g * x + h * y + i * z + zoff
-            return np.hstack(
-                (np.atleast_2d(xp).T, np.atleast_2d(yp).T, np.atleast_2d(zp).T)
-            )
+        return np.dot(A, coords.T).T + off
```
jorisvandenbossche, line 72: Would `np.dot(coords, A)` (without the double transpose) also work?

jorisvandenbossche (commented):
Thanks for looking at this!

shapely/affinity.py
mwtoews: that would raise (e.g.) "ValueError: shapes (2,2) and (3,2) not aligned: 2 (dim 1) != 3 (dim 0)". Transpose is an unchanged view of the original array, so it is a fast operation. I should also note that there are two alternative forms using matrix multiplcation:

1.  `np.matmul(A, coords.T).T + off`
2. `(A @ coords.T).T + off` for [PEP 465](https://peps.python.org/pep-0465/)

mwtoews:
I've switched out `np.dot` with `np.matmul`, as affine transform matrices use [matrix multiplication](https://en.wikipedia.org/wiki/Affine_transformation#Representation).

As for benchmarks, a few basic "%timeit" tests have approx 20% speedup from the previous implementation. There are no differences between `np.dot` vs `np.matmul` vs `@`.

tests/test_affinity.py
```diff
@@ -16,8 +16,8 @@ def test_affine_params(self):
         g = load_wkt('LINESTRING(2.4 4.1, 2.4 3, 3 3)')
         self.assertRaises(
             TypeError, affinity.affine_transform, g, None)
-        self.assertRaises(
-            TypeError, affinity.affine_transform, g, '123456')
+        # self.assertRaises(
+        #     TypeError, affinity.affine_transform, g, '123456')
```
jorisvandenbossche, line 20: One more question: this no longer raises an error? (or a different error?)

mwtoews: Fair question, it is not raised since numpy happily casts str to float (`np.array(list('123456'), dtype=float)`). This test was never really necessary, and on a dull day I would re-write all of the olden-days unittests to much more succinct pytest.

mwtoews:
I've done a bit more comprehensive benchmarks using the following ipython with a 1.8 MB shapefile:
```python
import geopandas
from shapely import affinity

gdf = geopandas.read_file(shp_fname)

%timeit gdf.geometry.apply(affinity.rotate, angle=12.3, origin=(0, 0))
```
- this PR (based on shapely 2): 65 ms
- shapely 2 (main): 85 ms; therefore this PR is 30% faster
- shapely 1.8: 100 ms; therefore this PR is 85% faster

tests/test_affinity.py
jorisvandenbossche: Yes, let's just remove the commented code then

jorisvandenbossche, line 20:
```suggestion
```

jorisvandenbossche (approved):

Issue #956, opened 2020-08-06
The emptyness of GeometryCollection with empty subparts


jorisvandenbossche:
Moving some discussion from a PR to a new issue: https://github.com/Toblerity/Shapely/pull/951#discussion_r465318489 (cc @tomplex @mwtoews)

```python
from shapely import geos, wkt 
g = wkt.loads("GEOMETRYCOLLECTION(LINESTRING EMPTY, POLYGON EMPTY)") 

>>> print(g)  
GEOMETRYCOLLECTION EMPTY
>>> g.is_empty 
True
>>> len(g)
0
>>> len(g.geoms)
0
```

So you can see that such a collection is considered "emtpy", but Shapely deviates from GEOS in that it also says the length is 0. GEOS does keeps the subparts:

```
>>> geos.lgeos.GEOSGetNumGeometries(g._geom)
2
>>> from shapely.geometry.base import geom_factory                                                                                                                                                            
>>> geom_factory(geos.lgeos.GEOSGetGeometryN(g._geom, 0)).wkt 
'LINESTRING EMPTY'
```

(although the WKT for the collection itself that GEOS creates is a bit confusing, since it just says "GEOMETRYCOLLECTION EMPTY", which is already reported by @mwtoews to GEOS: https://trac.osgeo.org/geos/ticket/1040).

PostGIS deviates slightly from the GEOS behaviour, as noted by @mwtoews in https://github.com/Toblerity/Shapely/pull/951/files#r465455890: it returns 0 for the number of geometries of a collection with empty parts (https://trac.osgeo.org/postgis/ticket/4736)

jorisvandenbossche:
Updated example with latest Shapely main branch:

```python
from shapely import geos, wkt 
g = wkt.loads("GEOMETRYCOLLECTION(LINESTRING EMPTY, POLYGON EMPTY)") 

>>> print(g)  
GEOMETRYCOLLECTION (LINESTRING EMPTY, POLYGON EMPTY)  # <-- this has been updated in latest GEOS
>>> g.is_empty 
True
>>> len(g.geoms)
0
>>> shapely.get_num_geometries(g)
2
```

So `.geoms` and `get_num_geometries` are contradictory. This is because in `.geoms` we currently have a special case check for empty geometries:

https://github.com/shapely/shapely/blob/48cf81e5ba3e48b779e81be543ab42858385c37d/shapely/geometry/base.py#L710-L714

PR #1420, opened 2022-06-13
API: all-empty GeometryCollection .geoms to return empty subgeoms

jorisvandenbossche:
Closes https://github.com/shapely/shapely/issues/956

Currently a GeometryCollection that consists of multiple empty geometries (which is considered as an empty geometry), does not return those geometries in the `.geoms` attribute (because of a special casing in our code for this, to specifically return an empty list).

Actually returning the empty subgeoms makes this more correct (the actual GEOS geometry has those empty subgeoms, also the WKT now displays them) and consistent with the top-level functions `get_num_geometries` / `get_geometry` (to index into the sub-geoms).

### DIFF
--- a/shapely/tests/geometry/test_collection.py
+++ b/shapely/tests/geometry/test_collection.py
@@ -33,15 +33,7 @@ def test_empty(geom):
 
 
 def test_empty_subgeoms():
-    geom = shape(
-        {
-            "type": "GeometryCollection",
-            "geometries": [
-                {"type": "Point", "coordinates": ()},
-                {"type": "LineString", "coordinates": (())},
-            ],
-        }
-    )
+    geom = GeometryCollection([Point(), LineString()])
     assert geom.type == "GeometryCollection"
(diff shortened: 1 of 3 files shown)


shapely/tests/geometry/test_collection.py
```diff
@@ -38,7 +29,24 @@ def test_empty(geom):
     assert geom.type == geom.geom_type
     assert geom.is_empty
     assert len(geom.geoms) == 0
-    assert geom.geoms == []
+    assert list(geom.geoms) == []
+
+
+def test_empty_subgeoms():
+    geom = shape(
```
brendan-ward, line 36: Seems like this could be somewhat more succinct as:

```Python
geom = GeometryCollection([Point(), LineString()])
```

(not seeing the added value of constructing from GeoJSON here since the focus is on the `.geoms` attribute)

mwtoews (approved):
This is fixes to the correct behaviour, thanks!

Issue #1280, opened 2022-01-03
Overview of new features in Shapely 2.0


jorisvandenbossche:
In addition to a rewrite of the internals of Shapely by merging in the pygeos code and the behaviour changes we added warnings for (hashability, iterability, etc), we also implicitly added quite some new features by merging pygeos if you compare it to Shapely 1.8 (which is the relevant comparison for shapely users). 

For the Shapely 2.0 release notes, we should have an overview of those new features. So starting to use this issue to keep track of those (so please add things along the way, and then eventually we should turn this into a PR adding the actual release notes of course).

* Vectorized functions in the top-level namespace ...
* Geometry classes in top-level namespace
* Support for fixed precision model for geometries and in overlay functions:
  * `set_precision` to conform a geometry to a certain grid size, and this will then also be used by subsequent overlay methods
  * `grid_size` keyword in overlay methods (`intersection`, `union`, `difference`, etc) 
* STRtree:
  * Specifying a `predicate` in the STRtree queries 
  * Bulk queries
* Bindings for new GEOS functionalities: 
  * `minimum_bounding_circle`, `minimum_bounding_radius`
  * `segmentize`
  * `reverse`
  * ...
* `dwithin`
* GeoJSON IO
* More informative repr with (truncated) WKT
* ...

jorisvandenbossche:
In addition we should also keep track of actual API changes (for which we didn't add warnings in 1.8). One potential example is https://github.com/shapely/shapely/issues/1023 (`.bounds` for empty geometries) and also https://github.com/shapely/shapely/issues/742 (empty constructors now return actual empty geometries instead of generic empty GeometryCollections)

PR #1421, opened 2022-06-13
CLN: removed unused shape_factory + HeterogeneousGeometrySequence

jorisvandenbossche:
This removes the no-longer used `shape_factory` method. Strictly speaking this is a public method, and so we could first deprecate it (in 2.0, or still in 1.8.3 if we do another release there). But since this is a method that was not mentioned in the docs (and not something people generally use, I think), it might also be OK to directly remove it.

Also removed the HeterogeneousGeometrySequence class, as its implementation became fully obsolete compared to GeometrySequence.


shapely/geometry/base.py
```diff
@@ -849,18 +836,6 @@ def __getitem__(self, key):
...
-class HeterogeneousGeometrySequence(GeometrySequence):
-    """
-    Iterative access to a heterogeneous sequence of geometries.
-    """
-
-    def __init__(self, parent):
-        super().__init__(parent, None)
-
-    def _get_geom_item(self, i):
-        return shapely.get_geometry(self.__p__, i)
```
jorisvandenbossche, line 861: This one overridden method was actually identical to the parent version.

mwtoews (approved):
Removing these looks good to me, as I believe they were only used internally, and was related to long deprecated and removed and forgotten functions like `geom_from_wkt`. I doubt they have any public use.

jorisvandenbossche:
>  it would be good to collect this as a breaking change to note in the 1.8 => 2.0 migration. Are you planning to do that at a later point, or would it be better to add to that with each PR that introduces a possibly breaking change?

For now I started keeping track of the items to list in https://github.com/shapely/shapely/issues/1280. Once we have some first content (I hope to start writing this soon), we could indeed also do it in each PR)

PR #1422, opened 2022-06-13
ENH: Update STRtree API (merge query / query-bulk, return only indices)

brendan-ward:
Progress toward #1386

This includes breaking changes for both Shapely 1.8 users and PyGEOS users; these are not yet noted in the docstrings, docs, or changelog.  Specifically:
* breaking from Shapely 1.8: ` query()` and `nearest()` now return indices only; no custom items or geometries.
* breaking from PyGEOS: `query_bulk()` is merged into `query`.  This requires only a tiny update for users because the overall API remains the same.

This focuses the base API for STRtree to always return indices.  It expands docstring examples to better highlight using these to retrieve geometries or custom items.

This combines `query_bulk()` into `query()`, so that shape of outputs is determined by the shape of the inputs:
- scalar query geometry returns array of shape (n, ) (from original `query()`)
- array of query geometries returns array of shape (2, n) (from `query_bulk()`)

This has a few nice effects:
- we struggled with the name of `query_bulk()`; this obviates needing to figure that out
- this is consistent with `nearest()` which makes it easier for us to communicate about the overall API of STRtree
- makes managing the test suite much easier
- eliminates the now redundant scalar-only query implementation in C; this now uses the original query_bulk implementation 

I tried to make the docstrings more clear and consistent:
- replaced use of `envelope` with `extent` throughout, since we were using a mix of both, and then having to define `envelope`.  My hope is that `extent` is a more ready accessible term to users (that GEOS uses an envelope data structure seems an internal implementation detail)
- replaced `indexes` with `indices`

I updated the docstrings and tests to use the specific geometry constructors, e.g., `Point()` (as per #1414) and also to avoid using `shapely` namespace for clarity.

I overhauled the STRtree tests, mostly to merge `query_bulk()` related tests into `query()` tests.  I added a few more unique test cases in some places.  I also added an array variant for almost every single scalar test case.  Possibly overkill, but thorough, and negligible overhead.

This includes minor updates to `nearest` to focus on returning indices only, but the more substantive overhaul to combine 'nearest` / `nearest_all` should happen in a follow-on PR immediately after this, since this changeset is already huge.

Since this is a big changeset that wasn't as easily doable in smaller easy-to-review chunks, I'll add notes throughout this PR to help guide review.


benchmarks/benchmarks.py
```diff
@@ -206,35 +206,35 @@ def time_tree_create(self):
         tree = pygeos.STRtree(self.polygons)
         tree.query(pygeos.points(0, 0))
 
-    def time_tree_query_bulk(self):
-        self.tree.query_bulk(self.polygons)
+    def time_tree_query(self):
```
brendan-ward, line 209: Benchmarks are untested; these changes are just to pick up the `query_bulk` => `query` transition.

shapely/strtree.py
```diff
@@ -34,317 +34,185 @@ class STRtree:
...
     ----------
     geoms : sequence
         A sequence of geometry objects.
-    items : sequence, optional
-        A sequence of objects which typically serve as identifiers in an
-        application. This sequence must have the same length as geoms.
     node_capacity : int, default 10
         The maximum number of child nodes per parent node in the tree.
 
-    Examples
```
brendan-ward, line 54: I think these examples are better handled in the specific methods, so I removed them from the class level.

shapely/strtree.py
```diff
@@ -34,317 +34,185 @@ class STRtree:
...
     def __init__(
         self,
         geoms: Iterable[BaseGeometry],
-        items: Iterable[Any] = None,
         node_capacity: int = 10,
     ):
         # Keep references to geoms
         self.geometries = np.asarray(geoms, dtype=np.object_)
+        # prevent modification
+        self.geometries.flags.writeable = False
```
brendan-ward, line 67: This is something to check against downstream implementations.  In theory, nobody _should_ be modifying these and expecting things to not break (we guard against segfaults related to doing so), but this sets a flag to help prevent them from attempting to modify the geometries.

shapely/strtree.py
```diff
@@ -34,317 +34,185 @@ class STRtree:
...
-        if self._has_custom_items:
-            return (STRtree, (self.geometries, self._items))
-        else:
-            return (STRtree, (self.geometries,))
+        return (STRtree, (self.geometries,))
 
-    def query_items(
-        self, geom: BaseGeometry, predicate=None, distance=None
-    ) -> Sequence[Any]:
+    def geometries(self):
```
brendan-ward, line 78: This is brought up to the top level so that we can provide a docstring for it; `tree.geometries` is used in the examples below, so it seemed important that this appear in the public API (which seems our overall intent with it?).

shapely/tests/test_strtree.py
```diff
@@ -65,151 +65,231 @@ def poly_tree():
...
+            [empty, empty_point, empty_line_string, point],
+            [empty, empty_point, empty_line_string, point],
+            [[3], [3]],
+        ),
     ],
 )
-def test_query(tree_geometry, geometry, expected):
-    tree = shapely.STRtree(np.array(tree_geometry))
-    assert_array_equal(tree.query_items(geometry), expected)
+def test_query_with_none_and_empty(tree_geometry, geometry, expected):
```
brendan-ward, line 170: This combines elements of the original `test_query_empty`, `test_query`, and similar tests from query_bulk.

shapely/tests/test_strtree.py
```diff
@@ -65,151 +65,231 @@ def poly_tree():
...
+    assert_array_equal(tree.query(geometry), expected)
 
 
 @pytest.mark.parametrize(
     "geometry,expected",
     [
         # points do not intersect
-        (shapely.points(0.5, 0.5), []),
+        (Point(0.5, 0.5), []),
+        ([Point(0.5, 0.5)], [[], []]),
```
brendan-ward, line 180: Pretty much every test case now has both a scalar and array variant.  In the case of no predicate, `interesects`, `dwithin`, these were migrated in from the associated `query_bulk` tests; in the other predicates they were added in this PR since they didn't have query_bulk specific tests.

shapely/tests/test_strtree.py
```diff
@@ -964,468 +1234,124 @@ def test_query_contains_properly_points(tree, geometry, expected):
...
     assert_array_equal(
-        poly_tree.query_items(geometry, predicate="contains_properly"), expected
+        poly_tree.query(geometry, predicate="contains_properly"), expected
     )
 
 
 ### predicate = 'dwithin'
 
 
-@pytest.mark.skipif(shapely.geos_version >= (3, 10, 0), reason="GEOS >= 3.10")
```
brendan-ward, line 1014: merged into other tests

shapely/tests/test_strtree.py
```diff
@@ -964,468 +1234,124 @@ def test_query_contains_properly_points(tree, geometry, expected):
...
-def test_query_dwithin_geos_version(tree):
-    with pytest.raises(UnsupportedGEOSVersionError, match="requires GEOS >= 3.10"):
-        tree.query_items(shapely.points(0, 0), predicate="dwithin", distance=1)
-
-
-@pytest.mark.skipif(shapely.geos_version < (3, 10, 0), reason="GEOS < 3.10")
```
brendan-ward, line 1020: merged into other tests

shapely/tests/test_strtree.py
```diff
@@ -964,468 +1234,124 @@ def test_query_contains_properly_points(tree, geometry, expected):
...
-@pytest.mark.parametrize(
-    "distance,match",
-    [
-        (None, "distance parameter must be provided"),
-        ("foo", "could not convert string to float"),
-        ([1.0], "distance must be a scalar value"),
```
brendan-ward, line 1026: The scalar distance error is no longer applicable under the combined `query()` / `query_bulk()` API (is always broadcast to match the input geometry array created from input geometries)

shapely/tests/test_strtree.py
```diff
@@ -964,468 +1234,124 @@ def test_query_contains_properly_points(tree, geometry, expected):
...
-        ([None], "distance must be a scalar value"),
-    ],
-)
-def test_query_dwithin_invalid_distance(tree, distance, match):
-    with pytest.raises(ValueError, match=match):
-        tree.query_items(shapely.points(0, 0), predicate="dwithin", distance=distance)
-
-
-@pytest.mark.skipif(shapely.geos_version < (3, 10, 0), reason="GEOS < 3.10")
```
brendan-ward, line 1035: These tests are mostly intact, but moved around just a little bit as part of merging them; unfortunate that these are showing up as being deleted rather than updated.  These are showing up a bit further down on the right in GH review panel (e.g., line 1330)

shapely/tests/test_strtree.py
```diff
@@ -1436,248 +1362,281 @@ def test_query_bulk_dwithin_invalid_distance(tree, distance, match):
...
 
 
-@pytest.mark.skipif(shapely.geos_version < (3, 6, 0), reason="GEOS < 3.6")
+@pytest.mark.skipif(geos_version < (3, 6, 0), reason="GEOS < 3.6")
 def test_nearest_empty_tree():
-    tree = shapely.STRtree([])
+    tree = STRtree([])
     # TODO what do we want here?
-    assert tree.nearest_item(point) is None
+    assert tree.nearest(point) is None
```
brendan-ward, line 1481: No substantive logic changes for `nearest` functions other than `nearest_item` => `nearest` and cleanup of geometry inputs / shapely namespace

src/strtree.c
```diff
@@ -326,97 +326,6 @@ static char evaluate_predicate(void* context, FuncGEOS_YpY_b* predicate_func,
   return errstate;
 }
 
-/* Query the tree based on input geometry and predicate function.
```
brendan-ward, line 329: This is no longer needed; the original `STRtree_query_bulk` did everything this did but for array inputs.

shapely/tests/test_strtree.py
```diff
@@ -65,151 +68,286 @@ def poly_tree():
...
+    tree = STRtree(arr)
     assert arr is tree.geometries
 
 
-def test_query_invalid_geometry(tree):
-    with pytest.raises(TypeError):
-        tree.query_items("I am not a geometry")
+# TODO(shapely-2.0) this fails on Appveyor, see
+# https://github.com/shapely/shapely/pull/983#issuecomment-718557666
+@pytest.mark.skipif(sys.platform.startswith("win32"), reason="does not run on Appveyor")
```
brendan-ward, line 134: This supersedes `tests/test_strtree.py::test_pickle_persistence` and instead of using a dedicated test script `unpickle-strtree.py` it writes it out temporarily.  This will make it easier to maintain alongside the tests.  The logic in the unpickle script is the same as before, except for removing unneeded print statements and comments.

tests/test_strtree.py
```diff
@@ -1,214 +0,0 @@
-import gc
```
brendan-ward, line 1: All tests are either superseded by more comprehensive test cases in `shapely/tests/test_strtree.py` or were migrated there (pickle, nearest exclusive).  Tests for custom items / geometries would be implemented if / when we implement specific classes for such (as per [#1386 comment](https://github.com/shapely/shapely/issues/1386#issuecomment-1147765811))

shapely/strtree.py
```diff
@@ -34,317 +34,185 @@ class STRtree:
...
+        can be extracted into arrays of the same length and order):
+        >>> records = [
+        ...     {"geometry": Point(0, 0), "value": "A"},
+        ...     {"geometry": Point(2, 2), "value": "B"}
+        ... ]
+        >>> tree = STRtree([record["geometry"] for record in records])
+        >>> import numpy as np
+        >>> items = np.array([record["value"] for record in records])
+        >>> items.take(tree.query(box(0, 0, 1, 1))).tolist()
+        ['A']
```
sgillies, line 204: @brendan-ward this is a very nice example, thank you!

sgillies:
@brendan-ward I'm now in favor of breaking the 1.8 STRtree API and supporting only storage of indexes. The API we have in this PR is more limited than 1.8's, but it does support higher level abstraction (like mapping arbitary Python objects to tree leaves), and a more limited design is easier to maintain over time.

shapely/strtree.py
```diff
@@ -34,317 +34,185 @@ class STRtree:
...
 
         Returns
         -------
-        ndarray with shape (2, n)
-            The first subarray contains input geometry indexes.
-            The second subarray contains tree geometry indexes.
+        ndarray with shape (n,) if geometry is a scalar
+            Contains tree geometry indices.
+
+        OR
```
jorisvandenbossche, line 141: (not fully sure if this will render nicely with sphinx, will do a check locally)

shapely/strtree.py
```diff
@@ -34,317 +34,185 @@ class STRtree:
...
+        >>> arr_indices.T.tolist()
+        [[0, 0], [0, 1], [1, 2], [1, 3]]
+
+        Retrieve the tree geometries by results of query:
+        >>> tree.geometries.take(indices).tolist()
+        [<shapely.Point POINT (0 0)>, <shapely.Point POINT (1 1)>]
+
+        Retrieve all pairs of input and tree geometries:
+        >>> np.array([boxes.take(arr_indices[0]),\
+tree.geometries.take(arr_indices[1])]).T.tolist()  # doctest: +NORMALIZE_WHITESPACE
```
jorisvandenbossche, line 175: Adding `NORMALIZE_WHITESPACE` shouldn't be needed, as this is a default flag added in setup.cfg

shapely/strtree.py
```diff
@@ -34,317 +34,185 @@ class STRtree:
...
+
+        Retrieve custom items associated with tree geometries (records can
+        be in whatever data structure so long as geometries and custom data
+        can be extracted into arrays of the same length and order):
+        >>> records = [
+        ...     {"geometry": Point(0, 0), "value": "A"},
+        ...     {"geometry": Point(2, 2), "value": "B"}
+        ... ]
+        >>> tree = STRtree([record["geometry"] for record in records])
+        >>> import numpy as np
```
jorisvandenbossche, line 201: numpy was already imported at the beginning of the examples section?

jorisvandenbossche (approved):
This is great, thanks for the PR!

Looking all good, just added a bunch of minor comments

shapely/strtree.py
jorisvandenbossche: With the new theme, this actually looks OK (it considers the "OR" also as a possible return value, which makes it bold, but visually that doesn't look problematic)

shapely/strtree.py
```diff
@@ -34,317 +34,185 @@ class STRtree:
...
-        [[0, 2], [0, 3], [0, 4], [1, 5], [1, 6]]
-        >>> # Query for tree geometries within 1 unit distance of input geometries
-        >>> tree.query_bulk([shapely.points(0.5, 0.5)], predicate='dwithin', distance=1.0).tolist()  # doctest: +SKIP
+        >>> from shapely import box, Point
+        >>> import numpy as np
+        >>> points = [Point(0, 0), Point(1, 1), Point(2,2), Point(3, 3)]
+        >>> tree = STRtree(points)
+
+        Query the tree using a scalar geometry:
+        >>> indices = tree.query(box(0, 0, 1, 1))
```
jorisvandenbossche, line 155:
```suggestion
        Query the tree using a scalar geometry:

        >>> indices = tree.query(box(0, 0, 1, 1))
```

All the example will need to be updated like this: the `>>>` block needs to be surrounded by a whitespace line to ensure sphinx recognizes it as a code block

shapely/strtree.py
```diff
@@ -34,317 +34,185 @@ class STRtree:
...
-        to the indexes of the input geometries and indexes of the tree geometries
-        associated with each.  To generate an array of pairs of input geometry
-        index and tree geometry index, simply transpose the results.
-
-        In the context of a spatial join, input geometries are the "left" geometries
-        that determine the order of the results, and tree geometries are "right" geometries
-        that are joined against the left geometries.  This effectively performs
-        an inner join, where only those combinations of geometries that can be joined
-        based on envelope overlap or optional predicate are returned.
+        In the context of a spatial join, input geometries are the "left"
```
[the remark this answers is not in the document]
brendan-ward: I agree that it would be good to split this out, but I'm unclear where the notes section would go if not in the docstring?

I'm thinking eventually we may want a handwritten page in the docs for STRtree rather than just the docstrings, where we can give more verbose examples around custom items and geometries; would this notes section go there?  (if so, I'd rather cut content from here and deal with that in a follow-on PR)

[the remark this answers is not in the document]
jorisvandenbossche: Ah, the "Notes" section is still part of the docstring (it's one of the recognized section titles by numpydoc), it's just a section that is typically put below the "Parameters" section (or also below the "Examples" section).

(and indeed, eventually it would be good to have a user guide page about the strtree)

shapely/strtree.py
```diff
@@ -34,317 +34,185 @@ class STRtree:
     at initialization. The tree is immutable and query-only, meaning that
     once created nodes cannot be added or removed.
 
-    An index is initialized from a sequence of geometry objects and
-    optionally a sequence of items. If items are not provided, the
-    indices of the geometry sequence will be used instead.
+    All operations return indices of the input geometries.  These indices
+    can be used to index into anything associated with the input geometries,
+    including the input geometries themselves, or custom items stored in
+    other object of the same length as the geometries.
```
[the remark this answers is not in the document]
brendan-ward: (native language doesn't make one an expert, mistakes still happen, like this one.  Language review is _always_ appreciated)

jorisvandenbossche:
Given that there are a couple of follow-up PRs that depend on this (updating nearest/nearest_all, updating the docstrings for the new repr, ..), let's merge this. But feel fee to still do a review of the changes here, then we can address those in a follow-up.

jorisvandenbossche:
Thanks @brendan-ward !

PR #1424, opened 2022-06-15
TST: Update benchmarks for pygeos=>shapely

brendan-ward:
This fixes the benchmark suite as part of the PyGEOS to Shapely 2.0 migration.  Other than minor whitespace changes, the only changes here are instructions at the top of `benchmarks.py` and find / replace `pygeos` => `shapely`.

Note: `STRtree` benchmarks are currently broken pending #1422 (they were written for the indices-only tree API).  The other benchmarks work properly.

### DIFF
--- a/asv.conf.json
+++ b/asv.conf.json
@@ -4,10 +4,10 @@
     "version": 1,
 
     // The name of the project being benchmarked
-    "project": "pygeos",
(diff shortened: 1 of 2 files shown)


jorisvandenbossche (approved):
Thanks!

Issue #1205, opened 2021-10-26
Drop Python 3.6 supported (end of life Dec 2021)


EwoutH:
The security-only support for Python 3.6 ends in Dec 2021, after Python 3.6 will be end-of-life (see [PEP 494](https://www.python.org/dev/peps/pep-0494/)). With 1.8 released as a stable version supporting Python 3.6, would it be time to drop Python 3.6 support, looking forward to the Shapely 2.0 release? Especially as Python 3.10 support is ramped up.

mwtoews:
I'm in two minds on when to drop Python 3.6 support, similar to @caspervdw's [comment](https://github.com/pygeos/pygeos/issues/414#issuecomment-954986554). On one end, I'd like to see a smaller CI footprint and would like to align the project closer to [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) (although NumPy is py38+). But on the other end, there is not much extra burden in maintaining older Python3 compatibility. Lots of folks, including myself, are still using Ubuntu 18.04 (with EOL April 2023), which natively supports Python 3.6, NumPy 1.13 and GEOS 3.6.

Any other thoughts on this? Does anyone plan to use shapely-2.0 with Python 3.6 for at least another year? Are there any planned development features that can only be implemented with Python 3.7 or later?

jorisvandenbossche:
I don't think we have much maintenance burden in supporting Python 3.6 for now (I don't recall an issue or difficulty while contributing related to this recently). So from that point of view, I don't feel strongly about dropping support sooner rather than later. And if there are people (even from the core team) still using it, that seems as a good reason to keep it for now.

I think we will mostly "miss out" on new features by keeping older Python support. But with Python 3.6 we eg already have f-strings and ordered dicts. And I don't directly see features in Pyton 3.7 I would really miss (we don't yet use type annotations anyway, and I don't think we have use for dataclasses)

mwtoews:
PR #1426 to move static project metadata to `pyproject.toml` ([PEP 621](https://peps.python.org/pep-0621/)) is blocked by Python 3.6 (because setuptools 61+ support starts from Python 3.7). This seems like a good reason to drop Python 3.6 for Shapely 2.0.

PR #1426, opened 2022-06-20
Move static metadata from setup.py to pyproject.toml, add sdist check

mwtoews:
This follows on from #1376 to move static metadata from `setup.py` to `pyproject.toml` as is supported by [setuptools](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) and described by [PEP 621](https://peps.python.org/pep-0621/). Note that with `python -m build` a message is shown:
> Support for `[tool.setuptools]` in `pyproject.toml` is still *beta*.

This also removes "wheel" from build system requirements, since it is exposed automatically by setuptools. More on that here: https://github.com/pypa/setuptools/commit/f7d30a9529378cf69054b5176249e5457aaf640a

Furthermore, after building the source distribution, use `twine check` to ensure that it looks adequate by PyPA.


mwtoews:
There seems to be some trouble with this move:

- Perhaps "wheel" is not automatically exposed by setuptools as though, so it should probably be re-added. Or it's due to the bogus version mismatch.
- `version` is required to be in `pyproject.toml`, even though it is dynamically done via versioneer in `setup.py`. Due to this requirement, a "0" placeholder was used, which is fine for sdist, but this version is used for the binary wheel (e.g. "shapely-0-cp38-cp38-linux_x86_64.whl"). Not sure if there is a workaround for projects that use versioneer and want to declare PEP 621 project metatdata. Note that pre shapely-2 uses [`__version__` in `shapely/__init__.py`](https://github.com/shapely/shapely/blob/maint-1.8/shapely/__init__.py), and @jorisvandenbossche provides some rational to prefer versioneer for shapely-2 [here](https://github.com/shapely/shapely/issues/962#issuecomment-961071354).

jorisvandenbossche:
@mwtoews you might need to make the `version` field in the toml table "dynamic": https://peps.python.org/pep-0621/#dynamic

Maybe you can allow the cibuildwheel step in release.yml workflow to fail, so that we still upload the other artifacts? (the mac wheels seems to be OK, its linux/windows where the tests are failing. And it seems to not find any tests)

mwtoews:
Thanks @jorisvandenbossche it make sense to use "dynamic" for this instance, and now the version works perfectly. The presence or absence of "wheel" does not seem to be a contributing factor to breaking cibuildwheel on Linux/Windows. Do you mean to add `continue-on-error: true` somewhere in relase.yml temporarily?

mwtoews:
The root cause is shown in the "Building wheel..." section:
```
      + python -m pip wheel /project --wheel-dir=/tmp/cibuildwheel/built_wheel --no-deps
  Processing /project
    Installing build dependencies: started
    Installing build dependencies: finished with status 'done'
    Getting requirements to build wheel: started
    Getting requirements to build wheel: finished with status 'done'
    Preparing metadata (pyproject.toml): started
    Preparing metadata (pyproject.toml): finished with status 'done'
  Building wheels for collected packages: UNKNOWN
    Building wheel for UNKNOWN (pyproject.toml): started
    Building wheel for UNKNOWN (pyproject.toml): finished with status 'done'
    Created wheel for UNKNOWN: filename=UNKNOWN-0+untagged.1.g32f5822-cp36-cp36m-linux_x86_64.whl size=771555 sha256=abebdb5777081ae5b911ab635b5c2f2f2cf987483bd6cb925b4d30a69cb5770c
    Stored in directory: /tmp/pip-ephem-wheel-cache-bghbt1yi/wheels/2e/83/83/fb24504ded536a3c91dd77593aa9037a98b0009fb86c7ebcc0
  Successfully built UNKNOWN
```
Looking at UNKNOWN-0 it seems some component does not know how to extract the project metadata following [PEP 621](https://peps.python.org/pep-0621/). I've tried a `pip install --upgrade pip` to see if that fixes it, but to no avail. I can't seem to reproduce this with a local docker image. It's possible that some component either needs to be upgraded, or it is not PEP 621 conformable yet.

pyproject.toml
```diff
@@ -1,7 +1,61 @@
 [build-system]
-requires = ["setuptools", "wheel", "cython", "oldest-supported-numpy"]
+requires = [
+    "cython",
+    "oldest-supported-numpy",
+    "setuptools==59.6.0; python_version=='3.6'",
```
jorisvandenbossche, line 5: But don't we need a more recent setuptools version to have this working with the metadata in pyproject.toml? 

(this might be another reason to drop Python 3.6 support)

jorisvandenbossche: (the wheel build is also still failing for Python 3.6 (the testing step))

jorisvandenbossche (commented):
The changes in pyproject.toml itself look good to me, but so it is still not fully working ..

jorisvandenbossche:
To resolve the "untagged" git version in `shapely-0+untagged.1.gb0d429d--....whl`, we might need (add the `fetch-depth`) in `release.yml`:

```
      - name: Checkout source
        uses: actions/checkout@v2
        with:
          fetch-depth: 0
```

to ensure that git has the tags available (it will typically work when doing a release, as then the last commit is also the one that is tagged, but it is probably good to have it working properly for non-tagged commits as well, so we actually see it is working before doing a release)

pyproject.toml
jorisvandenbossche: xref https://github.com/shapely/shapely/issues/1205 about Python 3.6 support

mwtoews:
Thanks for the tips @jorisvandenbossche but the same error [described above](https://github.com/shapely/shapely/pull/1426#issuecomment-1162511115) persists. Interesting that #1430 using `setup.cfg` was so much simpler. I'll see if I can start some discussions upstream to clear-up what is happening with the two approaches.

jorisvandenbossche:
> Interesting that https://github.com/shapely/shapely/pull/1430 using setup.cfg was so much simpler

One part of the complexity here is actually not related to this approach (we just happened to discuss it here): the untagged version for which you added fetch-depth also happens in the other PR and on main.

pyproject.toml
```diff
@@ -1,7 +1,59 @@
...
+]
+
+[project.optional-dependencies]
+test = ["pytest"]
+docs = ["sphinx", "numpydoc"]
+
+[project.urls]
+documentation = "https://shapely.readthedocs.io/"
+repository = "https://github.com/shapely/shapely"
+changelog = "https://github.com/shapely/shapely/blob/main/CHANGELOG.rst"
```
mwtoews, line 49: Just noticed this was for PyGEOS; shapely 1.8 has CHANGES.txt. See also #1442

pyproject.toml
```diff
@@ -1,7 +1,59 @@
...
+    "oldest-supported-numpy",
+    "setuptools>=61.0.0",
+]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "shapely"
+dynamic = ["version"]
+authors = [
+    {name = "Sean Gillies"},
```
mwtoews, line 13: @sgillies did you want your email here? It was there for 1.8 ([setup.py](https://github.com/shapely/shapely/blob/maint-1.8/setup.py#L175)).

sgillies: No, name only is fine :pray:

sgillies (approved):
Great! Thanks, @mwtoews !

snowman2:
Side note: I recommend waiting on this move to metadata in pyproject.toml. Debian builds are not ready for it yet: https://github.com/pyproj4/pyproj/discussions/1134#discussioncomment-3592703

mwtoews:
Thanks @snowman2 for the heads up, I didn't consider distribution packaging constraints! Any thoughts by @jorisvandenbossche and @sgillies to use to a more stable solution similar to #1430 ?

PR #1428, opened 2022-06-21
ENH: Update STRtree nearest neighbor API

### LINKED ISSUE #1423: ENH: STRtree nearest / nearest_all API
Migrating the discussion of the `nearest` / `nearest_all` API from [here](https://github.com/shapely/shapely/issues/1386#issuecomment-1152959469) to a dedicated issue so it doesn't get lost in the larger discussion of the STRtree API.

Background:
* `nearest` in Shapely pre-2.0 included an `exclusive` parameter, which if `True` did not return an output nearest geometry if it was equal (at the geometry level) to the input geometry
* `nearest` for Shapely 2.0 came from PyGEOS, does not have an `exclusive` parameter, always returns no more than one output geometry per input geometry (even if multiple intersecting / equidistant results exist), and was intended to be a lightweight wrapper around the underlying GEOS nearest function (so it lacks additional parameters)
* `nearest_all` for Shapely 2.0 came from PyGEOS and includes the ability return multiple equidistant results per input geometry, a `max_distance` parameter to constrain matches, and return distances.  The underlying C implementation is a bit more involved than the above (prescreens for geometries within `max_distance`, more involved callback, etc).
* using `nearest_all` with a large `max_distance` (rather than `None`) degrades performance because of the prescreening step; otherwise the additional overhead of `nearest_all` vs `nearest` is pretty negligible.

After studying the code a bit, it looks like the best way to include the `exclusive` parameter is to leverage the C implementation behind `nearest_all`, because the best place to check for equality of the geometries and force a broader search is within the C callback used with GEOS.

Thus it looks like there is an opportunity to combine `nearest` / `nearest_all` functions to provide a more feature-rich function that provides all the parameters (defaults listed first):

```Python
def nearest(
        geometries, exclusive=False|True,
        return_distance=False|True,
        max_distance=None|<float value>, 
        on_equidistant="first"|"all") 
              -> array[int] or tuple(array[int], array[float]) or array[2, int] or tuple(array[2, int], array[float])
```

Where shape of outputs is based on shape of inputs and `return_distance` parameter.

Note: whether this should be a single function where outputs are based on whether the input geometry is scalar or array of geometries vs splitting into separate methods is part of a larger discussion in #1386.

Also note that this is not a K nearest neighbors implementation; there is no utility in another variant of `on_equidistant` to get at some other number of matches, since it only applies to equidistant neighbors.

For scalar geometries:
* `return_distance=False`, `on_equidistant="first"` => index of nearest tree geometry: `array[int]` (array of length <= 1)
* `return_distance=False`, `on_equidistant="all"` => index of nearest tree geometries: `array[int]` (array of length n)
* `return_distance=True` =>  index of nearest tree geometry, distance to it:  `tuple(array[int], array[float])`

This mimics the API for `query`; results are always at least an array for scalar geometries (arrays may be empty, depending on parameters).

Here's how that would look (untested):

```Python
>>> tree = STRtree([Point(0, 0), Point(1, 1)])
>>> tree.nearest(Point(0.5, 0.5)).tolist()
[0]
>>> tree.nearest(Point(0.5, 0.5), on_equidistant="all").tolist()
[0, 1]
>>> tree.nearest(Point(0.5, 0.5), return_distance=True)
(array([0]), array([0.7071067811865476]))
>>> tree.nearest(Point(0.5, 0.5), return_distance=True, on_equidistant="all")
(array([0, 1]), array([0.7071..., 0.7071...]))
>>> tree.nearest(Point(0.5, 0.5), max_distance=0.1).tolist()
[]
>>> tree.nearest(Point(0.5, 0.5), max_distance=0.1, return_distance=True)
(array([]), array([]))
>>> tree.nearest(Point(0, 0), exclusive=True).tolist()
[1]
>>> tree.nearest(None, exclusive=True).tolist()
[]
```

For geometry arrays:
* `return_distance=False`, `on_equidistant="first"` => array of indexes of input geometries, indexes of nearest tree geometry per input geometry: `array[2, int]` (array is length of input geometries)
* `return_distance=False`, `on_equidistant="all"` => array of indexes of input geometries, indexes of nearest equidistant tree geometries per input geometry: `array[2, int]` (array is >= length of input geometries)
* `geometries`=array, `return_distance=True` =>  array of indexes of input geometries, indexes of nearest tree geometries per input geometry, distances to them `tuple(array[2, int], array[float])

Any input geometry that is `None`, empty, or doesn't have a nearest match within max distance (if set), distinct from itself (if `exclusive`) will be omitted from the outputs (same as for query / query_bulk).

Here's how that would look (untested):

```Python
>>> tree = STRtree([Point(0, 0), Point(1, 1)])
>>> tree.nearest([Point(0, 0), Point(0.5, 0.5)]).tolist()
[[0, 1], [0, 0]]
>>> tree.nearest([Point(0, 0), Point(0.5, 0.5)], return_distance=True)
(array([[0, 1], [0, 0]]), array([0., 0.7071...]))
>>> tree.nearest([Point(0, 0), Point(0.5, 0.5)], return_distance=True, on_equidistant="all")
(array([[0, 1, 1], [0, 0, 1]]), array([0., 0.7071..., 0.7071...]))
>>> tree.nearest([Point(0, 0), Point(1, 1)], exclusive=True).tolist()
[[0, 1], [1, 0]]
>>> tree.nearest([Point(0.5, 0.5)], max_distance=0.1).tolist()
[[], []]
```

The main decision point here is allowing the output per input geometry to always be an array instead of a scalar geometry; this keeps the outputs the same structure regardless of `on_equidistant` parameter and interactions with `max_distance` and `exclusive` parameters.

brendan-ward:
Resolves #1423

From discussion with @jorisvandenbossche @caspervdw @sgillies , the decision was made to keep the `nearest()` method as is (from PyGEOS), and handle all parameters and more advanced functionality in a second method called `query_nearest()`.

`query_nearest()` replaces the `nearest_all` method from PyGEOS and adds additional capabilities:
* output shapes are determined by the shape of the input geometry in the same way as `query()` and optionally wrapped into a tuple based on `return_distance` parameter: scalar input produces either `array[n]` or `tuple(tree_indices: array[n], distances: array[n])`; it produces either `array[2,n]` of input indices, tree indices, or `tuple(indices: array[2,n], distances: array[n])`
* `exclusive` parameter is added from Shapely 1.8 to prevent matching geometries that are equal to the input geometry
* `all_matches` parameter is added to control whether this method returns all equidistant / intersecting matches (default), or the first (both cases are used by GeoPandas `sjoin_nearest`; this simplifies that usage).

This should now present a signature similar to `query()`.

Both `exclusive` and `all_matches` involve low-complexity changes to the underlying C implementation: we add these as members of a struct passed to the GEOS nearest neighbor callback, and check geometric equality in the callback if `exclusive` is True, and add an epsilon to the returned distance if `all_matches` is True to force GEOS to continue finding nearest neighbors after finding the first.

Lots of test cases were added to handle both scalar and array input geometries.


shapely/strtree.py
```diff
@@ -342,28 +314,55 @@ def nearest(self, geom, exclusive: bool = False) -> Union[Any, None]:
         if self._tree.count == 0:
             return None
 
-        return self._nearest_idx(geom, exclusive)
+        geometry_arr = np.asarray(geometry, dtype=object)
+        # TODO those changes compared to _tree.nearest output should be pushed into C
```
brendan-ward, line 318: I'm unclear on the meaning of this comment (it came originally from `_nearest_idx` which is replaced here).  Is this something we should tackle in this PR too?

shapely/strtree.py
```diff
@@ -389,25 +401,86 @@ def nearest_all(self, geometry, max_distance=None, return_distance=False):
...
+         [<shapely.Point POINT (2.25 2.25)>, <shapely.Point POINT (2 2)>]]
+
+        All intersecting geometries in the tree are returned by default:
+
+        >>> tree.query_nearest(box(1,1,3,3)).tolist()
+        [1, 2, 3]
+
+        Set all_matches to False to to return a single match per input geometry:
+
+        >>> tree.query_nearest(box(1,1,3,3), all_matches=False).tolist()  # doctest: +SKIP
```
brendan-ward, line 445: I removed most of the `# doctest: +SKIP` now that the doctest infrastructure is using GEOS > 3.6.  This is retained here becuse the output is nondeterministic between GEOS versions; the output may be either `[0]` or `[1]`; both are equally valid.

shapely/strtree.py
jorisvandenbossche: I think that I wrote this, and that it meant that we can in principle let the C tree code handle this check. Because we currently have this in STRtree_nearest:

https://github.com/shapely/shapely/blob/9b3fa0ae3dfa802a56bb01b21ccf8411a20d5a9d/src/strtree.c#L680-L682

which has different behaviour (it doesn't raise an error but actually ignore missing or empty geometries). Although in practice we should also never end up in that if check now, given that we check here in advance for it (and keeping the check in python might be easier to raise a custom error).

jorisvandenbossche (commented):
Didn't yet look at the tests, but generally looks good to me!

shapely/strtree.py
brendan-ward: Doing it up front is maybe a little bit of extra overhead (up to 2 passes over the array) compared to moving into C as part of our main loop, but it provides the advantage of raising the error earlier rather than partway through finding nearest neighbors.

I think we can delete the comment?

jorisvandenbossche: > I think we can delete the comment?

Yes, sounds good

jorisvandenbossche:
@brendan-ward this is ready from your point of view? Will try to do a pass over the test changes shortly.

brendan-ward:
Yes, this is ready for review, though I'm just now seeing your comment about transpose above (will address shortly).

jorisvandenbossche (approved):
Test changes look good as well, and thanks for the updates!

jorisvandenbossche:
Given that we discussed this at the meeting, going to merge this to move forward. Thanks a lot @brendan-ward ! 

(@sgillies @caspervdw doing a check of the new API to ensure it matches with what we discussed is certainly still welcome!)

PR #1430, opened 2022-06-21
Move static metadata from setup.py to setup.cfg, add sdist check

mwtoews:
This is an alternative proposal to #1426 following [these setuptools instructions](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html) to move static metadata from `setup.py` to `setup.cfg`. Curious to see how the outcome of release.yml compares.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,8 +30,10 @@ jobs:
 
       - name: Build a source tarball
         run: |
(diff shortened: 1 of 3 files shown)


mwtoews:
Closing in favor of #1426, which embraces [PEP 621](https://peps.python.org/pep-0621/) to store project metadata in `pyproject.toml`.

Issue #1281, opened 2022-01-03
New methods to add to the Geometry classes?


jorisvandenbossche:
We now have bindings for a bunch of new GEOS functionalities (through the merger of pygeos), so we should also evaluate for which of those we want to add equivalent methods on the Geometry subclasses in addition to the top-level function. 

Potential candidates:

- [X] `dwithin` - xref #1496
- [x] `contains_properly` - #1605
- [x] `minimum_bounding_circle`, `minimum_bounding_radius`
- [X] `reverse` - xref #1457
- [X] `segmentize` - xref #1434

theroggy:
All methods listed in this issue have been implemented for some time...

PR #1434, opened 2022-06-26
ENH: Add segmentize method to BaseGeometry

gpapadok:
As requested in #1281.


shapely/geometry/base.py
```diff
@@ -745,6 +745,36 @@ def interpolate(self, distance, normalized=False):
...
+            Additional vertices will be added so that all line segments are no
+            greater than this value.  Must be greater than 0.
+        **kwargs
+            For other keyword-only arguments, see the
+            `NumPy ufunc docs <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs>`_.
+
+        Examples
+        --------
+        >>> line = shapely.Geometry("LINESTRING (0 0, 0 10)")
+        >>> line.segmentize(tolerance=5)  # doctest: +SKIP
```
jorisvandenbossche, line 770: I know the docstring of `shapely.segmentize` has this `# doctest: +SKIP` as well, but I _think_ we don't actually need it anymore (because we now build the docs / run the doctests using GEOS 3.10)

shapely/geometry/base.py
```diff
@@ -745,6 +745,36 @@ def interpolate(self, distance, normalized=False):
...
+        <shapely.LineString LINESTRING (0 0, 0 5, 0 10)>
```
jorisvandenbossche, line 771:
```suggestion
        <LINESTRING (0 0, 0 5, 0 10)>
```

We just changed the repr a bit (see https://github.com/shapely/shapely/pull/1412), sorry for the churn

shapely/geometry/base.py
```diff
@@ -745,6 +745,36 @@ def interpolate(self, distance, normalized=False):
...
+        >>> poly = shapely.Geometry("POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0))")
+        >>> poly.segmentize(tolerance=5)  # doctest: +SKIP
+        <shapely.Polygon POLYGON ((0 0, 5 0, 10 0, 10 5, 10 10, 5 10, 0 10, 0 5, 0 0))>
```
jorisvandenbossche, line 774:
```suggestion
        <POLYGON ((0 0, 5 0, 10 0, 10 5, 10 10, 5 10, 0 10, 0 5, 0 0))>
```

shapely/geometry/base.py
```diff
@@ -745,6 +745,36 @@ def interpolate(self, distance, normalized=False):
...
+        """
+        return shapely.lib.segmentize(self, tolerance, **kwargs)
```
jorisvandenbossche, line 776:
```suggestion
        return shapely.segmentize(self, tolerance, **kwargs)
```

To make it consistent with the other methods in this file, we use the top-level function here (in addition, this also ensures that we re-use the additional checks from that function compared to the `lib` version, for example in this `shapely.segmentize` checks for the GEOS version and raises an informative error message if it is too old)

jorisvandenbossche (commented):
@gpapadok thanks a lot for the contribution!

We don't yet do this consistently, but I think it would be a good idea that we start adding a small test for those methods, for example we can put that in `shapely/tests/geometry/test_geometry_base.py`. A small test that just confirms that the method is equivalent to the top-level function (which is tested more extensively) is more than enough (but it prevents that we accidentally remove the method or not pass through any arguments), something like:

```
def test_segmentize():
    line = LineString([(0, 0), (0, 10)])
    result = line.segmentize(tolerance=5)
    assert result.equals(shapely.segmentize(line, tolerance=5))
```

gpapadok:
All done. But pulling the main branch broke a lot of tests.

jorisvandenbossche:
> pulling the main branch broke a lot of tests.

Ah, that's actually an issue here: the `segmentize` method is only available starting with GEOS 3.10, so we need to skip the test on builds with older GEOS. 
Adding a `@pytest.mark.skipif(shapely.geos_version < (3, 10, 0), reason="GEOS < 3.10")` decorator to the test that you added should fix this.

shapely/tests/geometry/test_geometry_base.py
```diff
@@ -83,3 +84,10 @@ def test_GeometryType_deprecated():
         geom_type = geom.geometryType()
 
     assert geom_type == geom.geom_type
+
+
+@pytest.mark.skipif(shapely.geos_version < (3, 10, 0), reason="GEOS < 3.10")
+def test_segmentize():
+    line = LineString([(0, 0), (0, 10)])
+    result = line.segmentize(tolerance=5)
+    assert result.equals(shapely.segmentize(line, tolerance=5))
```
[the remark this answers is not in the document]
jorisvandenbossche: > the test result should be independent from the method:

Note that it was (intentionally) comparing to the _function_, not method (itself). Because in my thinking that was sufficient (we basically only want to test that the method uses the function under the hood, and it's the function that is more extensively tested for actual results)

Now, doesn't matter that much, the updated way is fine as well, I just wanted to note that the previous way was still testing something useful ;)

mwtoews:
Thanks, this looks good.

But a wee bit of discussion first. We like to compare the capabilities to PostGIS, which has [ST_Segmentize](https://postgis.net/docs/ST_Segmentize.html) (based on [internal](https://github.com/postgis/postgis/blob/6febe57e76d2239ea9779c489eff376f3ed76e72/postgis/lwgeom_functions_basic.c#L1861-L1926) [liblwgeom](https://github.com/postgis/postgis/blob/b6cce01cdaf6d31169d73871c374cae332f66992/liblwgeom/ptarray.c#L397-L473)), which uses "max_segment_length" instead of "tolerance". Should shapely also use "max_segment_length"? I've found "tolerance" to be slightly overloaded with other functions, and there could be benefits with parameter names that are more descriptive. But I also don't mind "tolerance", it just requires the user to look-up the docs.

Note that this functionality in the GEOS C API is called [`GEOSDensify`](https://github.com/libgeos/geos/blob/e3abcc8fc0f60035c7d8bda65e2417873f6da872/capi/geos_c.h.in#L3885-L3898) with "tolerance".

jorisvandenbossche:
@mwtoews moved your comment to https://github.com/shapely/shapely/issues/1454 (since it will require code changes not only the code touched here in this PR, I thought it would be better to leave for a separate PR, and so that this one can be merged)

jorisvandenbossche (approved):

jorisvandenbossche:
@gpapadok thanks again for the PR! 
If you would like to add other methods from https://github.com/shapely/shapely/issues/1281 as well, that's certainly welcome ;)

PR #1437, opened 2022-07-06
CI: add testing for 2022, including GEOS-3.11

mwtoews:
GEOS 3.11 was recently [released](https://lists.osgeo.org/pipermail/geos-devel/2022-July/010728.html), so adding it to the CI matrix.

According to [PEP 664](https://peps.python.org/pep-0664/), Python 3.11 will be released October 2022, so this "2022" matrix item should be updated when supported by GHA.


brendan-ward (approved):
Thanks @mwtoews

Issue #1436, opened 2022-07-05
Shapely relying on inconsistent GEOS behavior? (Using geos 3.11.0)


noamgat:
## Expected behavior and actual behavior.

We installed shapely 1.8.2 with geos 3.11.0 (we know that shapely was built with 3.10.2)
parallel_offset stopped producting results according to the behavior (not reverse direction) when offseting right.
""The side parameter may be ‘left’ or ‘right’. Left and right are determined by following the direction of the given geometric points of the LineString. Right hand offsets are returned in the reverse direction of the original LineString or LineRing, while left side offsets flow in the same direction."
We opened a ticket in GEOS : https://github.com/libgeos/geos/issues/650#issuecomment-1175202840 and got an answer that this is something that shapely encountered and documented, but it is not guaranteed behavior and not part of the contract.

I'm mainly opening this issue for the Shapely project to take note in the next GEOS update, to either update the documentation or add a post-process to parallel_offset to maintain this behavior.

We solved the issue by downgrading GEOS to 3.10.2, the same version that shapely 1.8.2 uses.

## Steps to reproduce the problem.

See expected / actual section

## Operating system

Windows and Linux

## Shapely version and provenance

1.8.2 with Geos 3.11.0

noamgat:
https://github.com/libgeos/geos/commit/0a32f150493afe818c126235be324ad6aa4bfecd#diff-d6792d71ab494d00c051c2dabbd8b55d2ee8d2ce542a175f07eea26c4ec01a5eR180

jorisvandenbossche:
@noamgat Thanks for the report and analysis!

Given the response of the GEOS developers, I think the best action to take is to update the Shapely documentation to remove this guarantee (and thus follow GEOS).

noamgat:
No problem.
While I have no problem with the decision, I think it requires treating this as a potentially breaking API change - as users of shapely (like myself) were relying on this behavior (and rightfully so, given the documentation).
So perhaps there should be an intermediate version with a deprecation warning on parallel_offset?

mwtoews:
Note that this issue was caught in CI when adding GEOS 3.11 for `maint-1.8`:
```
______________ OperationsTestCase.test_parallel_offset_linestring ______________
self = <tests.test_parallel_offset.OperationsTestCase testMethod=test_parallel_offset_linestring>
    def test_parallel_offset_linestring(self):
        line1 = LineString([(0, 0), (10, 0)])
[...]

PR #1357, opened 2022-04-01
TST: fix tests for GEOS main

jorisvandenbossche:

PR #1439, opened 2022-07-07
CI: add testing for 2022, including GEOS-3.11 (maint-1.8)

mwtoews:
Similar to #1437 but for the maintenance branch for shapely 1.8

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-2019]
         architecture: [x64]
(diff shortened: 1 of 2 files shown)


mwtoews:
Failures are related to #1436

Curious that these were not triggered with the main branch.

jorisvandenbossche:
> Curious that these were not triggered with the main branch.

It seems that is because I "fixed" those tests a while ago (to have the tests passing with GEOS main): https://github.com/shapely/shapely/pull/1357#pullrequestreview-929155296

(now, I did solve it differently than what I just pushed here)

mwtoews:
Thanks @jorisvandenbossche for "fixing" the CI test, for now at least. A proper fix will need to come later (if it's even possible).

PR #1448, opened 2022-07-13
Add perspective (homography) transform

Min-Sheng:
The change in this PR is to add perspective transform like cv2.warpPerspective.


idanmiara:
Hi, I'm not sure, but maybe if you added tests this would be more likely to get merged.

Min-Sheng:
> Hi, I'm not sure, but maybe if you added tests this would be more likely to get merged.

Hi, I have added an unittest for perspective (homography) transformation. Please check it out.

sgillies:
I'm uncertain about this one. It looks like the homography matrix needed by perspective_transform() would need to be computed by OpenCV, yes? And I think I see that OpenCV also has the function to make an affine transformation from the homography matrix, is that right? If that is so, I think this function could be eliminated and users could just use shapely's transform() with that OpenCV-produced affine transformation matrix.

Perspective transformations don't seem to be compatible with GEOS and its mathematical model. I think it would be weird to have a function that produced Shapely geometries that can't be operated on.

sgillies (changes requested):
I'm concerned that geometries returned from this function aren't compatible with GEOS.

Min-Sheng:
> I'm uncertain about this one. It looks like the homography matrix needed by perspective_transform() would need to be computed by OpenCV, yes? And I think I see that OpenCV also has the function to make an affine transformation from the homography matrix, is that right? If that is so, I think this function could be eliminated and users could just use shapely's transform() with that OpenCV-produced affine transformation matrix.
> 
> Perspective transformations don't seem to be compatible with GEOS and its mathematical model. I think it would be weird to have a function that produced Shapely geometries that can't be operated on.

Hi @sgillies,

I understand your concerns regarding the compatibility of the perspective (homography) transform with Shapely and GEOS, as well as the potential redundancy with existing OpenCV functions. However, I would like to present some arguments in favor of including this function in Shapely.

Consider a project that involves analyzing and manipulating geospatial data from aerial imagery. The aerial images require rectification due to distortions caused by camera perspective and terrain. Using Shapely with the perspective transform feature allows users to perform the following tasks seamlessly:

1. Load geospatial data, such as building footprints and road networks, into Shapely geometries.
2. Use  the homography matrix computed from OpenCV and then **apply the perspective transform on the Shapely geometries to align them with the aerial images**.
3. Perform further operations, such as calculating areas, intersections, and distances, using Shapely and GEOS functions on the transformed geometries.

By having the perspective transform in Shapely, users can perform these tasks within a single library, reducing complexity, avoiding potential data conversion issues, and offering a more cohesive workflow. This makes Shapely more versatile for a wider range of geospatial applications and analysis tasks.

Thanks again for your considering.

sgillies:
@Min-Sheng I really appreciate your explanation and I am so sorry that I haven't been able to respond until now. Limiting the scope of this project is necessary to its sustainability. As it is, we're struggling to keep up with changes coming up from GEOS and down from Python. I think it is best if we preserve the the boundaries between what shapely does (planar geometry for things on maps, viewed from directly overhead, essentially) and what OpenCV does (perspective, 3D surface modeling, and so much more). The three tasks you enumerated are interesting, for sure. I've done some related work and may again, and so I'm sympathetic. But I think the tasks are better left to a different Python package.

idanmiara:
@Min-Sheng this sounds like an interesting application for Shapely.
If this isn't approved to be included in Shapely, maybe there should be a `shapely-utils` package where it can be located (inspired by https://pypi.org/project/gdal-utils/)

sgillies:
@idanmiara I don't agree that there should be a shapely-utils package and this is not the place to socialize the idea.

caspervdw:
To me, anyone is free to start any package as long as it is clear that we (the shapely maintainers) are not related in any way. The name shapely-utils would be slightly misleading in that aspect so I would suggest a more specific name.

Anyway, that is besides the point here. @idanmiara I understand your frustration with the slow feedback on your PRs. The issue here (and also with a bunch of your PRs) is that the code is already there while there was no consensus about whether shapely actually needs the feature. So the PR review turns into a discussion with the PR author in a special role (which isn’t good).

So I propose the following for this PR (and likewise for some other PRs that have undiscussed new features): we’ll close any of them that have no related issue, are more than 2 months old, and have a disapproval from any maintainer. Then it is up to the PR author to make an issue starting a discussion on the new feature.

As we don’t have any governance (yet, see #1822) it is understandable that this is happening now.

PR #1451, opened 2022-07-16
Fix: update geometry repr used in docstrings to pass doctests

brendan-ward:
Updates to the latest geometry repr format throughout.

Also removes skipping of some of the tests now that tests are run on a more recent version of GEOS and a stable platform.

### DIFF
--- a/shapely/constructive.py
+++ b/shapely/constructive.py
@@ -1,6 +1,5 @@
 import numpy as np
 
-from . import Geometry  # NOQA
 from . import lib
 from .decorators import multithreading_enabled, requires_geos
 from .enum import ParamEnum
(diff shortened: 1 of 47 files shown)


jorisvandenbossche:
Won't this conflict with https://github.com/shapely/shapely/pull/1414? (as that PR also updates the repr output together with changing the construction in the examples)

brendan-ward:
Yes, forgot about that one.  I think the cleanest path forward is to merge this one first, then I'll overhaul #1414 or create a new PR with just the unique changes.  My hope is that once we have the same repr throughout, we'll have less churn / conflicts across these cleanup PRs...

brendan-ward:
Actually, #1414 should get merged first since it is already approved and you had already updated the repr (sorry I missed that) and I'll fixup this one as needed.

jorisvandenbossche:
Yep, I would also merge the other one first, as that's a big one and will be annoying to rebase ..

jorisvandenbossche (commented):
Thanks for updating this!

With those changes, the doctests now finally pass, so maybe you can remove the allow failures:

https://github.com/shapely/shapely/blob/fc611de55001d148ea87e751a08dbb253622ea8c/.github/workflows/tests.yml#L172-L174

shapely/geometry/base.py
```diff
@@ -414,11 +414,20 @@ def buffer(
...
+
+        128-gon approximation:
+
+        >>> g.buffer(1.0, 128).area  # doctest: +ELLIPSIS
+        3.141513801144...
+
+        triangle approximation:
+
+        >>> g.buffer(1.0, 3).area  # doctest: +ELLIPSIS
         3.0
```
mwtoews, line 431: Is this ever not 3.0? Given the simple trigonometry, it should be exact. (A related simple case is a square `g.buffer(1.0, 1)` with `.area` 2.0)

brendan-ward: Good catch, was likely a copy / paste without thinking it through.

brendan-ward:
Note: because there was inconsistency in the doctest results between GEOS 3.10 and 3.11, I opted to run doctests only on 3.11.  Seems like our general intent is for those to be based on the latest release of GEOS.

shapely/constructive.py
```diff
@@ -227,7 +227,7 @@ def offset_curve(
     >>> offset_curve(line, 2)
     <LINESTRING (-2 0, -2 2)>
     >>> offset_curve(line, -2)
-    <LINESTRING (2 2, 2 0)>
+    <LINESTRING (2 0, 2 2)>
```
mwtoews, line 230: noting this as a probable instance of #1436

mwtoews (approved):
Looks great, thanks! Agree to use GEOS 3.11 as baseline.

shapely/constructive.py
brendan-ward: Good catch.  Should we normalize here to avoid this biting us later?  Or so long as the order matches whatever a particular GEOS version produces, we leave as is until it breaks?

jorisvandenbossche: > Or so long as the order matches whatever a particular GEOS version produces, we leave as is until it breaks?

I would do like that for now. We should actually update that docstring to note that the exact order is GEOS-version dependent and not to be relied upon (or update our implementation to guarantee it ourselves), but that's for another PR.

jorisvandenbossche (approved):

PR #1452, opened 2022-07-16
DOC: fix repr in example

jorisvandenbossche:
Small left-over that I didn't fix in https://github.com/shapely/shapely/pull/1378


brendan-ward (approved):
Any reason this is still draft?

jorisvandenbossche:
> Any reason this is still draft?

No, just clicked that by accident (if you select that the previous time when opening a PR, it might remember that choice?)

PR #1457, opened 2022-07-22
[ENH] Add reverse method to BaseGeometry

gpapadok:
#1281

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -747,11 +747,10 @@ class BaseGeometry(shapely.Geometry):
 
         Examples
         --------
-        >>> line = shapely.Geometry("LINESTRING (0 0, 0 10)")
-        >>> line.segmentize(tolerance=5)
+        >>> from shapely import LineString, Polygon
+        >>> LineString([(0, 0), (0, 10)]).segmentize(tolerance=5)
         <LINESTRING (0 0, 0 5, 0 10)>
-        >>> poly = shapely.Geometry("POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0))")
-        >>> poly.segmentize(tolerance=5)
+        >>> Polygon([(0, 0), (10, 0), (10, 10), (0, 10), (0, 0)]).segmentize(tolerance=5)
         <POLYGON ((0 0, 5 0, 10 0, 10 5, 10 10, 5 10, 0 10, 0 5, 0 0))>
         """
         return shapely.segmentize(self, tolerance)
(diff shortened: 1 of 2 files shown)


shapely/geometry/base.py
```diff
@@ -759,6 +759,35 @@ def segmentize(self, tolerance, **kwargs):
...
+            For other keyword-only arguments, see the
+            `NumPy ufunc docs <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs>`_.
+
+        See also
+        --------
+        is_ccw : Checks if a Geometry is clockwise.
+
+        Examples
+        --------
+        >>> line = Geometry("LINESTRING (0 0, 1 2)")
```
jorisvandenbossche, line 782: On the latest main branch, those examples have been updated to no longer use `Geometry(..)`, see

https://github.com/shapely/shapely/blob/fc611de55001d148ea87e751a08dbb253622ea8c/shapely/constructive.py#L688-L694

jorisvandenbossche (commented):
Thanks for the PR!

gpapadok:
I also changed the required GEOS version for the test. I mindlessly set the same version as the previous PR.

shapely/tests/geometry/test_geometry_base.py
```diff
@@ -91,3 +91,10 @@ def test_segmentize():
     line = LineString([(0, 0), (0, 10)])
     result = line.segmentize(tolerance=5)
     assert result.equals(LineString([(0, 0), (0, 5), (0, 10)]))
+
+
+@pytest.mark.skipif(shapely.geos_version < (3, 7, 0), reason="GEOS < 3.7")
+def test_reverse():
+    line = LineString([(0, 0), (1, 2)])
+    result = line.reverse()
+    assert result.equals(LineString([(1, 2), (0, 0)]))
```
mwtoews, line 100: The "equals" predicate doesn't check the order of coordinates, so this would be True with `result.equals(line)` too. It's best to check the `coords` property like this
```suggestion
    assert result.coords[:] == [(1.0, 2.0), (0.0, 0.0)]
```

mwtoews (commented):
Thanks for the PR, it looks good!

The only other nit is that the three instances of "Geometry" in the docstring don't need to be capitalized, so just sub the three with "geometry".

mwtoews:
@jorisvandenbossche do you think this should be  defined as a property? There are no arguments. You may recall a similar discussion at #1090

gpapadok:
I think the conclusion of the other discussion applies here too. Property is designed for managed attributes. Reverse sounds more like an operation on the object so a method is more fitting. At least it feels this way for me.

mwtoews (approved):

jorisvandenbossche:
Thanks @gpapadok !

PR #1461, opened 2022-08-03
MAINT: remove workarounds when numpy was optional

mwtoews:
This cleans up some logic that was used when numpy was optional. As of shapely-2.0 it is required.


jorisvandenbossche (approved):
Thanks!

PR #961, opened 2020-08-11
[2.0] Remove deprecated asShape / adapter classes

jorisvandenbossche:

PR #1467, opened 2022-08-08
CLN: Rename `__p__`→`_parent`, clean-up `gtag`, `_factory` and `__rings__`

mwtoews:
This was partially discussed in #983 to rename a few variables. But also clean-up variables that are no longer used with shapely-2.

- Rename `__p__`→`_parent` to make it "private", and generally [avoid non-standard dunder names](https://dbader.org/blog/meaning-of-underscores-in-python). Another consideration is to rename it to `__parent` to enable name mangling for sub classes, see docs [Private Variable](https://docs.python.org/3/tutorial/classes.html#private-variables) and [Reserved classes of identifiers](https://docs.python.org/3/reference/lexical_analysis.html#reserved-classes-of-identifiers). Another consideration is with [`__del__`](https://docs.python.org/3/reference/datamodel.html#object.__del__), "Python guarantees that globals whose name begins with a single underscore are deleted from their module before other globals are deleted", implying that a double underscore variable would be deleted last.
- Remove unused `_factory`, which was formerly a callable that returned instances of Shapely geometries, i.e. `geom_factory`
- Remove unused `__rings__`, which was another former structure

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -824,22 +824,22 @@ class GeometrySequence:
 
     # Attributes
     # ----------
-    # __p__ : object
+    # _parent : object
     #     Parent (Shapely) geometry
-    __p__ = None
+    _parent = None
 
     def __init__(self, parent):
-        self.__p__ = parent
+        self._parent = parent
 
     def _get_geom_item(self, i):
-        return shapely.get_geometry(self.__p__, i)
+        return shapely.get_geometry(self._parent, i)
 
     def __iter__(self):
         for i in range(self.__len__()):
             yield self._get_geom_item(i)
 
     def __len__(self):
-        return shapely.get_num_geometries(self.__p__)
+        return shapely.get_num_geometries(self._parent)
 
     def __getitem__(self, key):
         m = self.__len__()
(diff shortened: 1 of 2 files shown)


mwtoews:
I'm tempted to also clean-up `_gtag` and `def gtag(self)` in this PR too. However, the later is public:
```python
>>> g = Point(0,1).buffer(1)
>>> g.interiors.gtag()
-7868044578161086364
```

jorisvandenbossche:
> I'm tempted to also clean-up `_gtag` and `def gtag(self)` in this PR too. However, the later is public:

I would personally fine with just removing it in shapely 2.0 (without having first deprecated it, we can list it in the whatsnew under the breaking changes). I would be surprised if many people are using this?

mwtoews:
Ok, I've update this to remove `_gtag` and `gtag()`. From what I see in shapely-1.8, this was used with `shapely/geometry/proxy.py` with the `CachingGeometryProxy` class, which was removed with #961. CI errors are unrelated.

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

PR #1468, opened 2022-08-08
ENH: expose fixed precision overlay (grid_size keyword) in the Geometry methods as well

jorisvandenbossche:
The fixed precision overlay operation (part of OverlayNG from GEOS) is a new feature in 2.0, and this is already available in the new top-level functions, but it would be good to expose this on the existing methods as well. This is adding the `grid_size` keyword and passing this through.

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -476,7 +476,7 @@ class BaseGeometry(shapely.Geometry):
         """
         return shapely.normalize(self)
 
-    # Binary (overlay) operations
+    # Overlay operations
     # ---------------------------
 
     def difference(self, other, grid_size=None):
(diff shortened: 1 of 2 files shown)


shapely/geometry/base.py
```diff
@@ -476,25 +476,40 @@ def normalize(self):
         """
         return shapely.normalize(self)
 
-    # Binary operations
-    # -----------------
+    # Binary (overlay) operations
```
brendan-ward, line 479:
```suggestion
    # Overlay operations
```

(binary does not seem helpful in this context)

brendan-ward (approved):
One minor comment but otherwise this looks good to me.  👍

PR #1476, opened 2022-08-14
CLN: actually remove the base class Geometry() constructor

### LINKED ISSUE #1330: What to do with the `Geometry()` class constructor?
In pygeos, there is a single `pygeos.Geometry` class (we didn't have the Point, LineString, etc subclasses), and this was exposed in the top-level namespace. It also provided an alternative constructor for creating scalar geometries, accepting WKT or WKB, in addition to the specialized and vectorized functions:

```python
# using one of the vectorized functions (with scalar inputs, it also gives a scalar geometry as result)
>>> pygeos.points(1, 1)
<pygeos.Geometry POINT (1 1)>
>>> pygeos.from_wkt("POINT (1 1)")
<pygeos.Geometry POINT (1 1)>

# or using the Geometry class constructor
>>> pygeos.Geometry("POINT (1 1)")
<pygeos.Geometry POINT (1 1)>
```

In Shapely, we have the geometry subclasses that also provide a class constructor:

```python
>>> shapely.geometry.Point(1, 1)
<shapely.geometry.Point POINT (1 1)>
```

But as a left-over from the pygeos merger, we now also have this top-level Geometry class constructor, that returns one of the subclasses:

```python
>>> shapely.Geometry("POINT (1 1)")
<shapely.geometry.Point POINT (1 1)>
```

Having this constructor might still be useful (it can be nicer than the `from_wkt`), but it is a bit strange that this lives in the top-level namespace, while the other subclasses live in `shapely.geometry`. It is also a bit strange that `BaseGeometry` is actually a subclass of `Geometry`

In the pygeos documentation, we made heavy use of this class constructor in the docstring examples (see eg https://pygeos.readthedocs.io/en/stable/predicates.html), and so this is also still present in the new shapely docstrings after the merger.

----

The above describes the current situation in the main branch and the history behind it, but we have to decide what to do with this situation for Shapely 2.0. 

One option is to keep it as is, but I would find it a bit strange (as mentioned above) to have a top-level `Geometry()` constructor while the other classes are in a submodule, especially if this is getting used heavily in the docs. 
Of course, now we have the subclass constructors (which didn't exist in pygeos), we can also make use of those more in the docstring examples.

But some other ideas to consider:

- Completely remove the `shapely.Geometry` class from the public (top-level) namespace, and only keep it as a private `shapely.lib.Geometry` as the parent class for the concrete subclasses.
  - This means that we would need to remove it from all docs (in favor of the subclass constructors or `from_wkt`)
- Move `shapely.Geometry` to the `shapely.geometry` submodule (so all geometry classes are consistently only available from that submodule). On the long term, we could maybe also have it take the role of BaseGeomety (switch the naming of those two)
- Make all geometry subclasses available in the top-level namespace (so all geometry classes are consistently available in the top-level namespace). For backwards compatibility we of course keep them in `shapely.geometry` as well.
  - This would actually make the subclasses more accessible and easier to use (doesn't need an additional `import shapely.geometry` or `from shapely.geometry import ..`, just `import shapely` is enough). 
  - Since we are also making all the vectorized functions available in the top-level namespace (which already breaks with the pre-2.0 pattern of Shapely, xref https://github.com/shapely/shapely/issues/1225), this is also somewhat consistent with that.
  - Extra bonus: this could also help with reducing the length of the repr (xref https://github.com/shapely/shapely/pull/1302#discussion_r799844326)

cc @sgillies @mwtoews @caspervdw @brendan-ward @shapely/shapely-2

jorisvandenbossche:
Closes #1330

In https://github.com/shapely/shapely/pull/1414 we already removed all usage of the base class Geometry() constructor in code, tests and documentation. But the constructor itself wasn't yet removed, so it could still be called. This PR also actually makes the base class not callable.

### DIFF
--- a/shapely/tests/test_creation.py
+++ b/shapely/tests/test_creation.py
@@ -504,37 +504,6 @@ def test_box_nan(coords):
     assert shapely.box(*coords) is None
 
 
-class BaseGeometry(shapely.Geometry):
-    @property
-    def type_id(self):
-        return shapely.get_type_id(self)
-
-
-class Point(BaseGeometry):
-    @property
-    def x(self):
-        return shapely.get_x(self)
-
-    @property
-    def y(self):
-        return shapely.get_y(self)
-
-
-@pytest.fixture
-def with_point_in_registry():
-    orig = shapely.lib.registry[0]
-    shapely.lib.registry[0] = Point
-    yield
-    shapely.lib.registry[0] = orig
-
-
-def test_subclasses(with_point_in_registry):
-    for _point in [Point("POINT (1 1)"), shapely.points(1, 1)]:
-        assert isinstance(_point, Point)
-        assert shapely.get_type_id(_point) == shapely.GeometryType.POINT
-        assert _point.x == 1
-
-
 def test_prepare():
     arr = np.array([shapely.points(1, 1), None, shapely.box(0, 0, 1, 1)])
     assert arr[0]._geom_prepared == 0
@@ -560,14 +529,6 @@ def test_destroy_prepared():
     assert arr[1] is None
     assert arr[2]._geom_prepared == 0
     shapely.destroy_prepared(arr)  # does not error
-
-
-def test_subclass_is_geometry(with_point_in_registry):
-    assert shapely.is_geometry(Point("POINT (1 1)"))
-
-
-def test_subclass_is_valid_input(with_point_in_registry):
-    assert shapely.is_valid_input(Point("POINT (1 1)"))
 
 
 @pytest.mark.parametrize("geom_type", [None, GeometryType.MISSING, -1])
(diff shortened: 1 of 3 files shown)


shapely/tests/test_creation.py
```diff
@@ -504,37 +504,6 @@ def test_box_nan(coords):
...
-@pytest.fixture
-def with_point_in_registry():
-    orig = shapely.lib.registry[0]
-    shapely.lib.registry[0] = Point
-    yield
-    shapely.lib.registry[0] = orig
-
-
-def test_subclasses(with_point_in_registry):
-    for _point in [Point("POINT (1 1)"), shapely.points(1, 1)]:
```
jorisvandenbossche, line 532: Those tests should now be superfluous, since we already use this registry for the actual subclasses in shapely. 

(alternatively I could also keep those tests by adding a `__new__` to the test subclass

caspervdw (approved):
This looks good to merge!

PR #1477, opened 2022-08-14
PERF: use GEOSGeom_getExtent for GEOS>=3.11 in bounds ufunc

jorisvandenbossche:
xref https://github.com/libgeos/geos/pull/554 / https://github.com/libgeos/geos/pull/555

This gives a modest speedup for simple geometries:

```
In [1]: import shapely

In [2]: polygons = shapely.polygons(np.random.randn(1_000_000, 3, 2))

In [3]: %timeit shapely.total_bounds(polygons)
64.9 ms ± 3.13 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)   # <- main
46.6 ms ± 2.31 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)   # <- PR
```


brendan-ward (approved):
Nice to see that the expected speedup is indeed a speedup.

Please add a note to the changelog, but otherwise this looks good, thanks!

jorisvandenbossche:
> Please add a note to the changelog, but otherwise this looks good, thanks!

Given that this is a new function in 2.0.0, is that worth including? (all top-level vectorized functions have a better performance compared to doing that before manually)

brendan-ward:
It is sometimes helpful to know when we've wrapped a particular GEOS function when tracing issues.

What I meant was something like this under 2.0b1:

```
- added use of GEOSGeom_getExtent to speed up bounds calculations for GEOS >= 3.11
```

But it is minor, so if you don't think it fits the intent of changes listed for 2.0 / 2.0b1, it is fine to leave it out.

jorisvandenbossche:
Thanks, I can certainly add such a sentence!

PR #1479, opened 2022-08-16
Replace c_void_p with c_geom_p

sgillies:
In an attempt to fix the issue reported in #1449 and perhaps solved in https://github.com/django/django/pull/15214.


jorisvandenbossche:
FYI the windows failures are not related, will be fixed by https://github.com/shapely/shapely/pull/1481

sgillies:
I believe we have parity with GeoDjango here. Tests pass. I'll merge, ask for some testing of the maint-1.8 dev wheels, and then make a 1.8.3 release later today.

PR #1480, opened 2022-08-16
Fix GEOSMinimumClearance signature.

### LINKED ISSUE #1473: Segfault in tests on aarch64-darwin
## Please note

If you are reporting an installation or module import issue, please note that
this project only accepts reports about problems with packages downloaded from
the Python Package Index. Conda users should take issues to one of the
following trackers:

- https://github.com/ContinuumIO/anaconda-issues/issues
- https://github.com/conda-forge/shapely-feedstock

Tests are failing due to a segmentation fault on Python 3.10.

See downstream report at https://github.com/NixOS/nixpkgs/issues/185996

## Operating system

macOS 12.4 on aarch64

Nix, Shapely 1.8.2

tjni:
This seems to fix a segmentation fault when calling this function in the test suite on aarch64-darwin. Fixes #1473.

### DIFF
--- a/shapely/ctypes_declarations.py
+++ b/shapely/ctypes_declarations.py
@@ -521,8 +521,8 @@ def prototype(lgeos, geos_version):
             c_void_p, py_object, c_void_p, lgeos.GEOSDistanceCallback, py_object]
         lgeos.GEOSSTRtree_nearest_generic.restype = c_void_p
 
-        lgeos.GEOSMinimumClearance.argtypes = [c_void_p]
-        lgeos.GEOSMinimumClearance.restype = c_double
+        lgeos.GEOSMinimumClearance.argtypes = [c_void_p, POINTER(c_double)]
+        lgeos.GEOSMinimumClearance.restype = c_int
 
     if geos_version >= (3, 8, 0):
         lgeos.GEOSMakeValid.restype = c_void_p


jorisvandenbossche (approved):
Looks good! That's in any case the correct signature for GEOSMinimumClearance ..

mwtoews:
Thanks, this looks like the [correct signature](https://github.com/libgeos/geos/blob/fc2e2e8c8745828ed05ea19f608fc6517422a3cb/capi/geos_c.h.in#L2806-L2823).

What I find confusing is that [GEOSArea](https://github.com/libgeos/geos/blob/fc2e2e8c8745828ed05ea19f608fc6517422a3cb/capi/geos_c.h.in#L2863-L2871) doesn't have a similar crash, because it has the same GEOS signature too, but is wrong in [ctypes_declarations.py](https://github.com/shapely/shapely/blob/ff9818d680669d6bac7f44108d2b7e746ae0599d/shapely/ctypes_declarations.py#L399-L400) in a similar way. Does ctypes_declarations.py need to be re-checked?

jorisvandenbossche:
> What I find confusing is that [GEOSArea](https://github.com/libgeos/geos/blob/fc2e2e8c8745828ed05ea19f608fc6517422a3cb/capi/geos_c.h.in#L2863-L2871) doesn't have a similar crash, because it has the same GEOS signature too, but is wrong in [ctypes_declarations.py](https://github.com/shapely/shapely/blob/ff9818d680669d6bac7f44108d2b7e746ae0599d/shapely/ctypes_declarations.py#L399-L400) in a similar way. Does ctypes_declarations.py need to be re-checked?

For GEOSArea, we _do_ specify two pointers as arguments, so here at least the number of arguments is correct (while for GEOSMinimumClearance even the number of arguments was incorrect). Maybe that makes the difference? 
(but it's true that the signature for GEOSArea seems also very much a mistake, as it uses c_double for return type)

tjni:
Your guess about the number of arguments mattering is more likely than not, because this is what I noticed:

The [routine](https://github.com/libffi/libffi/blob/95ef857d5c6ed15c6c6ba5f8a5b26c0a38b417ab/src/aarch64/sysv.S#L64) in `libffi` that calls the library function is quite different for aarch64 vs. other architectures, and even has differences for darwin. I find it hard to make complete sense of, but it's not too surprising to me that it only fails on one architecture.

in the failing test within `nixpkgs`, a EXC_BAD_ACCESS is triggered. It turns out that the memory address being accessed is also the representation of positive infinity, which is the value that is supposed to in that case be set on the double pointer being passed into `GEOSMinimumClearance`.

sgillies (approved):
@tjni thank you!

sgillies:
I'll address the issue with GEOSArea.

PR #1487, opened 2022-08-17
Revert "Replace c_void_p with c_geom_p"

sgillies:
Reverts shapely/shapely#1479, which didn't fix the issue it intended to fix.


jorisvandenbossche:
A subsequent commit also made use of this: https://github.com/shapely/shapely/commit/088ce60284a4bf31b7969cb365822e2576b8a8f4 (would need to be reverted as well)

Issue #1485, opened 2022-08-17
Error when translate a point


johnnv1:
## Expected behavior and actual behavior.

Something in the last release (1.8.3) seems to cause an error when translate a point. Working fine with polygons.

## Steps to reproduce the problem.

```python
from shapely.geometry import Point
from shapely import affinity
affinity.translate(Point(10, 10), 50, 50)
```
```output
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/tmp/t/venv/lib/python3.8/site-packages/shapely/affinity.py", line 264, in translate
    return affine_transform(geom, matrix)
  File "/tmp/t/venv/lib/python3.8/site-packages/shapely/speedups/__init__.py", line 73, in affine_transform
    return _speedups.affine_transform(geom, matrix)
  File "shapely/speedups/_speedups.pyx", line 521, in shapely.speedups._speedups.affine_transform
  File "shapely/speedups/_speedups.pyx", line 553, in shapely.speedups._speedups.affine_transform
TypeError: an integer is required
```

## Env
Ubuntu 20.04
Shapely version: 1.8.3

jorisvandenbossche:
@johnnv1 thanks for the report!

That's a regression in 1.8.3, I think caused by https://github.com/shapely/shapely/pull/1479 (cc @sgillies)

It's a bit strange that we didn't catch this in our CI

jorisvandenbossche:
> It's a bit strange that we didn't catch this in our CI

It seems that whether it errors or not depends on how the geometry was creates (and our affinity tests use WKT construction, which doesn't trigger the error, and thus this wasn't caught in our CI):

```
In [1]: import shapely

In [2]: shapely.__version__
Out[2]: '1.8.3'

In [3]: from shapely.geometry import Point
   ...: from shapely import affinity
   ...: from shapely import wkt

In [4]: p = Point(1, 1)

In [5]: affinity.translate(p)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [5], in <cell line: 1>()
----> 1 affinity.translate(p)

File ~/miniconda3/envs/shapely-183/lib/python3.10/site-packages/shapely/affinity.py:264, in translate(geom, xoff, yoff, zoff)
    259     return geom
    260 matrix = (1.0, 0.0, 0.0,
    261           0.0, 1.0, 0.0,
    262           0.0, 0.0, 1.0,
    263           xoff, yoff, zoff)
[...]

PR #1488, opened 2022-08-17
Ensure return type of _geom/__geom__ is integer

jorisvandenbossche:
xref https://github.com/shapely/shapely/issues/1485


jorisvandenbossche:
We reverted the original change instead (https://github.com/shapely/shapely/pull/1487), so this can be closed

PR #1496, opened 2022-08-22
ENH: expose dwithin on the Geometry subclasses

jorisvandenbossche:
xref https://github.com/shapely/shapely/issues/1281

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -618,6 +618,14 @@ class BaseGeometry(shapely.Geometry):
         """Returns True if geometry is within the other, else False"""
         return bool(shapely.within(self, other))
 
+    def dwithin(self, other, distance):
+        """
+        Returns True if geometry is within a given distance from the other, else False.
+
+        Refer to `shapely.dwithin` for full documentation.
+        """
+        return bool(shapely.dwithin(self, other, distance))
+
     def equals_exact(self, other, tolerance):
         """True if geometries are equal to within a specified
         tolerance.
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
Are we OK with also adding `contains_properly`?

shapely/geometry/base.py
```diff
@@ -618,6 +618,14 @@ def within(self, other):
         """Returns True if geometry is within the other, else False"""
         return bool(shapely.within(self, other))
 
+    def dwithin(self, other, distance):
+        """
+        Returns True if geometry is within a given distance from the other, else False.
+
+        Refer to `shapely.dwithin` for full documentation.
```
brendan-ward, line 625: One thing I noticed in the generated docs (unless I'm building them incorrectly) is that the Shapely docs do not add a note about the required version of GEOS, whereas pygeos included a note about this.

Does this also need the `@requires_geos("3.10.0")` annotation here?

brendan-ward (commented):
Thanks for working on this!

shapely/geometry/base.py
mwtoews: no annotation is required, a helpful exception is raised: "shapely.errors.UnsupportedGEOSVersionError: 'dwithin' requires at least GEOS 3.10.0"

mwtoews (approved):
Thanks, local testing looks good! I can't wait to test this out more thoroughly, as I consider [ST_DWithin](https://postgis.net/docs/ST_DWithin.html) one of the best PostGIS functions.

shapely/geometry/base.py
jorisvandenbossche: Ah, that's a good point about the docstrings not including those notes. 
@mwtoews from pygeos, we had an automated injection of a note into the docstring about the required GEOS version, see eg https://pygeos.readthedocs.io/en/stable/constructive.html#pygeos.constructive.build_area vs https://shapely.readthedocs.io/en/latest/constructive.html#shapely.constructive.build_area 

The code for this is still included, so not really sure why this is not working .. (we also have tests for this, and those are passing)

jorisvandenbossche: Fixing the docstring note in https://github.com/shapely/shapely/pull/1519

PR #1497, opened 2022-08-22
ENH: shapely.plotting module with basic matplotlib-based functionality (+ updated docs to use this)

### LINKED ISSUE #1285: Proposal: add basic plotting functionality to visualize geometries (matplotlib-based)
Currently, Shapely stays out of the business of plotting geometries (except for basic svg repr for the notebook), and you can use eg geopandas for visualizing shapely geometries. 

There was the `descartes` package which could plot shapely Polygons based on matplotlib (https://pypi.org/project/descartes/), however this project is stopped. For example, geopandas now uses matplotlib directly instead of relying on descartes. 

We certainly shouldn't go too far with fancy visualization options, but I think it would be useful to have some _basic_ visualization functionality included in Shapely, since it can be very useful to visualize geometries:

- We need some plotting code for our documentation examples anyway (eg the figures at https://shapely.readthedocs.io/en/latest/manual.html#Polygon). Currently that uses some custom code and descartes, and we would need to update those in any case to replace descartes (https://github.com/shapely/shapely/issues/1145). If we have code to create those visualizations anyway, we could as well factor out this code in some reusable functions in a public shapely module.
- When reporting or debugging an issue or potential bug, it can be very useful to have a quick visualization of the geometry in question. And it would be handy to be able to directly do this from python/shapely (instead of exporting it and using other tools like QGIS or JTS TestBuilder, or wrapping it in geopandas). See eg [#790 (comment)](https://github.com/shapely/shapely/issues/790#issuecomment-1004305587) (the plot I posted there is more or less what I imagine this would do by default, and was created with a first prototype):

  image

I think if we keep the functionality limited (just the ability to change some colors, and maybe a toggle to show/hide the vertices), this could be done without too much complexity, while adding quite some value.

jorisvandenbossche:
Closes #1285

As explained in https://github.com/shapely/shapely/issues/1285, this adds basic functionality to visualize single geometries, useful for debugging/reporting issues, and needed for our own docs as well. 
This PR adds a `shapely.plotting` module (currently with `plot_polygon`, `plot_line`, and `plot_points` functions), and updates our doc examples to use this public module (instead of descartes + bunch of private helpers).

You can see the built docs using this branch at https://shapely.readthedocs.io/en/plotting/manual.html, and you can compare the resulting figures with the ones on the stable docs: https://shapely.readthedocs.io/en/stable/manual.html

The default `plot_polygon` without any options gives (with the current version of the PR):

image

But for the doc examples we do some additional customizations (eg no grid, changing some colors, not always showing the points, etc).

Some general notes / questions:

- ~The current methods (`plot_polygon`, `plot_line`) don't yet handle their multi counterpart (so you have to loop through `.geoms` manually), but for convenience we should probably add that capability~
- What would be the best name for `plot_points`? `plot_coords` instead? (this function takes any kind of geometry, extract all coordinates, and plots each coordinate pair (the vertices) as points)
- ~In the current examples, when plotting a line or polygon, we typically plot the points (if they are plotted) always in grey, while in my current code those points get plotted in the same color as the line/polygon. Any preference here?~ -> kept original look of grey points in the docs (but the default is to use the same color)
- Some examples showing a line (eg https://shapely.readthedocs.io/en/stable/manual.html#LineString) highlight the start and end points of the line (black instead of grey). Do we want to keep that? 
  - And if so, is that useful enough to add this to the public function, or do we keep that as a customization in the doc example code?
  - -> added this manually to the code snippets (so not built-in)

~TODO: clean-up the public module (and improve docs) + add tests (for now I didn't add any explicit tests, except for running our own docs with it)~


jorisvandenbossche:
I updated this PR:

- Support for directly plotting a MultiLineString or MultiPolygon with the `plot_line` and `plot_polygon` functions
- Tweaked the figures code a bit more to ensure we now have more or less exactly the same figures as before (making points grey, start/end points black (where the figure caption in manual.rst mentioned this), etc)
- Expanded the docstrings and added a note that this is experimental

I am planning to merge this shortly since this PR is needed to get the figures in the doc build working again. It does add public API (so review would still be very welcome!), but so I added a disclaimer that this is experimental  / mostly meant for explorative use (so we can still further update the exact interface if needed)

mwtoews:
This will be a really useful feature! Answer to the question: `plot_points` is good, since it has closer alliance with matplotlib point plots with markers than the geometry type. I think other features, like point start/end colors etc. could be changed later.

One question is with the `add_points` option. The defaults are `add_points=True`, but there are 89 instances in the code with `add_points=False`. Should this keyword be removed? As demonstrated in the code, if a user wants to add points to a line or polygon, they can simply add `plot_points(a, ax=ax, ...)`.

jorisvandenbossche:
> One question is with the `add_points` option. The defaults are `add_points=True`, but there are 89 instances in the code with `add_points=False`. Should this keyword be removed? As demonstrated in the code, if a user wants to add points to a line or polygon, they can simply add `plot_points(a, ax=ax, ...)`.

Yeah, I have certainly been wondering that myself as well (or at least whether the default should be `add_points=True` if most of the time in our code snippets I turn it off).

The reason that I kept it for now (and with a default of True), is that I think it is the best default for interactive / debugging use cases. For example, see the plot of the polygon in the top post, which is what you get from a plain `plot_polygon(poly)`, and I personally quite like this for a quick introspection of a polygon. 

For making more fine-tuned figures for illustrations in the documentation, it appears that most of the time I didn't actually use this default. But in those code files I also always manually set up the matplotlib figure and axes, changed the default colors (regularly changed the default alpha), etc. So for those cases it is fine to manually do `plot_points` in addition to `plot_line`/`plot_polygon` to have greater control (in one of the later commits I added a lot of manual `add_points` to give them a different color as the line/polygon, just to preserve the look&feel of the figures we currently have)

jorisvandenbossche:
@mwtoews I am going to merge this, because I want to further get the docs fully ready for the 2.0 (beta) release. 

But I certainly would like to further discuss the default / the need for the `add_points` keyword. I opened https://github.com/shapely/shapely/issues/1550 for this.

martinfleis:
Is this intentionally left out of the API reference? It seems to be public API.

Issue #1196, opened 2021-10-03
Feature request: SVG as interoperability format?


NelsonMinar:
Is there any interest in supporting SVG as an input and output format, like WKT? I know Shapely objects have a `.svg()` method (which isn't in the docs, btw). But that method is export only and seems mostly aimed at making nice debug visualizations in Jupyter notebooks. Which is great! But I'm interested in working with SVG more abstractly as a vector shape format.

My use case is plotter art; I'm importing drawings from Inkscape into Shapely, manipulating them, and then writing them back out as SVG for the plotter. Right now I've written my own very limited import and export code. It's not hard and if SVG is out of scope for Shapely then an external add-on could provide what I'm asking. But maybe other Shapely users would benefit from SVG support too?

SVG is complicated with a lot of object types. Simple paths are the base data case. Even they are complicated; at the core they are LineStrings but the segments can be various kinds of curves. Also even basic path objects are implicitly MultiPolygons sometimes if you interpret the fill attribute. On top of that SVG has grouping, various object primitives like ellipses and text, etc. I don't know how much of that is worth supporting.

The best option for SVG import I've found so far is [svgpath2mpl](https://github.com/nvictus/svgpath2mpl), which interprets SVG paths and creates matplotlib objects. Those can then easily be input into Shapely, see [this discussion](https://gis.stackexchange.com/questions/301605/how-to-create-shape-in-shapely-from-an-svg-path-element) for an example. That library has some limits and is only for paths, not other SVG objects, but it could be a good start for import code. Note that importing will require linearizing all the curves.

For export the existing `svg()` support is pretty good; my main hassle using it in practice is that I have to override the styling attributes that Shapely is putting on them. The mixture of object types is also a bit awkward, simple path objects only would be a better match for my use case. I'm imagining a "pure" data-only SVG output which is then easily styled and manipulated by other code.

I'd totally understand if you think SVG is out of scope. But it could be useful! Honestly half the reason I'm writing this ticket is to document the state of things as it stands.

sgillies:
@NelsonMinar yes, I think better interoperability via SVG would be great for shapely. The manufacturing applications of shapely have always impressed me and are under-served compared to the GIS applications. Have you heard of the geomet project https://github.com/geomet/geomet? What would you think about a more general GeoJSON-SVG interchange?

sgillies:
[...]

PR #1499, opened 2022-08-22
[WIP] New SVGTree class

sgillies:
Towards resolving #1196.

In a nutshell, this is the draft API:

* SVGTree derives from xml.etree.ElementTree.
* It adds a `fromgeom` method which is a factory for `<g>` tag instances of `Element`. These instances contain one or more `<path>` sub-elements. In other words, `ET.tostring(SVGTree.fromgeom(LineString(...)))` returns `<g><path>...</path></g>`.
* Styling is someone else's problem :)
* If you want to control the precision of coordinates in SVG, call Shapely 2's new `set_precision()` function on the geometry object before passing it to `SVGTree.fromgeom()`.

Names of things in this API will hew close to etree (fromgeom, for example, in analogy to ET.fromstring).

@jorisvandenbossche @brendan-ward this is what I was referring to earlier today.

### DIFF
--- a/shapely/svgtree.py
+++ b/shapely/svgtree.py
@@ -1,0 +1,57 @@
+"""svgtree.py: ElementTree based geometry SVG"""
+
+import itertools
+from xml.etree.ElementTree import Element, ElementTree, SubElement
+
+import numpy
+
+
+class SVGTree(ElementTree):
+    @classmethod
+    def path(cls, coords):
+        return "M {} L {}".format(
+            f"{coords[0][0]},{coords[0][1]}",
+            " L ".join(f"{xy[0]},{xy[1]}" for xy in coords[1:]),
+        )
+
+    @classmethod
+    def fromgeom(cls, geom):
+        """Returns Element"""
+        if geom.geom_type == "Point":
+            return cls.frompoint(geom)
+        elif geom.geom_type == "LineString":
+            return cls.fromlinestring(geom)
+        elif geom.geom_type == "LinearRing":
+            return cls.fromlinearring(geom)
+        elif geom.geom_type == "Polygon":
+            return cls.frompolygon(geom)
+        else:
+            raise NotImplementedError
+
+    @classmethod
+    def frompoint(cls, geom):
+        group = Element("g")
+        SubElement(group, "circle", cx=str(geom.x), cy=str(geom.y), r="1.0")
+        return group
+
+    @classmethod
+    def fromlinestring(cls, geom):
+        group = Element("g")        
+        SubElement(group, "path", d=cls.path(geom.coords))
+        return group
+
+    @classmethod
+    def fromlinearring(cls, geom):
+        group = Element("g")        
+        SubElement(group, "path", d="{} Z".format(cls.path(geom.coords[:-1])))
+        return group
+
+    @classmethod
+    def frompolygon(cls, geom):
+        group = Element("g")        
+        path = "  ".join(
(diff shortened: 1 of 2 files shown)


tests/test_svgtree.py
```diff
@@ -0,0 +1,53 @@
...
+                [(0, 0), (100, 100), (100, 0)],
+                [Point(75, 25).buffer(20, quadsegs=2).exterior],
+            ),
+            0.01,
+        )
+    )
+    assert (
+        ET.tostring(elem, encoding="unicode")
+        == '<g><path d="M 100.0,100.0 L 100.0,0.0 L 0.0,0.0 Z  M 89.14,39.14 L 75.0,45.0 L 60.86,39.14 L 55.0,25.0 L 60.86,10.86 L 75.0,5.0 L 89.14,10.86 L 95.0,25.0 Z" /></g>'
+    )
```
sgillies, line 53: SVG coordinate precision to 0.01 (absolute precision only) thanks to `set_precision()`.

shapely/svgtree.py
```diff
@@ -0,0 +1,55 @@
...
+            return cls.fromlinestring(geom)
+        elif geom.geom_type == "LinearRing":
+            return cls.fromlinearring(geom)
+        elif geom.geom_type == "Polygon":
+            return cls.frompolygon(geom)
+        else:
+            raise NotImplementedError
+
+    @classmethod
+    def frompoint(cls, geom):
```
brendan-ward, line 30: Seems like `path` and the individual geometry functions should all be private in order to keep the public API smaller.

brendan-ward (commented):
Thanks for working on this @sgillies !

My first instinct on reviewing this is to keep more of this in the internal API.  It also seems like returning an ElementTree instance or text should be a parameter, and also consider bundling in ability to set precision as part of the call, instead of modifying geometries prior.

I'm thinking an interface that parallels `to_wkt` might be nice from a consistency standpoint; it could certainly leverage what you've worked out here under the hood.

Lastly, it might be a bit of an edge case and unless implemented in Cython / C could be computationally expensive, but handling arrays as input would be good from a consistency standpoint, returning an SVG output per input.

Here's a possible signature that gets at what I was thinking (just a suggestion): 

```
def to_svg(geometry, grid_size=None, return_xml=False):
```

Where `grid_size` is equivalent to the parameter of same name for `set_precision`; this is meant to be similar in intent to the `rounding_precision` parameter for `to_wkt` but not operate the same way.

`return_xml` of `True` would return what you have here, whereas perhaps by default returning a unicode string like in all your tests would be reasonable?  Or perhaps it is reasonable to always return only a unicode string?  Modifying the returned XML seems like it could be a special case.

Then your `SVGTree` would be a private API class used under that function.  I think this would make it easier to migrate the underlying implementation to Cython / C later.


I'm not an SVG expert, but I think there might need to be more coordinate transformations going on here before exporting to SVG that is easily usable:
- flipping Y values so that origin is upper left instead of bottom left
- extracting the bounding box of the coordinates (of a scalar geometry) and transforming all coordinates to relative within that bounding box; absolute coordinates seem especially problematic for use in SVG, especially those in a CRS that is based on meters / feet (large magnitude values).  This interacts with the `grid_size` parameter, so that would need some further thought.

sgillies:
@brendan-ward thanks for the comments! I'm not interested in vectorized SVG creation. I think I'll probably develop this elsewhere and close this PR.

PR #1501, opened 2022-08-23
ENH: Return empty geometry collection from union_all, intersection_all, symmetric_difference_all if result is otherwise None

### LINKED ISSUE #1332: ops.unary_union changed behavior from 1.8 to 2.0 when given an empty list, and now returns None
I'm not sure what the expected behavior is here, but I feel that always returning a shapely geometry is the safest as it means that the user doesn't always need to validate parameters/returns.

I found this by running the test suite for our project that uses shapely. We end up with quite a bit of strange/organic (and sometimes invalid!) geometry from user data so I have definitely found some edge cases (will be making more issues!).

### Expected (1.8)

```python
>>> from shapely import ops
>>> print(ops.unary_union([]))
GEOMETRYCOLLECTION EMPTY
```

### Actual (2.0)

```python
>>> from shapely import ops
>>> print(ops.unary_union([]))
None
```

```python
from shapely import ops
assert ops.unary_union([]) is not None
```

## Operating system

MacOS Catalina

```
Python 3.9.2 (default, Feb 28 2022, 13:37:28)
[Clang 12.0.0 (clang-1200.0.32.29)]
```

Installed using `python setup.py develop` for both versions

brendan-ward:
Resolves #1332 

These reduce style functions will now always return an empty geometry collection when the result of the reduce operation would otherwise return `None` along the returned axis.

```Python
>>> from shapely import intersection_all, symmetric_difference_all, union_all
>>> union_all([None])
<GEOMETRYCOLLECTION EMPTY>
>>> intersection_all([None, None])
<GEOMETRYCOLLECTION EMPTY>
>>> symmetric_difference_all([None, None])
<GEOMETRYCOLLECTION EMPTY>
```

There must be at least 2 non-None inputs to intersection_all and symmetric_difference_all, or they return an empty geometry collection; they previously would have returned `None` instead.  This does point out an inconsistency: now the *_all reduce functions return a different value for a pair of inputs (like example above) than the corresponding non-reduce functions: 

```Python
>>> from shapely import Point, intersection, intersection_all
>>> intersection(Point(), None) is None
True
>>> intersection_all([Point(), None])
<GEOMETRYCOLLECTION EMPTY>
```

This reverts [pygeos/pygeos#249](https://github.com/pygeos/pygeos/pull/249) which made `union_all` behave like the others and return `None` instead of an empty geometry collection.

Note: this does not yet add a parameter to control the output geometry type to use in place of `None` per [this suggestion](https://github.com/shapely/shapely/issues/1332#issuecomment-1223567525); that might be better in a follow-on PR.


jorisvandenbossche:
> This does point out an inconsistency: now the *_all reduce functions return a different value for a pair of inputs (like example above) than the corresponding non-reduce functions:
> 
> ```python
> >>> from shapely import Point, intersection, intersection_all
> >>> intersection(Point(), None) is None
> True
> >>> intersection_all([Point(), None])
> <GEOMETRYCOLLECTION EMPTY>
> ```

I _think_ this inconsistency can be OK, at least it is very similar how pandas works with missing values in numeric data:

```
>>> pd.Series([1]) + np.nan
0   NaN
dtype: float64

>>> pd.Series([1, np.nan]).sum()
1.0
```

In general in pandas, missing values propagate in (binary) operations, but are skipped in reductions (at least by default, there is still a keyword to _not_ skip them, with `skipna=False`). 

Numpy doesn't skip NaNs in reductions like sum (although you have `nansum` that does), but in numpy it's also not really used as missing values.

So if we see `None` as the missing value in shapely, I think the above behaviour that you call out is OK. Though it does bring up the question whether we want something like a `skip_none` keyword in those `.._all` methods?

shapely/set_operations.py
```diff
@@ -164,8 +166,19 @@ def intersection_all(geometries, axis=None, **kwargs):
     <LINESTRING (1 1, 2 2)>
     >>> intersection_all([[line1, line2, None]], axis=1).tolist()
     [<LINESTRING (1 1, 2 2)>]
+    >>> intersection_all([line1, None])
+    <GEOMETRYCOLLECTION EMPTY>
```
jorisvandenbossche, line 170: This rule of "at least two non-None geometries", that gives an inconsistency with having only one geometry?

```
In [31]: shapely.intersection_all([line1])
Out[31]: <LINESTRING (0 0, 2 2)>
```

(and I know this is already the case right now, but since we are discussing all those corner cases here ... ;))

Also, I would expect the following two to result in the same (the below is using main, not this PR):

```
In [32]: shapely.intersection_all([line1, None])   # returns None

In [33]: shapely.intersection_all([line1, None, None])
Out[33]: <LINESTRING (0 0, 2 2)>
```

It seems a bit strange that a single None or two or more Nones are treated differently here. 
With this PR, those both will give an empty geometry now? That at least makes it consistent, but I would personally still expect it to be the same as the first case (`shapely.intersection_all([line1])`), if we consider the logic to be "Nones are skipped".

jorisvandenbossche:
Thanks for the PR! Triggered by looking at your added examples and expanded docstring, I noticed some other inconsistencies that might influence the behaviour we actually want ...

shapely/set_operations.py
jorisvandenbossche: Testing with your PR, they do _not_ both give an empty geometry:

```
In [3]: shapely.intersection_all([line1, None, None])
Out[3]: <LINESTRING (0 0, 2 2)>

In [4]: shapely.intersection_all([line1, None])
Out[4]: <GEOMETRYCOLLECTION EMPTY>

In [5]: shapely.intersection_all([line1])
Out[5]: <LINESTRING (0 0, 2 2)>

```

_If_ we would want to make this consistent (all return the linestring?), I suppose that needs a change in the C code?

brendan-ward: The problem is that trying to reduce inputs of length 2 no longer fall into the `YY_Y_func_reduce` function in C, which specifically handles `None` and ignores cases.  They instead fall through to the regular `YY_Y_func` function in C, where we explicitly return `None` if either input is `None`.  Given that this is based on whatever is the inner axis of the operation, it is tough to special case this by filtering `None` out in advance.

Perhaps we also need to revisit if `intersection(Point(0, 0), None)` should return the non-`None` input, or always return `None` (as it does now).  This affects intersection, union, difference, symmetric_difference, and shared_paths.

caspervdw: I think that the bug in #1513 necessitated the `dimensions[0] > 1` check here. Then at least `intersection_all([geom])` does not segfault. Now that the bug is fixed I see no reason in keeping the `dimensions[0] > 1` check.

caspervdw:
Thanks for working on this @brendan-ward  . In my opinion this can however better be solved on the C side, something like

```
if (out == NULL) {
  out = GEOSGeom_createCollection_r(ctx, GEOS_GEOMETRYCOLLECTION, NULL, 0);
}
```

This prevents the additional loop that is done in the `np.where`.

brendan-ward:
We're still bumping into the edge case where we call reduce with arrays that are length 1 along their specified axis, because the underlying ufunc isn't called.

This means that by default numpy wants to return the original items:

```Python
>>> lib.intersection.reduce([None]) is None
True
>>> lib.intersection.reduce([[None]], axis=1)
array([None], dtype=object)
```

Is there a way to address this at a lower level, or is our only option to try and detect that the specified axis is length 1 and backfill None with empty geometry collection?  

Or we just state this in the docstring that if the specified axis is length 1 and contains None values, those will be returned?

jorisvandenbossche:
It might be possible to catch that in the ufunc code, if we add a special case for `dimensions==0` (in practice it only seems to happen in this case that the steps are not 0, I am wondering if this couldn't be considered as a bug in numpy).

I added some print statement, and then I see:

```python
In [1]: import shapely

In [2]: p = shapely.Point(1, 1)

In [3]: shapely.lib.intersection([], [])   # <-- only in the case of empty data, the dimension is 0
ufunc mode with args=[0x55fd5122a210, 0x55fd51b96040, 0x55fd51b8bcb0], steps=[8, 8, 8], dimensions=[0]
Out[3]: array([], dtype=object)

In [4]: shapely.lib.intersection([p], [p])
ufunc mode with args=[0x55fd51b69af0, 0x55fd51d05110, 0x55fd51b87200], steps=[8, 8, 8], dimensions=[1]
Out[4]: array([<POINT (1 1)>], dtype=object)

In [5]: shapely.lib.intersection.reduce([p])   # <-- in this case the steps are also 0
ufunc mode with args=[0x5639dce62af0, 0x5639dce42f00, 0x5639dce62af0], steps=[0, 0, 0], dimensions=[0]
using reduce version
Out[5]: <POINT (1 1)>

In [6]: shapely.lib.intersection.reduce([p, p])
ufunc mode with args=[0x5639dce62af0, 0x5639dc508198, 0x5639dce62af0], steps=[0, 8, 0], dimensions=[1]
using reduce version
Out[6]: <POINT (1 1)>

In [7]: shapely.lib.intersection.reduce([[p, p]])  # <-- but in this case the steps[2] is not 0, while dimension is 0
ufunc mode with args=[0x5639dc5331d0, 0x5639dc508190, 0x5639dc5331d0], steps=[8, 8, 8], dimensions=[0]
Out[7]: array([<POINT (1 1)>, <POINT (1 1)>], dtype=object)

In [8]: shapely.lib.intersection.reduce([[p, p]], axis=1)
ufunc mode with args=[0x5639dce62af0, 0x5639dc508198, 0x5639dce62af0], steps=[0, 8, 0], dimensions=[1]
using reduce version
Out[8]: array([<POINT (1 1)>], dtype=object)

In [9]: shapely.lib.intersection.reduce([[p, p]], axis=0)
ufunc mode with args=[0x5639dce55800, 0x5639dc508190, 0x5639dce55800], steps=[8, 8, 8], dimensions=[0]
Out[9]: array([<POINT (1 1)>, <POINT (1 1)>], dtype=object)

```

jorisvandenbossche:
By doing that, I also noticed that for `axis=0` for a 2D array, we actually never use the reduce version:

```
In [1]: import shapely

In [2]: p = shapely.Point(1, 1)

In [3]: shapely.lib.intersection.reduce([[p, p], [p, p], [p, p]])
ufunc mode with args=[0x55e015087be0, 0x55e014cab5a0, 0x55e015087be0], steps=[8, 8, 8], dimensions=[0]
ufunc mode with args=[0x55e015087be0, 0x55e014cab5b0, 0x55e015087be0], steps=[8, 8, 8], dimensions=[2]
ufunc mode with args=[0x55e015087be0, 0x55e014cab5c0, 0x55e015087be0], steps=[8, 8, 8], dimensions=[2]
Out[3]: array([<POINT (1 1)>, <POINT (1 1)>], dtype=object)
```

If that works in this case, why do we have the special ufunc inner loop for `reduce` at all? 
It does some different things regarding ownership and cloning if needed (but that doesn't seem to be needed above?), ànd also has different behaviour for None. So that's actually something you can see here leading to inconsistent behaviour:

```
In [4]: shapely.lib.intersection.reduce([[p, p], [p, None], [p, p]])
ufunc mode with args=[0x55e0150b2c20, 0x55e014cab5a0, 0x55e0150b2c20], steps=[8, 8, 8], dimensions=[0]
ufunc mode with args=[0x55e0150b2c20, 0x55e014cab5b0, 0x55e0150b2c20], steps=[8, 8, 8], dimensions=[2]
ufunc mode with args=[0x55e0150b2c20, 0x55e014cab5c0, 0x55e0150b2c20], steps=[8, 8, 8], dimensions=[2]
Out[4]: array([<POINT (1 1)>, None], dtype=object)

In [5]: shapely.lib.intersection.reduce([[p, p], [p, None], [p, p]], axis=1)
ufunc mode with args=[0x55e0159b9f80, 0x55e014cab5a8, 0x55e0159b9f80], steps=[0, 8, 0], dimensions=[1]
using reduce version
ufunc mode with args=[0x55e0159b9f88, 0x55e014cab5b8, 0x55e0159b9f88], steps=[0, 8, 0], dimensions=[1]
using reduce version
ufunc mode with args=[0x55e0159b9f90, 0x55e014cab5c8, 0x55e0159b9f90], steps=[0, 8, 0], dimensions=[1]
using reduce version
Out[5]: array([<POINT (1 1)>, <POINT (1 1)>, <POINT (1 1)>], dtype=object)
```

caspervdw:
> If that works in this case, why do we have the special ufunc inner loop for reduce at all?

Looking back at https://github.com/pygeos/pygeos/pull/156, the special version is there so that we can release the GIL. A reduce function creates/destroys PyObjects on the same place in memory (stride == 0). You can't do that without the GIL, hence the special version.

Apparently there are cases of `ufunc.reduce` where `stride != 0`? Probably has to do with the memory alignment of the reduction axis.

brendan-ward:
After struggling with this a bit more, I don't think we're going to be able to easily pull off having the plain ufunc (`YY_Y_func`) implement different behavior if one of the inputs is `None` than the reduce version, specifically because of cases like this:

```Python
>>> shapely.lib.intersection.reduce([[p, p], [p, None], [p, p]])
[<POINT (1 1)>, None]
```

This uses the plain ufunc (same result if axis=0).  Thus if we wait until after doing the calculation to try and backfill `None` with empty geometry collections, we produce a different result if we use the reduce variant, which ignores `None`s and returns a clone of the original geometry if needed:

```Python
>>> shapely.lib.intersection.reduce([p, None, p])
<POINT (1 1)>
```

We still have a special case when the reduction axis is length 1, but the above issue seems like a bigger deal at the moment.

Some options:
1) rework the underlying set operation ufuncs to return a clone of the original geometry if the other geometry input is `None`.  This feels like a bigger API change than we intended here, but we've always treated `None` (missing) geometries as their own special thing that we didn't expect to see often in practice.
2) use a different base ufunc when we are doing a reduce type operation; instead of calling `shapely.lib.intersection.reduce`, we'd have something like `shapely.lib.intersection_reducer.reduce`, which would pretty much be a copy of `lib.intersection` but one that clones the non-`None` input if either of them is `None` instead of only ever returning `None`.

At the moment I'm leaning toward option 2.

caspervdw:
@brendan-ward  Thanks for you efforts here! I reflected a bit on this and wrote it down in #1332 .

We should separate the issue in two parts:

- ignoring None, yes or no
- setting the return value for empty slices.

For the first one, I am all for the separate implementation (your option 2). The second one should be solved by the `initial` kwarg I guess?

jorisvandenbossche:
Superseded by https://github.com/shapely/shapely/pull/1562, so closing

PR #1502, opened 2022-08-25
DOC: update migration guide to use 'packaging' instead of 'distutils'

mwtoews:
Distutils is deprecated, so the documentation should suggest an alternative method to compare versions (one of [many](https://stackoverflow.com/q/11887762)).

Also, this approach needed `version.parse("2.0a1")` to be compared to work with the existing pre-release.

### DIFF
--- a/docs/migration.rst
+++ b/docs/migration.rst
@@ -261,9 +261,9 @@ manager that can be copied into your project::
     import contextlib
     import shapely
     import warnings
-    from distutils.version import LooseVersion
+    from packaging import version  # https://packaging.pypa.io/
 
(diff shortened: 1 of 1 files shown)


jorisvandenbossche (approved):
Thanks!

jorisvandenbossche:
Ah, didn't notice this was for the maint-1.8 branch. Maybe we should get into the habit of doing such changes first for the main branch, and then backport to maint-1.8 instead of the other way around?

jorisvandenbossche:
Cherry-picked to main: https://github.com/shapely/shapely/commit/73ff48c528d466651c590ba856f69e9274467572

Issue #1454, opened 2022-07-18
Rename segmentize() tolerance argument to max_segment_length


jorisvandenbossche:
We like to compare the capabilities to PostGIS, which has [ST_Segmentize](https://postgis.net/docs/ST_Segmentize.html) (based on [internal](https://github.com/postgis/postgis/blob/6febe57e76d2239ea9779c489eff376f3ed76e72/postgis/lwgeom_functions_basic.c#L1861-L1926) [liblwgeom](https://github.com/postgis/postgis/blob/b6cce01cdaf6d31169d73871c374cae332f66992/liblwgeom/ptarray.c#L397-L473)), which uses "max_segment_length" instead of "tolerance". Should shapely also use "max_segment_length"? I've found "tolerance" to be slightly overloaded with other functions, and there could be benefits with parameter names that are more descriptive. But I also don't mind "tolerance", it just requires the user to look-up the docs. At a minimum, the docstrings should be rewritten to "distance tolerance" instead of "tolerance".

Note that this functionality in the GEOS C API is called [`GEOSDensify`](https://github.com/libgeos/geos/blob/e3abcc8fc0f60035c7d8bda65e2417873f6da872/capi/geos_c.h.in#L3885-L3898) with "tolerance".

_Originally posted by @mwtoews in https://github.com/shapely/shapely/issues/1434#issuecomment-1186617986_

jorisvandenbossche:
I find "max_segment_length" in any case quite a bit more descriptive

Issue #1500, opened 2022-08-23
Add a pygeos migration guide for Shapely 2.0 


caspervdw:
The Shapely 2.0 migration guide is now focused on shapely users, but we need to inform existing pygeos users as well.

@jorisvandenbossche

akadouri:
The change of segmentize's `tolerance` parameter to `max_segment_length` is one item that needs to be documented. https://github.com/shapely/shapely/pull/1506

jorisvandenbossche:
The relevant changes from https://github.com/shapely/shapely/issues/1276 are:

- `preserve_topology=True` now default in `simplify` function(https://github.com/shapely/shapely/pull/1392)
- renamed `apply` to `transform` (https://github.com/shapely/shapely/pull/1393)
- The buffer keyword `quadsegs` -> `quad_segs` (https://github.com/shapely/shapely/pull/1512)

PR #1506, opened 2022-08-27
Update segmentize tolerance parameter to max segment length

akadouri:
- Renamed parameter
- Updated function description
- Updated tests

Closes https://github.com/shapely/shapely/issues/1454

### DIFF
--- a/shapely/constructive.py
+++ b/shapely/constructive.py
@@ -703,7 +703,7 @@ def segmentize(geometry, max_segment_length, **kwargs):
     """Adds vertices to line segments based on maximum segment length.
 
     Additional vertices will be added to every line segment in an input geometry
-    so that segments are longer than the provide maximum segment length.  New vertices
+    so that segments are longer than the provided maximum segment length.  New vertices
     will evenly subdivide each segment.
 
     Only linear components of input geometries are densified; other geometries
--- a/shapely/constructive.py
+++ b/shapely/constructive.py
@@ -703,8 +703,8 @@ def segmentize(geometry, max_segment_length, **kwargs):
     """Adds vertices to line segments based on maximum segment length.
 
     Additional vertices will be added to every line segment in an input geometry
-    so that segments are longer than the provided maximum segment length.  New vertices
-    will evenly subdivide each segment.
+    so that segments are no longer than the provided maximum segment length. New
+    vertices will evenly subdivide each segment.
 
     Only linear components of input geometries are densified; other geometries
     are returned unmodified.
--- a/shapely/constructive.py
+++ b/shapely/constructive.py
@@ -699,12 +699,12 @@ def reverse(geometry, **kwargs):
 
 @requires_geos("3.10.0")
 @multithreading_enabled
-def segmentize(geometry, tolerance, **kwargs):
-    """Adds vertices to line segments based on tolerance.
+def segmentize(geometry, max_segment_length, **kwargs):
+    """Adds vertices to line segments based on maximum segment length.
 
     Additional vertices will be added to every line segment in an input geometry
-    so that segments are no greater than tolerance.  New vertices will evenly
-    subdivide each segment.
+    so that segments are longer than the provide maximum segment length.  New vertices
+    will evenly subdivide each segment.
 
     Only linear components of input geometries are densified; other geometries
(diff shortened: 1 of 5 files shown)


jorisvandenbossche (commented):
Looking good, thanks!

One more thing: there is also a `segmentize` method in shapely/geometry/base.py, that needs to be updated as well

akadouri:
Did not realize re-requesting a review would remove the other reviewer 🤷

@jorisvandenbossche Is there somewhere I should note this change? It will break older implementations that use `segmentize` with `tolerance=` (at least those using [2.0a1](https://github.com/shapely/shapely/releases/tag/2.0a1))

shapely/constructive.py
```diff
@@ -699,22 +699,22 @@ def reverse(geometry, **kwargs):
...
 @multithreading_enabled
-def segmentize(geometry, tolerance, **kwargs):
-    """Adds vertices to line segments based on tolerance.
+def segmentize(geometry, max_segment_length, **kwargs):
+    """Adds vertices to line segments based on maximum segment length.
 
     Additional vertices will be added to every line segment in an input geometry
-    so that segments are no greater than tolerance.  New vertices will evenly
-    subdivide each segment.
+    so that segments are longer than the provided maximum segment length.  New vertices
```
brendan-ward, line 706:
```suggestion
    so that segments are no longer than the provided maximum segment length.  New vertices
```

shapely/geometry/base.py
```diff
@@ -753,31 +753,31 @@ def interpolate(self, distance, normalized=False):
...
 
-    def segmentize(self, tolerance):
-        """Adds vertices to line segments based on tolerance.
+    def segmentize(self, max_segment_length):
+        """Adds vertices to line segments based on maximum segment length.
 
         Additional vertices will be added to every line segment in an input geometry
-        so that segments are no greater than tolerance.  New vertices will evenly
-        subdivide each segment.
+        so that segments are longer than the provided maximum segment length.  New
```
brendan-ward, line 760:
```suggestion
        so that segments are no longer than the provided maximum segment length.  New
```

brendan-ward (changes requested):
@akadouri thanks for working on this!  

Since this is targeting a major version release and only showed up briefly in shapely, I think we can skip a deprecation period here.

Agreed on need to document this as a breaking change for both shapely 2.0a1 and also for pygeos under #1500; I don't think we have a place yet for either.  @jorisvandenbossche can advise, but I think so long as we make sure to document this change, this PR can get merged once the typos are fixed.

jorisvandenbossche:
> Since this is targeting a major version release and only showed up briefly in shapely, I think we can skip a deprecation period here.

To be clear, the PR isn't adding deprecations and just changing, so I assume you are agreeing ;)

> Agreed on need to document this as a breaking change for both shapely 2.0a1 and also for pygeos under #1500; I don't think we have a place yet for either. @jorisvandenbossche can advise, but I think so long as we make sure to document this change, this PR can get merged once the typos are fixed.

Yeah, for pygeos it can go in the to-be-written migration guide for pygeos. For shapely 2.0a1 -> 2.0b1, we don't have a place. And I don't want to include it in the main changelog for 2.0 (as it is just a distraction later on for people reading the 2.0.0 release notes). Maybe we can include those actual releases (a1, b1, etc) in CHANGES.txt and so mention it there? (although that might also get a bit messy later on to know what needs to be documented where)

(to be honest, I also don't think this particular case is very important. Exactly because it is a new function, I don't think that many people will already started using it in shapely 2.0a1, unless they were migrating from pygeos, for which we certainly need a guide as well)

akadouri:
Thanks for the feedback. Yeah that makes sense with how few are likely already relying on it in shapely 2.0a1.

jorisvandenbossche:
@brendan-ward I added a small note for this change in CHANGES.txt. Together with the changes in https://github.com/shapely/shapely/pull/1520, I think we can keep such a note in CHANGES.txt, while the release docs just focus on the full 2.0.0 release. 

For pygeos, we can include this later when tackling https://github.com/shapely/shapely/issues/1500

jorisvandenbossche:
Thanks @akadouri !

Issue #1507, opened 2022-08-27
Consolidate buffer enum classes


jorisvandenbossche:
Enums for buffer keyword values:

* Original shapely classes: https://github.com/shapely/shapely/blob/03284740014ba0d8b1917752a6b5dcbfd4ffeb39/shapely/geometry/base.py#L48-L57
* New ones from pygeos: https://github.com/shapely/shapely/blob/03284740014ba0d8b1917752a6b5dcbfd4ffeb39/shapely/constructive.py#L36-L45

Decision: alias the new (from pygeos) enums with the original shapely enum names

PR #1508, opened 2022-08-27
Alias geometry.base buffer styles to geometry.constructive enums

akadouri:
- Update `CAP_STYLE` and `JOIN_STYLE` in `shapely/geometry/base.py` to alias `BufferCapStyles` and `BufferJoinStyles` enums in `shapely/constructive.py`
- Update docs to encourage use of new enums
- Refactor `tests/test_styles.py` tests into `tests/test_buffer.py`
- Add test to ensure they are the same

geometry.base styles are now just used in https://github.com/shapely/shapely/blob/main/shapely/geometry/__init__.py#L26-L27

Closes https://github.com/shapely/shapely/issues/1507


jorisvandenbossche (commented):
Thanks for the PR! 
Can you also add a test where we pass the old class value to the `buffer` method? (in `tests/test_buffer.py`). It seems we currently only test those CAP_STYLE classes superficially in test_styles.py)

jorisvandenbossche:
@caspervdw something I am wondering from reviewing this PR: the current enum classes use a plural form like `BufferCapStyles`. Would the singular form make more sense, like `BufferCapStyle` ?

The enum of course is to group _multiple_ style options, but when using it, you use it specify a _single_ style (the examples in the python docs also seem to use a singular form: https://docs.python.org/3/library/enum.html#creating-an-enum)

akadouri:
I believe the python style standard is to have the enum class name singular as well as the values capitalized, so `BufferCapStyles.round` should be `BufferCapStyle.ROUND`. I can make that change if that's wanted.

tests/test_styles.py
```diff
@@ -1,15 +1,21 @@
 from . import unittest
 from shapely.geometry import CAP_STYLE, JOIN_STYLE
+from shapely.constructive import BufferCapStyles, BufferJoinStyles
 
 
 class StylesTest(unittest.TestCase):
-
     def test_cap(self):
```
mwtoews, line 7: the tests below can be improved to simplify (e.g.):
```diff
-        self.assertEqual(CAP_STYLE.round, 1)
+        assert CAP_STYLE.round == 1
```
(the `self.assert*` functions were used with unittest, prior to pytest)

akadouri: updated to use `assert`, also moved the tests into `tests/test_buffer.py`

docs/manual.rst
```diff
@@ -1524,13 +1524,13 @@ derived from set-theoretic analysis.
 
   The styles of caps are specified by integer values: 1 (round), 2 (flat),
   3 (square). These values are also enumerated by the object
-  :class:`shapely.geometry.CAP_STYLE` (see below).
+  :class:`shapely.constructive.BufferCapStyles` (see below).
```
jorisvandenbossche, line 1527:
```suggestion
  :class:`shapely.BufferCapStyles` (see below).
```

We import it in the top-level namespace, so we can refer to it like that here as well

docs/manual.rst
```diff
@@ -1552,10 +1552,10 @@ derived from set-theoretic analysis.
 
 .. code-block:: pycon
 
-  >>> from shapely.geometry import CAP_STYLE, JOIN_STYLE
-  >>> CAP_STYLE.flat
+  >>> from shapely.constructive import BufferCapStyles, BufferJoinStyles
```
jorisvandenbossche, line 1555:
```suggestion
  >>> from shapely import BufferCapStyles, BufferJoinStyles
```

docs/manual.rst
jorisvandenbossche: and the same for the few other occurrences below

tests/test_buffer.py
```diff
@@ -60,19 +70,78 @@ def test_line_single_sideded_right(self):
...
+        assert CAP_STYLE.square == BufferCapStyles.square
+
+        assert JOIN_STYLE.round == 1
+        assert JOIN_STYLE.round == BufferJoinStyles.round
+        assert JOIN_STYLE.mitre == 2
+        assert JOIN_STYLE.mitre == BufferJoinStyles.mitre
+        assert JOIN_STYLE.bevel == 3
+        assert JOIN_STYLE.bevel == BufferJoinStyles.bevel
+
+    def test_styles(self):
```
brendan-ward, line 123: Suggest splitting this into 2 more focused tests, one for cap styles and one for join styles.

tests/test_buffer.py
```diff
@@ -60,19 +70,78 @@ def test_line_single_sideded_right(self):
...
+        g = geometry.LineString([[0, 0], [1, 0]])
+        h = g.buffer(1, cap_style=CAP_STYLE.round)
```
brendan-ward, line 125: I'd suggest flipping this, so that the expected value `h` is based on the currently recommended enum value, then the 2 assertions are based on the other forms (string, `CAP_STYLE.*`) to prove that they are equivalent to the recommended value.

brendan-ward (commented):
Thanks for working on this @akadouri !  A few comments to consider

caspervdw:
> enum

That makes sense, I am 👍 on the singular class name and all-caps values.

shapely/geometry/base.py
```diff
@@ -351,8 +352,8 @@ def buffer(
         distance,
         resolution=16,
         quadsegs=None,
-        cap_style=CAP_STYLE.round,
-        join_style=JOIN_STYLE.round,
+        cap_style="round",
```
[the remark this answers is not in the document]
brendan-ward: > I actually asked to use "round"

Sorry, I didn't see that here.

The problem with strings is that they are out of sync with the parameter descriptions below:

```Python
cap_style : int, optional
```

Maybe changing that to 

```Python
cap_style : shapely.BufferCapStyles, optional
```

Would make it more clear that either a string that matches an enum, or the enum value, or an integer that matches an enum value all work?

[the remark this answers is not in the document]
jorisvandenbossche: > Sorry, I didn't see that here.

Yeah, sorry, I did that in person ;) (just wanted to indicate that @akadouri didn't make that change out of the blue)

> The problem with strings is that they are out of sync with the parameter descriptions below:

That parameter description is out of date anyway, since we now do allow more with the new implementation. The top-level function has `cap_style : {'round', 'square', 'flat'}, default 'round'`, but so actually also doesn't mention the enum and is also not fully complete ..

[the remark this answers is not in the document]
akadouri: In updating the enums to be in line with python styles, those values now become `ROUND`, `SQUARE`, etc. I think it makes sense to direct people to using the enums, though I guess this will break backwards compatibility...

One option is to update https://github.com/shapely/shapely/blob/main/shapely/enum.py#L16 to `upper()` the input values. I'm not sure what other libraries do to allow people to avoid using enums?

shapely/geometry/base.py
```diff
@@ -351,8 +352,8 @@ def buffer(
         distance,
         resolution=16,
         quadsegs=None,
-        cap_style=CAP_STYLE.round,
-        join_style=JOIN_STYLE.round,
+        cap_style=BufferCapStyle.ROUND,
+        join_style=BufferJoinStyle.ROUND,
```
akadouri, line 356: Should geometry.base's buffer have references to the new enums or preserve the old ones?

shapely/constructive.py
```diff
@@ -108,16 +108,17 @@ def buffer(
...
-        Specifies the shape of buffered line endings. 'round' results in
-        circular line endings (see ``quadsegs``). Both 'square' and 'flat'
-        result in rectangular line endings, only 'flat' will end at the
-        original vertex, while 'square' involves adding the buffer width.
-    join_style : {'round', 'bevel', 'mitre'}, default 'round'
-        Specifies the shape of buffered line midpoints. 'round' results in
-        rounded shapes. 'bevel' results in a beveled edge that touches the
-        original vertex. 'mitre' results in a single vertex that is beveled
-        depending on the ``mitre_limit`` parameter.
+    cap_style : shapely.BufferCapStyle, shapely.BufferCapStyle.ROUND
```
mwtoews, line 111:
```suggestion
    cap_style : shapely.BufferCapStyle, default shapely.BufferCapStyle.ROUND
```

shapely/constructive.py
```diff
@@ -108,16 +108,17 @@ def buffer(
...
+        Specifies the shape of buffered line endings. BufferCapStyle.ROUND results in
+        circular line endings (see ``quadsegs``). Both BufferCapStyle.SQUARE and
+        BufferCapStyle.FLAT result in rectangular line endings, only BufferCapStyle.FLAT
+        will end at the original vertex, while BufferCapStyle.SQUARE involves adding the
+        buffer width.
+    join_style : shapely.BufferJoinStyle, shapely.BufferJoinStyle.ROUND
```
mwtoews, line 117:
```suggestion
    join_style : shapely.BufferJoinStyle, default shapely.BufferJoinStyle.ROUND
```

shapely/constructive.py
```diff
@@ -129,41 +130,41 @@ def buffer(
...
     <POLYGON ((2 2, 2 8, 8 8, 8 2, 2 2))>
-    >>> buffer(square, -5, join_style="mitre")
+    >>> buffer(square, -5, join_style=BufferJoinStyle.MITRE)
     <POLYGON EMPTY>
     >>> buffer(line, float("nan")) is None
     True
     """
     if isinstance(cap_style, str):
-        cap_style = BufferCapStyles.get_value(cap_style)
+        cap_style = BufferCapStyle.get_value(cap_style)
```
mwtoews, line 165:
```suggestion
        cap_style = BufferCapStyle.get_value(cap_style.upper())
```

shapely/constructive.py
```diff
@@ -129,41 +130,41 @@ def buffer(
...
     if isinstance(join_style, str):
-        join_style = BufferJoinStyles.get_value(join_style)
+        join_style = BufferJoinStyle.get_value(join_style)
```
mwtoews, line 167:
```suggestion
        join_style = BufferJoinStyle.get_value(join_style.upper())
```

mwtoews (commented):
A few more suggestions. The docstring descriptions could possibly be shortened from using `BufferCapStyle.ROUND` to simply `ROUND`.

The suggestions `get_value(join_style.upper())` will allow (e.g.) "round" and "ROUND".

jorisvandenbossche:
Personally, I am not fully enthusiastic about strongly favoring using `cap_style=shapely.BufferCapStyle.ROUND` in the docstring (at least we should make it very clear that passing a string is also fine). While having enums can certainly be nice, many of our users will also not be very familiar with enums, and IMO something like `cap_style="round"` is still the easiest and most readable option (that I would personally use in code).

jorisvandenbossche:
What are other people's thoughts here?

I would personally prefer to keep using `cap_style="round"` instead of `cap_style=shapely.BufferCapStyle.ROUND` (in the _signature_, while of course clearly documenting the enums in both the docstring parameters section as in the manual).  
But if others prefer using the enum in the signature as well, it's certainly not a hill I want to die on ;)

brendan-ward:
Strings are fine by me, and are nicer for the reasons you list above, so long as in the docstring we make sure to list valid values, and also allow enum values.  The enums are perhaps more useful for validation than for input.

akadouri:
Make sense to me to keep things simple with string params. I'll update the doc strings. Thanks for all the feedback everyone.

docs/manual.rst
```diff
@@ -1710,8 +1710,8 @@ linestring feature (right).
   parameterized as in the :meth:`buffer` method.
 
   The `join_style` is for outside corners between line segments. Accepted integer
-  values are 1 (round), 2 (mitre), and 3 (bevel). See also
-  :data:`shapely.geometry.JOIN_STYLE`.
+  values are 1 (ROUND), 2 (MITRE), and 3 (BEVEL). See also
+  :data:`shapely.constructive.BufferJoinStyle`.
```
jorisvandenbossche, line 1714:
```suggestion
  :data:`shapely.BufferJoinStyle`.
```

shapely/constructive.py
```diff
@@ -108,16 +108,17 @@ def buffer(
...
-        result in rectangular line endings, only 'flat' will end at the
-        original vertex, while 'square' involves adding the buffer width.
-    join_style : {'round', 'bevel', 'mitre'}, default 'round'
-        Specifies the shape of buffered line midpoints. 'round' results in
-        rounded shapes. 'bevel' results in a beveled edge that touches the
-        original vertex. 'mitre' results in a single vertex that is beveled
-        depending on the ``mitre_limit`` parameter.
+    cap_style : shapely.BufferCapStyle or {'round', 'square', 'flat'}, default 'round'
+        Specifies the shape of buffered line endings. BufferCapStyle.ROUND ('round')
+        results in circular line endings (see ``quadsegs``). Both BufferCapStyle.SQUARE
```
jorisvandenbossche, line 113:
```suggestion
        results in circular line endings (see ``quad_segs``). Both BufferCapStyle.SQUARE
```

(this has changed in the meantime)

shapely/constructive.py
```diff
@@ -129,41 +130,41 @@ def buffer(
...
+    >>> from shapely import LineString, Point, Polygon, BufferCapStyle, BufferJoinStyle
     >>> buffer(Point(10, 10), 2, quadsegs=1)
     <POLYGON ((12 10, 10 8, 8 10, 10 12, 12 10))>
     >>> buffer(Point(10, 10), 2, quadsegs=2)
     <POLYGON ((12 10, 11.414 8.586, 10 8, 8.586 8.586, 8 10, 8.5...>
     >>> buffer(Point(10, 10), -2, quadsegs=1)
     <POLYGON EMPTY>
     >>> line = LineString([(10, 10), (20, 10)])
-    >>> buffer(line, 2, cap_style="square")
+    >>> buffer(line, 2, cap_style='square')
```
jorisvandenbossche, line 141: It also doesn't matter too much, but so this change is not needed (I suppose it is a left-over from changing this and then changing back)

shapely/constructive.py
```diff
@@ -188,7 +189,7 @@ def buffer(
 
 @multithreading_enabled
 def offset_curve(
-    geometry, distance, quadsegs=8, join_style="round", mitre_limit=5.0, **kwargs
+    geometry, distance, quadsegs=8, join_style="ROUND", mitre_limit=5.0, **kwargs
```
jorisvandenbossche, line 192:
```suggestion
    geometry, distance, quadsegs=8, join_style="round", mitre_limit=5.0, **kwargs
```

(to be consistent with buffer())

shapely/tests/test_constructive.py
```diff
@@ -261,7 +261,7 @@ def test_offset_curve_distance_array():
 def test_offset_curve_kwargs():
     # check that kwargs are passed through
     result1 = shapely.offset_curve(
-        line_string, -2.0, quadsegs=2, join_style="mitre", mitre_limit=2.0
+        line_string, -2.0, quadsegs=2, join_style="MITRE", mitre_limit=2.0
```
jorisvandenbossche, line 264:
```suggestion
        line_string, -2.0, quadsegs=2, join_style="mitre", mitre_limit=2.0
```

(let's leave some as lower case, to ensure this is covered that it works)

jorisvandenbossche (approved):
Thanks for the updates! (just a few minor comments)

akadouri:
Thanks for all the feedback everyone, @jorisvandenbossche I made those changes. I think the tests are failing now due to unrelated issues with the geos main build.

jorisvandenbossche:
@mwtoews @caspervdw @brendan-ward more comments here?

> I think the tests are failing now due to unrelated issues with the geos main build.

Yes, that's unrelated, see https://github.com/shapely/shapely/issues/1541

shapely/tests/test_constructive.py
```diff
@@ -105,12 +105,12 @@ def test_float_arg_nan(geometry, func):
 
 
 def test_buffer_cap_style_invalid():
-    with pytest.raises(ValueError, match="'invalid' is not a valid option"):
+    with pytest.raises(ValueError, match="'INVALID' is not a valid option"):
```
brendan-ward, line 108: This seems to point out an inconsistency: we instruct users to provide these in lowercase strings, which we automatically uppercase before checking against the enums.  `ParamEnum` handles the validation, but has no idea that the inputs are originally lowercase.


While minor, this might be confusing to users, since it returns a different casing on exception than they provided on input.

Perhaps it would be better if the entries were lowercase?

(note: the other enums `BinaryPredicate` and `DecodingErrorOptions` have lowercase entries)

akadouri: Back in these comments we decided to uppercase the enums since I believe that's the python style guide.
- https://github.com/shapely/shapely/pull/1508#issuecomment-1229214042
- https://github.com/shapely/shapely/pull/1508#issuecomment-1229277562

We could change things to use regular Enums with the values as the lowercase strings. (Though this would drop support for using 1/2/3 as inputs.)

```
class BufferCapStyle(Enum):
    ROUND = "round"
    FLAT = "flat"
    SQUARE = "square"


class BufferJoinStyle(Enum):
    ROUND = "round"
    MITRE = "mitre"
    BEVEL = "bevel"

>>> BufferCapStyle['ROUND']
<BufferCapStyle.ROUND: 'round'>
```

Then run input validation in `buffer` like
```
try:
   buffer_cap_style = BufferCapStyle[input.upper()]
except KeyError:
   raise ValueError(f"'{input"}' not in BufferCapStyle, options: {[e.value for e in BufferCapStyle]}")
```

which would return
```
'invalid' not in BufferCapStyle, options: ['round', 'flat', 'square']
```


Could also just do the input validation in `buffer` with the existing ParamEnum. Or could return the entries back to lowercase. I don't have a strong preference.

akadouri: After writing that all out I checked ParamEnum and realized it already does that validation that prints the enum options, the test just only checked the first half. The rest of that output would be `'Invalid' is not a valid option, must be one of 'SQUARE', 'ROUND', 'FLAT'`. That is kind of confusing.

brendan-ward: I think in this case we could opt for lowercase enum entries and deviate a bit from the Python style guide; that keeps us consistent with the other enums here, allows us to continue leveraging the validation messages from `ParamEnum` with no changes, and does not require a casing change to use incoming values.

jorisvandenbossche: Changing back to lower case enum values is OK for me. 
Alternatively, we could also convert the values to lower case when printing them in the error message.

akadouri: I changed the enums to lower case, changing the case of the input to lowercase may be confusing as well.

jorisvandenbossche:
@akadouri thanks for the update!

I am going to merge this. The main part here is that the old enums are made compatible. We can still further tweak the docstring or value we put in the signature in follow-up PRs if needed (but that shouldn't affect behaviour for people testing shapely 2)

PR #1510, opened 2022-08-27
Make parallel_offset an alias for offset_curve

enricofer:
See https://github.com/shapely/shapely/issues/1276

ref foss4g2022 codesprint
parallel_offset as ALIAS OF offset_curve new method

### DIFF
--- a/shapely/geometry/linestring.py
+++ b/shapely/geometry/linestring.py
@@ -110,7 +110,7 @@ class LineString(BaseGeometry):
         """
         return self.coords.xy
 
-    def parallel_offset(
+    def offset_curve(
         self,
         distance,
         side="right",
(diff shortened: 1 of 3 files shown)


shapely/geometry/linestring.py
```diff
@@ -145,5 +145,30 @@ def parallel_offset(
...
+        The join style is for outside corners between line segments. Accepted
+        values are JOIN_STYLE.round (1), JOIN_STYLE.mitre (2), and
+        JOIN_STYLE.bevel (3).
+
+        The mitre ratio limit is used for very sharp corners. It is the ratio
+        of the distance from the corner to the end of the mitred offset corner.
+        When two line segments meet at a sharp angle, a miter join will extend
+        far beyond the original geometry. To prevent unreasonable geometry, the
+        mitre limit allows controlling the maximum length of the join corner.
+        Corners with a ratio which exceed the limit will be beveled.
```
jorisvandenbossche, line 169: On second thought, maybe we can remove this documentation here (to not have it duplicated with the other method, since it is quite long), and just refer to `offset_curve` to see the documentation.

jorisvandenbossche (approved):
Looks good, thanks for the contribution!

PR #1511, opened 2022-08-28
PERF: use GEOSCoordSeq_copyToBuffer for get_coordinates (GEOS >= 3.10)

jorisvandenbossche:
Similar as https://github.com/pygeos/pygeos/pull/436, but now for getting coordinates.

Small example showing the speed-up (10,000 polygons of 100 vertices):

```python
geoms = shapely.polygons(np.random.randn(10000, 100, 2))

In [10]: %timeit shapely.get_coordinates(l)
13.6 ms ± 314 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)   # <-- main
6.46 ms ± 322 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)   # <-- PR
```

### DIFF
--- a/src/geos.c
+++ b/src/geos.c
@@ -950,20 +950,18 @@ GEOSCoordSequence* coordseq_from_buffer(GEOSContextHandle_t ctx, const double* b
  *
  * Returns 0 on error, 1 on success.
  */
-char coordseq_to_buffer(GEOSContextHandle_t ctx, const GEOSCoordSequence* coord_seq,
-                        double* buf, unsigned int size, unsigned int dims) {
-  char *cp1, *cp2;
-  unsigned int i, j;
+int coordseq_to_buffer(GEOSContextHandle_t ctx, const GEOSCoordSequence* coord_seq,
+                       double* buf, unsigned int size, unsigned int dims) {
 
 #if GEOS_SINCE_3_10_0
 
   int hasZ = dims == 3;
-  if (!GEOSCoordSeq_copyToBuffer_r(ctx, coord_seq, buf, hasZ, 0)) {
-    return 0;
-  }
-  return 1;
+  return GEOSCoordSeq_copyToBuffer_r(ctx, coord_seq, buf, hasZ, 0);
 
-#endif
+#else
+
+  char *cp1, *cp2;
+  unsigned int i, j;
 
   cp1 = (char*)buf;
   for (i = 0; i < size; i++, cp1 += 8 * dims) {
@@ -975,4 +973,6 @@ char coordseq_to_buffer(GEOSContextHandle_t ctx, const GEOSCoordSequence* coord_
     }
   }
   return 1;
+
+#endif
 }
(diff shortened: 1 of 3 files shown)


src/geos.c
```diff
@@ -942,3 +942,37 @@ GEOSCoordSequence* coordseq_from_buffer(GEOSContextHandle_t ctx, const double* b
...
+ */
+char coordseq_to_buffer(GEOSContextHandle_t ctx, const GEOSCoordSequence* coord_seq,
+                        double* buf, unsigned int size, unsigned int dims) {
+  char *cp1, *cp2;
+  unsigned int i, j;
+
+#if GEOS_SINCE_3_10_0
+
+  int hasZ = dims == 3;
+  if (!GEOSCoordSeq_copyToBuffer_r(ctx, coord_seq, buf, hasZ, 0)) {
```
brendan-ward, line 961: What about changing the return value to `int` and then 

```C
return GEOSCoordSeq_copyToBuffer_r(ctx, coord_seq, buf, hasZ, 0);
```

src/geos.c
```diff
@@ -942,3 +942,37 @@ GEOSCoordSequence* coordseq_from_buffer(GEOSContextHandle_t ctx, const double* b
...
+/* Copy coordinates of a GEOSCoordSequence to an array
+ *
+ * Note: this function assumes that the buffer is from a C-contiguous array,
+ * and that the dimension of the buffer is only 2D or 3D.
+ *
+ * Returns 0 on error, 1 on success.
+ */
+char coordseq_to_buffer(GEOSContextHandle_t ctx, const GEOSCoordSequence* coord_seq,
+                        double* buf, unsigned int size, unsigned int dims) {
+  char *cp1, *cp2;
```
brendan-ward, line 955: can move these two declaration lines (955-956) to closer to where they are needed (line 967), since they are not used under GEOS 3.10

src/geos.c
```diff
@@ -942,3 +942,37 @@ GEOSCoordSequence* coordseq_from_buffer(GEOSContextHandle_t ctx, const double* b
...
+
+#if GEOS_SINCE_3_10_0
+
+  int hasZ = dims == 3;
+  if (!GEOSCoordSeq_copyToBuffer_r(ctx, coord_seq, buf, hasZ, 0)) {
+    return 0;
+  }
+  return 1;
+
+#endif
```
brendan-ward, line 966: This needs to be an `#else` and `#endif` moved below line 977, otherwise we get compiler warnings when building against GEOS 3.10:

```
src/geos.c:968:16: warning: code will never be executed [-Wunreachable-code]
```

brendan-ward (commented):
Nice to see a 2x speedup.  A few minor suggestions, but otherwise this looks good!

src/geos.c
jorisvandenbossche: Changed to int (I used char because some of our other helper functions were using that, in general we don't seem to be very consistent about using `int` vs `char` as return type)

src/geos.c
jorisvandenbossche: Ah, I don't get that warning, but easy change to do

PR #1512, opened 2022-08-28
API: use quad_segs for buffer keyword instead of resolution

jorisvandenbossche:
xref https://github.com/shapely/shapely/issues/1276, making the buffer keyword consistent. We thought `quad_segs` might be better (`resolution` does not really signal what the value actually means, and `quad_segs` is also consistent with GEOS/PostGIS).

This PR does:

- changed `quadsegs` to `quad_segs` in the `buffer`/`offset_curve` top-level functions (those are new, so didn't add any compatibility for old names)
- added `quad_segs` to the `buffer` method, while for now keep allowing `resolution` as well. I did not (yet) deprecate `resolution`, as that makes it easier to write code that is compatible with both <=1.8 and >=2.0 (but we should maybe do that in the future)
- changed `resolution` to `quad_segs` in the `offset_curve` method (this is a new method, so didn't add any compatibility for old names)
- kept the `resolution` keyword in `parallel_offset` as is, since we now favor `offset_curve` alias (which already uses the new `quad_segs` keyword)
  - Although I could also do the same as in the `buffer` method and just allow both resolution and quad_segs?

The above changes shouldn't break anyone's code (except if migrating from pygeos, then you might need to update the keyword name).

There is one additional change that I want to highlight (because I don't think we discussed this before):

- the `offset_curve` function uses a positive/negative distance to indicate a left/right offset (which is consistent with how `buffer` does this in the single-sided case). However, the `parallel_offset` method is using a `side="left"|"right"` keyword instead.  
  I changed the new `offset_curve` method we added in https://github.com/shapely/shapely/pull/1510 to be consistent with the `offset_curve` top-level function (so using sign of the distance), instead of being consistent with the existing `parallel_offset` method. Since `buffer` also uses the sign (both the top-level function and the geometry method), that seems the best option, but wanted to flag this explicitly.

cc @caspervdw @sgillies @brendan-ward @mwtoews


sgillies:
@jorisvandenbossche I'm :+1: on signed distance for offset_curve.

docs/manual.rst
```diff
@@ -1699,22 +1699,25 @@ linestring feature (right).
   Returns a LineString or MultiLineString geometry at a distance from the
   object on its right or its left side.
 
-  Alias method to :meth:`offset_curve` method.
+  Alias method to :meth:`offset_curve` method, but uses `resolution` instead
```
brendan-ward, line 1702: Aliases with different parameters feels like a weird pattern.  Instead, it seems better to have this as a pure alias, and then having `resolution` and `'left'` / `'right'` values as deprecated for shapely 2.0.

shapely/geometry/base.py
```diff
@@ -435,20 +435,29 @@ def buffer(
...
             warn(
-                "The `quadsegs` argument is deprecated. Use `resolution`.",
-                DeprecationWarning,
+                "The `quadsegs` argument is deprecated. Use `quad_segs` instead.",
+                FutureWarning,
             )
-            resolution = quadsegs
+            quad_segs = quadsegs
+
+        # TODO deprecate `resolution` keyword in the future as well
```
brendan-ward, line 446: It seems like the time to go ahead and deprecate this under shapely 2.0 instead of doing so later.

brendan-ward (commented):
Thanks for working on this @jorisvandenbossche .  A few comments to consider re: deprecations, but otherwise this looks good.

shapely/geometry/base.py
jorisvandenbossche: > It seems like the time to go ahead and deprecate this under shapely 2.0 instead of doing so later.

I don't feel very strong about this, but so I gave some rationale for not yet deprecating it in the top post: keeping it working (for now) without warning makes it easier to write code that is compatible with both <=1.8 and >=2.0. If we start deprecating it, any other package that uses this and wants to initially support both 1.8 and 2.0 needs to do an if/else based on the shapely version to avoid showing the warning to users.

(one might think that there are already other cases that would need such version checks with the many changes in shapely 2.0, but for example all breaking changes for which we warned in 1.8 can all be solved by using already existing APIs)

docs/manual.rst
jorisvandenbossche: But shouldn't we then just deprecate the whole method? Because if we ask users to no longer pass the `side` keyword, they need to update their code anyway, and they can just as well change the method name?

brendan-ward: Deprecating the method seems reasonable.  To be clear: I don't have strong opinions about this, only that aliases that are not identical to the thing they are aliasing are not really aliases...

jorisvandenbossche: >  only that aliases that are not identical to the thing they are aliasing are not really aliases...

I can also not call it an "alias" ;)

jorisvandenbossche:
> > It seems like the time to go ahead and deprecate this under shapely 2.0 instead of doing so later.
> 
> I don't feel very strong about this, but so I gave some rationale for not yet deprecating it in the top post: keeping it working (for now) without warning makes it easier to write code that is compatible with both <=1.8 and >=2.0. If we start deprecating it, any other package that uses this and wants to initially support both 1.8 and 2.0 needs to do an if/else based on the shapely version to avoid showing the warning to users.

Are people OK with punting on the deprecation for now?

jorisvandenbossche:
I opened https://github.com/shapely/shapely/issues/1531 to further track the possible deprecation of the `resolution` keyword and `parallel_offset` method.

PR #1514, opened 2022-08-30
TST: assert rewriting in tests

enricofer:
ref FOSS4G2022 codesprint

PR #1516, opened 2022-09-01
Raise ValueError for non-finite distance to buffer/parallel_offset (maint-1.8)

### LINKED ISSUE #1483: segmentation fault using float("nan") in parallel_offset
Using a nan as a parallel_offset causes a segmentation fault.

Expected the function to return e.g. a ctypes.ArgumentError: argument 3: <class 'TypeError'>: wrong type (as happens if you pass a string "nan")

from shapely.geometry import shape
shape({"type": "LineString", "coordinates": [[0,0],[0,1]]}).parallel_offset(float("nan"))

## Operating system

Linux 

1.8.2 installed form pypi using pip

### LINKED ISSUE #1515: Segfault on .buffer(np.nan)
Expect: Not a Segfault, anything one can catch in a python program.
Actual: Segfault. Cannot catch.

```
import numpy as np
Point(1,1).buffer(np.nan)
```

## Operating system
Debian 10

shapely==1.8.2 installed from PyPI using pip

mwtoews:
Fixes #1483 and fixes #1515 for the shapely 1.8.x maintenance branch.

I'll apply a similar approach for the main branch soon (sometime after #1514 is merged).

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -622,6 +622,8 @@ class BaseGeometry:
         if mitre_limit == 0.0:
             raise ValueError(
                 'Cannot compute offset from zero-length line segment')
+        elif not math.isfinite(distance):
+            raise ValueError("buffer distance must be finite")
 
         if 'buffer_with_params' in self.impl:
             params = self._lgeos.GEOSBufferParams_create()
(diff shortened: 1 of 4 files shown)


jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

> I'll apply a similar approach for the main branch soon (sometime after https://github.com/shapely/shapely/pull/1514 is merged).

See also https://github.com/shapely/shapely/issues/1483#issuecomment-1217510186, it might be that we only need to add tests on the main branch

Issue #1296, opened 2022-01-26
Feature: concave hulls


mwtoews:
The GEOS C API main branch has a new feature [`GEOSConcaveHull`](https://github.com/libgeos/geos/blob/70d2111bd66ea848809e5738896256199d9a7cba/capi/geos_c.h.in#L2585-L2608) which was [ported from JTS](https://github.com/locationtech/jts/pull/823) to evaluate concave hulls [described here](https://lin-ear-th-inking.blogspot.com/2022/01/concave-hulls-in-jts.html) by Martin Davis.

See also [PostGIS docs for `ST_ConcaveHull`](https://postgis.net/docs/ST_ConcaveHull.html).

jorisvandenbossche:
There are still some other "new" features from previous GEOS versions as well that are not yet exposed in Shapely. We should probably move over a cleaned-up version of https://github.com/pygeos/pygeos/issues/126 to the Shapely repo.

PR #1518, opened 2022-09-02
ENH: Add concave_hull() function

gahjelle:
GEOS introduces a `GEOSConcaveHull()` function in v3.11. This PR adds a `shapely` binding.

(Contributed during EuroSciPy 2022 sprints in Basel)

Closes https://github.com/shapely/shapely/issues/1296


gahjelle:
Docs are still missing.

jorisvandenbossche:
@gahjelle I rebased this and added a test for the keyword arguments (and a small fix for it)

I seem to remember you started on a section for the user guide, but you didn't finish / push that yet? 
Now, I think this is certainly already in a good state (it has API docs), so I would propose to merge this as is, and certainly feel free to still open a follow-up PR with those docs (even if you don't find it finished).

gahjelle:
@jorisvandenbossche Thanks a lot for cleaning this up and bringing it to completion!

I started looking at the extra documentation but didn't really do much there yet.

jorisvandenbossche:
Thanks again!

PR #1519, opened 2022-09-03
DOC/CLN: clean-up conf.py and unused docs files

jorisvandenbossche:
This PR:

- Cleans up the conf.py file: I generated a new one based on the latest version of sphinx-quickstart, and then filled in with what we had in the previous file (this removes a lot of commented out options, giving a shorter file)
- Remove the no longer used files in `docs/sphinxext` and `docs/images` directories

This also fixes an issue with the `requires_geos` decorator not working to inject a note in the docstring (https://github.com/shapely/shapely/pull/1496#discussion_r952025357, we were importing shapely before setting the env variable)


docs/conf.py
```diff
@@ -1,38 +1,33 @@
...
 
-# If extensions (or modules to document with autodoc) are in another directory,
-# add these directories to sys.path here. If the directory is relative to the
-# documentation root, use os.path.abspath to make it absolute, like shown here.
-sys.path.append(os.path.abspath('sphinxext'))
+# -- Project information -----------------------------------------------------
 
-import shapely
+project = 'Shapely'
+copyright = '2011-2022, Sean Gillies and Shapely contributors'
```
jorisvandenbossche, line 15: I also updated this to match what we have in pyproject.toml now

docs/conf.py
```diff
@@ -44,98 +39,13 @@
...
-#today = ''
-# Else, today_fmt is used as the format for a strftime call.
-#today_fmt = '%B %d, %Y'
-
-# List of documents that shouldn't be included in the build.
-#unused_docs = []
-
-# List of directories, relative to source directory, that shouldn't be searched
-# for source files.
-exclude_trees = ['_build']
```
jorisvandenbossche, line 85: This config option no longer exists

docs/conf.py
```diff
@@ -44,98 +39,13 @@
...
-
 html_theme = "sphinx_book_theme"
 
-# Theme options are theme-specific and customize the look and feel of a theme
-# further.  For a list of options available for each theme, see the
-# documentation.
-#html_theme_options = {}
-
-# Add any paths that contain custom themes here, relative to this directory.
-html_theme_path = ['themes']
```
jorisvandenbossche, line 121: We don't have a theme directory, so not needed

docs/conf.py
```diff
@@ -149,87 +59,9 @@
...
-# If true, an OpenSearch description file will be output, and all pages will
-# contain a <link> tag referring to it.  The value of this option must be the
-# base URL from which the finished HTML is served.
-#html_use_opensearch = ''
-
-# If nonempty, this is the file name suffix for HTML files (e.g. ".xhtml").
-#html_file_suffix = ''
-
-# Output file base name for HTML help builder.
-htmlhelp_basename = 'Shapelydoc'
```
jorisvandenbossche, line 189: We don't build "html help" output ourselves, so I don't think it is needed to keep this?

brendan-ward (approved):
Thanks @jorisvandenbossche , this appears to work properly, and it is good to have the GEOS version note autogenerated for methods / functions tied to specific GEOS versions.

I do see that there are a lot of warnings emitted during doc building, but I'm assuming we'll need to make a pass through these before 2.0 in order to address those that might impact the docs.

jorisvandenbossche:
> I do see that there are a lot of warnings emitted during doc building, but I'm assuming we'll need to make a pass through these before 2.0 in order to address those that might impact the docs.

Yes, there are indeed still quite some warnings (and errors as well, those will be fixed in https://github.com/shapely/shapely/pull/1497). Most of the warnings seem to be somewhat harmless (based on looking at the output, for example there are a whole bunch of warnings about no stub files), but we should still try to fix those indeed.

PR #1442, opened 2022-07-11
DOC: add changelog for Shapely 2.0

jorisvandenbossche:
Closes #1280

~This is a start for a changelog overview for Shapely 2.0. Many of the section still need to be written out. For now I mostly tried to start a possible outline. And I already copied all relevant parts of the PyGEOS changelog here (and removed that file), but those entries will still need to be rewritten for Shapely / integrated.~
We also still need to decide where/how to integrate this in the existing CHANGES.txt or docs.

PR #1520, opened 2022-09-03
DOC/RLS: update CHANGES.txt with last releases, move old releases to docs

jorisvandenbossche:
This implements the suggestion from https://github.com/shapely/shapely/pull/1442#issuecomment-1223326351, following up on the changelog PR:

- Include the content of CHANGES.txt verbatim into the docs (`docs/release/1.x.rst`), instead of using the include directive
- Remove the old releases (before 1.8) from CHANGEST.txt, and refer to the docs instead (this keeps that file shorter, plus reduces the duplication from the previous point)
- Include the changes from recent 1.8.x releases that were currently only present on the maint-1.8 branch
- For 2.0, only add a summary to CHANGES.txt, referring to the docs for the full changelog. This way, we can at least include the dates of the releases, and also include the changes in consecutive alpha/beta/rc releases that are not relevant for the final release notes in the docs.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,84 @@
 Changes
 =======
 
+2.0a1 (2022-08-03)
+------------------
+
+Shapely version 2.0 alpha 1 is the first of a major release featuring a
+complete refactor of the internals and new vectorized (element-wise) array
+operations providing considerable performance improvements. For a full
+changelog, see https://shapely.readthedocs.io/en/latest/release/2.x.html#version-2-0-0
+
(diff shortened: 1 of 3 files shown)


brendan-ward (commented):
Thanks @jorisvandenbossche !

Overall this is looking great!  However, I'm wondering if we want to limit the subheadings visible when looking at the release notes; right now you get:

image


But perhaps for consistency this should be limited to just a list of the releases?

jorisvandenbossche:
Good idea, this now looks like this:

image

PR #1522, opened 2022-09-05
Raise ValueError for non-finite distance to buffer/offset_curve

mwtoews:
Similar to #1516 for `maint-1.8`, raise ValueError for non-finite distance values passed to buffer or offset_curve (aka parallel_offset) functions.

Previous behavior for shapely-2.0 did not crash (as it did for 1.8), but would be inconsistent:

- distance float('inf') would raise "GEOSException: IllegalArgumentException: CGAlgorithmsDD::orientationIndex encountered NaN/Inf numbers"
- distance float("nan") would not raise or warn, but would return None

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -464,6 +464,8 @@ class BaseGeometry(shapely.Geometry):
 
         if mitre_limit == 0.0:
             raise ValueError("Cannot compute offset from zero-length line segment")
+        elif not np.isfinite(distance):
+            raise ValueError("buffer distance must be finite")
 
         return shapely.buffer(
             self,
--- a/shapely/geometry/linestring.py
+++ b/shapely/geometry/linestring.py
@@ -1,5 +1,6 @@
 """Line strings and related utilities
 """
+import numpy as np
 
 import shapely
 from shapely.geometry.base import BaseGeometry, JOIN_STYLE
@@ -146,6 +147,8 @@ class LineString(BaseGeometry):
         """
         if mitre_limit == 0.0:
             raise ValueError("Cannot compute offset from zero-length line segment")
+        elif not np.isfinite(distance):
+            raise ValueError("offset_curve distance must be finite")
         return shapely.offset_curve(self, distance, quad_segs, join_style, mitre_limit)
 
     def parallel_offset(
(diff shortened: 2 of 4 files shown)


jorisvandenbossche:
> Previous behavior for shapely-2.0 did not crash (as it did for 1.8), but would be inconsistent:

On the other hand, now it is inconsistent with the top-level vectorized functions .. (I think it can certainly be fine to deviate in some cases in the Geometry subclasses, since it can be specialized for the scalar use case, but it should be a conscious decision)

mwtoews:
Oh right, [here it is](https://github.com/shapely/shapely/blob/73ff48c528d466651c590ba856f69e9274467572/shapely/constructive.py#L82-L#L186). A few observations: the second parameter is either `radius` or `width`, while shapely 1.8, GEOS and [JTS](https://locationtech.github.io/jts/javadoc/org/locationtech/jts/operation/buffer/BufferOp.html) use `distance`. I should probably change this in this PR to `distance` for consistency.

Also one of the examples has `buffer(line, float("nan")) is None`, as noted at top. Should vectorized forms continue to return None on error (and be different than from base geometry), or change to raise ValueError? Is now a good time to introduce (e.g.) `on_error="null"` options?

Lastly, forgot to mention upstream https://github.com/libgeos/geos/pull/663 which would change messaging/behavior.

jorisvandenbossche:
> the second parameter is either `radius` or `width`, while shapely 1.8, GEOS and [JTS](https://locationtech.github.io/jts/javadoc/org/locationtech/jts/operation/buffer/BufferOp.html) use `distance`. I should probably change this in this PR to `distance` for consistency.

Ah, good catch, that's another inconsistency I didn't note in https://github.com/shapely/shapely/issues/1276. Using `distance` seems fine for me.

> Should vectorized forms continue to return None on error (and be different than from base geometry), or change to raise ValueError? Is now a good time to introduce (e.g.) on_error="null" options?

My preference is that we keep returning None. In general when adding those to pygeos, we have been following the idea that `None` represents missing values for geometry data, and `NaN` for float data, and that we try to propagate missing values (instead of erroring) in the element-wise funcs. 
(of course, NaN for float data is strictly speaking not a missing value (like "null" or "NA"), but in practice in the python ecosystem often treated that way given the limitation of numpy of not having actual nulls)

But at the same time, I think we could also certainly consider 1) being more strict for scalar geometry class methods, and 2) making the above behaviour configurable with some keyword.

mwtoews:
I think this is ready to go, as it addresses single geometry operations which should behave differently than the vectorized functions with respect to NaN values (used for missing).

Further work with the vectorized functions should follow, as discussed above.

jorisvandenbossche (approved):
Thanks!

Issue #1503, opened 2022-08-25
Add another level to API docs contents?


ivirshup:
I have a feature request for the documentation.

Could another level be added to the table of contents for the API docs? I am having trouble navigating the new docs when I'm not able to easily see a listing of classes or jump to any particular one.

As an example, I would like to see what the various Geometry types are at a glance, but only see this:

![image](https://user-images.githubusercontent.com/8238804/186664025-33e6ea66-653b-46b2-a017-2d248846892e.png)

jorisvandenbossche:
@ivirshup Thanks for opening the issue!

I agree that it would be nice to have an easier overview of those in the left or right sidebar (similarly on the other pages, eg to get a quick overview of all predicates functions on that page: https://shapely.readthedocs.io/en/latest/predicates.html)

However, this is a bit of a limitation of how we use sphinx / autodoc in this case. The content that is being generated by autodoc doesn't create any titles or sections or similar, and the content in the sidebar of the sphinx theme is based on those (it are just some `<dl>` (definition list) html objects, although they have an `id` to link to, they don't make the structure of the document).

jorisvandenbossche:
One alternative I have been thinking about is that we could also create a single page per class or per function. And then we could use `autosummary` to create an easier overview of all functions. 

Similarly to how eg numpy does it here: https://numpy.org/doc/stable/reference/arrays.indexing.html (and then a single function getting a page of its own, for example https://numpy.org/doc/stable/reference/generated/numpy.take.html).

This is actually also similar to the PostGIS function documentation (for example https://postgis.net/docs/ST_Contains.html).

@caspervdw since you added those reference docs initially, what would you think about such an approach?

ivirshup:
If you wanted to get around `autodoc` here, I think you could also create a template which creates headers for each item you'd like to have in the table of contents. Here's [a template](https://github.com/scverse/scvi-tools/blob/master/docs/_templates/autosummary/class.rst) which does this, and [the resulting docs](https://docs.scvi-tools.org/en/stable/api/reference/scvi.model.PEAKVI.html#scvi.model.PEAKVI).

jorisvandenbossche:
Yes, but so as far as I understand, that's a template for autosummary when creating a page per function, and this template is not used by autodoc when bundling them all on a single page. 

So what you mention is exactly what I wanted to suggest above as well to use autosummary with separate pages per function (in numpy/pandas, we are using a similar template).

jorisvandenbossche:
[...]

PR #1529, opened 2022-09-10
DOC: autogenerate separate reference pages per function with autosummary

jorisvandenbossche:
Related to https://github.com/shapely/shapely/issues/1503

Preview of the docs with this branch can be seen at: https://shapely.readthedocs.io/en/reference-docs/index.html

What this PR changes:

- Instead of having all docstrings of a group of functions on the same page (eg the constructive methods: https://shapely.readthedocs.io/en/latest/constructive.html), this now makes one page per function (eg preview for buffer: https://shapely.readthedocs.io/en/reference-docs/reference/shapely.buffer.html), while keeping a table overview of all those functions on the original page (preview for the constructive methods: https://shapely.readthedocs.io/en/reference-docs/constructive.html)
  (this is also similar to how PostGIS has a page per function (eg https://postgis.net/docs/ST_Buffer.html), and similar to for example how numpy and pandas do this, xref https://github.com/shapely/shapely/issues/1503#issuecomment-1237363952)
- While doing the above, I also changed the actual code path (and url) of how a function is documented to be the public one (so eg `shapely.buffer` instead of `shapely.constructive.buffer`)
- I added some code to conf.py to automate this, to keep the benefit of the `.. automodule:: ` we used previously to document all functions in a certain submodule, without having to list them explicitly again in the reference documentation. But to reduce the complexity, I can also certainly just harcode lists of functions in the `.. autosummary::` directives instead of using the jinja templating to list them.

What do people think of this?

A possible next step could also be to think about whether we want to keep the separate pages for constructive, predicates, measurement, etc methods, or rather have one page that lists them all (eg like https://postgis.net/docs/reference.html)

cc @caspervdw


docs/conf.py
```diff
@@ -74,3 +76,57 @@
...
+
+"""
+
+modules = ["_geometry", "creation", "constructive", "coordinates", "io", "linear", "measurement", "predicates", "set_operations"]
+functions = [func for mod in modules for func in get_module_functions(mod)]
+
+template += "   " + "\n   ".join(functions)
+
+with open("_reference.rst", "w") as f:
+    f.write(template)
```
jorisvandenbossche, line 128: The reason for creating this otherwise ignored `_reference.rst` file is because the autogeneration of the reference/ stub files by `autosummary` happens _before_ updates the source with jinja rendering (and so at that point, the `.. autosummary::` directives do not yet contain the methods, and otherwise the individual pages would not be autogenerated)

docs/constructive.rst
```diff
@@ -1,9 +1,11 @@
...
-   :inherited-members:
-   :show-inheritance:
+.. currentmodule:: shapely
+
+.. autosummary::
+   :toctree: reference/
+
+{% for function in get_module_functions("constructive") %}
+   {{ function }}
+{% endfor %}
```
jorisvandenbossche, line 11: Alternatively to this templating, we could also just hardcode this list here, which would look like:

```
.. autosummary::
   :toctree: reference/

   boundary
   buffer
   offset_curve
   centroid
   clip_by_rect
   convex_hull
   delaunay_triangles
   segmentize
   envelope
   extract_unique_points
   build_area
   make_valid
   normalize
   point_on_surface
   polygonize
   polygonize_full
   reverse
   simplify
   snap
   voronoi_polygons
   oriented_envelope
   minimum_rotated_rectangle
   minimum_bounding_circle
```

That's easy to do once. But afterwards, we will have to remember to add an entry here when adding a new function (also not that much of work, so maybe this is not worth the complexity of the templating)
```

File: docs/geometry.rst [hunk 3]
```diff
@@ -88,9 +88,16 @@ This is the same as using :func:`shapely.predicates.equals_exact` with a ``toler
...
-  :exclude-members: GeometryType
-  :special-members:
-  :inherited-members:
-  :show-inheritance:
+.. currentmodule:: shapely
+
+.. autosummary::
+    :toctree: reference/
+
+    Point
```
brendan-ward, line 96: Given the importance of the individual geometry types, it seems like it would be helpful to put them above Construction / Pickling / Hashing.

docs/conf.py
brendan-ward: Suggest adding a brief comment to the code to note this.

brendan-ward (commented):
Thanks @jorisvandenbossche - the results look great!

A few minor comments to consider.

brendan-ward (approved):
Thanks!

jorisvandenbossche:
Any comments on this aspect:

> A possible next step could also be to think about whether we want to keep the separate pages for constructive, predicates, measurement, etc methods, or rather have one page that lists them all (eg like https://postgis.net/docs/reference.html)

brendan-ward:
> A possible next step could also be to think about whether we want to keep the separate pages for constructive, predicates, measurement, etc methods, or rather have one page that lists them all

I don't have strong opinions about this.   I do occasionally get confused about which function falls within a particular grouping, which might be reduced by having them all on the same page.  Presumably you'd keep the categories we have now as headings?

jorisvandenbossche:
> Presumably you'd keep the categories we have now as headings?

Yes, that would be the idea.

(I would also not put all things on a single page. I would certainly keep the Geometry classes separately, and maybe also STRtree and its methods, or the geometry creation methods)

>  I do occasionally get confused about which function falls within a particular grouping, which might be reduced by having them all on the same page.

That's also for me the main reason that I would put them on a single page: it makes it easier to search / find one, without having to know in which groups it lives (in general IMO we should hide as much as possible in which submodule a certain function resides, as this is an internal detail users don't need to know, for that reason this PR also removes the submodule from the function name in the docs).

Now, this can certainly be done as a follow-up. I can already merge this one, and then do a next PR to combine them on a single page, and then we can further discuss there.

Issue #1523, opened 2022-09-05
PERF: Reintroduce a global (or threadlocal) GEOS context


jorisvandenbossche:
_Originally reported by @caspervdw in https://github.com/pygeos/pygeos/issues/192_

In https://github.com/pygeos/pygeos/pull/113, the global GEOS context was eliminated in favor of a context that was initialized and destroyed in every function. This was necessary for multithreading.

However, in https://github.com/pygeos/pygeos/pull/182, @jorisvandenbossche encountered a performance hit coming from this intialization/finalization of the context.

I am thinking especially of the *deallocation* of Geometry objects. It would probably be hard to time this, but I think that destroying a array with lots of geometries has taken a performance hit after #113. 

Maybe the functions that need the GIL anyway can benefit from keeping a global GEOS context around. I would like to explore this, and maybe we can also use a global WKB reader / writer and WKT reader / writer for speeding up single-geometry construction, and `repr` and `str` functions.

jorisvandenbossche:
I suppose deallocation is one of the main cases where a method of the geometry gets called many times individually and where we don't have control to be able to call a vectorized function? (and pickling or hashing an array of geometries being other cases)

But indeed, also not directly an idea how to time this ..

The performance hit we saw for creation of simple points in https://github.com/pygeos/pygeos/pull/182 was a 50% slowdown. I suppose for deallocation it will be in a similar order of magnitude? Given that this is 50% for an operation that is generally cheap (compared to other ufuncs like overlays or buffers etc that create geometries), that is maybe acceptable? 

That said, functions needing the GIL could indeed use a global context as we had before. That doesn't sound as a huge complication to also keep a global one (and in principle the macro GEOS_INIT vs GEOS_INIT_THREADS could take care of assigning the global one or a newly initialized one to the `ctx` variable?)

_Originally posted by @jorisvandenbossche in https://github.com/pygeos/pygeos/issues/192#issuecomment-693610318_

jorisvandenbossche:
This may or may not be helpful, but pyproj allows users to toggle on using the global context at their own risk: https://pyproj4.github.io/pyproj/stable/api/global_context.html.

_Originally posted by @snowman2 in https://github.com/pygeos/pygeos/issues/192#issuecomment-846662160_

jorisvandenbossche:
[...]

Issue #1521, opened 2022-09-03
performance regression and future of shapely.vectorized?


mathause:
## Expected behavior and actual behavior.

I noticed that `shapely.vectorized.contains` is considerably slower (~5x) in shapely 2.0 than 1.8. Now this function may be superfluous in the new version but then maybe it should be deprecated?

(I do plan to use an STRtree but still wanted to bring it up.) 

## Steps to reproduce the problem.

```python
import shapely
import numpy as np
import shapely.geometry
import shapely.vectorized

p = shapely.geometry.box(30, 20, 60, 60)

lon = np.arange(0, 360, 1)
lat = np.arange(90, -91, -1)
LON, LAT = np.meshgrid(lon, lat)

print(shapely.__version__)
%timeit shapely.vectorized.contains(p, LON.flatten(), LAT.flatten())
```

```
1.8.2
4.76 ms ± 49.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
```
2.0a1
25.5 ms ± 891 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

## Operating system

Linux Mint 20.3

## Shapely version and provenance

conda install

jorisvandenbossche:
@mathause thanks for testing shapely 2.0 !

This indeed gives a slowdown. The reason is that, before, those functions were specialized to handle x/y input, while now it's a wrapper around the vectorized predicate functions that expect geometries as input. 
So the slowdown is almost entirely due to creating the temporary Point geometries from the x/y values. When profiling your snippet above, the actual contains predicate evaluation is only 6% of the total time, while all the rest is creating and descroying the points (for shapely 2.0).

When you do already have Points to start with, the top-level `contains` it's actually faster than before (just the predicate evaluation):

```
In [2]: points = shapely.points(LON.flatten(), LAT.flatten())

In [3]: shapely.prepare(p)

In [4]: %timeit shapely.contains(p, points)
2.9 ms ± 90.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

This is of course assuming you already have points to start with. 

For the case of starting with x/y values, the wrapper using the standard predicate based on geometry objects will always be slower. We certainly could in theory keep the specialized versions taking x/y values, but it's a trade-off with the maintenance effort (and assuming the majority use case will be where people already have geometries).

jorisvandenbossche:
[...]

PR #1530, opened 2022-09-10
PERF: reintroduce global context and use in GeometryObject dealloc

jorisvandenbossche:
xref https://github.com/shapely/shapely/issues/1523, https://github.com/shapely/shapely/issues/1521

This creates one global GEOS context object, which can be used in functions that hold the GIL (so we know it will always be used from only a single thread at the same time). And for now using this for the deallocation method for the GeometryObjects. 

This does not destroy this context (`GEOS_finish_r`), which theoretically should happen on module cleanup, which will typically happen at Python interpreter termination (this is certainly possible, for example with https://docs.python.org/3.10/c-api/sys.html#c.Py_AtExit). But we actually also didn't do this originally in pygeos (before moving to a short-lived context in https://github.com/pygeos/pygeos/pull/113), and it's only about little data that gets clean-up anyway if the process stops? (cc @caspervdw)

See the timing at https://github.com/shapely/shapely/issues/1521#issuecomment-1237340989 which illustrates the impact when creating/destroying many objects.


src/geos.c
```diff
@@ -17,6 +20,11 @@ int init_geos(PyObject* m) {
   geos_exception[0] = PyErr_NewException("shapely.errors.GEOSException", base_class, NULL);
   PyModule_AddObject(m, "GEOSException", geos_exception[0]);
   return 0;
+
+  void* context_handle = GEOS_init_r();
+  // GEOSContext_setErrorMessageHandler_r(context_handle, geos_error_handler, last_error);
```
brendan-ward, line 25: Is there a reason not to hook up the error handler to the global context handle?  If so, a comment would be helpful to indicate why.

jorisvandenbossche: Sorry, I should have added an inline comment here on github to ask about that.

So the problem is that this needs a `last_error`, a variable that is otherwise set up in `GEOS_INIT`. Should we have a "global" `last_error` variable as well? But then we also need another helper to check for the error (the current macro is set up to work with GEOS_INIT/GEOS_FINISH). 

Currently this global context is only use in `GeometryObject_dealloc`, where we actually don't do any error handling at the moment. But if we want to use this global context in other places as well (eg repr, hashing), then we will certainly need to set up the error handler.

brendan-ward: Thanks for the explanation.  I think a comment indicating it is intentionally not added right now since global context handle is only used where error handling is not used.  Maybe also a TODO?  Adding more macros can happen as part of a later PR if / when integrating the global context handle into more places.

caspervdw (approved):

PR #1532, opened 2022-09-10
CLN: remove custom minimum_rotated_rectangle implementation + add oriented_envelope alias

jorisvandenbossche:
This PR removes the custom implementation we had for `minimum_rotated_rectangle`. The existing tests for the GEOS-based implementation would pass with this old custom implementation (except for coordinate ordering), so that indicates that replacing this custom implementation with the GEOS one should preserve behaviour.

I also added `oriented_envelope` as a property on the geometry class to be consistent with the top-level namespace.


shapely/tests/geometry/test_geometry_base.py
```diff
@@ -135,3 +135,13 @@ def test_dwithin():
...
+
+
+@pytest.mark.parametrize(
+    "op", ["convex_hull", "envelope", "oriented_envelope", "minimum_rotated_rectangle"]
+)
+def test_constructive_properties(op):
+    geom = LineString([(0, 0), (0, 10), (10, 10)])
+    result = getattr(geom, op)
+    expected = getattr(shapely, op)(geom)
+    assert result == expected
```
jorisvandenbossche, line 147: I added a basic test just to ensure we have coverage for the geometry class properties and that they are equivalent with the top-level function.

tests/test_minimum_rotated_rectangle.py
```diff
@@ -1,30 +0,0 @@
...
-        rect = geometry.Point((0, 1)).minimum_rotated_rectangle
-        assert isinstance(rect, geometry.Point)
-        assert len(rect.coords) == 1
-        assert rect.coords[0] == (0, 1)
-
-        rect = geometry.LineString([(0, 0), (2, 2)]).minimum_rotated_rectangle
-        self.assertIsInstance(rect, geometry.LineString)
-        assert isinstance(rect, geometry.LineString)
-        assert rect.coords[0] == (0, 0)
-        assert rect.coords[1] == (2, 2)
```
jorisvandenbossche, line 30: I removed those tests because they seemed fully superfluous with the tests we now have in `test_constructive.py::test_oriented_envelope`, which also tests for the degenerate case of point/linestring output, and is actually more explicit (tests actual output geometry, instead of only the type and number of coords)

shapely/geometry/base.py
```diff
@@ -307,44 +304,33 @@ def envelope(self):
...
-                # the axes-parallel bounding rectangle.
-                transf_rect = affine_transform(hull, (ux, uy, vx, vy, 0, 0)).envelope
-                # yield the transformed rectangle and a matrix to
-                # transform it back to the original coordinate system.
-                yield (transf_rect, (ux, vx, uy, vy, 0, 0))
-
-        # check for the minimum area rectangle and return it
-        transf_rect, inv_matrix = min(_transformed_rects(), key=lambda r: r[0].area)
-        return affine_transform(transf_rect, inv_matrix)
+        Unlike envelope this rectangle is not constrained to be parallel to the
```
brendan-ward, line 327:
```suggestion
        Unlike `envelope` this rectangle is not constrained to be parallel to the
```

brendan-ward (approved):
Looks good, thanks!

PR #1533, opened 2022-09-11
Move project metadata to setup.cfg

mwtoews:
This PR is for @sebastic to endorse or say it is not needed, following a recent discussion https://github.com/pyproj4/pyproj/discussions/1134#discussioncomment-3592703 . This applies only to shapely-2.0, which has been under development for a few years, but will hopefully reach a final release perhaps this year (although no fixed deadline has been formally discussed). Is there an approximate date when setuptools>=61 will be in debian stable?

While modernizing project metadata, I had picked #1426 instead of #1430 considering recent PyPA developments, but forgetting the constraints for binary packaging (i.e. fixed package versions, no pip/network abilities). It is a bit annoying, as I had thought that `pyproject.toml` was the final spot to put project metadata.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,7 @@ on:
       - "ci/*"
       - "MANIFEST.in"
       - "pyproject.toml"
-      - "setup.py"
+      - "setup.*"
 
 jobs:
   build_sdist:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,56 +2,11 @@
 requires = [
     "cython",
     "oldest-supported-numpy",
-    "setuptools>=61.0.0",
+    "setuptools",
 ]
 build-backend = "setuptools.build_meta"
 
-[project]
-name = "shapely"
-dynamic = ["version"]
-authors = [
-    {name = "Sean Gillies"},
-]
-maintainers = [
-    {name = "Shapely contributors"},
-]
-description = "Manipulation and analysis of geometric objects"
-readme = "README.rst"
-keywords = ["geometry", "topology", "gis"]
-license = {text = "BSD 3-Clause"}
-classifiers = [
-    "Development Status :: 5 - Production/Stable",
-    "Intended Audience :: Developers",
-    "Intended Audience :: Science/Research",
-    "License :: OSI Approved :: BSD License",
-    "Operating System :: Unix",
-    "Operating System :: MacOS",
-    "Operating System :: Microsoft :: Windows",
-    "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.7",
-    "Programming Language :: Python :: 3.8",
-    "Programming Language :: Python :: 3.9",
-    "Programming Language :: Python :: 3.10",
-    "Topic :: Scientific/Engineering :: GIS",
-]
-requires-python = ">=3.7"
-dependencies = [
-    "numpy>=1.14",
-]
-
-[project.optional-dependencies]
-test = ["pytest"]
-docs = ["sphinx", "numpydoc"]
-
-[project.urls]
-Documentation = "https://shapely.readthedocs.io/"
-Repository = "https://github.com/shapely/shapely"
-
-[tool.setuptools.packages.find]
-include = [
-    "shapely",
-    "shapely.*",
-]
+# [project] -- see setup.cfg for project metadata
 
 [tool.coverage.run]
 source = ["shapely"]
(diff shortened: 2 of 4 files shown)


sebastic:
Using `setup.cfg` for the time being until downstreams have setuptools 61+ is the way to go if you ask me.

Because Shapely 2.0 will break dependent projects like Cartopy, it will be uploaded to Debian experimental first until most of the rdeps have releases which support it. The freeze in preparation for the next stable release is scheduled for [February 2023](https://release.debian.org/#release-dates), so not a whole lot of time for rdeps to get their Shapely 2.0 support released.

We can keep python-shapely 2.0 in experimental until setuptools 61+ is in unstable and the Shapely rdeps support 2.0, and move it to unstable sometime during the trixie development cycle (after the bookworm release).

> Is there an approximate date when setuptools>=61 will be in debian stable?

Let's ask the maintainer: @doko42, do you plan get setuptools 61+ into bookworm?

jorisvandenbossche:
> Because Shapely 2.0 will break dependent projects like Cartopy, it will be uploaded to Debian experimental first until most of the rdeps have releases which support it.

I took a quick look at how it breaks Cartopy, and my hope is that it could be fixed (with some small fixes on both sides, https://github.com/SciTools/cartopy/issues/2076#issuecomment-1243294605). 
Do you know of other rdeps that break because of shapely 2.0? (is there a way for me to check which are the rdeps? (I don't see this at https://packages.debian.org/sid/python3-shapely)

@mwtoews I think this PR is certainly fine for the time being.

jorisvandenbossche (approved):

sebastic:
> Do you know of other rdeps that break because of shapely 2.0?

I'm not aware of others at the moment, but the new major version suggests that all rdeps will need changes.

> (is there a way for me to check which are the rdeps? (I don't see this at https://packages.debian.org/sid/python3-shapely)

The rdeps in Debian are included in the transition tracker:

 https://linuxminded.nl/debian/gis-transitions/html/python-shapely.html#good,bad,partial,unknown,!notintesting

jorisvandenbossche:
> the new major version suggests that all rdeps will need changes.

There are actually not that many breaking changes (except for the STRtree interface), _assuming_ the package has been updated for those changes that 1.8 warned about (but yeah, I know that this assumption will not be true in many cases). This are the known breaking changes: https://shapely.readthedocs.io/en/latest/release/2.x.html#breaking-api-changes (although testing cartopy with shapely 2.0 revealed we also broke weakref).

(cartopy was actually already updated to fix all the things shapely 1.8 warned about, but was still breaking with shapely 2.0 because it uses some internal shapely APIs)

> The rdeps in Debian are included in the transition tracker:
> 
> https://linuxminded.nl/debian/gis-transitions/html/python-shapely.html#good,bad,partial,unknown,!notintesting

Thanks, that's useful! Checking those will be interesting to see if we have more (unknown) breaking changes.

jorisvandenbossche:
@mwtoews I think this can be undrafted?

mwtoews:
My only hesitation is the timing of shapely 2.0's release (or release candidate) and if setuptools 61+ gets into debian bookworm. This is because I'd rather not unnecessarily move this metadata more than a few times.

sebastic:
setuptools 65.3.0 is now in unstable, but is prevented from migrating to testing (bookworm) due to breaking rdeps:

 https://tracker.debian.org/pkg/setuptools

EwoutH:
Any update on this effort? It looks like it blocks #1564, so I would like to know if we can move this forward!

mwtoews:
Hi @EwoutH it seems that last week [setuptools in debian testing was upgraded](https://tracker.debian.org/news/1374817/setuptools-6530-11-migrated-to-testing/), so this PR may not be needed. However, testing with a docker image, I see it does not work without possible modifications to `debian/rules`. Either way, we'll make a call before the first release candidate two weeks from now, on November 14.

jorisvandenbossche:
@sebastic FYI, shapely 2.0 no longer depends on descartes, so that could be removed as a dependency from https://packages.debian.org/source/experimental/python-shapely

sebastic:
`requirements-dev.txt` still includes `descartes`:
```
$ grep descartes requirements-dev.txt 
descartes==1.0.1
```

jorisvandenbossche:
Yes, that was just fixed -> https://github.com/shapely/shapely/pull/1591

mwtoews:
Closing this PR as it's fortunately not needed! It was a close one too!

Issue #1443, opened 2022-07-11
Expose a public C API


caspervdw:
Following https://github.com/shapely/shapely/issues/782 shapely 2.0 will have an (internal) C API. This C API is already used in a few cython functions.

If we would make this C API public then users may be able to iterate over GEOSGeometry objects without the overhead of the Python interpreter. Maybe we could even allow third party libraries linking against pygeos.
There are already some instances of custom Python methods that would benefit from this. For instance subdivide (https://github.com/pygeos/pygeos/pull/256), my own line_substring (https://github.com/nens/threedigrid-builder/blob/bebc3fda5ed873af6505994925b59b03dc77ad6d/threedigrid_builder/grid/linear.py#L511). Also @dhodcz2 reached out with questions about using the C API in one of his projects.

I think adding this would be a good thing. We would have to be clear on several points:

- ABI compatibility: public C functions cannot change their call signature across versions. We might add a “void *” argument to all functions to have some space for additional options.
- Reference counting
- Thread safety (are you allowed to use the method without a GIL?)
- GEOS compatibility: this is a tricky one. For GEOSGeometry objects to be usable, one needs GEOS functions. The GEOS functions used by a dependent library should come from the exact same GEOS binary that shapely is using. However when using binary wheels, it is easy to link a dependent library against a different GEOS binary. On conda you are probably fine. I think we should recommend users to use conda here, or compile pygeos themselves.

@sgillies @jorisvandenbossche @brendan-ward @mwtoews

jorisvandenbossche:
I think the last point (GEOS compatibility) will be the biggest hurdle for making this usable in practice for general usage (I don't see an easy way you could do this with wheels). Of course that doesn't necessarily mean that we can't provide such an API, since it can still be useful to prototype or for smaller / internal projects (where you might care less about portability for a broad user base).

Practically speaking, is this mostly about making the content of [`_pygeos_api.pxd`](https://github.com/shapely/shapely/blob/1a97fc7890b1440b018e3b164f2618e9de1532dd/shapely/_pygeos_api.pxd) public? (+ `get_geos_handle` which is currently exposed in `_geos.pxd`) 
We could for example put that in an `__init__.pxd` ?

jjyyxx:
From #962, I realized that `_geom` is still usable and my code could work as is. But I found working with `char get_geom(GeometryObject* obj, GEOSGeometry** out)` (if possible) may be more concise and robust.

sgillies:
[...]

PR #1534, opened 2022-09-12
COMPAT: keep (deprecated) geom_factory for downstream packages (cartopy)

jorisvandenbossche:
For context, see https://github.com/SciTools/cartopy/issues/2076#issuecomment-1243294605. Cartopy is using `geom_factory` in their cython code to create a shapely geometry from a GEOS geometry object pointer they create in cython. This PR adds back a (deprecated) `geom_factory` function that did this before in <=1.8. 
Of course this is not really safe to use (it requires that you are using the exact same GEOS version and build), but this caveat already applies to cartopy+shapely now as well, so this just keeps that situation a bit longer. Longer term, cartopy will ideally stop passing pointers to shapely, for example by using shapely more instead of interfacing directly with GEOS (cfr https://github.com/SciTools/cartopy/issues/805)

Also related to https://github.com/shapely/shapely/issues/1443

### DIFF
--- a/shapely/_geometry_helpers.pyx
+++ b/shapely/_geometry_helpers.pyx
@@ -3,6 +3,7 @@
 cimport cython
 from cpython cimport PyObject
 from cython cimport view
+from libc.stdint cimport uintptr_t
 
 import numpy as np
 
@@ -384,3 +385,11 @@ def collections_1d(object geometries, object indices, int geometry_type = 7, obj
             geom_idx_1 += collection_size[coll_idx]
 
     return out
+
+
+def _geom_factory(uintptr_t g):
+
+    with get_geos_handle() as geos_handle:
+        geom = PyGEOS_CreateGeometry(<GEOSGeometry *>g, geos_handle)
+
+    return geom
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -12,6 +12,7 @@ from warnings import warn
 import numpy as np
 
 import shapely
+from shapely._geometry_helpers import _geom_factory
 from shapely.affinity import affine_transform
 from shapely.coords import CoordinateSequence
 from shapely.errors import GeometryTypeError, GEOSException, ShapelyDeprecationWarning
@@ -26,6 +27,16 @@ GEOMETRY_TYPES = [
     "MultiPolygon",
     "GeometryCollection",
 ]
+
+
+def geom_factory(g, parent=None):
+    warn(
+        "The 'geom_factory' function is deprecated in Shapely 2.0, and will be "
+        "removed in a future version",
+        DeprecationWarning,
+        stacklevel=2,
+    )
+    return _geom_factory(g)
 
 
 def dump_coords(geom):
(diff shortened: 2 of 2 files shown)


jorisvandenbossche:
For context, see https://github.com/SciTools/cartopy/issues/2076#issuecomment-1243294605. Cartopy is using `geom_factory` in their cython code to create a shapely geometry from a GEOS geometry object pointer they create in cython. This PR adds back a (deprecated) `geom_factory` function that did this before in <=1.8. 
Of course this is not really safe to use (it requires that you are using the exact same GEOS version and build), but this caveat already applies to cartopy+shapely now as well, so this just keeps that situation a bit longer. Longer term, cartopy will ideally stop passing pointers to shapely, for example by using shapely more instead of interfacing directly with GEOS (cfr https://github.com/SciTools/cartopy/issues/805)

Also related to https://github.com/shapely/shapely/issues/1443

shapely/_geometry_helpers.pyx
```diff
@@ -384,3 +385,11 @@ def collections_1d(object geometries, object indices, int geometry_type = 7, obj
             geom_idx_1 += collection_size[coll_idx]
 
     return out
+
+
+def _geom_factory(uintptr_t g):
+
+    with get_geos_handle() as geos_handle:
+        geom = PyGEOS_CreateGeometry(<GEOSGeometry *>g, geos_handle)
```
brendan-ward, line 393: We probably should rename the `PyGEOS_*` Cython helpers to `Shapely_*` (another PR).

shapely/geometry/base.py
```diff
@@ -28,6 +29,16 @@
 ]
 
 
+def geom_factory(g, parent=None):
+    warn(
```
brendan-ward, line 33: `parent` is unused?  Can it be removed, or does it need to be retained for backward compatibility?

Should this also include a brief docstring, something like

```Python
"""Creates a Shapely geometry instance from a pointer to a GEOS geometry.

WARNING: the GEOS library used to create the the GEOS geometry pointer and the GEOS library used by Shapely must be exactly the same, or unexpected results or segfaults may occur.

Deprecated in Shapely 2.0

"""
```

jorisvandenbossche: Yes, I only added it back for backwards compatibility. Apart from keeping a reference to that parent object, it didn't really do anything (https://github.com/shapely/shapely/blob/b6aa26a0bcd9cb1f0767d5d80df0fe32eacd5b31/shapely/geometry/base.py#L70-L91)

jorisvandenbossche: > Should this also include a brief docstring, something like

Thanks! That's a good addition

brendan-ward (approved):
Thanks!

shapely/_geometry_helpers.pyx
jorisvandenbossche: > We probably should rename the `PyGEOS_*` Cython helpers to `Shapely_*` (another PR).

For some that might indeed make sense, for others where it's a Python C wrapper ("Py") around something GEOS related, it could also makes sense to keep it (but didn't check the different cases in detail)

PR #1535, opened 2022-09-12
Make Geometry objects weakref-able

jorisvandenbossche:
For context, see https://github.com/SciTools/cartopy/issues/2076#issuecomment-1243294605. 

While trying to make cartopy run with Shapely 2.0, I discovered that they relied on geometry objects being weakref-able. And in Shapely 2.0 we broke that (first by making it an extension type in C (in pygeos), and then afterwards by adding `__slots__` to the python subclasses, https://github.com/shapely/shapely/pull/1181)

This PR makes the geometry objects again weakref-able. 

There is the question whether we want to do this, with the main tradeoff being that this adds 8 bytes of memory to each geometry object (although, given that each coordinate is also 8 bytes, this is only somewhat significant for Point objects)

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -63,7 +63,7 @@ class BaseGeometry(shapely.Geometry):
 
     """
 
-    __slots__ = []
+    __slots__ = ["__weakref__"]
 
     def __new__(self):
         warn(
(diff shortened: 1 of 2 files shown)


caspervdw (approved):
I never worked with weakref before, but if this fixes cartopy then I think we need to do this.

jorisvandenbossche:
Yes, it fixes current cartopy, but it also seems they were probably using weakref only because the geometries were not hashable before (and couldn't be used as dict keys)

jorisvandenbossche:
Let's merge this for compatibility with shapely<2

PR #1536, opened 2022-09-12
Add unary_union alias for top-level union_all

jorisvandenbossche:
xref https://github.com/shapely/shapely/issues/1276

Since we already have `unary_union`, and this is also used in GeoPandas (and PostGIS also uses this), we thought it best to keep that name. But also keeping `union_all` because this is consistent with `intersection_all` etc (and we don't want to rename the `_all` functions to `unary_`, since that is also not a super clear name if you are not familiar with it)


brendan-ward (approved):
Thanks!

PR #1537, opened 2022-09-12
DOC: update offset_curve (parallel_offset) documentation regarding direction of the resulting line

jorisvandenbossche:
Closes https://github.com/shapely/shapely/issues/1436


shapely/geometry/linestring.py
```diff
@@ -120,11 +120,10 @@ def offset_curve(
         """Returns a LineString or MultiLineString geometry at a distance from
         the object on its right or its left side.
 
-        The side parameter may be 'left' or 'right' (default is 'right'). The
-        resolution of the buffer around each vertex of the object increases by
-        increasing the resolution keyword parameter or third positional
-        parameter. Vertices of right hand offset lines will be ordered in
-        reverse.
+        The side is determined by the sign of the `distance` parameter
+        (negative for right side offset, positive for left side offset). The
```
brendan-ward, line 124: Should this also include the note re: GEOS version?

jorisvandenbossche: Added (because this was not mentioning anything about the orientation yet, I was thinking to not add it, but easy to just be consistent and have the same note everywhere)

PR #1540, opened 2022-09-18
[Done] Consistently skip (or don't) skip None in YY_Y functions

caspervdw:
Trial implementation following up on #1332 
Ref #1501 

One thing I noted is that you can't simply use `intersection.reduce(..., initial=GeometryCollection([]))`. That would lead to `intersection_all` always returning an empty geometry. Instead you should use the ufuncs 'identity' (see https://numpy.org/doc/stable/reference/c-api/ufunc.html#c.PyUFunc_FromFuncAndDataAndSignatureAndIdentity) to define behaviour on reductions on empty lists.

However (as also discussed in #1501) a reduction on a length-1 axis directly returns the value in that axis. Our registered ufunc isn't even called. So we have to do some post-correction anyway. Which is also done by numpy: https://github.com/numpy/numpy/blob/54c52f13713f3d21795926ca4dbb27e16fada171/numpy/lib/nanfunctions.py#L344. This takes away the reason for bothering with the custom Identity.


src/ufuncs.c
```diff
@@ -826,12 +848,16 @@ static void YY_Y_func_reduce(char** args, npy_intp* dimensions, npy_intp* steps,
...
+                      void* data) {
+  FuncGEOS_YY_Y* func = ((FuncGEOS_YY_Y**)data)[0];
+  int skip_na = (int)((int**)data)[1];
+
   // A reduce is characterized by multiple iterations (dimension[0] > 1) that
   // are output on the same place in memory (steps[2] == 0).
   if ((steps[2] == 0) && (dimensions[0] > 1)) {
     if (args[0] == args[2]) {
-      YY_Y_func_reduce(args, dimensions, steps, data);
+      YY_Y_func_reduce(args, dimensions, steps, func, skip_na);
```
jorisvandenbossche, line 860: Would it be an option to stop trying to "detect" whether we are in the reduce context and then call the reduce version, and instead just register them as two different ufuncs, one that is called for normal binary versions, and one that is only called for the reducing `_all` versions? 
(then the standard ufunc also doesn't need to care about skipna)

caspervdw: I think that would simplify things indeed. We do have to carry out this check though, in both versions. I'll give it a try.

caspervdw: @jorisvandenbossche  Although it was a logical suggestion, I now realize this can't work if the reduction axis is not contiguous in memory.

I implemented `intersection_reduce`, which directly ties into the "reduction" implementation. Then, when testing I got this:

```
> shapely.lib.intersection_reduce.reduce([[None] * 2] * n, axis=1)  # OK
> shapely.lib.intersection_reduce.reduce([[None] * 2] * n, axis=0)  # error
NotImplementedError: Unknown ufunc mode with args=[0x55d240df39c0, 0x55d240743c00, 0x55d240df39c0], steps=[8, 8, 8], dimensions=[2].
```

So it seems that the "reduction" version occasionally will need the normal impelmentation.

jorisvandenbossche: > I now realize this can't work if the reduction axis is not contiguous in memory.

That isn't covered by a different stride that is passed? (in general our ufuncs can handle input that are not strictly contiguous?)

> So it seems that the "reduction" version occasionally will need the normal impelmentation.

Yeah, from my previous testing on Brendan's PR, it was indeed clear that sometimes the reduce version doesn't always get called (at least based on our stride / dimension checks), and I couldn't directly understand the logic of when you got which version. 

Now, I had another idea: we could also directly implement the "reduce" logic in the ufunc, instead of implementing a "normal" ufunc that we call with `.reduce`. 
And we actually already do this (so that would keep the implementation similar to existing patterns), for example for unary_union/union_all, or for generalized ufuncs such as `polygonize`. 

I quickly experimented with both just to get an idea:

* Option to mimic `unary_union`, by also collecting the geometries (over the correct axis) into collections, and then implementing a reducing intersection function as a simple Y_Y func: https://github.com/jorisvandenbossche/Shapely/commit/4903951e1ac2f7fad73c616a44b8e0cddf1d2666
* Option to use generalized ufunc with a reducing axis (like `polygonize`): https://github.com/jorisvandenbossche/Shapely/commit/e144885b20a1a9c65326200a7f6f8bd3320dec50

Code can use some clean-up for both, but both do pass our tests (except for returning empty geometry instead of None), and from a quick test give the same performance as the existing implementation (but with just a 1D array). I suppose the time is dominated by the actual intersection.

Sorry to keep tinkering on this one. I am just a bit uneasy with the complexity of the current "reduce" version (the fact that we try to detect when this is the case while I can't fully wrap my head around what numpy is doing, that it complicates the normal ufunc as well, the reduce version is non-trivial (given we had segfaults until recently), ...). In contrast one of both options from above might potentially be simpler to follow.

caspervdw: Thanks for joining the effort in getting this right. I agree with  you on the point that the current reduce is too complex. I actually the second implementation you did (https://github.com/jorisvandenbossche/Shapely/commit/e144885b20a1a9c65326200a7f6f8bd3320dec50) over all the options for that passed until now.

In that case we should still raise in the normal implementation if steps[0] == 0, else you get in trouble with memory management.

I do think that having both the “skip_na=False” and “skip_na=True” would be a good thing to support. 

Can you make a PR out of your generic ufunc implementation?

jorisvandenbossche: Opened https://github.com/shapely/shapely/pull/1562

> In that case we should still raise in the normal implementation if steps[0] == 0, else you get in trouble with memory management.

Yes, I think that is what our standard macro `CHECK_NO_INPLACE_OUTPUT` is doing?

caspervdw:
Superseded by #1562

Issue #1399, opened 2022-05-22
Feature: Remove repeated points


mwtoews:
Upcoming GEOS 3.11 (currently in main branch) has a new feature [`GEOSRemoveRepeatedPoints`](https://github.com/libgeos/geos/blob/2a82afc1ac2327966f56bdf35a6d88aaadb81330/capi/geos_c.h.in#L2795-L2812) (https://github.com/libgeos/geos/issues/599):
> Works from start of each coordinate sequence in the geometry, retaining points that are further away from the previous retained point than the tolerance value.
> Removing repeated points with a non-zero tolerance may result in an invalid geometry being returned. Be sure to check and repair validity.

PR #1546, opened 2022-09-22
ENH: Remove Repeated Points

akadouri:
Closes https://github.com/shapely/shapely/issues/1399

Add support for [GEOSRemoveRepeatedPoints](https://github.com/libgeos/geos/blob/2a82afc1ac2327966f56bdf35a6d88aaadb81330/capi/geos_c.h.in#L2795-L2812), introduced in GEOS 3.11.0


shapely/constructive.py
```diff
@@ -662,6 +663,30 @@ def polygonize_full(geometries, **kwargs):
...
+    """Returns a copy of a Geometry with repeated points removed.
+
+    From the start of the coordinate sequence, each next point within the tolerance is removed.
+
+    Removing repeated points with a non-zero tolerance may result in an invalid geometry being returned.
+
+    Parameters
+    ----------
+    geometry : Geometry or array_like
+    tolerance : float or array_like
```
mwtoews, line 677: Consider adding `tolerance` description from [here](https://github.com/libgeos/geos/blob/8bb8356694a57bd31a19e13d0c9ed77df7b5005e/capi/geos_c.h.in#L2868).

mwtoews (commented):
Thanks, looks good!

I think tolerance should have a default 0.0, which is the simplest forms of this function. PostGIS has a default tolerance 0.0 too.

akadouri:
@mwtoews good point, updated to include the default tolerance.

shapely/constructive.py
```diff
@@ -662,6 +663,33 @@ def polygonize_full(geometries, **kwargs):
...
+    Parameters
+    ----------
+    geometry : Geometry or array_like
+    tolerance : float or array_like, default=0.0
+        Use 0.0 to remove only exactly repeated points.
+
+    Examples
+    ----------
+    >>> from shapely import LineString, Polygon
+    >>> remove_repeated_points(LineString([(0,0), (0,0), [1,0]]), tolerance=0)
```
jorisvandenbossche, line 685:
```suggestion
    >>> remove_repeated_points(LineString([(0,0), (0,0), (1,0)]), tolerance=0)
```

shapely/constructive.py
```diff
@@ -662,6 +663,33 @@ def polygonize_full(geometries, **kwargs):
     return lib.polygonize_full(geometries, **kwargs)
 
 
+@requires_geos("3.11.0")
```
jorisvandenbossche, line 666:
```suggestion
@requires_geos("3.11.0")
@multithreading_enabled
```

(for functions that release the GIL in ufuncs.c (which is the case here, since it uses `GEOS_INIT_THREADS` and not `GEOS_INIT`), we also have this decorator on the Python side to ensure the arrays are protected from being written to)

jorisvandenbossche (commented):
Thanks for the PR! Looks good, just a few minor comments

Can you also add this to the list of new GEOS exposed functions in docs/release/2.x.rst? (https://shapely.readthedocs.io/en/latest/release/2.x.html#bindings-for-new-geos-functionalities)

shapely/constructive.py
akadouri: Are you referring to this line https://github.com/shapely/shapely/blob/main/src/ufuncs.c#L636? Is `Y_func` the only type that uses GEOS_INIT (and should not have `@multithreading_enabled` on it?) which looks like only `prepare`/`destroy_prepared`.

jorisvandenbossche: > Are you referring to this line

Yes

> Is `Y_func` the only type that uses GEOS_INIT (and should not have `@multithreading_enabled` on it?) which looks like only `prepare`/`destroy_prepared`.

No, if you search for `GEOS_INIT;` in ufuncs.c, you will see a few others using it as well (eg to/from_geojson and wkt, relate, is_valid_reason, all because those interact with Python string objects). `Y_func` actually seems like a case where we could actually use `GEOS_INIT_THREADS`

jorisvandenbossche (approved):
Thanks!

PR #1549, opened 2022-09-24
DOC/CI: fix doc build for latex + re-enable epub and htmlzip

jorisvandenbossche:
After https://github.com/shapely/shapely/pull/1529, the doc build on readthedocs was failing for latex (pdf). Although if this rendering can't be done for pdf, then I am not sure that the approach of https://github.com/shapely/shapely/pull/1529 will actually result in proper pdf docs .. (I am also not sure how important this is)

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -16,8 +16,7 @@ sphinx:
   fail_on_warning: false
 
 # Optionally build your docs in additional formats such as PDF and ePub
-formats:
-  - pdf
+formats: all
 
 conda:
   environment: docs/environment.yml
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
This fixed the doc build: https://readthedocs.org/projects/shapely/builds/18158708/

And it seems that the PDF is actually fine, it includes all the reference docs as well: https://shapely.readthedocs.io/_/downloads/en/fix-pdf-docs/pdf/ (so not really an idea how this works, since the templating doesn't seem to be called when running the latex docs; unless that is build after html, and uses the cached environment from that? Anyway, it works ;))

PR #1554, opened 2022-09-27
CI: add PyPI token for travis wheel uploads (splitted secret)

jorisvandenbossche:
The linux aarch64 wheels were not yet uploaded for the previous alpha (see https://github.com/shapely/shapely/issues/962#issuecomment-1203723128), because the PyPI authentication wasn't yet set up in travis.yml

However I ran into the issue that the token is "too large" for travis to encrypt (apparently something for older repos ..), see https://travis-ci.community/t/travis-encrypt-data-too-large-for-pypi-tokens-with-older-repos/5792/10 or https://github.com/travis-ci/travis.rb/issues/687 for context.

I followed a workaround mentioned in those threads (and similarly as done in eg https://github.com/rafguns/linkpred/commit/f0a4ebdfa4532f519e5e5c74d17b98d12e58cdc3) to split the secret. No idea if I did it correctly, I suppose we will only see with the next time we try to upload .. 

Basically what I did was creating a token on PyPI to use for uploading from Travis, and then encrypting this using Travis' cli tool, but cut the token into three pieces: `travis encrypt PYPI_TOKEN1=<part of the token> --com`

(I also sent Travis a support request to fix this)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ if: (branch = main OR tag IS present) AND (type = push)
 env:
   global:
   - GEOS_VERSION=3.11.0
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Thanks to Travis support, it seems you can regenerate the key pair yourself through the API (https://developer.travis-ci.com/resource/key_pair_generated#Key%20pair%20(generated)), so did that, and now we have a longer key and encrypting the token works without having to cut it into pieces.

PR #1556, opened 2022-10-03
ENH: add `__format__` specification for geometry types

mwtoews:
This enhancement allows a format string to be passed to Python's [`format()`](https://docs.python.org/3/library/functions.html#format) built-in function to modify the output format of a geometry using specific options from GEOS' WKT and WKB writers.

For example:
```python
from shapely import Point

pt = Point(1.23456789, -9.87654321)

# use fixed-precision to three decimal places
print(f"Treasure close to {pt:.3f}")
# Treasure close to POINT (1.235 -9.877)

print(f"or in hex-encoded WKB: {pt:x}")
# or in hex-encoded WKB: 01010000001bde8342cac0f33f33f68845cac023c0
```

From the WKT writer features, I've only exposed rounding precision and trim, as these fit [Pythons format specification mini-language](https://docs.python.org/3/library/string.html#formatspec). The hex-encoded WKB writer is enabled with "x" or "X" format codes.

As for documentation, I'm not sure where else it should go. Currently, it's hidden in (e.g.) `help(Point.__format__)`. Is there a better way to expose or move it?


mwtoews:
Clearly my testing was done with an older GEOS before Ryu integration. With the older GEOS, the "g" format code paired nicely with setting GEOS' `trim=True`. However, now `trim=True` it has completely different behaviour, where not even scientific representation is possible (it is always fixed-point notation). I'm not sure how well this feature integrates with newer GEOS versions, except to remove "g" and "G" format codes.

jorisvandenbossche:
> As for documentation, I'm not sure where else it should go. Currently, it's hidden in (e.g.) `help(Point.__format__)`. Is there a better way to expose or move it?

Not that it makes it much more discoverable, but maybe you can add a "Formatting" section on https://shapely.readthedocs.io/en/latest/geometry.html?

mwtoews:
Thanks @jorisvandenbossche that's a good spot to expand the docs.

Also, I've decided to keep the `g` option, which toggles the trim option.

jorisvandenbossche (approved):
Looks good to me!

PR #1557, opened 2022-10-03
DOC: add migration guide for PyGEOS users

### LINKED ISSUE #1500: Add a pygeos migration guide for Shapely 2.0

jorisvandenbossche:
Closes #1500

@caspervdw @brendan-ward a review especially of the STRtree changes would be welcome

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -12,6 +12,7 @@
    installation
    User Manual <manual>
    migration
+   migration_pygeos
    release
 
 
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
@brendan-ward going to merge this, but if you can still take a look alter (as avid pygeos user), that would certainly be welcome!

PR #1559, opened 2022-10-03
ENH: convert to / construct from ragged array (flat coordinates and offset arrays)

### LINKED ISSUE #1498: Expose method to get / construct from coordinates and offset arrays
In the GeoArrow initiative, we defined a way how to store geometries in a Arrow compatible memory format (using a more "native" storage of the data (using the nested types available in Arrow) compared to using binary WKB): https://github.com/geopandas/geo-arrow-spec/blob/main/format.md (https://github.com/geopandas/geo-arrow-spec/issues/4)
Essentially, this uses a "nested list" logical type (like GeoJSON), or "ragged array", which physically means a single flat array of coordinates with zero or more additional offset arrays to keep track of where each new ring, new geometry or new geometry part starts in the coordinate array. 

This was discussed before in pygeos as well, see:

* [ENH: get coordinates + offset arrays for "ragged array" representation - pygeos/pygeos#75](https://github.com/pygeos/pygeos/issues/75)
* [Add cython algo to get offset arrays for flat coordinates representation - pygeos/pygeos#93 ](https://github.com/pygeos/pygeos/pull/93)
* [Add “offsets” as an alternative to “indices” - pygeos/pygeos#438](https://github.com/pygeos/pygeos/issues/438)

jorisvandenbossche:
Fixes #1498

This still needs some more docs and tests.

Some general comments:

* Input welcome on **naming**!
  * I currently called this `to_coordinates_offsets` and `from_coordinates_offsets` (I used `to_` because there is a matching `from_`, although otherwise a `get_..` would also be consistent with the other `get_coordinates`/`get_parts`/.. functions we already have)
  * Some other ideas that have floated before: `to/from_flatcoords`. In geoarrow I used ``flat_coords_offset_arrays``
* **Performance**: 
  * Currently this is not an optimal implemenation (`from_coordinates_offsets` is about the same speed as `from_wkb` for a largish test case with (multi)polygons). This is mostly because of doing this with our existing APIs in multiple passes (for example in the `from_` method, we first create linearrings, which we then clone in to create polygons, and clone those again to create multipolygons. This cloning takes a significant amount of time, and is not needed in this case)
  * We could do better with a custom implementation (eg https://github.com/pygeos/pygeos/pull/93 in cython), but I would first like to focus on the exact API to implement and the correctness, so that we could include this in a Shapely release faster. And then later for subsequent releases we can still improve the performance of the function.

Additionally, there are some API questions:

* For 3D geometries, should we automatically include z in the coordinates? `get_coordinates` has a `include_z=False` keyword and so doesn't "infer" this itself
* For roundtripping mixed single and multi geometries, would we like to have an option in the `from_` function to return Multi geometries with 1 part as the singular version? (that doesn't guarantee a faithful rountrip, since you can have eg MultiPolygons with a single sub-polygon in your input, but I do think it would cover the large majority of use cases where someone would prefer to preserve the mixture). 
* Currently, this function returns the coordinates as a single, interleaved (xyxy) array (which matches our `get_coordinates` function). There is still some discussion about whether we should rather prefer separate arrays for x and y (xx and yy) (https://github.com/geopandas/geo-arrow-spec/pull/26/), but I think we can easily add an `interleaved=True/False` keyword (or other name) later to support both.


thomcom:
I don't think you need to include `z` at this time. GeoArrow spec iirc says that we use interleaved xy (which may be on its way out) and that if z is present it'll be in its own ListArray. 

Again referencing GeoArrow, a Multi geometry with only a single member is still a multi geometry. I'm not familiar with a use case where users might like to implicitly convert Multis to singles.

kylebarron:
> GeoArrow spec iirc says that we use interleaved xy (which may be on its way out) and that if z is present it'll be in its own ListArray.

Does it? Interleaved xy and separate z would be an interesting approach (and would align geoarrow with FlatGeobuf), but my read of the [current spec page](https://github.com/geoarrow/geoarrow/blob/f2eb98e7e2f60ed409ed88078015a9c45f469604/format.md) is that xyz would be in a `FixedSizeList` column of width 3. 

> **Point**: `FixedSizeList<double>[n_dim]`
> 
> For an array of Point geometries, only a single array is used of interleaved
> coordinates. `n_dim` can be 2, 3, or 4 depending on the dimensionality of the
> geometries, and the field name of the list should be "xy", "xyz" or "xyzm",
> reflecting the dimensionality.

thomcom:
Looks like your reading is the correct one! We had talked about leaving z in its own column and I assumed that's what was decided.

jorisvandenbossche:
Yes, the current spec specifies all interleaved, and if we allow structs as well (as we are currently planning), it would be all three separate fields (although I raised the option of separate xy and z in https://github.com/geoarrow/geoarrow/pull/26#issuecomment-1251535340, but that wasn't further discussed). 

But for the current PR, I am always returning a single coordinates array (we could provide an option later to not have interleaved, or to return it as Fortran ordered array, which would basically allow seeing it as separate arrays as well). And so if returning 3D data, it would simply be a (n, 3) array instead of (n, 2) array. 
My main question is about what API we want here: I could detect from the first geometry in the array whether it is 2D or 3D, and then automatically return xy or xyz. Alternatively, this could require an argument specified by the user explicitly (this is what we do in `get_coordinates`, which by default always returns only xy, but has a `include_z=True` option to return xyz)

jorisvandenbossche:
Another thought about the API: for the `to_..` function, currently I always return a tuple of 2 elements: the coordinates array + a variable length tuple of offset arrays (or empty tuple in case of points) -> eg `return coords, (offset1, offset2)`. 
But alternatively, I could also return this as a flat (variable-length) tuple, where you know the first element is always the coordinates array, and then the remainder of the tuple the offset arrays -> eg `return coords, offset1, offset2`

caspervdw:
I intuitively prefer the first proposal API (`return coords, (offset1, offset2)`)

It also fits better into PEP484: `Tuple[ndarray[float], Tuple[ndarray[int], ...]]`

brendan-ward:
`to/from_ragged_array` seems like a reasonable naming convention here.

jorisvandenbossche:
cc @ianthomas23 this will probably also be interesting for spatialpandas (conversion between geopandas/shapely geometries and  spatialpandas/awkward array like representation, which currently exist in spatialpandas but this should optimize that conversion)

ianthomas23:
@jorisvandenbossche Thanks for the heads-up here. I'll need to look into the details and see what overlap there is with what we have been considering.

jorisvandenbossche:
I updated the PR: updated the name to `to_ragged_array`/`from_ragged_array`, fixed some bugs in handling empty geometries, and also added some explanation and expanded docstring.

There is one more API question that I bumped into (while trying to add optional support for returning 3D coordinates with an `include_z` like in `get_coordinates`). 
Currently, I was always returning a flattened (`np.ravel()`) 1D array of coordinates, instead of the original 2D array of coordinates you get from `shapely.get_coordinates` (i.e. `coords = shapely.get_coordinates(arr).ravel()`), because this is what the list array in Arrow under the hood is. But for shapely's point of view, it's probably more logical to leave out the `ravel()` (it's still easy and cheap for the user to do this if they want to have the flattened array). Because then you could for example also pass those coordinates directly to eg `shapely.points(..)` without having to reshape them (`coords.reshape((-1, 2))`). 

Doing this would also help when the function can optionally return z coordinates, because then you can see this from the coordinates array shape (`(n, 2)` vs `(n, 3)`), instead of having to keep track of this manually what you passed for `include_z`.

jorisvandenbossche:
I think this is ready, and unless there are major comments, am planning to merge this in one of the next days (to have it included in 2.0)

shapely/_flatcoords.py
```diff
@@ -0,0 +1,426 @@
...
+           [ 2.,  3.],
+           [ 2.,  2.]])
+
+    >>> offsets
+    (array([0, 5, 9]), array([0, 2]))
+
+    As an example how to interpret the offsets: the i-th ring in the
+    coordinates is represented by ``offsets[0][i]`` to ``offsets[0][i+1]``:
+
+    >>> ring1_start, ring1_end = offsets[0][0], offsets[0][1]
```
brendan-ward, line 220:
```suggestion
    >>> exterior_ring_start, exterior_ring_end = offsets[0][0], offsets[0][1]
```

shapely/_flatcoords.py
```diff
@@ -0,0 +1,426 @@
...
+        The type of the input geometries (required information for rountrip).
+    coords : np.ndarray
+        Contiguous array of shape (n, 2) or (n, 3) of all coordinates of
+        all input geometries.
+    offsets: tuple of np.ndarray
+        Offset arrays that allow to reconstruct the geometries based on the
+        flat coordinates array. The number of offset arrays depends on the
+        geometry type. See
+        https://github.com/geoarrow/geoarrow/blob/main/format.md for details.
+
```
brendan-ward, line 169: Suggestion, add:

```Python
Mixed singular and multi geometry types of the same basic type are allowed
(e.g., Point and MultiPoint) and all singular types will be treated as multi types.
GeometryCollections and other mixed geometry types are not supported.
```

shapely/_flatcoords.py
```diff
@@ -0,0 +1,426 @@
...
+    result = np.empty(len(offsets3) - 1, dtype=object)
+    result = creation.multipolygons(polygons, indices=multipolygon_indices, out=result)
+    result[multipolygon_parts == 0] = creation.empty(
+        1, geom_type=GeometryType.MULTIPOLYGON
+    ).item()
+
+    return result
+
+
+def from_ragged_array(typ, coords, offsets=None):
```
brendan-ward, line 376:
```suggestion
def from_ragged_array(geometry_type, coords, offsets=None):
```

shapely/tests/test_flatcoords.py
```diff
@@ -0,0 +1,351 @@
...
+
+
+@pytest.mark.parametrize("geom", all_types_not_supported)
+def test_raise_geometry_type(geom):
+    with pytest.raises(ValueError):
+        shapely.to_ragged_array([geom, geom])
+
+
+def test_points():
+    arr = shapely.from_wkt(
```
brendan-ward, line 85: Suggest using `shapely.points` constructor here instead:

```
shapely.points([[0, 0], [1, 1], [np.nan, np.nan], ...])
```

(makes it a little easier to compare to `expected` below)

In fact, you could probably move `expected` up here, then

```Python
points = shapely.points(expected)
```

shapely/tests/test_flatcoords.py
```diff
@@ -0,0 +1,351 @@
...
+            "POLYGON EMPTY",
+            "POLYGON EMPTY",
+            "POLYGON ((30 10, 40 40, 20 40, 10 20, 30 10))",
+            "POLYGON EMPTY",
+        ]
+    )
+    typ, coords, offsets = shapely.to_ragged_array(arr)
+    expected = np.array(
+        [
+            [30.0, 10.0],
```
brendan-ward, line 157: Trailing 0's can be removed, and suggest grouping these to make it easier to compare to the input geometries, e.g.,

```Python
expected = np.array(
    [
        # first polygon
        [30, 10], [40, 40], [20, 40], ...
        # second polygon (exterior ring, interior ring)
        [35, 10], ...
        [20, 30], ...
        # empty polygons omitted from coordinates
        # fifth polygon
        [30, 10], ...
```

brendan-ward (commented):
Thanks for your work on this @jorisvandenbossche !  Overall it looks good to me.

Added some comments re: docstrings, variable naming, and additional comments in the tests; pretty much all of them are personal preferences, so feel free to ignore the suggestions if you don't agree.

shapely/_flatcoords.py
```diff
@@ -0,0 +1,426 @@
...
+        elif typ == GeometryType.MULTIPOINT:
+            coords, offsets = _get_arrays_multipoint(arr, include_z)
+        elif typ == GeometryType.MULTILINESTRING:
+            coords, offsets = _get_arrays_multilinestring(arr, include_z)
+        elif typ == GeometryType.MULTIPOLYGON:
+            coords, offsets = _get_arrays_multipolygon(arr, include_z)
+        else:
+            raise ValueError(f"Geometry type {typ.name} is not supported")
+
+    elif len(geom_types) == 2:
```
caspervdw, line 250: Is there a specific use case to be that tolerant here? In general I would say no to these kind of exceptions.

If there is a use case, then I'd prefer having a separate "force_collection" function to make this conversion explicit (instead of silently converting here)

shapely/_flatcoords.py
```diff
@@ -0,0 +1,426 @@
...
+
+    This follows the in-memory layout of the variable size list arrays defined
+    by Apache Arrow, as specified for geometries by the GeoArrow project:
+    https://github.com/geoarrow/geoarrow.
+
+    Parameters
+    ----------
+    arr : array_like
+        Array of geometries (1-dimensional).
+    include_z : bool, default False
```
caspervdw, line 153: Inserting NaNs may result into trouble when reconstructing the geometry.

I think that this function should expect homogeneous typed geometries. 

Mostly it is better to start out strict with input parsing. You can be nice to the user and fix inhomogeneously typed geometries, but maybe the user rather sees an error in that case so that the user is able to fix problems earlier in its process.

So I'd prefer a check on uniqueness of the `coordinate_dimension` as well (probably excluding the empties).

There are `force_2d` and `force_3d` functions if the user wants to fix inhomogeneous dimensionality

caspervdw (commented):
Two remarks both about the user input parsing.

I'd prefer having a strict homogeneous geometry type & dimensionality input policy,  and (if at all necessary) leave normalization (`force_2d`, `force_3d`, a possible `force_collection`) out of scope of the `to_ragged_array`.

shapely/tests/test_flatcoords.py
jorisvandenbossche: > Suggest using `shapely.points` constructor here instead:

Unfortunately that doesn't work reliably for empty geometries (you would have to use NaNs, but that depends on the GEOS version whether they get turned into empty geometries or not). 
(I would prefer to keep `expected` independent from the input)

I have been trying some different ways, but in the end landed on WKT seemingly being the most robust way to represent all kinds of geometries in a human readable way. 
I am also planning to create a standard test suite in geoarrow/geoarrow and use WKT there as well. And then it would be nice to keep the data consistent.

shapely/tests/test_flatcoords.py
jorisvandenbossche: That won't work with black, but I suppose I can use comments to turn off/on formatting around those blocks

jorisvandenbossche:
@brendan-ward @caspervdw thanks a lot for the feedback!

shapely/_flatcoords.py
jorisvandenbossche: I think a very concrete use case is that many shapely functions that create geometries will actually give you a mixture (GEOS is typically not strict about the type it returns). For example if you are overlaying polygons, you will almost always get a mix of Polygons and MultiPolygons. So my experience is that it is very normal to have such a mixture. 

The result would also be exactly the same, so there is no ambiguity in doing it (except that users might unexpectedly get multipolygon-formatted results if they thought to only have polygons).
(it also follows naturally from other functions we have like `get_num_geometries` which also work with mixtures)

I would also like that tools that use this function can enable such automatic behaviour (eg geopandas using this to convert geometries to arrow to then write to file). Of course, they could make use of a `force_collection` function, but that does require another pass over the data and cloning of geometries. While if it is integrated in this function, it can be done in one pass.

So I would personally like to have at least the option to do it within `to_ragged_array` (to be able to it as efficiently as possible). But it could be behind a flag? (in pyogrio we have a `promote_to_multi` keyword, which is True by default (if the file format requires it))

jorisvandenbossche: And BTW, we should certainly have a "force_collection" like function, that would be handy, regardless of the above discussion (PostGIS has this as https://postgis.net/docs/ST_Multi.html)

shapely/_flatcoords.py
jorisvandenbossche: Yeah, I would generally be fine with being strict about the dimensionality. 

A problem for implementing that, however, is that GEOS itself is not very strict about it, since they basically always store 3D coords, but then just NaNs for z if 2D. So there are some inconsistencies with empty geometries and checking `has_z`. 

And I don't want to exclude empties, since I want to use those functions to be able to store geometry data to disk, and empties are a fact of life ..

shapely/_flatcoords.py
caspervdw: I am :+1: on making this an option (and also on defaulting this to `True`)

I am mildly concerned that this doesn't roundtrip and that the outcome of a roundtrip of for instance a Polygon will depend on other elements in the array being a Multi- or not.

shapely/_flatcoords.py
caspervdw: We could keep the empties, just without asserting their dimensionality.

I just realized that I am contradicting myself (a few years ago); in the `coordinates.py` module I made the `include_z` exactly like you have here.

caspervdw (approved):

shapely/_flatcoords.py
jorisvandenbossche: Sorry, re-reading your comment, you meant "excluding the empties" for when checking the dimensionality (exactly because empties are a bit inconsistent there, as I was unnecessarily explaining ;)). Yes, that could indeed work. 

One question though: currently I default to outputting 2D, even if you have 3D data (which is consistent with `get_coordinates`). Do we keep that (and thus the `include_z` keyword)? Or do we check the first non-empty geometry to "infer" the dimensionality? 
(we probably need to keyword anyway, to be robust against corner cases of all missing / empty arrays where the user wants to have an ensured shape of the result)

shapely/_flatcoords.py
jorisvandenbossche: > I am mildly concerned that this doesn't roundtrip and that the outcome of a roundtrip of for instance a Polygon will depend on other elements in the array being a Multi- or not.

Yes, rountripping a mixture of singular/multi geoms is indeed not exact. I asked about this in the top post:

> For roundtripping mixed single and multi geometries, would we like to have an option in the `from_` function to return Multi geometries with 1 part as the singular version? (that doesn't guarantee a faithful rountrip, since you can have eg MultiPolygons with a single sub-polygon in your input, but I do think it would cover the large majority of use cases where someone would prefer to preserve the mixture).

We also had a long discussion about this in pyogrio: some GDAL-supported file formats requires a single geometry type, and so we convert singular/multi of the same type when writing, but with the consequence that reading back doesn't exactly preserve the types. I think among others Levi argues that this isn't typically that important (as long as it are always single/multi of the same type)

caspervdw: Ok 👍

shapely/_flatcoords.py
caspervdw: I think I would prefer inferring the dimensionality except when the user explicitly provides it (so `include_z=None` as default). Indeed to deal with all empty/all missing.

shapely/_flatcoords.py
jorisvandenbossche: If we add a `promote_to_multi` keyword, then that gives also some more complication / questions: what does `promote_to_multi=True` mean if you have only Polygons? Also output as MultiPolygon? 
And if so, we would have a default of `promote_to_multi=None` that means "only promote if you actually have a mixture", allowing the user to override with True or False to always be sure about what happens with singular types?

caspervdw: > And if so, we would have a default of promote_to_multi=None that means "only promote if you actually have a mixture", allowing the user to override with True or False to always be sure about what happens with singular types?

Yes I think that is precisely what should happen; an "automatic promotion if necessary" by default (like you implemented in this PR) and if it is True/False, make sure the output is always a collection (or not).

But to be clear , that effort can be separate from this PR, because it can be done in a backwards compatible way.

jorisvandenbossche: > But to be clear , that effort can be separate from this PR, because it can be done in a backwards compatible way.

Yes, that's right. If we have the default to autopromote if necessary/possible (current behaviour), then this can be done as an addition later on.

I opened https://github.com/shapely/shapely/issues/1617 to discuss the "force_collection" idea.

shapely/_flatcoords.py
jorisvandenbossche: So it's quite easy to infer the dimensionality for the default case of `include_z=None` (and raise an error if there is a mixed dimensionality). 

But I would prefer that the user can avoid the performance cost of doing this check by explicitly specifying `include_z=True/False`. But if we still want to be strict about homogeneous dimensionality even when not guessing if it's 2D or 3D, we still need to query the dimensionality anyway to be able to be strict .. (since `get_coordinates` handles a mixture fine, and I think that's also fine for that function)

caspervdw: I was typing a reply that such a check would be hardly noticable, but then I did a benchmark, and in fact a dimensionality check would have a significant impact on the `to_ragged_array`:

```
In [5]: x = shapely.linestrings(np.random.random((1000, 10, 2)))
In [6]: %timeit shapely.to_ragged_array(x)
174 µs ± 141 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
In [8]: %timeit np.unique(shapely.get_coordinate_dimension(x))
31.9 µs ± 423 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

The type check has similar impact:

```
In [9]: %timeit np.unique(shapely.get_type_id(x))
31.7 µs ± 833 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

jorisvandenbossche: > So I think I would go for:

OK, updated the PR to follow that logic.

For the inference, I went for setting `include_z` to `True` once there is at least one 3D geometry, in which case I indeed don't have to do anything special for the empties

jorisvandenbossche: Of course, since you mentioned yourself that this is only consistent since GEOS 3.9, this is failing for older GEOS versions.

So two options:

- Skip the failing tests related to include_z inference for older GEOS and leave as is
- Do actually add the `is_empty` check and only infer dimensionality on non-empties (that would have a corner case if you only have singular empty geometries, which then would not infer potentially 3D geometries, but that seems very much a corner case)

caspervdw: let’s leave this as is and drop a comment in the docstring for the include_z behaviour.

jorisvandenbossche: OK, I added a note about this in the `include_z` entry of the docstring. 

I think we can also certainly still fine-tune this in the future.

PR #1562, opened 2022-10-10
Refactor reducing set operations as gufuncs + return emtpy collection for empty/all-None input

jorisvandenbossche:
Next iteration on https://github.com/shapely/shapely/pull/1540 and https://github.com/shapely/shapely/pull/1501 
Addresses https://github.com/shapely/shapely/issues/1332

See https://github.com/shapely/shapely/pull/1540#discussion_r990693467 for details. This PR implements `intersection_all` and `symmetric_difference_all` as a generalized ufunc with the last dimension of the array that gets reduced. This way we can implement the reducing set operations without relying on the `.reduce` mode of standard ufuncs. 
To achieve the same functionality from before, we only have to reshape the input array to move the axis to be reduced (based on the `axis` keyword) to the last dimension.

I still need to add some more tests for the cases brought up in the various issues and PRs (and copy from https://github.com/shapely/shapely/pull/1540). 
I also didn't yet add a keyword to control the behaviour of skipping None or not, but that should be straightforward to add on top of the current changes (also a keyword to control the "initial" value (in case of empty or all None) that is returned could be added)

### DIFF
--- a/src/ufuncs.c
+++ b/src/ufuncs.c
@@ -984,6 +984,66 @@ static void YY_Y_func(char** args, npy_intp* dimensions, npy_intp* steps, void* 
 }
 static PyUFuncGenericFunction YY_Y_funcs[1] = {&YY_Y_func};
 
+/* Define the reducing geoms -> geom functions (Y_Y_reduce) */
+static void* intersection_all_data[1] = {GEOSIntersection_r};
+static void* symmetric_difference_all_data[1] = {GEOSSymDifference_r};
+static char Y_Y_reduce_dtypes[2] = {NPY_OBJECT, NPY_OBJECT};
+static void Y_Y_reduce_func(char** args, npy_intp* dimensions, npy_intp* steps,
+                                   void* data) {
+  FuncGEOS_YY_Y* func = (FuncGEOS_YY_Y*)data;
+  GEOSGeometry* geom = NULL;
+  GEOSGeometry* ret_ptr = NULL;
+  int n_geoms;
+
+  GEOS_INIT;
+
+  GEOSGeometry** geoms = malloc(sizeof(void*) * dimensions[1]);
+  if (geoms == NULL) {
+    errstate = PGERR_NO_MALLOC;
+    goto finish;
+  }
+
+  SINGLE_COREDIM_LOOP_OUTER {
+    CHECK_SIGNALS(i);
+    if (errstate == PGERR_PYSIGNAL) {
+      goto finish;
+    }
+    n_geoms = 0;
+    SINGLE_COREDIM_LOOP_INNER {
+      if (!get_geom(*(GeometryObject**)cp1, &geom)) {
+        errstate = PGERR_NOT_A_GEOMETRY;
+        goto finish;
+      }
+      if (geom == NULL) {
+        continue;
+      }
+      geoms[n_geoms] = geom;
+      n_geoms++;
+    }
+    if (n_geoms == 0) {
+      ret_ptr = GEOSGeom_createEmptyCollection_r(ctx, 7);
+    } else {
+      ret_ptr = GEOSGeom_clone_r(ctx, geoms[0]);
+      int j;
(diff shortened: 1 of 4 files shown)


src/ufuncs.c
```diff
@@ -3580,6 +3540,9 @@ int init_ufuncs(PyObject* m, PyObject* d) {
   DEFINE_YY_Y_REORDERABLE(union);
```
jorisvandenbossche, line 3540: @caspervdw this `_REORDERABLE` was to enable those to be used as reduction? In which case we can simpify this and make all set operations just `YY_Y`?

src/ufuncs.c
```diff
@@ -984,6 +875,68 @@ static void YY_Y_func(char** args, npy_intp* dimensions, npy_intp* steps, void*
...
+static void Y_Y_reduce_func(char** args, npy_intp* dimensions, npy_intp* steps,
+                                   void* data) {
+  FuncGEOS_YY_Y* func = (FuncGEOS_YY_Y*)data;
+  GEOSGeometry* geom = NULL;
+  GEOSGeometry* ret_ptr = NULL;
+  int n_geoms;
+
+  CHECK_NO_INPLACE_OUTPUT(1);
+
+  GEOS_INIT;
```
jorisvandenbossche, line 891: I started from the `polygonize` implementation, which is using OUTPUT_Y and thus not releasing the GIL. But we can use the `geom_arr` temporary array to store output GEOSGeometries trick here as well to be able to release the GIL.

src/ufuncs.c
```diff
@@ -984,6 +875,68 @@ static void YY_Y_func(char** args, npy_intp* dimensions, npy_intp* steps, void*
...
+      geoms[n_geoms] = geom;
+      n_geoms++;
+    }
+    if (n_geoms == 0) {
+      ret_ptr = GEOSGeom_createEmptyCollection_r(ctx, 7);
+    } else {
+      ret_ptr = GEOSGeom_clone_r(ctx, geoms[0]);
+      int j;
+      for (j = 1; j < n_geoms; j++) {
+        ret_ptr = func(ctx, ret_ptr, geoms[j]);
```
caspervdw, line 922: Here the old geometry should be destroyed before overwriting the `ret_ptr`.

src/ufuncs.c
```diff
@@ -984,6 +875,68 @@ static void YY_Y_func(char** args, npy_intp* dimensions, npy_intp* steps, void*
...
+    n_geoms = 0;
+    SINGLE_COREDIM_LOOP_INNER {
+      if (!get_geom(*(GeometryObject**)cp1, &geom)) {
+        errstate = PGERR_NOT_A_GEOMETRY;
+        goto finish;
+      }
+      if (geom == NULL) {
+        continue;
+      }
+      geoms[n_geoms] = geom;
```
caspervdw, line 913: Why did you choose to first collect the geometries in an array, and then reiterate to do the reduction? I think the `func` could be executed here directly as follows:

```
ret_ptr = NULL;  // before the loop
(...)
if (ret_ptr == NULL) {
    ret_ptr = GEOSGeom_clone_r(ctx, geom);
} else {
    GEOSGeom_destroy_r(ctx, ret_ptr);
    ret_ptr = func(ctx, ret_ptr, geom);
}
// error handling
```

src/ufuncs.c
caspervdw: Yes that's right.

src/ufuncs.c
jorisvandenbossche: Good point!

caspervdw (approved):
Looks good!
Let’s run valgrind before tagging 2.0.

caspervdw:
Two afterthoughts:

- this should be listed in the pygeos migration guide
- I am thinking about doing “ignore nan” also for the geometry-creation functions. There are two advantages here: 1) avoid having nans in geometries and 2) provide another way of bulk-constructing geometries with different lengths.

PR #1564, opened 2022-10-10
Move dev requirements to pyproject.toml 

EwoutH:
Removes `requirements-dev.txt` file and consolidates those requirements as optional dependencies in `pyproject.toml` with the [dev] extra.

Also requires a newer version of Cython for Python 3.11, and specify minimum instead of specific versions for Cython and Descartes.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -79,8 +79,7 @@ using ``venv``:
 
     $ python3 -m venv .
     $ source bin/activate
-    (env) $ pip install -r requirements-dev.txt
-    (env) $ pip install -e .
+    (env) $ pip install -e .[dev]
 
 Or using ``conda``:
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,8 +1,8 @@
 [build-system]
 requires = [
-    "cython",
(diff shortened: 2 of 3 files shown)


jorisvandenbossche:
FYI we might temporarily move away again from having the project details in pyproject.toml because it is still experimental in setuptools (and not supported everywhere, such as debian), see https://github.com/shapely/shapely/pull/1533. So I would keep this on hold until we made a decision there.

EwoutH:
In June [they removed](https://github.com/pypa/setuptools/pull/3347) the experimental status for a lot of parts of the pyproject.toml configuration, including project metadata (see [current docs](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html)).

> It is been a few months since the implementation of PEP 621 support landed in setuptools (after a long period of feedback), and it seems now that support for project metadata in pyproject.toml has stabilised. Moreover the existing implementation is backed by a PEP, therefore it makes sense to consider this part of the feature stable.

mwtoews:
Apologies @EwoutH as I've prepared a similar #1606

Issue #1575, opened 2022-10-17
BUG: weakref support broke PyPy support


jorisvandenbossche:
We didn't catch this because we don't have CI running PyPy (something to fix?), but when building packages on conda-forge for 2.0beta1, this was failing on PyPy when running the tests with the following [error](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=590249&view=logs&j=4f922444-fdfe-5dcf-b824-02f86439ef14&t=b2a8456a-fb11-5506-ca32-5ccd32538dc0&l=926):

```
===== testing package: shapely-2.0b1-py38h4ff34cb_0 =====
running run_test.py
implementation: PyPy
target platform: linux-64
python version: (3, 8)
ImportError while loading conftest '/home/conda/feedstock_root/build_artifacts/shapely_1666013794866/test_tmp/tests/conftest.py'.
tests/__init__.py:12: in <module>
    from shapely.geos import geos_version_string
../_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/lib/pypy3.8/site-packages/shapely/__init__.py:14: in <module>
    from .strtree import *  # NOQA
../_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/lib/pypy3.8/site-packages/shapely/strtree.py:8: in <module>
    from .geometry.base import BaseGeometry
../_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/lib/pypy3.8/site-packages/shapely/geometry/__init__.py:4: in <module>
    from .base import CAP_STYLE, JOIN_STYLE
../_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/lib/pypy3.8/site-packages/shapely/geometry/base.py:78: in <module>
    class BaseGeometry(shapely.Geometry):
E   TypeError: __weakref__ slot disallowed: we already got one
Traceback (most recent call last):
[...]

PR #1577, opened 2022-10-17
BUG: move weakref-able implementation to base C extension type (fix PyPy)

jorisvandenbossche:
Closes https://github.com/shapely/shapely/issues/1575

I haven't yet checked if this actually fixes it on PyPy, but it's a potential fix for the issue. My understanding is that on PyPy also extension types defined in C support weakref by default, so therefore adding this manually in the subclasses as we did here is failing (https://foss.heptapod.net/pypy/pypy/-/issues/2670#note_45293)

The implementation of adding a  `weakreflist` slot is mimicked from numpy (how they do that for the ndarray extension type)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -58,6 +58,11 @@ jobs:
             python: 3.9
             geos: 3.10.3
             numpy: 1.19.5
+          # pypy
+          - os: ubuntu-latest
+            python: "pypy3.9"
+            geos: 3.11.0
+            numpy: 1.23.3
 
     env:
       GEOS_VERSION: ${{ matrix.geos }}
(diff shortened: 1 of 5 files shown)


EwoutH:
Out of curiosity, did you figure out why PyPy 3.9 didn't work in the CI but 3.8 did?

jorisvandenbossche:
> Out of curiosity, did you figure out why PyPy 3.9 didn't work in the CI but 3.8 did?

Sorry, missed that question. I am not really sure anymore why I switched from 3.9 to 3.8 for PyPy, but so the reason the CI passed in the end is because the commit after that to skip some tests that don't work on PyPy, not because of the 3.9->3.8 switch.

PR #1585, opened 2022-10-27
Resolve pytest warnings (maint-1.8)

mwtoews:
Two files (`tests/test_doctests.py` and `tests/test_hash.py`) are refactored from older testing paradigms into pytests to resolve a few warnings from tests that returned non-None objects.

```
=============================== warnings summary ===============================
tests/test_doctests.py::test_suite
  $PREFIX/lib/pypy3.8/site-packages/_pytest/python.py:201: PytestReturnNotNoneWarning: Expected None, but tests/test_doctests.py::test_suite returned <unittest.suite.TestSuite tests=[<doctest._DocTestSuite tests=[$SRC_DIR/tests/test_binascii_hex.txt]>]>, which will be an error in a future version of pytest.  Did you mean to use `assert` instead of `return`?
    f"Expected None, but {pyfuncitem.nodeid} returned {result!r}, which will be an error in a "

tests/test_hash.py::test_point
  $PREFIX/lib/pypy3.8/site-packages/_pytest/python.py:201: PytestReturnNotNoneWarning: Expected None, but tests/test_hash.py::test_point returned True, which will be an error in a future version of pytest.  Did you mean to use `assert` instead of `return`?
    f"Expected None, but {pyfuncitem.nodeid} returned {result!r}, which will be an error in a "

tests/test_hash.py::test_multipoint
  $PREFIX/lib/pypy3.8/site-packages/_pytest/python.py:201: PytestReturnNotNoneWarning: Expected None, but tests/test_hash.py::test_multipoint returned True, which will be an error in a future version of pytest.  Did you mean to use `assert` instead of `return`?
    f"Expected None, but {pyfuncitem.nodeid} returned {result!r}, which will be an error in a "

tests/test_hash.py::test_polygon
  $PREFIX/lib/pypy3.8/site-packages/_pytest/python.py:201: PytestReturnNotNoneWarning: Expected None, but tests/test_hash.py::test_polygon returned True, which will be an error in a future version of pytest.  Did you mean to use `assert` instead of `return`?
    f"Expected None, but {pyfuncitem.nodeid} returned {result!r}, which will be an error in a "

tests/test_hash.py::test_collection
  $PREFIX/lib/pypy3.8/site-packages/_pytest/python.py:201: PytestReturnNotNoneWarning: Expected None, but tests/test_hash.py::test_collection returned True, which will be an error in a future version of pytest.  Did you mean to use `assert` instead of `return`?
[...]

PR #1586, opened 2022-10-27
TST: clean-up test_doctests (with binascii_hex), refactor test_hash

mwtoews:
Based on #1585 from maint-1.8, but targeted for main.

`test_hash.py` did not have any warnings, but was re-written using parameters, similar to the other PR.


jorisvandenbossche:
Thanks!

Sidenote: we should probably also just remove this test_doctest.py (it's only used for a single file `test_binascii_hex.txt` I think, and I don't see a reason why those can't be normal pytest tests (if it's not already covered in existing tests)

mwtoews:
The aims of this PR have changed, as it removes `test_doctests.py` and `test_binascii_hex.txt`, which is more-or-less covered by [`test_wkb.py`](https://github.com/shapely/shapely/blob/main/tests/test_wkb.py). The re-written `test_hash.py` remains.

jorisvandenbossche (approved):
Thanks!

PR #1589, opened 2022-10-28
API: rename 'radius' -> 'distance' keyword in shapely.buffer() for consistency

jorisvandenbossche:
See https://github.com/shapely/shapely/issues/1276#issuecomment-1295427157 for context

### DIFF
--- a/shapely/constructive.py
+++ b/shapely/constructive.py
@@ -83,7 +83,7 @@ MultiLineString, MultiPoint, Point, Polygon
 @multithreading_enabled
 def buffer(
     geometry,
-    radius,
+    distance,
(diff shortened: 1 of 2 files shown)


shapely/constructive.py
```diff
@@ -95,16 +95,16 @@ def buffer(
     Computes the buffer of a geometry for positive and negative buffer radius.
```
mwtoews, line 95:
```suggestion
    Computes the buffer of a geometry for positive and negative buffer distance.
```

mwtoews (approved):
Thanks, looks good, just one minor edit.

Issue #1587, opened 2022-10-28
Point no longer accepts arrays of shape (1,)


martinfleis:
## Expected behavior and actual behavior.

Shapely 2.0 no longer accepts arrays of shape (1,) as a valid input in `Point` constructor, while shapely 1.8.5 does and does not emit any warning while doing that.

## Steps to reproduce the problem.

shapely 2.0b1:

```py
In [1]: import numpy
   ...: import shapely
   ...: from shapely.geometry import Point
   ...: shapely.__version__
Out[1]: '2.0b1'

In [2]: x, y = numpy.array([1]), numpy.array([1])
   ...: x, y
Out[2]: (array([1]), array([1]))

In [3]: Point(x, y)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [3], in <cell line: 1>()
----> 1 Point(x, y)

File ~/mambaforge/envs/geo_dev/lib/python3.10/site-packages/shapely/geometry/point.py:77, in Point.__new__(self, *args)
     74     geom = shapely.points(np.array(args))
     76 if not isinstance(geom, Point):
---> 77     raise ValueError("Invalid values passed to Point constructor")
     78 return geom

ValueError: Invalid values passed to Point constructor
```

The same works correctly with 1.8.5:
```py
In [3]: Point(x, y)
Out[3]: <shapely.geometry.point.Point at 0x10679b730>
```

## Operating system

macOS 13.0, Ubuntu

## Shapely version and provenance

2.0b1 installed from PyPI as well as from conda-forge

xref https://github.com/pysal/spopt/pull/288#issuecomment-1294392000

jorisvandenbossche:
Good catch, thanks for the report. I think it is fine to allow that again.

jorisvandenbossche:
-> https://github.com/shapely/shapely/pull/1590

jorisvandenbossche:
A beta 2 is out with this fix

PR #1590, opened 2022-10-28
Allow passing len-1 array for x and y coordinates in Point(..) (compat with 1.8)

jorisvandenbossche:
Closes https://github.com/shapely/shapely/issues/1587

### DIFF
--- a/shapely/geometry/point.py
+++ b/shapely/geometry/point.py
@@ -71,10 +71,13 @@ class Point(BaseGeometry):
             geom = shapely.points(coords)
         else:
             # 2 or 3 args
-            geom = shapely.points(np.array(args))
+            geom = shapely.points(np.array(args).T)
 
         if not isinstance(geom, Point):
-            raise ValueError("Invalid values passed to Point constructor")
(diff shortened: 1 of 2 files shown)


martinfleis:
I think that this has introduced a regression geopandas CI is now picking - https://github.com/geopandas/geopandas/actions/runs/3351235289/jobs/5552549617

martinfleis:
GeoPandas CI fails on this

```py
Point([(2, 3), (11, 4), (7, 2), (8, 9), (1, 13)])
```

Which I'd say is a correct behaviour. I don't really understand what is the intention of this line. Shapely 1.8 creates `POINT (2 3)` but I'd say that raising `ValueError: Invalid values passed to Point constructor`, which is the behaviour of 2.0b2 is correct. But it should be noted, that it is a change of behaviour.

jorisvandenbossche:
@martinfleis thanks for the note. I am also unsure why this behaviour existed that way, and I think it is a good change. But a change indeed, so will add it to the release notes.

jorisvandenbossche:
-> https://github.com/shapely/shapely/pull/1600

PR #1591, opened 2022-10-29
MAINT: Remove descartes from requirements-dev.txt

### LINKED ISSUE #1145: Future of descartes documentation dependency?
I am one of the maintainers of the `python-shapely` package in Fedora Linux. We build the HTML documentation and make it available for installation in an RPM subpackage. Unfortunately, the `python-descartes` package in Fedora, needed for a complete build of the documentation, is in bad shape and likely to be retired because the tests fail with current Shapely, and there is no longer an upstream.

See https://pypi.org/project/descartes/, where the last release was in 2017, and the BitBucket home page has gone away with no apparent replacement.

Is there a plan to replace `descartes` in the documentation source with something that is maintained, or to fork and maintain `descartes`?

musicinmybrain:
It is no longer needed since the `shapely.plotting` module was added in cbf28cc.

Fixes #1145.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,4 @@
 cython==0.29.21
-descartes==1.0.1
 matplotlib
 numpy>=1.4.1
 pytest


jorisvandenbossche (approved):
Thanks!

Issue #1570, opened 2022-10-17
Functions to implement from GEOS 


jorisvandenbossche:
Opening a similar issue here as we had at https://github.com/pygeos/pygeos/issues/126.  
GEOS changelog at https://github.com/libgeos/geos/blob/master/NEWS.md

This issue lists GEOS functions that could be exposed in Shapely as new public functions as well.

GEOS <= 3.8:

* [ ] GEOSOrientationIndex
* [ ] GEOSNode -> https://github.com/shapely/shapely/pull/1431
* [ ] GEOSisValidDetail
* [ ] GEOSGeomGetEndPoint
* [ ] GEOSGeomGetStartPoint
* [ ] GEOSMinimumClearanceLine
* [ ] GEOSMinimumWidth
* [ ] GEOSRelateBoundaryNodeRule (GEOSRelatePatternMatch)
* [ ] GEOSDistanceIndexed
* [ ] GEOSSegmentIntersection
* [ ] GEOSPolygonize_valid

GEOS 3.9:

* [ ] GEOSPreparedDistance
* [ ] GEOSMaximumInscribedCircle
* [ ] GEOSLargestEmptyCircle

GEOS 3.10:

* [ ] GEOSMakeValidWithParams
* [ ] GEOSWKBWriter_setFlavor
* [ ] GEOSConstrainedDelaunayTriangulation

GEOS 3.11:

* [ ] GEOSGeom_transformXY (https://github.com/shapely/shapely/issues/1400)
* [ ] GEOSHilbertCode (https://github.com/shapely/shapely/issues/1398)
* [ ] GEOSConcaveHull (https://github.com/shapely/shapely/issues/1296) -> https://github.com/shapely/shapely/pull/1518
* [ ] GEOSConcaveHullOfPolygons
* [ ] GEOSPolygonHullSimplify
* [ ] GEOSPolygonHullSimplifyMode
* [ ] GEOSLineMergeDirected
* [ ] GEOSWKBReader_setFixStructure
* [ ] GEOSWKTReader_setFixStructure

GEOS 3.12:

* [ ] GEOSPreparedIntersectsXY and GEOSPreparedContainsXY -> https://github.com/shapely/shapely/pull/1548

---

If other functions with the same signature are already exposed, adding a new GEOS function to Shapely can be quite straightforward, see https://github.com/shapely/shapely/pull/1431/ as an example

akadouri:
[...]

PR #1592, opened 2022-11-01
ENH: Add directed parameter to line merge

akadouri:
GEOS 3.11.0 adds [GEOSLineMergeDirected](https://github.com/libgeos/geos/blob/main/capi/geos_c.h.in#L3925-L3934) which is similar to `line_merge` but only returns a merged geometry if possible without merging points.

Does it make sense to implement this as an additional parameter to `line_merge` instead of a separate function?
If so is it possible to specify a geos version restriction on the parameter?

One of the functions in #1570

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -11,6 +11,7 @@ Relevant changes in behaviour compared to 2.0b1:
   coordinates (fix compatibility with Shapely 1.8).
 - Raise ValueError for non-finite distance in the ``buffer()`` and ``offset_curve()``
   methods on the Geometry classes (consistent with Shapely 1.8).
+- Add `directed` parameter to `line_merge` for GEOS >= 3.11.0
 
 2.0b1 (2022-10-17)
 ------------------
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -11,7 +11,6 @@ Relevant changes in behaviour compared to 2.0b1:
   coordinates (fix compatibility with Shapely 1.8).
 - Raise ValueError for non-finite distance in the ``buffer()`` and ``offset_curve()``
   methods on the Geometry classes (consistent with Shapely 1.8).
-- Add `directed` parameter to `line_merge` for GEOS >= 3.11.0
 
 2.0b1 (2022-10-17)
 ------------------
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -5,7 +5,7 @@ The Shapely User Manual
 =======================
 
 :Author: Sean Gillies, <sean.gillies@gmail.com>
-:Version: 2.0b1
+:Version: 1.7.0
 :Date: |today|
 :Copyright:
   This work is licensed under a `Creative Commons Attribution 3.0
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -5,7 +5,7 @@ The Shapely User Manual
 =======================
 
 :Author: Sean Gillies, <sean.gillies@gmail.com>
-:Version: 1.7.0
+:Version: 2.0b1
 :Date: |today|
 :Copyright:
   This work is licensed under a `Creative Commons Attribution 3.0
--- a/docs/release/2.x.rst
+++ b/docs/release/2.x.rst
@@ -317,7 +317,7 @@ Several (new) functions from GEOS are now exposed in Shapely:
 * :func:`.segmentize` (GEOS >= 3.10)
 * :func:`.dwithin` (GEOS >= 3.10)
 * :func:`.remove_repeated_points` (GEOS >= 3.11)
-* :func:`.line_merge`  (GEOS > 3.11) added `directed` parameter.
+* :func:`.line_merge` added `directed` parameter (GEOS > 3.11)
 
 In addition some aliases for existing methods have been added to provide a
 method name consistent with GEOS or PostGIS:
--- a/docs/release/2.x.rst
+++ b/docs/release/2.x.rst
@@ -317,6 +317,7 @@ Several (new) functions from GEOS are now exposed in Shapely:
 * :func:`.segmentize` (GEOS >= 3.10)
 * :func:`.dwithin` (GEOS >= 3.10)
 * :func:`.remove_repeated_points` (GEOS >= 3.11)
+* :func:`.line_merge`  (GEOS > 3.11) added `directed` parameter.
 
 In addition some aliases for existing methods have been added to provide a
 method name consistent with GEOS or PostGIS:
@@ -330,9 +331,6 @@ method name consistent with GEOS or PostGIS:
 * :func:`.voronoi_polygons` (``ops.voronoi_diagram``)
 * :func:`.shortest_line` (``ops.nearest_points``)
 * :func:`.is_valid_reason` (``validation.explain_validity``)
-
-Functions with additional parameters introduced with new GEOS versions:
-* :func:`.line_merge`  (GEOS > 3.11) added `directed` parameter.
 
 
 Getting information / parts / coordinates from geometries
--- a/docs/release/2.x.rst
+++ b/docs/release/2.x.rst
@@ -331,6 +331,9 @@ method name consistent with GEOS or PostGIS:
 * :func:`.shortest_line` (``ops.nearest_points``)
 * :func:`.is_valid_reason` (``validation.explain_validity``)
 
+Functions with additional parameters introduced with new GEOS versions:
+* :func:`.line_merge`  (GEOS > 3.11) added `directed` parameter.
+
 
 Getting information / parts / coordinates from geometries
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/shapely/linear.py
+++ b/shapely/linear.py
@@ -92,12 +92,19 @@ def line_locate_point(line, other, normalized=False, **kwargs):
 
 @multithreading_enabled
 def line_merge(line, directed=False, **kwargs):
-    """Returns (multi)linestrings formed by combining the lines in a
+    """Returns (Multi)LineStrings formed by combining the lines in a
     MultiLineString.
 
-Lines are joined together at their endpoints in case two lines are intersecting. Lines are not joined when 3 or more lines are intersecting at the endpoints. Line elements that cannot be joined are kept as is in the resulting MultiLineString.
+    Lines are joined together at their endpoints in case two lines are
+    intersecting. Lines are not joined when 3 or more lines are intersecting at
+    the endpoints. Line elements that cannot be joined are kept as is in the
+    resulting MultiLineString.
 
-The direction of each merged LineString will be that of the majority of the LineStrings from which it was derived. Except if ``directed=True`` is specified, then the operation will not change the order of points within lines and so only lines which can be joined with no change in direction are merged.
+    The direction of each merged LineString will be that of the majority of the
+    LineStrings from which it was derived. Except if ``directed=True`` is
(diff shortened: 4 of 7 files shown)


jorisvandenbossche:
> Does it make sense to implement this as an additional parameter to `line_merge` instead of a separate function?

I think that makes sense, yes. And checking, PostGIS ([ST_LineMerge](https://postgis.net/docs/ST_LineMerge.html)) also has this as a keyword in a single function.

> If so is it possible to specify a geos version restriction on the parameter?

That's not possible with the current decorator. But I think you can do a similar check as what the decorator does under the hood manually here in case the keyword is specified (`if lib.geos_version < (3, 11, 0): UnsupportedGEOSVersionError("... requires at least GEOS 3.11.0")`

docs/manual.rst
```diff
@@ -5,7 +5,7 @@ The Shapely User Manual
 =======================
 
 :Author: Sean Gillies, <sean.gillies@gmail.com>
-:Version: 1.7.0
+:Version: 2.0b1
```
akadouri, line 8: Unrelated to this PR but this number is shown on `https://shapely.readthedocs.io/en/latest/manual.html` and needed to be bumped.

jorisvandenbossche: That reminded me we can actually let this fill in by sphinx -> https://github.com/shapely/shapely/pull/1595

shapely/tests/test_linear.py
```diff
@@ -136,6 +137,22 @@ def test_line_merge_geom_array():
     assert_geometries_equal(actual[1], LineString([(0, 0), (1, 2)]))
 
 
+@pytest.mark.skipif(shapely.geos_version < (3, 11, 0), reason="GEOS < 3.11.0")
+def test_line_merge_directed():
+    lines = MultiLineString([[(0, 0), (1, 0)], [(0, 0), (3, 0)]])
+    result = shapely.line_merge(lines)
+    assert_geometries_equal(result, LineString([(1, 0), (0, 0), (3, 0)]))
+    result = shapely.line_merge(lines, directed=True)
+    assert_geometries_equal(result, lines)
```
jorisvandenbossche, line 146: What happens if it is not possible to merge the lines (without changing point order)? Does GEOS raise an exception? Can we add a test for that as well?

CHANGES.txt
```diff
@@ -11,6 +11,7 @@ Relevant changes in behaviour compared to 2.0b1:
   coordinates (fix compatibility with Shapely 1.8).
 - Raise ValueError for non-finite distance in the ``buffer()`` and ``offset_curve()``
   methods on the Geometry classes (consistent with Shapely 1.8).
+- Add `directed` parameter to `line_merge` for GEOS >= 3.11.0
```
jorisvandenbossche, line 14: We can maybe add it in this list instead: https://shapely.readthedocs.io/en/latest/release/2.x.html#bindings-for-new-geos-functionalities (that is the detailed changelog for the final 2.0 release)

src/ufuncs.c
```diff
@@ -622,6 +622,7 @@ static void* segmentize_data[1] = {GEOSDensify_r};
 
 #if GEOS_SINCE_3_11_0
 static void* remove_repeated_points_data[1] = {GEOSRemoveRepeatedPoints_r};
+static void* line_merge_directed_data[1] = {GEOSLineMergeDirected_r};
```
jorisvandenbossche, line 625: Since this is a `Y_Y` function, can you move the above to where those are defined (above `FuncGEOS_Y_Y`, while here it is `FuncGEOS_Yd_Y`). 
Strictly speaking it doesn't matter where the variable is defined (also given this is working right now), but just for code organisation (you will need to add a `#if GEOS_SINCE_3_11_0` if/endif block above `FuncGEOS_Y_Y`, I think)

jorisvandenbossche (commented):
Thanks a lot for the PR! Added a few comments

shapely/tests/test_linear.py
akadouri: This is what the test is showing, that `result` is the same as the input `lines`.

akadouri: The input is returned if it can't merge the lines.

CHANGES.txt
akadouri: Updated the guide, let me know if the change makes sense. It doesn't really fit into the existing groups.

src/ufuncs.c
akadouri: Currently all the methods that require a GEOS version check are grouped together in those version checks, https://github.com/shapely/shapely/blob/main/src/ufuncs.c#L3593-L3633

Might make sense to reorganize them separately and keep this for consistency now.

shapely/tests/test_linear.py
jorisvandenbossche: Ah, yes, that's indeed what the test is showing here. I didn't get that from the docstring. I would maybe explicitly mention that?

src/ufuncs.c
jorisvandenbossche: Yes, they are put together at the bottom for the `DEFINE_..` calls, but what I meant is here where the `_data` object is defined. If you look at where `FuncGEOS_Y_Y` is defined (line 479), you can see that just above that there are also already #if / #endif blocks with geos versions for defining certain `.._data` functions.

shapely/tests/test_linear.py
jorisvandenbossche: Now, the fact that it returns the input unchanged if it was not possible to combine them, that's actually also the behaviour of the original (non-directed) `line_merge` function, so nothing changed here. 
So we don't need to specify this specifically for `directed`, but maybe it might still be useful to mention in general.

docs/release/2.x.rst
```diff
@@ -331,6 +331,9 @@ method name consistent with GEOS or PostGIS:
 * :func:`.shortest_line` (``ops.nearest_points``)
 * :func:`.is_valid_reason` (``validation.explain_validity``)
 
+Functions with additional parameters introduced with new GEOS versions:
+* :func:`.line_merge`  (GEOS > 3.11) added `directed` parameter.
```
jorisvandenbossche, line 335: I think it is fine to add it to the first list of new functions from GEOS (it's indeed not a new function in shapely, but is _is_ exposing a new GEOS function ..). If it is worded as "added directed parameter" then that should be clear I think

shapely/linear.py
```diff
@@ -110,7 +114,17 @@ def line_merge(line, **kwargs):
     <MULTILINESTRING ((0 2, 0 10), (0 11, 5 10))>
     >>> line_merge(MultiLineString())
     <GEOMETRYCOLLECTION EMPTY>
+    >>> line_merge(MultiLineString([[(0, 0), (1, 0)], [(0, 0), (3, 0)]]), directed=True)
+    <MULTILINESTRING ((0 0, 1 0), (0 0, 3 0))>
```
jorisvandenbossche, line 118: To illustrate the contrast between directed=True vs False, maybe you could show this exact example with the default behaviour as well (so that you can see the direction was changed to form a single line)

shapely/linear.py
```diff
@@ -90,13 +91,16 @@ def line_locate_point(line, other, normalized=False, **kwargs):
 
 
 @multithreading_enabled
-def line_merge(line, **kwargs):
+def line_merge(line, directed=False, **kwargs):
     """Returns (multi)linestrings formed by combining the lines in a
-    multilinestrings.
+    MultiLineString.
```
jorisvandenbossche, line 97:
```suggestion

Lines are joined together at their endpoints in case two lines are intersecting. Lines are not joined when 3 or more lines are intersecting at the endpoints. Line elements that cannot be joined are kept as is in the resulting MultiLineString.

The direction of each merged LineString will be that of the majority of the LineStrings from which it was derived. Except if ``directed=True`` is specified, then the operation will not change the order of points within lines and so only lines which can be joined with no change in direction are merged.

```

<- attempt to expand the docstring with some more context / explanation

docs/release/2.x.rst
```diff
@@ -317,6 +317,7 @@ Several (new) functions from GEOS are now exposed in Shapely:
 * :func:`.segmentize` (GEOS >= 3.10)
 * :func:`.dwithin` (GEOS >= 3.10)
 * :func:`.remove_repeated_points` (GEOS >= 3.11)
+* :func:`.line_merge`  (GEOS > 3.11) added `directed` parameter.
```
jorisvandenbossche, line 320:
```suggestion
* :func:`.line_merge` added `directed` parameter (GEOS > 3.11)
```

(just for consistency with the others above)

jorisvandenbossche (approved):
Thanks for the updates! Looking good!

akadouri:
Thanks for the review!

jorisvandenbossche:
Thanks again!

idan-miara:
Hi, 
Is there a simpler variation of `line_merge_directed` that merges/concats all the given lines in a row, keeping the order of the given lines, regardless of existence of connecting points in between (without duplicating the connecting points, if they exist)?
maybe something like the second example below:
```
import shapely
from shapely import MultiLineString, LineString
from shapely.testing import assert_geometries_equal

lines = MultiLineString([[(0, 0), (0, 1), (0, 2)], [(0, 2), (0, 3), (0, 4)], [(0, 4), (0, 5), (0, 6)]])
result = shapely.line_merge(lines, directed=True)
assert_geometries_equal(result, LineString([(0, 0), (0, 1), (0, 2), (0, 3), (0, 4), (0, 5), (0, 6)]))

# no connecting point (0,2) between 2nd and 3rd lines
lines = MultiLineString([[(0, 0), (0, 1), (0, 2)], [(0, 3), (0, 4)], [(0, 4), (0, 5), (0, 6)]])
result = shapely.line_concat(lines)
assert_geometries_equal(result, LineString([(0, 0), (0, 1), (0, 2), (0, 3), (0, 4), (0, 5), (0, 6)]))
```

This is similar to the proposed solution here,
https://gis.stackexchange.com/a/341776
but with removing duplicated connected points

If this is not available - I could make a PR to add this, would that be useful?

idanmiara:
Hi,
Is there a simpler variation of `line_merge_directed` that merges/concats all the given lines in a row, keeping the order of the given lines, regardless of existence of connecting points in between (without duplicating the connecting points, if they exist)?
maybe something like the second example below:

```
import shapely
from shapely import MultiLineString, LineString
from shapely.testing import assert_geometries_equal

lines = MultiLineString([[(0, 0), (0, 1), (0, 2)], [(0, 2), (0, 3), (0, 4)], [(0, 4), (0, 5), (0, 6)]])
result = shapely.line_merge(lines, directed=True)
assert_geometries_equal(result, LineString([(0, 0), (0, 1), (0, 2), (0, 3), (0, 4), (0, 5), (0, 6)]))

# no connecting point (0,2) between 2nd and 3rd lines
lines = MultiLineString([[(0, 0), (0, 1), (0, 2)], [(0, 3), (0, 4)], [(0, 4), (0, 5), (0, 6)]])
result = shapely.line_concat(lines)
assert_geometries_equal(result, LineString([(0, 0), (0, 1), (0, 2), (0, 3), (0, 4), (0, 5), (0, 6)]))
```

This is similar to the proposed solution here,
https://gis.stackexchange.com/a/341776
but also removing duplicated connected points.

If this is not available - I could make a PR to add `line_concat`, would that be useful?

akadouri:
@idanmiara you could do what the stackexchange answer says and then run [remove_repeated_points.](https://shapely.readthedocs.io/en/latest/reference/shapely.remove_repeated_points.html) I'm not sure something you're looking for exists in geos, where most of the shapely methods are pulled from. It may help to start a new issue here for visibility.

Issue #1573, opened 2022-10-17
Prevent NaN in geometries


caspervdw:
Discussed with @jorisvandenbossche . Port of Pygeos ticket: https://github.com/pygeos/pygeos/issues/323

Having NaN in geometries is always a bad idea. GEOS will respond unpredictably in this situation. We would like to prevent this *by default* in shapely.

- raise on NaN when constructing points
- ignore NaN when constructing linear geometries (linestring/linearring),
- raise when setting NaN in `set_coordinates`
- for io (WKB/WKT/geojson), we have no control

We'd like to expose this as an additional kwarg "`handle_nans`", which may have 2 or 3 options:  `{"ignore", "allow", "raise"}`.

caspervdw:
Some work in progress over here: https://github.com/caspervdw/Shapely/tree/casper-handle-nans

PR #1594, opened 2022-11-01
Add handle_nans to linestrings and linearrings

caspervdw:
This is a draft implementing the ideas in #1573 for ignoring / raising / allowing NaN values in geometry constructors.

Please let me know what you think about this.

### DIFF
--- a/shapely/_geometry_helpers.pyx
+++ b/shapely/_geometry_helpers.pyx
@@ -58,7 +58,7 @@ def _check_out_array(object out, Py_ssize_t size):
  
 @cython.boundscheck(False)
 @cython.wraparound(False)
-def simple_geometries_1d(object coordinates, object indices, int geometry_type, char handle_nans = 1, object out = None):
+def simple_geometries_1d(object coordinates, object indices, int geometry_type, int handle_nans = 1, object out = None):
     cdef Py_ssize_t idx = 0
     cdef unsigned int coord_idx = 0
     cdef Py_ssize_t geom_idx = 0
--- a/shapely/_geometry_helpers.pyx
+++ b/shapely/_geometry_helpers.pyx
@@ -58,7 +58,7 @@ def _check_out_array(object out, Py_ssize_t size):
  
 @cython.boundscheck(False)
 @cython.wraparound(False)
-def simple_geometries_1d(object coordinates, object indices, int geometry_type, object out = None):
+def simple_geometries_1d(object coordinates, object indices, int geometry_type, char handle_nans = 1, object out = None):
     cdef Py_ssize_t idx = 0
     cdef unsigned int coord_idx = 0
     cdef Py_ssize_t geom_idx = 0
@@ -133,7 +133,7 @@ def simple_geometries_1d(object coordinates, object indices, int geometry_type, 
                     # the error equals PGERR_LINEARRING_NCOORDS (in shapely/src/geos.h)
                     raise ValueError("A linearring requires at least 4 coordinates.")
 
-            seq = PyGEOS_CoordSeq_FromBuffer(geos_handle, &coord_view[idx, 0], geom_size, dims, ring_closure)
+            seq = PyGEOS_CoordSeq_FromBuffer(geos_handle, &coord_view[idx, 0], geom_size, dims, ring_closure, handle_nans)
             if seq == NULL:
                 return  # GEOSException is raised by get_geos_handle
             idx += geom_size
--- a/shapely/tests/test_creation.py
+++ b/shapely/tests/test_creation.py
@@ -150,6 +150,17 @@ def test_linestrings_invalid_ndim():
         shapely.linestrings(coords)
 
 
+
+def test_linestrings_ignore_nan():
+    actual = shapely.linestrings([[0, 1], [2, float("nan")], [2, 3]], handle_nans="ignore")
+    assert_geometries_equal(actual, LineString([(0, 1), (2, 3)]))
+
+
+def test_linestrings_error_nan():
+    with pytest.raises(shapely.GEOSException):
+        shapely.linestrings([[0, 1], [2, float("nan")], [2, 3]], handle_nans="error")
+
+
 def test_linearrings():
     actual = shapely.linearrings(box_tpl(0, 0, 1, 1))
     assert_geometries_equal(
--- a/src/geos.c
+++ b/src/geos.c
@@ -901,12 +901,14 @@ unsigned int count_finite(const double* buf, unsigned int size, unsigned int dim
   char *cp1, *cp2;
   unsigned int result = 0;
   char has_non_finite = 0;
+  unsigned int i, j;
 
   cp1 = (char*)buf;
   for (i = 0; i < size; i++, cp1 += cs1) {
+    cp2 = cp1;
     has_non_finite = 0;
     for (j = 0; j < dims; j++, cp2 += cs2) {
-      if !(npy_isfinite(*(double*)cp2)) {
+      if (!(npy_isfinite(*(double*)cp2))) {
         has_non_finite = 1;
         break;
       }
@@ -928,15 +930,16 @@ unsigned int count_finite(const double* buf, unsigned int size, unsigned int dim
  */
 GEOSCoordSequence* coordseq_from_buffer(GEOSContextHandle_t ctx, const double* buf,
                                         unsigned int size, unsigned int dims,
-                                        char ring_closure, char handle_nans, npy_intp cs1,
+                                        char ring_closure, int handle_nans, npy_intp cs1,
                                         npy_intp cs2) {
   GEOSCoordSequence* coord_seq;
   char *cp1, *cp2;
-  unsigned int i, j, actual_size;
-  double first_coord;
+  unsigned int i, j, current, first, actual_size;
+  double coord;
+  char all_finite;
 
   switch (handle_nans)
-  ​{
+  {
       case PYGEOS_HANDLE_NANS_ALLOW:
         actual_size = size;
         break;
@@ -949,6 +952,8 @@ GEOSCoordSequence* coordseq_from_buffer(GEOSContextHandle_t ctx, const double* b
           return NULL;  // TODO better error msg
         }
         break;
+      default:
+        return NULL;
   }
 
 #if GEOS_SINCE_3_10_0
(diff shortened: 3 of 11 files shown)


jorisvandenbossche:
This generally looks good to me!

While working on https://github.com/shapely/shapely/pull/1559, I realized that recent versions of GEOS actually convert NaN coordinates to empty Point (for point construction):

```
In [28]: shapely.points([[1, 2], [np.nan, np.nan]])
Out[28]: array([<POINT (1 2)>, <POINT EMPTY>], dtype=object)
```

vs

```
In [29]: shapely.linestrings([[1, 2], [np.nan, np.nan]])
Out[29]: <LINESTRING (1 2, NaN NaN)>
```

So that also makes that the "allow" option is still somewhat different depending on the actual geometry type.

jorisvandenbossche:
Ah, and I suppose we should also add `handle_nans` option to the `points` ufunc? (at least to allow the "raise" option, "ignore" is probably not an option since that changes output size)

src/geos.c
```diff
@@ -890,34 +891,84 @@ GEOSGeometry* PyGEOSForce3D(GEOSContextHandle_t ctx, GEOSGeometry* geom, double
...
+  return result;
+}
+
 /* Create a GEOSCoordSequence from an array
  *
  * Note: this function assumes that the dimension of the buffer is already
  * checked before calling this function, so the buffer and the dims argument
  * is only 2D or 3D.
+ *
+ * handle_nans: 0 means 'raise', 1 means 'allow', 2 means 'ignore'
```
jorisvandenbossche, line 929:
```suggestion
 * handle_nans: 0 means 'allow', 1 means 'ignore', 2 means 'raise'
```

caspervdw:
@jorisvandenbossche  I dusted of this old PR, it should be good to merge now (just for the linearrings and linestrings)

src/geos.h
```diff
@@ -100,6 +102,17 @@ enum {
       PyErr_SetString(PyExc_ValueError,                                                  \
                       "A linearring requires at least 4 coordinates.");                  \
       break;                                                                             \
+    case PGERR_NAN_COORD:                                                                \
+      PyErr_SetString(PyExc_ValueError,                                                  \
+                      "A NaN, Inf or -Inf coordinate was supplied. Remove the "          \
+                      "coordinate or adapt the 'handle_nans' parameter.");               \
+      break;                                                                             \
+    case PGWARN_NAN_COORD:                                                               \
```
jorisvandenbossche, line 110: This doesn't seem to be used anymore. Idea initially was to also have the option to raise a warning instead of error?

jorisvandenbossche (commented):
Thanks!

Didn't take a detailed look, so just some quick comments.

In the C code, there seem to be quite some unrelated formatting changes. Maybe you could do a separate PR doing a pass with the formatter on all C files? (we could maybe also add a clang format check in CI / in pre-commit, to enforece this) That would make the diff here a bit smaller.

For the tests, do we already have sufficient coverage for the default of "allow"? (it seems most of the new tests are for the other options).  
Also, it might be worth adding a test case where the NaN is in the z dimension? (for this case, users might also want to to ignore just that specific dimension, but I assume you can do that manually by just only passing x and y)

caspervdw:
Thanks for your quick reply @jorisvandenbossche !


> In the C code, there seem to be quite some unrelated formatting changes. Maybe you could do a separate PR doing a pass with the formatter on all C files? (we could maybe also add a clang format check in CI / in pre-commit, to enforece this) That would make the diff here a bit smaller.

Of course, sorry about the noise

> 
> For the tests, do we already have sufficient coverage for the default of "allow"? (it seems most of the new tests are for the other options). Also, it might be worth adding a test case where the NaN is in the z dimension? (for this case, users might also want to to ignore just that specific dimension, but I assume you can do that manually by just only passing x and y)

We didn't actually have tests that checked for the default `handle_nans="allow"`. I added them. There was a bit of a surprise in the 3D NaN coordinates, which probably came up earlier: the behaviour for 3D NaN coordinates is currently such that if you pass a (N x 3) coordinate array with the first z coordinate being NaN, the geometry will be considered 2D (`has_z == 2`), but all the z coordinates are actually there if you fetch them using `get_coordinates(..., include_z=True)`.

The new `handle_nans="ignore"` option doesn't special case the Z coordinate, so if you supply only-NaN z coordinates, you will get an empty geometry (and not a 2D geometry). And you will get an error if you do `handle_nans="error"`. We *could* detect all-NaN z coordinates and assume 2D in that case? Not sure about that.

jorisvandenbossche:
> There was a bit of a surprise in the 3D NaN coordinates, which probably came up earlier: the behaviour for 3D NaN coordinates is currently such that if you pass a (N x 3) coordinate array with the first z coordinate being NaN, the geometry will be considered 2D (`has_z == 2`), but all the z coordinates are actually there if you fetch them using `get_coordinates(..., include_z=True)`.

Didn't test right now, but I _think_ this has been improved on latest GEOS (currently main branch, so will only be in 3.12). Dan Baston did quite some work on refactoring the CoordSequence (also better support Z, M and ZM coords), and I think a CoordinateSequence should now correctly set its dimension when it is being constructed, without relying on checking a coordinate for NaN (it will fall back to this check if it was not constructed with explicit dimensionality).

jorisvandenbossche:
> I _think_ this has been improved on latest GEOS

To illustrate this, using released GEOS

```
In [2]: line = shapely.linestrings([[0, 1, float("nan")], [2, 2, float("nan")]])

In [3]: line
Out[3]: <LINESTRING (0 1, 2 2)>

In [4]: line.has_z
Out[4]: False

In [5]: shapely.get_coordinate_dimension(line)
Out[5]: 2
```

vs latest GEOS main:

```
In [2]: line = shapely.linestrings([[0, 1, float("nan")], [2, 2, float("nan")]])

In [3]: line
Out[3]: <LINESTRING Z (0 1 NaN, 2 2 NaN)>

In [4]: line.has_z
Out[4]: True

In [5]: shapely.get_coordinate_dimension(line)
Out[5]: 3

In [6]: shapely.geos_version
Out[6]: (3, 12, 0)
```

shapely/tests/test_creation.py
```diff
@@ -245,6 +296,56 @@ def test_linearrings_buffer(dim, order):
...
+    ],
+)
+def test_linearrings_ignore_nan(x, y):
+    actual = shapely.linearrings(x, y, handle_nans="ignore")
+    assert_geometries_equal(actual, LinearRing([(0, 3), (1, 4), (2, 5), (0, 3)]))
+
+
+def test_linearrings_ignore_nan_invalid():
+    with pytest.raises(ValueError):
+        shapely.linearrings([1, 2], [np.nan, 2], handle_nans="ignore")
```
jorisvandenbossche, line 332:
```suggestion
        shapely.linearrings([0, float("nan"), 0], [3, 4, 3], handle_nans="ignore")
```

So that it covers the corner case of being long enough when allowing NaNs but not when ignoring them

EDIT: changed the suggestion to use `shapely.linearrings([0, float("nan"), 0], [3, 4, 3], handle_nans="ignore")` instead of `shapely.linearrings([0, 1, float("nan"), 0], [3, 4, 5, 3], handle_nans="ignore")`

jorisvandenbossche: Actually, running that on this branch, this seems to do something wrong:

```
In [16]: shapely.linearrings([0, 1, float("nan"), 0], [3, 4, 5, 3], handle_nans="ignore")
Out[16]: <LINEARRING (0 3, 1 4, 0 3, 0 3)>
```

But, that seems to be our current behaviour anyway, if you indeed ignore the NaN in the above:

```
In [17]: shapely.linearrings([0, 1, 0], [3, 4, 3])
Out[17]: <LINEARRING (0 3, 1 4, 0 3, 0 3)>
```

So even if first and last coordinates are already equal, we do "close" the ring once more if it has a length of 3? (i.e. if it has 3 coordinates, we just always assume the ring has to be closed)

shapely/tests/test_creation_indices.py
```diff
@@ -174,6 +174,43 @@ def test_linestrings_out(indices, expected):
     assert actual is out
 
 
+@pytest.mark.parametrize(
+    "coordinates,indices,expected",
+    [
+        ([[1, 1], [1, float("nan")], [2, 2]], [0, 0, 0], [lstrs([[1, 1], [2, 2]])]),
+    ],
+)
+def test_linestrings_ignore_nan(coordinates, indices, expected):
```
jorisvandenbossche, line 183: Add the "allow_nan" variant of this test here as well?

src/geos.c
```diff
@@ -892,67 +894,205 @@ GEOSGeometry* PyGEOSForce3D(GEOSContextHandle_t ctx, GEOSGeometry* geom, double
...
+  }
 
-#if GEOS_SINCE_3_10_0
+  if (actual_size == 0) {
+    *coord_seq = GEOSCoordSeq_create_r(ctx, 0, dims);
+    return (*coord_seq != NULL) ? PGERR_SUCCESS : PGERR_GEOS_EXCEPTION; 
+  }
+
+  /* Rings automatically get an extra (closing) coordinate if they have
+     only 3 or if the first and last are not equal. */
```
jorisvandenbossche, line 1041: As you can see from my earlier comment, I was initially surprised by this "always close if they have only 3". But this is the current behaviour of course and can be left for a separate issue. But we _could_ also only do that if the first/last are not equal?

shapely/_geometry_helpers.pyx
```diff
@@ -118,24 +122,22 @@ def simple_geometries_1d(object coordinates, object indices, int geometry_type,
...
-                    ring_closure = 1
-                else:
-                    for coord_idx in range(dims):
-                        if coord_view[idx, coord_idx] != coord_view[idx + geom_size - 1, coord_idx]:
-                            ring_closure = 1
-                            break
-                # check the resulting size to prevent invalid rings
-                if geom_size + ring_closure < 4:
+            errstate = PyGEOS_CoordSeq_FromBuffer(geos_handle, &coord_view[idx, 0], geom_size, dims, is_ring, handle_nans, &seq)
+            if errstate == 9:  # PGERR_NAN_COORD (in shapely/src/geos.h)
```
jorisvandenbossche, line 126: One thing I am wondering is if we should "expose" those exact numbers from our C code to cython. As long as we are only using this ourselves, that is fine of course (we can always update the above code if that number would change). 

But if we want to make it official that third party packages can use the C/cython API, we could also change `PyGEOS_CoordSeq_FromBuffer` to catch the return value of `coordseq_from_buffer` and translate it into something more generic (0/1/2 or 0/1/-1)

jorisvandenbossche (commented):
This looks great! Did a detailed review of the code now, which generated some comments, but to clear generally looks good / fully on board with the changes.

jorisvandenbossche:
> * Do you think that we should consider linestring with all-NaN z values as 2 dimensional?

Not once they are created that way, I would say. I think the new behaviour with latest GEOS is better (and more consistent, not depending on checking the first coordinate), keeping track of the dimensionality how it was created. 
So the question is then if we would want to detect all-NaN z values inside the construction code, and in that case create a 2D coordseq. I am not sure that is something we should do (and would it then be another option, or fall under one of the existing ones?)

The question is when users would get such coordinates with all-NaN z values. You can easily get that using `get_coordinates` if you specify `include_z=True` for 2D geometries. But I think the answer here is: don't do that? (and we should maybe change the default for that to depend on the dimensionality of the geometry, like we did for `to_ragged_array`) 
Although one case I can think of: if you have an array of mixed 3D and 3D geometries, and call `get_coordinates` with `include_z=True`, the subset for 2D geometries will have all-NaN z values. So if you want to roundtrip that (preserving the mixed dimensionality), having the option to ignore the all-NaN z values could be useful (eg in `transform`)

> * Should we indeed also use this new parameter to handle Inf values? NaN and inf are conceptually different and it seems a bit random to me to disallow Inf and allow 1.4E+308.

No strong opinion. If treating NaNs, my feeling is that it makes sense to also handle Inf, although the NaN case is probably more common. 
(I am wondering if you can get Inf values eg when reprojecting?)

> * What do you think about the naming of the new argument. Maybe “handle_nonfinite”? Singular “handle_nan” or plural “handle_nans”?

Yeah, I had been thinking the same while reviewing. If we keep it handling both NaN and Inf, that would make the keyword more "correct". But on the other hand, NaN will be the general case, and so using a name that reflects that might be more user friendly.

> * The “ignore” enum value actually confused me when I looked at the PR after a few months. It could be interpreted as “ignore the fact that a value is NaN, just insert all of them into the linestring”. I think “skip” is slightly better.

Indeed, "ignore" could easily be interpreted like that. +1 on using "skip" instead.

shapely/tests/test_creation.py
caspervdw: This is backwards compatibility. It is there in master, and it was put there to remain compatible with shapely 1.x

Although it is better to just check if the coordiantes are equal, I don't like to be breaking for a shapely 2.1 release.

https://github.com/shapely/shapely/blob/cde4016dfb8d232d548087fe8338ec42a2174e20/src/ufuncs.c#L2596

src/geos.c
```diff
@@ -892,67 +894,205 @@ GEOSGeometry* PyGEOSForce3D(GEOSContextHandle_t ctx, GEOSGeometry* geom, double
...
+        return PGERR_NAN_COORD;
+      }
+      break;
+    default:
+      return PGERR_NAN_COORD;
+  }
 
-#if GEOS_SINCE_3_10_0
+  if (actual_size == 0) {
+    *coord_seq = GEOSCoordSeq_create_r(ctx, 0, dims);
```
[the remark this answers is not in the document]
caspervdw: I think it is OK to start returning empty linearrings?

src/geos.c
caspervdw: See comment above.

caspervdw:
@jorisvandenbossche  I pushed changes per your suggestions, I think it is ready for merge now.

caspervdw:
@jorisvandenbossche OK to merge?

PR #1596, opened 2022-11-02
CI: ensure doctests are running

jorisvandenbossche:
Because we limit the doctests to a single python/GEOS version combo, we need to remember updating the check if the python or GEOS version gets updated.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -37,7 +37,7 @@ jobs:
             geos: 3.10.3
             numpy: 1.21.3
           # 2022
-          - python: "3.11-dev" # switch to "3.11" when released (PEP 664)
+          - python: "3.11"
             geos: 3.11.0
             numpy: 1.23.3
             matplotlib: true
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -58,8 +59,8 @@ jobs:
             python: 3.9
             geos: 3.10.3
             numpy: 1.19.5
-          # pypy
-          - os: ubuntu-latest
+          # pypy (don't use ubuntu-latest to not overwrite existing geos 3.11.0 build)
```
akadouri, line 62: nit: double negative, suggest rephrasing to "Pin ubuntu version to ensure GEOS 3.11.0 build"

akadouri (approved):

.github/workflows/tests.yml
jorisvandenbossche, line 62:
```suggestion
          # pypy (use explicit ubuntu version to not overwrite existing ubuntu-latest + geos 3.11.0 build)
```

jorisvandenbossche: Thanks, reworded a bit to remove the double negative!

Issue #1588, opened 2022-10-28
ENH: allow arrays as argument to Geometry methods


jorisvandenbossche:
Currently, the Geometry scalars typically accept only scalar arguments in their methods. Take for example the `contains` method:

```python
import shapely
polygon = shapely.Polygon([(0, 0), (0, 1), (1, 1), (1, 0), (0, 0)])
point = shapely.Point(0.5, 0.5)
points = shapely.points([(0, 0), (0.5, 0.5), (1, 1)])
```

We can check this using the method for a single point:

```python
>>> polygon.contains(point)
True
```

But when passing the array of points (which might seem a logical thing to do for the user), we get an error at the moment:

```python
>>> polygon.contains(points)
...
~/scipy/repos/shapely/shapely/geometry/base.py in contains(self, other)
    596     def contains(self, other):
    597         """Returns True if the geometry contains the other, else False"""
--> 598         return bool(shapely.contains(self, other))
    599 
    600     def crosses(self, other):

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

And for this case you need to use the top-level function instead:

```python
>>> shapely.contains(polygon, points)
array([False,  True, False])
```

This of course works fine, but I think it can be surprising for users that it doesn't work for the Polygon method, and in any case the current error you get is quite unclear. 
The only reason we get the error, is because inside the method we call `bool(..)` on result of `shapely.contains(..)`, with the goal to convert the numpy boolean scalar to a Python bool. 

But we could easily relax this and do this conversion to a Python bool _only if_ the result is a scalar, and not an array. For example, if the implementation would look like:

```python
def contains(self, other):
    result = shapely.contains(self, other)
    return bool(result) if result.ndim == 0 else result
```

then `polygon.contains(points)` would also work and return the array `array([False,  True, False])`.

I think this would be a nice convenience for users (also especially for GeoPandas users since we have methods on GeoSeries that also accept array likes of geometries), and would also make it consistent for _all_ methods, since the above actually already works for methods that return geometries (where we don't have this conversion step from numpy to python scalar):

```python
# for methods that return geometries, this already works
>>> polygon.intersection(points)
array([<POINT (0 0)>, <POINT (0.5 0.5)>, <POINT (1 1)>], dtype=object)
```

cc @shapely/shapely-2

martinfleis:
Is there a plan to make this part of 2.0 release? I think it is really handy and the implementation overhead is quite small.

[...]

PR #1599, opened 2022-11-09
ENH: allow Geometry methods to return arrays if passed array as argument

jorisvandenbossche:
Closes https://github.com/shapely/shapely/issues/1588

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -64,9 +64,10 @@ def dump_coords(geom):
         raise GeometryTypeError("Unhandled geometry type: " + repr(geom.geom_type))
 
 
-def _maybe_unpack(result, typ):
+def _maybe_unpack(result):
     if result.ndim == 0:
-        return typ(result)
+        # convert numpy 0-d array / scalar to python scalar
+        return result.item()
     else:
         # >=1 dim array
         return result
@@ -311,11 +312,11 @@ class BaseGeometry(shapely.Geometry):
 
     def distance(self, other):
         """Unitless distance to other geometry (float)"""
-        return _maybe_unpack(shapely.distance(self, other), float)
+        return _maybe_unpack(shapely.distance(self, other))
 
     def hausdorff_distance(self, other):
         """Unitless hausdorff distance to other geometry (float)"""
-        return _maybe_unpack(shapely.hausdorff_distance(self, other), float)
+        return _maybe_unpack(shapely.hausdorff_distance(self, other))
 
     @property
     def length(self):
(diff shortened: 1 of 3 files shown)


shapely/tests/geometry/test_geometry_base.py
```diff
@@ -157,3 +158,104 @@ def test_constructive_properties(op):
...
+        "intersection",
+        "symmetric_difference",
+        "union",
+    ],
+)
+def test_array_argument_binary_geo(op):
+    box = Polygon([(0, 0), (0, 1), (1, 1), (1, 0), (0, 0)])
+    polygons = shapely.buffer(shapely.points([(0, 0), (0.5, 0.5), (1, 1)]), 0.5)
+
+    result = getattr(box, op)(polygons)
```
jorisvandenbossche, line 220: I added tests for the overlay set ops as wel, but just to note that those actually already behaved this way (didn't change the implementation)

shapely/tests/geometry/test_geometry_base.py
```diff
@@ -157,3 +158,110 @@ def test_constructive_properties(op):
...
     expected = getattr(shapely, op)(geom)
     assert result == expected
+
+
+@pytest.mark.parametrize(
+    "op",
+    [
+        "crosses",
+        "contains",
+        # "contains_properly",
```
mwtoews, line 168: to get `contains_properly` to work, perhaps add a try / except AttributeError with:
```python
fn = getattr(shapely.predicates, op)
result = fn(polygon, points)
...
expected = np.array([fn(polygon, p) for p in points], dtype=bool)
```
or perhaps just use that since each of these fn(a, b) signatures appear in the `shapely.predicates` submodule.

mwtoews (approved):
This looks well-thought out, and I can't see any glaring omissions to any other function signature. I only have a suggestion for "contains_properly".

shapely/tests/geometry/test_geometry_base.py
jorisvandenbossche: To be clear, the reason `contains_properly` can't be tested here, is because that's not a method on the Geometry class (only a top-level function), xref https://github.com/shapely/shapely/issues/1281

So I should probably just do a PR to add `contains_properly` as a method.

shapely/geometry/base.py
```diff
@@ -64,6 +64,14 @@ def dump_coords(geom):
         raise GeometryTypeError("Unhandled geometry type: " + repr(geom.geom_type))
 
 
+def _maybe_unpack(result, typ):
+    if result.ndim == 0:
+        return typ(result)
```
caspervdw, line 69: Did you try `result.item()`? That would prevent the separate type administration.

caspervdw (approved):
Looks good!

shapely/geometry/base.py
jorisvandenbossche: Good point, that's indeed simpler!

PR #1600, opened 2022-11-11
Document Point(..) constructor change no longer allowing coordinate sequence of len > 1

jorisvandenbossche:
See https://github.com/shapely/shapely/pull/1590#issuecomment-1295823911 for context

With shapely 1.8:

```
In [1]: from shapely.geometry import Point

In [2]: Point([(1, 2), (3, 4), (5, 6)]).wkt
Out[2]: 'POINT (1 2)'
```

One 2.x, this raises an error instead of ignoring the other coordinates. This PR adds an explicit test for this and a note in the release notes.

### DIFF
--- a/docs/release/2.x.rst
+++ b/docs/release/2.x.rst
@@ -158,6 +158,9 @@ Some additional backwards incompatible API changes were included in Shapely
 * A ``GeometryCollection`` that consists of all empty sub-geometries now
   returns those empty geometries from its ``.geoms`` attribute instead of
   returning an empty list (#1420).
+* The ``Point(..)`` constructor no longer accepts a sequence of coordinates
+  consisting of more than one coordinate pair (previously, subsequent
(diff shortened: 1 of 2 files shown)


caspervdw (approved):

docs/release/2.x.rst
```diff
@@ -158,6 +158,9 @@ Some additional backwards incompatible API changes were included in Shapely
 * A ``GeometryCollection`` that consists of all empty sub-geometries now
   returns those empty geometries from its ``.geoms`` attribute instead of
   returning an empty list (#1420).
+* The ``Point(..)`` constructor no longer accepts a sequence of coordinates
+  consisting of more than one coordinate pair (previously, subsequent
+  coordinates would be ignored) (#1600).
```
caspervdw, line 163:
```suggestion
  coordinates were ignored) (#1600).
```

PR #1604, opened 2022-11-14
DEP: change `almost_equals()` to be removed from version 2.0 to 2.1

mwtoews:
Xref to #889 and 26cafd027ab26a6b728ce8a3d26f0b9c8b114bf1

This function was deprecated for 1.8 and planned to be removed for 2.0. But willing to hold off for another release if there is any compelling reason.

---

On a related topic, these signatures have different `tolerance` defaults:

- [`shapely.geometry.base.BaseGeometry.equals_exact(self, other, tolerance)`](https://github.com/shapely/shapely/blob/bc7fa916fc680b43cf58fc206b039e2563ef7a7f/shapely/geometry/base.py#L703)
- [`shapely.predicates.equals_exact(a, b, tolerance=0.0, **kwargs)`](https://github.com/shapely/shapely/blob/bc7fa916fc680b43cf58fc206b039e2563ef7a7f/shapely/predicates.py#L944)

It was noted in https://github.com/shapely/shapely/issues/889#issuecomment-609093066 that tolerance should have a default of 0. Should it?

### DIFF
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -1097,10 +1097,9 @@ differently.
   >>> b == c
   False
 
-.. method:: object.almost_equals(other[, decimal=6])
+.. method:: object.equals_exact(other, tolerance)
 
-  Returns ``True`` if the object is approximately equal to the `other` at all
-  points to specified `decimal` place precision.
+  Returns ``True`` if the object is within a specified `tolerance`.
 
 .. method:: object.contains(other)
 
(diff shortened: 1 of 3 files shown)


jorisvandenbossche:
>  But willing to hold off for another release if there is any compelling reason.

We are still in the process of deprecating it in geopandas, so that might be a reason to keep a bit longer (eg remove in 2.1)

mwtoews:
While shapely has no formal policy, [NEP 23](https://numpy.org/neps/nep-0023-backwards-compatibility.html) suggests that deprecated functions should be removed after 2 releases, therefore removing for 2.1 makes sense.

jorisvandenbossche (approved):
Thanks!

PR #1606, opened 2022-11-14
MAINT: remove requirements-dev.txt; expand optional dependencies in pyproject.toml

mwtoews:
This PR has a few related aims:

- Remove requirements-dev.txt
- Expand the optional dependencies (in pyproject.toml) for "test" and "docs"

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -79,8 +79,7 @@ using ``venv``:
 
     $ python3 -m venv .
     $ source bin/activate
-    (env) $ pip install -r requirements-dev.txt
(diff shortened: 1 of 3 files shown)


jorisvandenbossche:
There was also already https://github.com/shapely/shapely/pull/1564, but I think updating the existing "test" and "docs" extras as you do here is better.

mwtoews:
I thought it was familiar! I don't think "cython>=0.29.32" is important, as typically this is done in an isolated environment were the latest versions are retrieved from PyPI.

jorisvandenbossche (approved):

PR #1607, opened 2022-11-14
MAINT: Add Python 3.11 classifier and upgrade to GEOS 3.11.1; other CI upgrades

mwtoews:
Add a Python 3.11 classifier.

This follows on from #1596 to test Python 3.11 for the "dev" builds (although presumably 3.12-dev could be considered), and also bump NumPy to 1.23.4.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,7 +46,7 @@ jobs:
     name: Build ${{ matrix.arch }} wheels on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     env:
-      GEOS_VERSION: "3.11.0"
+      GEOS_VERSION: "3.11.1"
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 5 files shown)


jorisvandenbossche (approved):
Thanks!

jorisvandenbossche:
Those lines need to be updated:

https://github.com/shapely/shapely/blob/e027985cef648c5ce07c26a06dbd986c222e107e/tests/test_locale.py#L38-L42

`resetlocale` is deprecated starting with Python 3.11. The message says "Use locale.setlocale(locale.LC_ALL, "") instead", so what we already do for windows.

mwtoews:
Also upgrading [GEOS 3.11.1](https://github.com/libgeos/geos/releases/tag/3.11.1) released earlier today.

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

Issue #1531, opened 2022-09-10
Deprecate the parallel_offset alias of offset_curve


jorisvandenbossche:
Follow up items from https://github.com/shapely/shapely/pull/1512:

- Deprecate the `resolution` keyword in the `buffer()` geometry class method
- Deprecate the `parallel_offset` alternative for `offset_curve` (given that it is not an exact alias, but has a different API, to avoid confusing it's maybe better to remove this)

In https://github.com/shapely/shapely/pull/1512 I argued to leave those deprecations for after 2.0.0, so it is easier for people to write code that is compatible with both 1.8 and 2.0 (without having to do if/else or catch deprecation warnings). But that's certainly open for discussion.

jorisvandenbossche:
The `resolution` keyword already got deprecated in https://github.com/shapely/shapely/pull/2243

PR #1611, opened 2022-11-15
DOC: clean-up docstrings and Sphinx markup for deprecated functions

mwtoews:
This PR attends to a TODO to deprecate `resolution` from `buffer()`.

Also a few clean-ups, updating docstrings, using Sphinx directives, etc.

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -32,11 +32,13 @@ def geom_factory(g, parent=None):
     """
     Creates a Shapely geometry instance from a pointer to a GEOS geometry.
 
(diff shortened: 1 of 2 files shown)


mwtoews:
See also #1512 and discussion in https://github.com/shapely/shapely/pull/1512#discussion_r957734475

#1531 suggests that this PR could be held until after 2.0 is released, which is fine with me.

mwtoews:
Re-setting 2.0 milestone as a bit of a clean-up

jorisvandenbossche (approved):
Looks good, thanks!

PR #1612, opened 2022-11-15
DEP: remove 'preserve_topology' from 'set_precision()'

mwtoews:
This cleans up a deprecation from PyGEOS, that should not apply to shapely 2.0


jorisvandenbossche (approved):
Good catch!

PR #1619, opened 2022-11-22
CLN: clean c code - remove unused variables / functions

jorisvandenbossche:
Cleaning up some compilation warnings.


src/geos.c
```diff
@@ -19,14 +19,14 @@ int init_geos(PyObject* m) {
...
   PyModule_AddObject(m, "GEOSException", geos_exception[0]);
-  return 0;
 
   void* context_handle = GEOS_init_r();
   // TODO: the error handling is not yet set up for the global context (it is right now
   // only used where error handling is not used)
   // GEOSContext_setErrorMessageHandler_r(context_handle, geos_error_handler, last_error);
   geos_context[0] = context_handle;
 
+  return 0;
```
jorisvandenbossche, line 29: The reason this didn't actually cause issues (i.e. the global `geos_context` actually being null) is because this is right now only used for `GEOSGeom_Destroy`, and that function only needs the context in case an error happens when deallocating, which I suppose doesn't happen in standard usage)

caspervdw (approved):

src/ufuncs.c
```diff
@@ -2780,7 +2777,6 @@ static void bounds_func(char** args, npy_intp* dimensions, npy_intp* steps, void
   GEOSGeometry *envelope = NULL, *in1;
   const GEOSGeometry* ring;
   const GEOSCoordSequence* coord_seq;
-  int size;
```
caspervdw, line 2783: This is actually used for GEOS < 3.7

PR #1625, opened 2022-11-24
CI/TST: fix tests for voronoi_diagram for latest GEOS

jorisvandenbossche:
The return value has been changed to always be a MultiLineString, instead of either LineString or MultiLineString, depending on the number of input points -> https://github.com/libgeos/geos/pull/702 
This updates the test to reflect that.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -166,7 +166,7 @@ jobs:
         # Enable this if we have failures on GEOS main
         # continue-on-error: ${{ matrix.geos == 'main' }}
         run: |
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
Some more errors turned up compared to what we had on the main branch. The new ones might be related to a numpy change?

mwtoews:
See https://github.com/numpy/numpy/releases/tag/v1.24.0rc1

jorisvandenbossche:
Yes, I suppose because of "Ragged array creation will now always raise a ValueError unless dtype=object is passed. This includes very deeply nested sequences.", this now raises with a different error (message). Will handle this in a separate PR.

PR #1626, opened 2022-11-24
DOC: Improve STRtree docstrings

### LINKED ISSUE #1622: DOC: Enhance description of STRtree to indicate that geometries are indexed and queried based on extents

`STRTree` returns a collision where there is none.
A minimal case:
image

```python
from shapely import wkt
from shapely.strtree import STRtree

poly_query = wkt.loads('POLYGON ((58.27 44.72, 56.96 47.18, '
                       '58.47 47.98, 59.78 45.52, '
                       '58.27 44.72))')
poly_in_tree = wkt.loads('POLYGON ((57.94 44.73, 57.94 44.42, '
                         '56.02 44.42, 56.02 44.73, '
                         '57.94 44.73))')

query_tree = STRtree([poly_in_tree])
collide_objects = query_tree.query_geoms(poly_query)

assert not poly_in_tree.intersects(poly_query)
assert 0 == len(collide_objects)
```
## Operating system
Windows

Shapely 1.8.2 via `conda`.

brendan-ward:
Resolves #1622 

Improves STRtree docstrings in an attempt to make it more clear that the tree is based on 2-dimensional bounding boxes.  Also added a note about tree efficiency, but that might be overkill for the docstring and more appropriate for dedicated documentation on STRtree - which we have not yet created.

/cc @adishavit

### DIFF
--- a/shapely/strtree.py
+++ b/shapely/strtree.py
@@ -30,9 +30,20 @@ class STRtree:
     A query-only R-tree spatial index created using the
     Sort-Tile-Recursive (STR) [1]_ algorithm.
 
-    For two-dimensional spatial data. The tree is constructed directly
-    at initialization. The tree is immutable and query-only, meaning that
-    once created nodes cannot be added or removed.
+    The tree indexes two-dimensional axis-aligned bounding boxes of each
+    geometry; any Z values present in geometries are ignored for purposes of
+    indexing within the tree.  The tree is constructed directly at
+    initialization and nodes cannot be added or removed after it has been
+    created.
+
+    The tree is more efficient for querying when there are fewer geometries that
+    have overlapping bounding boxes and where there is greater similarity
+    between the outer boundary of a geometry and its bounding box.  For example,
+    a MultiPolygon composed of widely-spaced individual Polygons will have a
+    large overall bounding box compared to the boundaries of its individual
+    Polygons, and the bounding box may also potentially overlap many other
+    geometries within the tree.  This means that the resulting tree may be less
+    efficient to query than a tree constructed from individual Polygons.
 
     All operations return indices of the input geometries.  These indices
     can be used to index into anything associated with the input geometries,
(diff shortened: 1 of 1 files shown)


brendan-ward:
Test failures are unrelated, this is a doc only change

jorisvandenbossche:
> Test failures are unrelated, this is a doc only change

Yes, I am trying to fix those in https://github.com/shapely/shapely/pull/1625

mwtoews (approved):
Nicely written, and good to distinguish between a geometry's "extent" vs "bounding box", among several other details added.

jorisvandenbossche:
> good to distinguish between a geometry's "extent" vs "bounding box",

Apart from that "bounding box" is a clearer name (and I assume more people will understand it), what do you understand to be the difference between both? (my assumption was that GEOS uses "extent" term for bounding box)

jorisvandenbossche:
Actually, in the STRtree API, they actually use "envelope", not extent (and also the property in shapely is `envelope`), so yet another term .. :)

shapely/strtree.py
```diff
@@ -94,8 +105,8 @@ def geometries(self):
     def query(self, geometry, predicate=None, distance=None):
         """
         Return the integer indices of all combinations of each input geometry
-        and tree geometries where the extent of each input geometry intersects
-        the extent of a tree geometry.
+        and tree geometries where the two-dimensional axis-aligned bounding box
+        of each input geometry intersects the bounding box of a tree geometry.
```
jorisvandenbossche, line 109: Is the "axis-aligned" needed here? Or can we understand "bounding box" to always be axis aligned? 

The reason I am asking is that it, while maybe strictly more correct, also makes this summary paragraph more complex to read. 

Wondering if we could leave the "two dimensional" and "axis-aligned" for a second (or other) paragraph where details are given what kind of bounding box is used and that z dimensions are ignored, and keeping the first summary paragraph simpler.

jorisvandenbossche (commented):
Thanks for the clarifications! That are nice improvements, just one nitpicky comment

adishavit:
I think it helps. That was my original misunderstanding. I would go further
and mention `.bounds` too.

mwtoews:
> GEOS uses "extent" term for bounding box.

[Yes](https://github.com/libgeos/geos/blob/3.11.0/capi/geos_c.h.in#L2565), but as mentioned previously it is not widely understood as bounding box. Agree there are many terms for the same thing.

jorisvandenbossche (approved):
Thanks!

PR #1627, opened 2022-11-24
CI/TST: fix tests for changed numpy error from creating ragged array

jorisvandenbossche:
xref https://github.com/shapely/shapely/pull/1625#issuecomment-1326685177

### DIFF
--- a/shapely/tests/test_strtree.py
+++ b/shapely/tests/test_strtree.py
@@ -177,7 +177,7 @@ print("done")
 )
 @pytest.mark.filterwarnings("ignore:Creating an ndarray from ragged nested sequences:")
 def test_query_invalid_geometry(tree, geometry):
-    with pytest.raises(TypeError, ValueError):
+    with pytest.raises((TypeError, ValueError)):
         tree.query(geometry)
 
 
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
```
>>> np.__version__
'1.23.4'
>>> np.array([[1], [2, 3]])
VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple 
  of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do 
  this, you must specify 'dtype=object' when creating the ndarray.
array([list([1]), list([2, 3])], dtype=object)
```
vs

```
>>> np.__version__
'1.24.0rc1'
>>> np.array([[1], [2, 3]])
...
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous 
shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part.
```

Now this raises an error. Before, we would also get some (other) error in a later stage because of the object dtype array (when expecting a numerical array of coordinates)

mwtoews (approved):
Looks good!

PR #1628, opened 2022-11-24
ENH: expose flavor keyword in to_wkb

jorisvandenbossche:
Adding a `flavor` keyword for WKB writing (added in GEOS 3.10.0)


mwtoews:
This looks good so far! One thing that I find puzzling about the [GEOS docs on wkb](https://libgeos.org/specifications/wkb/) describe three "flavours" (Canadian devs 'eh?) but [enum GEOSWKBFlavors](https://github.com/libgeos/geos/blob/3.10.0/capi/geos_c.h.in#L228-L242) only has two, as it does not have "standard". Perhaps @pramsey could say if there is a "standard" flavour? Would "standard" just force `output_dimension=2`? If so, should that logic be embed here as a third choice?

jorisvandenbossche:
Yes, AFAIU the extended vs ISO only _matters_ when you have >2D geometries and/or want to include SRID. For 2D geometries without SRID, extended or ISO WKB are identical:

> Standard WKB supports two-dimensional geometry, and is a proper subset of both extended WKB and ISO WKB.

GEOS itself doesn't do any validation of "wrong" combinations. For example, it will ignore the flag about setting SRID if you specify ISO; that's the reason for this specific case I added a validation check for this on the python side.

> Would "standard" just force `output_dimension=2` and `include_srid=False`? If so, should that logic be embed here as a third choice?

I am not sure that is worth adding, since that is already the default behaviour anyway. We could use such a value (or `flavor=None`) if we would want to force the user to make an explicit choice between exended/iso if you ask for `output_dimension=3`, but that would not be backwards compatible (GEOS has always been using extended before / by default).

mwtoews:
> Yes, AFAIU the extended vs ISO only _matters_ when you have >2D geometries and/or want to include SRID. For 2D geometries without SRID, extended or ISO WKB are identical.

Yes, that's correct.

> I am not sure it is worth adding a `flavor="standard"` option, since that is already the default behaviour anyway (and it is controlled by the other keywords). We could use such a value (or maybe rather `flavor=None`) if we would want to force the user to make an explicit choice between exended/iso if you ask for `output_dimension=3` (i.e. that in such a case you would always need to also specify `flavor=..`), but that would not be backwards compatible (GEOS has always been using extended before / by default).

Oh right, the existing default `output_dimension=3` would be incompatible with `flavor="standard"`. Then perhaps it's just a note to add to the docstring to say that:
> Standard WKB is returned by setting `output_dimension=2` and `include_srid=False`, regardless of flavor choice.

jorisvandenbossche:
Ah, sorry I was wrong about 2D being the default. We actually have `output_dimension=3` as the default, but that still only affects if you actually have 3D geometries. In case we have 2D geometries, we still just output 2D "standard" WKB. 
So that makes it indeed even harder to add such an option. 

I can certainly add a note.

mwtoews (approved):

shapely/io.py
```diff
@@ -138,10 +147,23 @@ def to_wkb(
     include_srid : bool, default False
         If True, the SRID is be included in WKB (this is an extension
         to the OGC WKB specification).
```
brendan-ward, line 149:
```suggestion
        to the OGC WKB specification).
       Not allowed when flavor is "iso".
```

shapely/io.py
```diff
@@ -138,10 +147,23 @@ def to_wkb(
...
+        embedding). ISO flavor is "more standard" for 3D output.
+        The `from_wkb` function can read both flavors.
     **kwargs
         For other keyword-only arguments, see the
         `NumPy ufunc docs <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs>`_.
 
+    Notes
+    -----
+
+    Standard WKB is returned by setting ``output_dimension=2`` and
```
brendan-ward, line 163: The term "Standard" is potentially confusing here, as both ISO and EWKB are also standards.

Maybe rephrase to

```Python
Both flavors are equivalent when ``output_dimension=2`` and ``include_srid=False``.
```

brendan-ward (approved):
Thanks for working on this @jorisvandenbossche.

A couple minor clarifying comments, but otherwise this looks good to go.

shapely/io.py
jorisvandenbossche: Yeah, I agree that "standard" not a great term. It is used by the GEOS docs, though (https://libgeos.org/specifications/wkb/), and so that was the reason for Mike to suggest it. 

But I like your phrasing, and that can maybe be moved into the flavor parameter explanation.

PR #1632, opened 2022-11-27
BLD: pin to Cython~=0.29, ignore .c and .pyx files in wheels

mwtoews:
This PR ignores a few files from binary wheels that should not be there. The .c, .pxd and .pyx files were not present before #1426.

The curious change of packaging is because the default for include_package_data is different between `setup.py` and `pyproject.toml` metadata structures. Before #1426 the default for `setup.py`'s `include_package_data` was False, but the default for `pyproject.toml`'s `include-package-data` is true ([ref](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#include-package-data)), so it needs to be explicitly set back to false.

Only a few .c, .pxd and .pyx files in .whl files are excluded. Sdist contents are unchanged.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-    "cython",
+    "Cython~=0.29",
     "oldest-supported-numpy",
     "setuptools>=61.0.0",
 ]
@@ -57,11 +57,14 @@ docs = [
 Documentation = "https://shapely.readthedocs.io/"
 Repository = "https://github.com/shapely/shapely"
 
+[tool.setuptools]
+include-package-data = false
+
 [tool.setuptools.packages.find]
-include = [
-    "shapely",
-    "shapely.*",
-]
+include = ["shapely", "shapely.*"]
+
+[tool.setuptools.package-data]
+"shapely" = ["*.pxd"]
 
 [tool.coverage.run]
 source = ["shapely"]


jorisvandenbossche:
> Before #1426 the default for `setup.py`'s `include_package_data` was False, but the default for `pyproject.toml`'s `include-package-data` is true ([ref](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#include-package-data)), so it needs to be explicitly set back to false.

Looking at the diff of that PR, we actually explicitly had `include_package_data=True,` before moving to pyproject.toml. So to keep it the same as before that PR, we should actually set it to True. 
I suppose this was something that was introduced when merging the setup.py's from shapely and pygeos, since this is also done in pygeos' setup.py.

I think it would be good to keep including .pxd files (so you can write your own cython code building against shapely), but it should be fine to exclude the .pyx files. 
The cython documentation (https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#distributing-cython-modules) mentions to use `package_data` for this purpose. In setup.py syntax, that would be something like:

```
setup(
    package_data = {
        'shapely': ['*.pxd'],
    },
    ...
)
```

mwtoews:
Thanks for the corrections @jorisvandenbossche . For some reason I had assumed they were not packaged before, but after inspecting the contents of a few .whl files from PyPI they have always existed for binary wheels for the 1.8 series (speedups only), and for the first 2.0 alpha. PyGEOS' .whl files also include these files too.

I'll close this PR, since no change is needed since the default is already `include-package-data = true`.

mwtoews:
Um, closed too early. I'll take your recommendation to keep .pxd and exclude .pyx files. I'll re-create the commit.

mwtoews:
PR updated, also to pin to Cython ~=0.29 for future-proofing. I can't see when Cython 3 is anticipated (it's been in the works for a few years; [docs](https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html)), but it is not tested to work with Shapely (yet?).

jorisvandenbossche:
Cython 3 has a very uncertain timeline AFAIK.

Thanks for the update, from looking at the wheels built on CI, this looks to do what it is supposed to do ;)

jorisvandenbossche (approved):

PR #1645, opened 2022-12-01
CLN: remove shapely/examples submodule

jorisvandenbossche:
Those examples are not kept up to date, and AFAIK not included anywhere. We could also move them to the documentation, but in any case I would say they shouldn't be included in the shapely _package_ itself (currently one can do `from shapely.examples import dissolve`).

cc @mwtoews @sgillies


jorisvandenbossche:
I am going to merge this, but if there was something useful in those examples, we can always recover them and move them to the documentation (but in any case I want to avoid we have the submodule in shapely 2.0)

PR #1647, opened 2022-12-03
TST: fix tests for GEOS changes in M handling

jorisvandenbossche:
GEOS is improving its support for M values (it can now store either XYZ or XYZM buffers under the hood, https://github.com/libgeos/geos/pull/721), and for example can (I think) now also parse WKT or WKB with M values. We will need to update our code to actually make use of those abilities(-> opened https://github.com/shapely/shapely/issues/1648 for this), but this PR just fixes a few test failures because of those changes in GEOS.

### DIFF
--- a/shapely/tests/test_io.py
+++ b/shapely/tests/test_io.py
@@ -183,7 +183,13 @@ def test_from_wkb_exceptions():
         shapely.from_wkb(1)
 
     # invalid WKB
-    with pytest.raises(shapely.GEOSException, match="Unexpected EOF parsing WKB"):
+    with pytest.raises(
+        shapely.GEOSException,
+        match=(
+            "Unexpected EOF parsing WKB|"
+            "ParseException: Input buffer is smaller than requested object size"
+        ),
+    ):
         result = shapely.from_wkb(b"\x01\x01\x00\x00\x00\x00")
         assert result is None
 
@@ -287,7 +293,7 @@ def test_to_wkt_exceptions():
         shapely.to_wkt(1)
 
     with pytest.raises(shapely.GEOSException):
-        shapely.to_wkt(point, output_dimension=4)
+        shapely.to_wkt(point, output_dimension=5)
 
 
 def test_to_wkt_point_empty():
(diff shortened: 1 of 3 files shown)


src/ufuncs.c
```diff
@@ -2455,6 +2455,14 @@ static void points_func(char** args, npy_intp* dimensions, npy_intp* steps, void
   GEOSCoordSequence* coord_seq = NULL;
   GEOSGeometry** geom_arr;
 
+  // check the ordinate dimension before calling GEOSCoordSeq_create_r
+  if (dimensions[1] < 2 || dimensions[1] > 3) {
+    PyErr_Format(PyExc_ValueError,
+                 "The ordinate (last) dimension should be 2 or 3, got %ld",
+                 dimensions[1]);
+    return;
```
jorisvandenbossche, line 2463: Currently we are just passing through whatever dimension we get through to GEOS. But `GEOSCoordSeq_create_r` is documented to only accept dimension 2 and 3. And so when now passing dimension 4, we actually get a wrong result instead of an error (with previous versions of GEOS). See https://github.com/libgeos/geos/pull/721#discussion_r1038752015

And even if GEOS would change that to raise an error again, checking this on our side is better I think. It is consistent with other creation ufuncs (eg for `linestrings()` we have this check on our side as well), and it avoids that you can accidentally create points for 1D coords (`shapely.points([1])` currently creates POINT (1 0), so filling in with zero; with this change that will raise an error).

shapely/tests/test_io.py
```diff
@@ -183,7 +183,13 @@ def test_from_wkb_exceptions():
         shapely.from_wkb(1)
 
     # invalid WKB
-    with pytest.raises(shapely.GEOSException, match="Unexpected EOF parsing WKB"):
+    with pytest.raises(
+        shapely.GEOSException,
+        match=(
+            "Unexpected EOF parsing WKB|"
+            "ParseException: Input buffer is smaller than requested object size"
```
jorisvandenbossche, line 190: This is just an error message that changed

shapely/tests/test_io.py
```diff
@@ -287,7 +293,7 @@ def test_to_wkt_exceptions():
         shapely.to_wkt(1)
 
     with pytest.raises(shapely.GEOSException):
-        shapely.to_wkt(point, output_dimension=4)
+        shapely.to_wkt(point, output_dimension=5)
```
jorisvandenbossche, line 296: GEOS now allows to specify an output dimension of 4, for the case your geometry is value 4. We don't yet have ways to create such geometries on our side, but so changing the value here to trigger the same error as before on GEOS side.

jorisvandenbossche:
I want to do a rc2 to test the latest packaging changes, so going to merge this already to ensure we have green CI

Issue #1650, opened 2022-12-03
TST: Intermittent failure in test_remove_repeated_points_invalid_result on Linux


jorisvandenbossche:
The Linux builds for GEOS 3.11 and main started to regularly fail (but not always) in the `test_remove_repeated_points_invalid_result` test (`remove_repeated_points` is new in GEOS 3.11, so those two builds are the only ones that run this test):

```
 =================================== FAILURES ===================================
_____________ test_remove_repeated_points_invalid_result[geom0-2] ______________

geom = <POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0))>, tolerance = 2

    @pytest.mark.skipif(shapely.geos_version < (3, 11, 0), reason="GEOS < 3.11")
    @pytest.mark.parametrize(
        "geom, tolerance", [[Polygon([(0, 0), (1, 0), (1, 1), (0, 1), (0, 0)]), 2]]
    )
    def test_remove_repeated_points_invalid_result(geom, tolerance):
>       with pytest.raises(shapely.GEOSException, match="Invalid number of points"):
            shapely.remove_repeated_points(geom, tolerance)
E       Failed: DID NOT RAISE <class 'shapely.errors.GEOSException'>
```

I can't reproduce it locally.

jorisvandenbossche:
Adding some print debug statements in https://github.com/shapely/shapely/pull/1651 indicates that _if_ it fails to raise the expected error, then it returns the (invalid) polygon `POLYGON ((0 0, 0 0, 0 0))` (which is strange that this could be returned, because normally GEOS should error when trying to construct a LinearRing of 3 points)

I opened an issue about this on the GEOS side: https://github.com/libgeos/geos/issues/759

mwtoews:
One consideration is to use a [flaky plug-in](https://github.com/box/flaky) for [flaky tests](https://docs.pytest.org/en/7.1.x/explanation/flaky.html).

jorisvandenbossche:
To not have red CI on our side, you mean? (so we can ignore it for now) I could also add an xfail strict=False (to not have to add an additional test dependency)

mwtoews:
Yes, either approach to "ignore it for now", since it's an upstream issue.

jorisvandenbossche:
This is already fixed upstream in GEOS main, so I will just skip the test for now for 3.11 and only run it with GEOS main (if 3.11 gets a bugfix release with this included, we could revert that) -> https://github.com/shapely/shapely/pull/1656

PR #1651, opened 2022-12-03
TEMP: debug test_remove_repeated_points_invalid_result

jorisvandenbossche:
Trying to debug https://github.com/shapely/shapely/issues/1650


jorisvandenbossche:
The result it prints when the test fails is `POLYGON ((0 0, 0 0, 0 0))`, which is clearly not a valid polygon, since a linearring requires at least 4 coordinates pairs.

jorisvandenbossche:
Fixing in https://github.com/shapely/shapely/pull/1656

Issue #1653, opened 2022-12-06
Version missing in 2.0rc2


sebastic:
## Expected behavior and actual behavior.

geoalchemy2 fails because the version comparison fails:
```python
if parse_version(shapely.__version__) < parse_version("1.7"):  # pragma: no cover
```

Shapely doesn't provide a usable version:
```
>>> shapely.__version__ 
'0+unknown'
>>> shapely._version.get_versions()
{'date': None, 'dirty': None, 'error': 'unable to compute version', 'full-revisionid': None, 'version': '0+unknown'}
```

`shapely.__version__` should return `2.0rc2`.

`cfg.versionfile_source` in `_version.py` is wrong:
```python
$ grep cfg.versionfile_source shapely/_version.py 
    cfg.versionfile_source = "pygeos/_version.py"
        for i in cfg.versionfile_source.split("/"):
```
`pygeos/_version.py` does not exist in the source tree.

## Steps to reproduce the problem.

On Debian unstable with experimental sources:
```
apt install -t experimental python3-shapely
python3 -c 'import shapely; print(shapely._version.get_versions())'
```

## Operating system

Debian unstable

## Shapely version and provenance

2.0rc2 Debian package.

sebastic:
versioneer doesn't work for package builds, it doesn't provide a way to set the version like setuptools_scm:

 https://github.com/python-versioneer/python-versioneer/labels/fallback

mwtoews:
I was scheming to propose setuptools_scm a few weeks ago (https://github.com/mwtoews/shapely/tree/setuptools-scm), but held back as to accumulate too many changes / delay any further. Is it worthwhile pushing this thru for another rc?

sebastic:
Yes, rc3 should include a fix this issue.

The changes in #1654 may suffice.

The _version.py in the GitHub tarball should be correct then.

jorisvandenbossche:
Yes, certainly will do a rc3 with this fix

jorisvandenbossche:
@sebastic I suppose the reason you run into this is because you use the github created archive for debian packaging? (for example on https://github.com/shapely/shapely/releases/tag/2.0rc2, the "Source code (tar.gz)" download link, not the sdist that was uploaded (first link "shapely-2.0rc2.tar.gz")).

Is the reason for that because there are additional files you need that are not included in the sdist? 
(I would always recommend to start from the sdist if possible, since that is the official release artifact, and also has proper versions without relying on the file name)

[...]

PR #1654, opened 2022-12-06
Fix versioneer configuration

sebastic:
Update versioneer configuration like geopandas for which it does work in the context of package builds.

Fixes: #1653


mwtoews (approved):
Thanks, looks good.

This also reminds me to sift thru `git grep -i pygeos`, so I might have another PR soon.

jorisvandenbossche:
Thanks @sebastic !

PR #1655, opened 2022-12-06
PERF: Fix double check of is_empty in MultiPolygon() constructor (because truthiness implies that)

### LINKED ISSUE #1615: MultiPolygon does some slow checks on construction
It's me again! back with another performance issue.

MultiPolygon does some slow checks on creation (not sure if they are strictly necessary, but that's a different point), and they are done twice!

Ideally, MultiPolygon() should only check `is_empty` once for each polygon passed to it, but in reality because it checks `__bool__` and `is_empty`, and `__bool__` calls `is_empty`, it effectively checks it twice.

I did found this while doing profiling on our code base with py-spy. Let me know if you want me to write down exactly how I did that, but here's an image from the profiling: 
Screen Shot 2022-11-16 at 12 32 56 PM

## Operating system

MacOS

`2.0b2+9.ge027985` installed from source against GEOS 3.11

Lyle-Alloy:
Closes #1615


idanmiara:
https://github.com/shapely/shapely/actions/runs/3630132715/jobs/6126743563#logs
`RuntimeWarning: overflow encountered in remove_repeated_points` 
Seems like a bug in GEOS main ?

Lyle-Alloy:
Yeah I'm really not sure how my change would have caused that test to fail. Any ideas? Should we just rerun and hope they fixed something upstream?

idanmiara:
> Yeah I'm really not sure how my change would have caused that test to fail. Any ideas? Should we just rerun and hope they fixed something upstream?

Sounds like a good idea. Or just merge regardless as this failure obviously has nothing to do with your PR.

Lyle-Alloy:
@mwtoews could you weigh in? I think we need you or another maintainer to kick [this](https://github.com/shapely/shapely/actions/runs/3630132715/jobs/6126743563) job off again

idanmiara:
> @mwtoews could you weigh in? I think we need you or another maintainer to kick [this](https://github.com/shapely/shapely/actions/runs/3630132715/jobs/6126743563) job off again

You could trigger it by rebasing and force pushing:
```
git switch fix_multipolygon_double_is_empty
git remote add upstream https://github.com/shapely/shapely.git
git fetch upstream
git rebase upstream/main
git push --force
```

shapely/geometry/multipolygon.py
```diff
@@ -55,11 +55,7 @@ def __new__(self, polygons=None):
             return polygons
 
         polygons = getattr(polygons, "geoms", polygons)
-        polygons = [
-            p
-            for p in polygons
-            if p and not (isinstance(p, polygon.Polygon) and p.is_empty)
-        ]
+        polygons = [p for p in polygons if p]
```
jorisvandenbossche, line 58: Let's just add a small comment about why we are doing this (now it is less explicit)
```suggestion
        # remove None and empty polygons from list of Polygons
        polygons = [p for p in polygons if p]
```

jorisvandenbossche (approved):
Looks good, and sorry for the very late review!

jorisvandenbossche:
It seems this was actually not really covered, so added a small test.

jorisvandenbossche:
Thanks @Lyle-Alloy!

PR #1657, opened 2022-12-08
Add compat for unpickling shapely<2 geometries

### LINKED ISSUE #1613: Can shapely 2.0 support unpickling of shapely 1 object?
shapely 2.0 can't unpickle objects generated from shapely 1.8. I am guessing this is intentional because shapely 2.0 objects are immutable. But we have many existing pickles that's also difficult to convert to new shapely 2.0 geometry class. Wondering if the library can provide a backward compatible method to load old pickles? 

in a shapely 1.8 environment:
```py
import shapely.geometry
import pickle
shape = shapely.geometry.Polygon([[0,0],[3,3],[2,2]])
bytes = pickle.dumps(shape)
```

in a shapely 2.0 enviroment
```py
import shapely
import pickle
pickle.loads(<the bytes from above>)
```
And it will throw error `_pickle.UnpicklingError: state is not a dictionary`, because `__setstate__` is not implemented. 

## Operating system
Ubuntu 20

2.0b2 installed from PyPI using pip

caspervdw:
Closes #1613

I think it is good to have a temporary compatibility fix for shapely <2.0 pickles.

This in essence makes the `Geometry` subclasses mutable through their `__setstate__`, which I think is OK when only done right after construction. As a temporary measure.


caspervdw:
@zhengwy888  maybe you can test with your nested pickle objects and see if it works for you.

jorisvandenbossche:
Nice! That looks like a good solution for this upgrade issue.

caspervdw:
OK! I’d like to have a test for each geometry type, before merging.
Unittest failure seems unrelated btw.

jorisvandenbossche:
I can also add some additional tests if you didn't start yet

caspervdw:
Yes please! I want to focus on the pygeos py3.11 release (if I can find the time)

jorisvandenbossche:
I updated the test file so you can use it to create pickle files for all geometry types and for different versions of shapely. I then ran this for shapely 1.8.5 (with GEOS 3.10.3), shapely 1.8.0 (with GEOS 3.9.1) and shapely 1.7.1 (with GEOS 3.8.0), all installed using pip wheels (so the GEOS versions are the ones included in those wheels). 

In practice, the pickled bytes turned out to be exactly equal bit-for-bit, except for empty polygon for shapely 1.7.1 (at which point empty polygon was converted to WKB for empty 3D polygon). So I only included the files for 1.8.5, except for empty polygon.

Now, in practice, since 1.7.1 was saving empty polygon as 3D WKB, I don't think we can "recover" that correctly, since we only have the WKB to use. So will probably relax the test in that case, or just remove it. 

---

I did include a fix to correctly unpickle LinearRings (so they come back as LinearRing instead of LineString)

tests/test_pickle.py
```diff
@@ -1,24 +1,93 @@
...
+
+    geom_type = fname.name.split("_")[0]
+    expected = TEST_DATA[geom_type]
+    
+    with open(fname, "rb") as f:
+        with pytest.warns(UserWarning):
+            result = pickle.load(f)
+    
+    if geom_type == "emptypolygon" and "1.7.1" in fname.name:
+        expected = wkt.loads("POLYGON Z EMPTY")
```
jorisvandenbossche, line 75: I could also just leave out this case from the files (so just test with the generated pickle data from 1.8.5)

caspervdw: That would be fine with me! I think these tests are already pretty thorough.
And my test above appears to be obsolete now? Feel free to remove it

jorisvandenbossche: Yes, cleaned up

PR #1665, opened 2022-12-13
DOC: fix default for preserve_topology in simplify docstring

eripe970:
Update docs.


mwtoews (approved):
Thanks for the catch, it was missed from #1392

jorisvandenbossche:
Thanks @eripe970 !

Issue #1662, opened 2022-12-12
REGR: Constucting Polygon from Point list fails


snowman2:
## Expected behavior and actual behavior.

Error message:

```
self = <class 'shapely.geometry.polygon.Polygon'>
shell = array([<POINT (-95 41)>, <POINT (-95 42)>, <POINT (-94 42)>,
       <POINT (-94 41)>], dtype=object)
holes = None
    def __new__(self, shell=None, holes=None):
        if shell is None:
            # empty geometry
            # TODO better way?
            return shapely.from_wkt("POLYGON EMPTY")
        elif isinstance(shell, Polygon):
            # return original objects since geometries are immutable
            return shell
        # else:
        #     geom_shell = LinearRing(shell)
        #     if holes is not None:
        #         geom_holes = [LinearRing(h) for h in holes]
    
        if holes is not None:
            if len(holes) == 0:
                # shapely constructor cannot handle holes=[]
                holes = None
            else:
                holes = [LinearRing(ring) for ring in holes]
    
        if not isinstance(shell, BaseGeometry):
            if not isinstance(shell, (list, np.ndarray)):
                # eg emtpy generator not handled well by np.asarray
                shell = list(shell)
            shell = np.asarray(shell)
    
            if len(shell) == 0:
                # empty geometry
                # TODO better constructor + should shapely.polygons handle this?
                return shapely.from_wkt("POLYGON EMPTY")
    
            if not np.issubdtype(shell.dtype, np.number):
                # conversion of coords to 2D array failed, this might be due
                # to inconsistent coordinate dimensionality
>               raise ValueError("Inconsistent coordinate dimensionality")
E               ValueError: Inconsistent coordinate dimensionality
.../lib/python3.9/site-packages/shapely/geometry/polygon.py:251: ValueError

```
## Steps to reproduce the problem.

This fails:
```python
from shapely.geometry import Point, Polygon

geom = Polygon([Point(-95, 41), Point(-95, 42), Point(-94, 42), Point(-94, 41)])
```

This works:
```python
from shapely.geometry import LinearRing, LineString, Point, Polygon

geom = Polygon([(-95, 41), (-95, 42), (-94, 42), (-94, 41)])
geom = Polygon(LinearRing([Point(-95, 41), Point(-95, 42), Point(-94, 42), Point(-94, 41)]))
geom = Polygon(LineString([(-95, 41), (-95, 42), (-94, 42), (-94, 41)]))
[...]

PR #1671, opened 2022-12-17
REGR: fix Polygon() constructor from sequence of Points

jorisvandenbossche:
Closes https://github.com/shapely/shapely/issues/1662

### DIFF
--- a/shapely/geometry/polygon.py
+++ b/shapely/geometry/polygon.py
@@ -90,7 +90,11 @@ class LinearRing(LineString):
                     else:
                         return o
 
-                coordinates = [_coords(o) for o in coordinates]
+                coordinates = np.array([_coords(o) for o in coordinates])
+                if not np.issubdtype(coordinates.dtype, np.number):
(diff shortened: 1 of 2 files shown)


shapely/geometry/polygon.py
```diff
@@ -234,24 +236,6 @@ def __new__(self, shell=None, holes=None):
...
-                # empty geometry
-                # TODO better constructor + should shapely.polygons handle this?
-                return shapely.from_wkt("POLYGON EMPTY")
-
-            if not np.issubdtype(shell.dtype, np.number):
-                # conversion of coords to 2D array failed, this might be due
-                # to inconsistent coordinate dimensionality
-                raise ValueError("Inconsistent coordinate dimensionality")
-        elif not isinstance(shell, LinearRing):
-            shell = LinearRing(shell)
```
jorisvandenbossche, line 253: With the small addition in `LinearRing` (to raise about the inconsistent dimensionality), this is now all handled by that constructor, so we can just call `shell = LinearRing(shell)` and simplify the code here.

PR #1673, opened 2022-12-17
DOC: `snap`: merge documentation and examples from `ops` and `constructive`

idanmiara:
shapely/ops.py, shapely/constructive.py - `snap`: merge documentation and examples and examples and add some more details from `libgeos` docs.

### DIFF
--- a/shapely/constructive.py
+++ b/shapely/constructive.py
@@ -914,7 +914,7 @@ def snap(geometry, reference, tolerance, **kwargs):
 
     Examples
     --------
-    >>> from shapely import snap, distance, LineString, Point, Polygon, MultiPolygon, MultiPoint, box
+    >>> from shapely import snap, distance, LineString, Point, Polygon, MultiPoint, box
 
     >>> point = Point(0.5, 2.5)
     >>> target_point = Point(0, 2)
--- a/shapely/constructive.py
+++ b/shapely/constructive.py
@@ -887,9 +887,21 @@ def simplify(geometry, tolerance, preserve_topology=True, **kwargs):
 def snap(geometry, reference, tolerance, **kwargs):
     """Snaps an input geometry to reference geometry's vertices.
 
-    The tolerance is used to control where snapping is performed.
+    Vertices of the first geometry are snapped to vertices of the second.
+    geometry, returning a new geometry; the input geometries are not modified.
     The result geometry is the input geometry with the vertices snapped.
     If no snapping occurs then the input geometry is returned unchanged.
+    The tolerance is used to control where snapping is performed.
+
+    Snapping one geometry to another can improve robustness for overlay
+    operations by eliminating nearly-coincident edges
+    (which cause problems during noding and intersection calculation).
+
+    Too much snapping can result in invalid topology being created,
+    so the number and location of snapped vertices is decided using
+    heuristics to determine when it is safe to snap.
+    This can result in some potential snaps being omitted, however,
+    if they would lead to an invalid geometry.
 
     Parameters
     ----------
@@ -902,18 +914,54 @@ def snap(geometry, reference, tolerance, **kwargs):
 
     Examples
     --------
-    >>> from shapely import LineString, Point, Polygon
+    >>> from shapely import snap, distance, LineString, Point, Polygon, MultiPolygon, MultiPoint, box
+
     >>> point = Point(0.5, 2.5)
     >>> target_point = Point(0, 2)
     >>> snap(point, target_point, tolerance=1)
     <POINT (0 2)>
     >>> snap(point, target_point, tolerance=0.49)
     <POINT (0.5 2.5)>
+
     >>> polygon = Polygon([(0, 0), (0, 10), (10, 10), (10, 0), (0, 0)])
     >>> snap(polygon, Point(8, 10), tolerance=5)
     <POLYGON ((0 0, 0 10, 8 10, 10 0, 0 0))>
     >>> snap(polygon, LineString([(8, 10), (8, 0)]), tolerance=5)
     <POLYGON ((0 0, 0 10, 8 10, 8 0, 0 0))>
+
+    You can snap one line to another, for example to clean imprecise coordinates:
+
+    >>> line1 = LineString([(0.1, 0.1), (0.49, 0.51), (1.01, 0.89)])
+    >>> line2 = LineString([(0, 0), (0.5, 0.5), (1.0, 1.0)])
+    >>> snap(line1, line2, 0.25)
+    <LINESTRING (0 0, 0.5 0.5, 1 1)>
+
+    Snapping also supports Z coordinates:
+
+    >>> point1 = Point(0.1, 0.1, 0.5)
+    >>> multipoint = MultiPoint([(0, 0, 1), (0, 0, 0)])
+    >>> snap(point1, multipoint, 1)
+    <POINT Z (0 0 1)>
+
+    Snapping to an empty geometry has no effect:
+
+    >>> snap(line1, LineString([]), 0.25)
+    <LINESTRING (0.1 0.1, 0.49 0.51, 1.01 0.89)>
+
+    Snapping to a non-geometry (None) will always return None:
+
+    >>> snap(line1, None, 0.25) is None
+    True
+
+    Only one vertex of a polygon is snapped to a target point,
+    even if all vertices are equidistant to it,
+    in order to prevent collapse of the polygon:
+
+    >>> poly = box(0, 0, 1, 1)
+    >>> poly
+    <POLYGON ((1 0, 1 1, 0 1, 0 0, 1 0))>
+    >>> snap(poly, Point(0.5, 0.5), 1)
+    <POLYGON ((0.5 0.5, 1 1, 0 1, 0 0, 0.5 0.5))>
     """
     return lib.snap(geometry, reference, tolerance, **kwargs)
 
--- a/shapely/constructive.py
+++ b/shapely/constructive.py
@@ -897,7 +897,7 @@ def snap(geometry, reference, tolerance, **kwargs):
     however, it does not guarantee that output geometries will be valid.  It is
     the responsibility of the caller to check for and handle invalid geometries.
 
-    Because too much snapping can result in invalid  being created,
+    Because too much snapping can result in invalid geometries being created,
     heuristics are used to determine the number and location of snapped
     vertices that are likely safe to snap. These heuristics may omit
     some potential snaps that are otherwise within the tolerance.
--- a/shapely/constructive.py
+++ b/shapely/constructive.py
@@ -893,15 +893,14 @@ def snap(geometry, reference, tolerance, **kwargs):
     If no snapping occurs then the input geometry is returned unchanged.
     The tolerance is used to control where snapping is performed.
 
-    Snapping one geometry to another can improve robustness for overlay
-    operations by eliminating nearly-coincident edges
-    (which cause problems during noding and intersection calculation).
+    Where possible, this operation tries to avoid creating invalid geometries;
+    however, it does not guarantee that output geometries will be valid.  It is
+    the responsibility of the caller to check for and handle invalid geometries.
 
-    Too much snapping can result in invalid topology being created,
-    so the number and location of snapped vertices is decided using
-    heuristics to determine when it is safe to snap.
-    This can result in some potential snaps being omitted, however,
-    if they would lead to an invalid geometry.
+    Because too much snapping can result in invalid  being created,
+    heuristics are used to determine the number and location of snapped
+    vertices that are likely safe to snap. These heuristics may omit
+    some potential snaps that are otherwise within the tolerance.
 
     Parameters
     ----------
--- a/shapely/ops.py
+++ b/shapely/ops.py
@@ -317,11 +317,8 @@ def nearest_points(g1, g2):
 
 
 def snap(g1, g2, tolerance):
-    """Snap one geometry to another with a given tolerance
-
-    Vertices of the first geometry are snapped to vertices of the second
-    geometry. The resulting snapped geometry is returned. The input geometries
-    are not modified.
+    """
+    Snaps an input geometry (g1) to reference (g2) geometry's vertices.
 
     Parameters
     ----------
@@ -332,14 +329,9 @@ def snap(g1, g2, tolerance):
     tolerance : float
         The snapping tolerance
 
-    Example
-    -------
-    >>> square = Polygon([(1,1), (2, 1), (2, 2), (1, 2), (1, 1)])
-    >>> line = LineString([(0,0), (0.8, 0.8), (1.8, 0.95), (2.6, 0.5)])
-    >>> result = snap(line, square, 0.5)
-    >>> result.wkt
-    'LINESTRING (0 0, 1 1, 2 1, 2.6 0.5)'
+    Refer to :func:`shapely.snap` for full documentation.
     """
+
     return shapely.snap(g1, g2, tolerance)
 
 


shapely/constructive.py
```diff
@@ -887,10 +887,23 @@ def simplify(geometry, tolerance, preserve_topology=True, **kwargs):
...
     """Snaps an input geometry to reference geometry's vertices.
 
-    The tolerance is used to control where snapping is performed.
+    Vertices of the first geometry are snapped to vertices of the second.
+    geometry. The input geometries are not modified.
     The result geometry is the input geometry with the vertices snapped.
     If no snapping occurs then the input geometry is returned unchanged.
 
+    The tolerance is used to control where snapping is performed.
+    A snap distance tolerance is used to control where snapping is performed.
```
jorisvandenbossche, line 896: Those two sentences are a bit duplicative?

shapely/constructive.py
```diff
@@ -914,6 +927,11 @@ def snap(geometry, reference, tolerance, **kwargs):
     <POLYGON ((0 0, 0 10, 8 10, 10 0, 0 0))>
     >>> snap(polygon, LineString([(8, 10), (8, 0)]), tolerance=5)
     <POLYGON ((0 0, 0 10, 8 10, 8 0, 0 0))>
+    >>> square = Polygon([(1,1), (2, 1), (2, 2), (1, 2), (1, 1)])
+    >>> line = LineString([(0,0), (0.8, 0.8), (1.8, 0.95), (2.6, 0.5)])
+    >>> result = snap(line, square, 0.5)
+    >>> result.wkt
+    'LINESTRING (0 0, 1 1, 2 1, 2.6 0.5)'
```
jorisvandenbossche, line 934: I know you copied this from the existing docstrings, but do we think this one adds value to the other examples that are already present? 
If not, I think we can also just leave it out.

shapely/constructive.py
```diff
@@ -887,10 +887,23 @@ def simplify(geometry, tolerance, preserve_topology=True, **kwargs):
...
+    A snap distance tolerance is used to control where snapping is performed.
+
+    Snapping one geometry to another can improve robustness for overlay
+    operations by eliminating nearly-coincident edges
+    (which cause problems during noding and intersection calculation).
+
+    Too much snapping can result in invalid topology being created,
+    so the number and location of snapped vertices is decided using
+    heuristics to determine when it is safe to snap.
+    This can result in some potential snaps being omitted, however.
```
jorisvandenbossche, line 905: Does this mean that snapping will not always happen, even when within the tolerance distance, to avoid invalid geometries? That's something the underlying algorithm does automatically?

jorisvandenbossche (commented):
Thanks a lot! (there are indeed quite some docstrings that can be consolidated, now we added the new top-level functions)

shapely/constructive.py
idanmiara: This is copy pasted from geos docs:
https://libgeos.org/doxygen/classgeos_1_1operation_1_1overlay_1_1snap_1_1GeometrySnapper.html#details
I'm not sure I fully understand the logic of the tolerance. I couldn't find a better documentation of what exactly it means.

shapely/constructive.py
idanmiara: oh right, I'll remove that

shapely/constructive.py
idanmiara: Maybe these are more helpful examples to add instead:
https://postgis.net/docs/ST_Snap.html

idanmiara: how about now?

idanmiara:
Hi @jorisvandenbossche
Is this PR ready ?

idanmiara:
@jorisvandenbossche do we have more actionable items for this PR?

shapely/constructive.py
brendan-ward: This docstring originally came from [JTS](https://locationtech.github.io/jts/javadoc/org/locationtech/jts/operation/overlay/snap/GeometrySnapper.html) 

My understanding is that the snapping algorithm will avoid using potential snap targets if they would lead to an invalid geometry.

shapely/ops.py
```diff
@@ -319,29 +319,12 @@ def nearest_points(g1, g2):
...
+    Snaps an input geometry (g1) to reference (g2) geometry's vertices.
 
-    Example
-    -------
-    >>> square = Polygon([(1,1), (2, 1), (2, 2), (1, 2), (1, 1)])
-    >>> line = LineString([(0,0), (0.8, 0.8), (1.8, 0.95), (2.6, 0.5)])
-    >>> result = snap(line, square, 0.5)
-    >>> result.wkt
-    'LINESTRING (0 0, 1 1, 2 1, 2.6 0.5)'
+    For full details, see documentation of: shapely.snap()
```
brendan-ward, line 325:
```suggestion
    Refer to `shapely.snap` for full documentation.
```

(small precedent for phrasing set in `geometry.base.py`)

shapely/ops.py
```diff
@@ -319,29 +319,12 @@ def nearest_points(g1, g2):
...
-    Parameters
-    ----------
-    g1 : geometry
-        The first geometry
-    g2 : geometry
-        The second geometry
-    tolerance : float
-        The snapping tolerance
+    """
+    Snaps an input geometry (g1) to reference (g2) geometry's vertices.
```
brendan-ward, line 323: Suggest changing the parameters to `geometry`, `reference` to match the underlying function instead of `g1` and g2`.

Should probably still retain the parameters part of the docstring, but the more verbose description and examples can be dropped in favor of `shapely.snap`.

shapely/constructive.py
```diff
@@ -914,6 +925,21 @@ def snap(geometry, reference, tolerance, **kwargs):
     <POLYGON ((0 0, 0 10, 8 10, 10 0, 0 0))>
     >>> snap(polygon, LineString([(8, 10), (8, 0)]), tolerance=5)
     <POLYGON ((0 0, 0 10, 8 10, 8 0, 0 0))>
+    >>> # example from: https://postgis.net/docs/ST_Snap.html
```
brendan-ward, line 928: What is the intent of the additional examples here?  I'm finding the MultiPolygon examples here a bit hard to follow and not clarifying behavior, which means we should seek out simpler examples unless there is a point you are trying to illustrate regarding MultiPolygons specifically.

I suggest dropping the MultiPolygon and LineString examples added here, and add a simpler example of snapping a line to line (demonstrates cleaning imprecise coordinates): 

```Python
>>> line1 = LineString([(0.1, 0.1), (0.49, 0.51), (1.01, 0.89)])
>>> line2 = LineString([(0, 0), (0.5, 0.5), (1.0, 1.0)])
>>> snap(line1, line2, 0.25)
<LINESTRING (0 0, 0.5 0.5, 1 1)>
```

We also occasionally like to demonstrate behavior for `None` or empty geometries if they lead to unexpected behavior.

For example, I'd suggest adding:

```Python
>>> snap(line1, None, 0.25) is None
True
```

Because without noting it here, a user might instead expect to get back the input geometry (snap to nothing => no snapping).


Likewise, though perhaps more obvious, is behavior for an empty geometry:

```Python
>>> snap(line1, LineString([]), 0.25)
<LINESTRING (0.1 0.1, 0.49 0.51, 1.01 0.89)>
```

Might also be good to demonstrate a very simple 3D example:

```Python
>>> point1 = Point(0.1, 0.1, 0.5)
>>> multipoint = MultiPoint([(0, 0, 1), (0, 0, 0)])
>>> snap(point1, multipoint, 1)
<POINT Z (0 0 1)>
```

I'm not sure if this is deterministic across GEOS versions, for me it picks the first point that is within tolerance.

This might be now getting a bit carried away, but here is an example that shows how snapping only updates one of the coordinates in a Polygon even though all points in the polygon are equidistant, because this would cause the polygon to collapse:

```Python
>>> poly = box(0, 0, 1, 1)
>>> snap(poly, Point(0.5, 0.5), 1)
<POLYGON ((0.5 0.5, 1 1, 0 1, 0 0, 0.5 0.5))>
```



I also discovered an unexpected promotion of dimension that we should probably track down (don't use as example, just demonstrating it here):

```Python
>>> snap(LineString([]), LineString([]), 0.25)
<LINESTRING Z EMPTY>
```

brendan-ward (commented):
Thanks for working on this @idanmiara .

A few comments to consider.  

I'm also a bit torn between wanting to retain the original docstring text from GEOS / JTS versus trying to rewrite it a bit for clarity.  Will ponder that a bit.

shapely/ops.py
idanmiara: Changing parameter names would be a breaking change for users who used kwargs.

shapely/constructive.py
idanmiara: @brendan-ward Thanks very much. I've added your examples.

> I suggest dropping the MultiPolygon and LineString examples added here
I kept the `MultiPolygon` example because there is nice graphics for it in JTS.

>I also discovered an unexpected promotion of dimension that we should probably track down (don't use as example, just demonstrating it here):
https://github.com/shapely/shapely/issues/1686

idanmiara:
@jorisvandenbossche @brendan-ward 
I've consolidated the examples, added the examples from JTS (because they have neat graphics that explains them) and @brendan-ward examples.
What do you think?

shapely/ops.py
brendan-ward: These are required positional parameters; hopefully the risk of changing the parameter names is very low in this case?  i.e., it would be weird to call it like

```
shapely.ops.snap(tolerance=12, g2=Point(0, 1), g1=Point(10, 10)
```


I agree that changing optional named parameters is a breaking change; I'm not sure if we are as strict about simpler cases of required positional parameters (can't remember a recent precedent either way)

shapely/constructive.py
brendan-ward: Thanks for the updates!

I still think including the PostGIS example is too much for the examples here; they are not easy to follow without the graphic.  We definitely need more illustrative graphics, but that is out of scope here.

Instead, one option would be to put the link in the descriptive part of the docstring instead, and drop the multipolygon example (I'm not keen on this option though).

```Python
... if they would lead to an invalid geometry.

Also see https://postgis.net/docs/ST_Snap.html for additional examples.

Parameters
-----------
```

However, it should be noted that there does not appear to be an existing precedent for linking to PostGIS for examples, and the examples there are in SQL wrapped WKT, and may not be as easily accessible to Shapely users (the visuals yes, the SQL less so).

At the moment, I think it is just fine to leave out both the example and the crosslink to a different toolkit, and leave it as a follow-up to generate graphics in the docs via `shapely.plotting` rather than in the examples here.  I think the examples should be easy to follow just from looking at the code, and we have some work to do to simplify them further across the board (also out of scope here).

shapely/constructive.py
```diff
@@ -902,18 +914,63 @@ def snap(geometry, reference, tolerance, **kwargs):
...
     >>> snap(point, target_point, tolerance=0.49)
     <POINT (0.5 2.5)>
+
     >>> polygon = Polygon([(0, 0), (0, 10), (10, 10), (10, 0), (0, 0)])
     >>> snap(polygon, Point(8, 10), tolerance=5)
     <POLYGON ((0 0, 0 10, 8 10, 10 0, 0 0))>
     >>> snap(polygon, LineString([(8, 10), (8, 0)]), tolerance=5)
     <POLYGON ((0 0, 0 10, 8 10, 8 0, 0 0))>
+
+    # snapping a line to line (demonstrates cleaning imprecise coordinates)
```
brendan-ward, line 932:
```suggestion
    You can snap one line to another, for example to clean imprecise coordinates:
    
```

You can just use text, so long as it is separated by newlines.  See [examples in `strtree.py`](https://github.com/shapely/shapely/blob/2.0.0/shapely/strtree.py#L168) for instance

Then the text blocks can be more descriptive of the behavior that follows (sorry, mine didn't do this well when I added them in review comments previously)

e.g.,

```
You can snap all 3 coordinates in 3D geometries:

>>> ...
```

```
Snapping to an empty geometry has no effect:

>>> ...
```

etc.

shapely/constructive.py
```diff
@@ -887,9 +887,21 @@ def simplify(geometry, tolerance, preserve_topology=True, **kwargs):
 def snap(geometry, reference, tolerance, **kwargs):
     """Snaps an input geometry to reference geometry's vertices.
 
-    The tolerance is used to control where snapping is performed.
+    Vertices of the first geometry are snapped to vertices of the second.
+    geometry. The input geometries are not modified.
```
brendan-ward, line 891:
```suggestion
    geometry, returning a new geometry; the input geometries are not modified.
```

brendan-ward:
In trying to understand how this works for multigeometries, I think I found some cases where GEOS isn't behaving as I'd expect given the docstring.  I need to investigate a bit more, but if these are expected behavior we may want to note these issues in the docstring / examples (just fine for follow-up PR)

```Python
from shapely import MultiPolygon, LineString, is_valid

>>> boxes = MultiPolygon([box(0, 0, 1, 1), box(3, 0, 4, 1)])
>>> boxes
<MULTIPOLYGON (((1 0, 1 1, 0 1, 0 0, 1 0)), ((4 0, 4 1, 3 1, 3 0, 4 0)))>
>>> snapped = snap(boxes, LineString([(2, 0), (2, 1)]), tolerance=2)
>>> snapped
<MULTIPOLYGON (((2 1, 1 1, 0 1, 0 0, 2 0, 2 1)), ((4 0, 4 1, 2 1, 2 0, 4 0)))>
>>> shapely.is_valid(snapped)
False
```

Edges of polygons in a MultiPolygon are not allowed to touch; both Polygons snapped to the LineString here, which isn't valid.

shapely/constructive.py
idanmiara: @brendan-ward thanks for your review!
Regarding the JTS/PostGIS examples:

> the visuals yes, the SQL less so

I also think that the visuals help and the SQL is hard to follow for a shapely user, that's why I've translated the examples from SQL to shapely, so a shapely user can follow the graphics with the shapely examples.
Although I think they are helpful - if you are not inclined to include them I don't mind removing them or leaving this discussion for a follow-up PR as I agree this is already out of scope of this PR.

shapely/ops.py
idanmiara: @brendan-ward I'm marking this as resolved and will open a follow-up issue to discuss parameter renaming as I feel this is out of scope of this PR.

shapely/ops.py
```diff
@@ -332,14 +329,9 @@ def snap(g1, g2, tolerance):
...
         The snapping tolerance
 
-    Example
-    -------
-    >>> square = Polygon([(1,1), (2, 1), (2, 2), (1, 2), (1, 1)])
-    >>> line = LineString([(0,0), (0.8, 0.8), (1.8, 0.95), (2.6, 0.5)])
-    >>> result = snap(line, square, 0.5)
-    >>> result.wkt
-    'LINESTRING (0 0, 1 1, 2 1, 2.6 0.5)'
+    Refer to `shapely.snap` for full documentation.
```
mwtoews, line 332: the usual way to cross-reference another function is to do this:
```suggestion
    Refer to :func:`shapely.snap` for full documentation.
```

jorisvandenbossche:
> Edges of polygons in a MultiPolygon are not allowed to touch; both Polygons snapped to the LineString here, which isn't valid.

Something to report to GEOS? (I can imagine that right now the snapping is done polygon by polygon in a collection)

shapely/constructive.py
jorisvandenbossche: @brendan-ward JTS is BSD-style licensed, so if this text also is present there, I don't think we have to worry too much about licensing

brendan-ward:
> This is something to report to GEOS?

Likely yes, need to investigate more and come up with more test cases.  But that shouldn't hold up this PR, I just wanted to make sure the docstring wasn't misleading about validity.

shapely/constructive.py
```diff
@@ -902,18 +914,47 @@ def snap(geometry, reference, tolerance, **kwargs):
...
     <POLYGON ((0 0, 0 10, 8 10, 8 0, 0 0))>
+
+    You can snap one line to another, for example to clean imprecise coordinates:
+
+    >>> line1 = LineString([(0.1, 0.1), (0.49, 0.51), (1.01, 0.89)])
+    >>> line2 = LineString([(0, 0), (0.5, 0.5), (1.0, 1.0)])
+    >>> snap(line1, line2, 0.25)
+    <LINESTRING (0 0, 0.5 0.5, 1 1)>
+
+    # XYZ geometry:
```
brendan-ward, line 939:
```suggestion
    Snapping also supports Z coordinates:
    
```

shapely/constructive.py
```diff
@@ -902,18 +914,47 @@ def snap(geometry, reference, tolerance, **kwargs):
...
+    >>> point1 = Point(0.1, 0.1, 0.5)
+    >>> multipoint = MultiPoint([(0, 0, 1), (0, 0, 0)])
+    >>> snap(point1, multipoint, 1)
+    <POINT Z (0 0 1)>
+
+    # empty geometry:
```
brendan-ward, line 945:
```suggestion
    Snapping to an empty geometry has no effect:
    
```

shapely/constructive.py
```diff
@@ -902,18 +914,47 @@ def snap(geometry, reference, tolerance, **kwargs):
...
+    >>> snap(line1, LineString([]), 0.25)
+    <LINESTRING (0.1 0.1, 0.49 0.51, 1.01 0.89)>
+
+    >>> snap(line1, None, 0.25) is None
```
brendan-ward, line 949:
```suggestion
    Snapping to a non-geometry (None) will always return None:
    
    >>> snap(line1, None, 0.25) is None
```

shapely/constructive.py
```diff
@@ -902,18 +914,47 @@ def snap(geometry, reference, tolerance, **kwargs):
...
+    True
+
+    # example that shows how snapping only updates one of the coordinates in a Polygon
```
brendan-ward, line 952:
```suggestion
    Only one vertex of a polygon is snapped to a target point, even if all vertices are equidistant to it, in order to prevent collapse of the polygon:

    >>> poly = box(0, 0, 1, 1)
    >>> poly
    <POLYGON ((1 0, 1 1, 0 1, 0 0, 1 0))>
    >>> snap(poly, Point(0.5, 0.5), 1)
    <POLYGON ((0.5 0.5, 1 1, 0 1, 0 0, 0.5 0.5))>
```

However, this might be an awkward example to communicate, so it may be OK to leave it out.

brendan-ward (commented):
@idanmiara thanks for the updates here.  A few more minor changes would help, and then I think this is ready.

Let's leave mention of issues related to snapping multi geometries out of this for now; I'll raise those on the GEOS side.

jorisvandenbossche:
@idanmiara as a general workflow comment: could you try to add _new_ commits when making changes instead of squashing them into the previous one? That makes it a lot easier to see what has changed since a previous review (and making reviewing ieasier). We squash on merge anyway, so it doesn't matter how many commits there are on the PR's branch.

idanmiara:
> @idanmiara thanks for the updates here. A few more minor changes would help, and then I think this is ready.
> 
> Let's leave mention of issues related to snapping multi geometries out of this for now; I'll raise those on the GEOS side.

@brendan-ward thanks! I think this is ready now, please approve.

shapely/constructive.py
```diff
@@ -887,9 +887,21 @@ def simplify(geometry, tolerance, preserve_topology=True, **kwargs):
...
-    The tolerance is used to control where snapping is performed.
+    Vertices of the first geometry are snapped to vertices of the second.
+    geometry, returning a new geometry; the input geometries are not modified.
     The result geometry is the input geometry with the vertices snapped.
     If no snapping occurs then the input geometry is returned unchanged.
+    The tolerance is used to control where snapping is performed.
+
+    Snapping one geometry to another can improve robustness for overlay
+    operations by eliminating nearly-coincident edges
+    (which cause problems during noding and intersection calculation).
```
jorisvandenbossche, line 898: @brendan-ward this paragraph has been removed on the GEOS side (after your PR), so remove it here as well?

brendan-ward: Yes; was waiting for that to get updated before following up here.  I think lines 896 - 904 should be replaced with:

```
Where possible, this operation tries to avoid creating invalid geometries;
however, it does not guarantee that output geometries will be valid.  It is
the responsibility of the caller to check for and handle invalid geometries.

Because too much snapping can result in invalid
topology being created, heuristics are used to determine the number and
location of snapped vertices that are likely safe to snap. These heuristics
may omit some potential snaps that are otherwise within the tolerance.
```

shapely/constructive.py
```diff
@@ -887,9 +887,20 @@ def simplify(geometry, tolerance, preserve_topology=True, **kwargs):
...
+    geometry, returning a new geometry; the input geometries are not modified.
     The result geometry is the input geometry with the vertices snapped.
     If no snapping occurs then the input geometry is returned unchanged.
+    The tolerance is used to control where snapping is performed.
+
+    Where possible, this operation tries to avoid creating invalid geometries;
+    however, it does not guarantee that output geometries will be valid.  It is
+    the responsibility of the caller to check for and handle invalid geometries.
+
+    Because too much snapping can result in invalid  being created,
```
brendan-ward, line 900:
```suggestion
    Because too much snapping can result in invalid geometries being created,
```

brendan-ward (approved):
Thanks for your patience working this out @idanmiara 

Approved contingent on typo fix (will be away for a few days and don't want to hold this up further)

idanmiara:
> Thanks for your patience working this out @idanmiara

Thanks for your thorough review!

idanmiara:
@jorisvandenbossche I think this PR might be finally ready 😄

jorisvandenbossche:
Thanks @idanmiara and @brendan-ward !

Issue #1668, opened 2022-12-14
Make `transform` twice as fast for `pyproj.transforer.transform`-like functions


idanmiara:
## Expected behavior and actual behavior.

I am trying to optimize geometry transformations. 
In a few tests I ran, I could get ±x2.2 speedup for transform functions like `pyproj.transforer.transform` using the code below.
Before I make a PR I wanted to hear your thoughts.

## Steps to reproduce:

add this code to `shapely.ops`
```
def transform_coords_inline(func, coords):
    xxyy = np.array(coords).transpose().copy()
    func(*xxyy, inplace=True)
    return xxyy.transpose()

# new part is following the first `try`. other parts are copied from `transform`
def transform_fast(func, geom):
    if geom.is_empty:
        return geom
    if geom.geom_type in ("Point", "LineString", "LinearRing", "Polygon"):

        # First we try to apply func to x, y, z sequences. 
        # When func is optimized for inplace numpy arrays 
        # (i.e. transform: pyproj.transformer.transform)
        try:
            if geom.geom_type in ("Point", "LineString", "LinearRing"):
                return type(geom)(transform_coords_inline(func, geom.coords))
            elif geom.geom_type == "Polygon":
                shell = type(geom.exterior)(transform_coords_inline(func, geom.exterior.coords))
                holes = list(
                    type(ring)(transform_coords_inline(ring.coords))
                    for ring in geom.interiors
                )
                return type(geom)(shell, holes)
        
        # We try to apply func to x, y, z sequences. When func is
        # optimized for sequences, this is the fastest, though zipping
        # the results up to go back into the geometry constructors adds
        # extra cost.
        except TypeError:
            try:
                if geom.geom_type in ("Point", "LineString", "LinearRing"):
                    return type(geom)(zip(*func(*zip(*geom.coords))))
                elif geom.geom_type == "Polygon":
                    shell = type(geom.exterior)(zip(*func(*zip(*geom.exterior.coords))))
                    holes = list(
                        type(ring)(zip(*func(*zip(*ring.coords))))
                        for ring in geom.interiors
                    )
                    return type(geom)(shell, holes)

            # A func that assumes x, y, z are single values will likely raise a
            # TypeError, in which case we'll try again.
            except TypeError:
                if geom.geom_type in ("Point", "LineString", "LinearRing"):
                    return type(geom)([func(*c) for c in geom.coords])
                elif geom.geom_type == "Polygon":
[...]

PR #1676, opened 2022-12-18
ENH:consolidate `shapely.ops.transform` (+speedup) with `shapely.transform`

idanmiara:
Closes https://github.com/shapely/shapely/issues/1668

### DIFF
--- a/shapely/coordinates.py
+++ b/shapely/coordinates.py
@@ -1,11 +1,26 @@
+from typing import Optional
+
 import numpy as np
 
+import shapely
 from shapely import lib
+from shapely.lib import Geometry
 
-__all__ = ["transform", "count_coordinates", "get_coordinates", "set_coordinates"]
+__all__ = [
+    "transform",
+    "transform_extra",
+    "transform_planar",
+    "transform_rebuild_planar",
+    "transform_rebuild",
+    "count_coordinates",
+    "get_coordinates",
+    "set_coordinates",
+]
+
+from shapely.errors import GeometryTypeError
 
 
-def transform(geometry, transformation, include_z=False):
+def transform(geometry, transformation, include_z: bool = False):
     """Returns a copy of a geometry array with a function applied to its
     coordinates.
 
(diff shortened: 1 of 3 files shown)


idanmiara:
all tests pass both with `shapely.ops.transform` and `shapely.ops.transform2`. 
I couldn't find a better name for `shapely.ops.transform2` (or shapely.transform2_coords`) and I'm sure if there is a point of keeping it anyway, as I think it's only advantage is that is a bit faster on some cases I tested with transformation functions that accepted only scalars (see https://github.com/shapely/shapely/issues/1668)
Maybe keep it so users can try both and find out if there are actually any other advantages to it?

Also - I do think that the situation that `shapely.ops.transform` == `shapely.ops.transform_coords` != `shapely.transform` could be misleading, but otherwise we would break backwards compatibility (maybe rename `shapely.transform` as this is a new for shapely users? but then it's not new for `pygeos` users...), not sure what's better here.

idanmiara:
hmm.. it seems that `Point(0, 1, np.nan)` in `GEOS` `main` now returns `Point(0, 1, nan)`, and before it returned `Point(0, 1)`.

idanmiara:
Hi, 
This PR ready for review.
I'll squash and cleanup commit history when it's approved.
Thanks!

shapely/coordinates.py
```diff
@@ -1,30 +1,76 @@
...
+    include_z : bool or int, default False
+        If True | 1, include the third dimension in the coordinates array
+            that is passed to the ``transformation`` function. If a
+            geometry has no third dimension, the z-coordinates passed to the
+            function will be NaN.
+        If False | 0, The third-dimension of the geometries (if exist),
+            would not be included in the input to transform function.
+        If -1 Include the third dimension in the coordinates array that is passed to the
+            ``transformation`` function only if the geoemtries are three-dimensional.
+    single_vec: bool, default True
```
jorisvandenbossche, line 61: In https://github.com/shapely/shapely/pull/1559, I mentioned `interleaved=True/False` as an optional name (it wasn't included in the PR, only mentioned as possible future enhancement). 

That_might_ be a more descriptive name for the keyword (although `single_vec` is more explicit about the _effect_ of using interleaved coordinates or not, i.e. whether the coordinates are passed as a single array or not)

shapely/coordinates.py
```diff
@@ -1,30 +1,76 @@
...
+            (2D array in case of multiple coordinates)
+        Otherwise, Pass them as separate x, y(, z) arguments
+            (separate 1D arrays in case of multiple coordinates)
+    rebuild: bool, default False
+        If False, output geometries will preserve the coordinate number and
+            dimnensionality as the input geometries (but in include_z=False,
+            two-dimension geometries are always returned).
+        Otherwise, output geometries will get rebuilt from the transformation
+            output, so they might have different number of coordinate or
+            a different coordiante-dimensionality.
```
jorisvandenbossche, line 73: General note on formatting: this way of indenting paragraphs doesn't work for restructuredtext (it would be interpreted as "definition list" (https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#lists-and-quote-like-blocks), but that would render incorrectly)

shapely/coordinates.py
```diff
@@ -44,7 +90,55 @@ def transform(geometry, transformation, include_z=False):
...
+        transform_wrapper = _transform2 if single_vec else _transform2_multi_vec
+        vectorize = not isinstance(geometry, Geometry)
+    else:
+        transform_wrapper = _transform if single_vec else _transform_multi_vec
+        vectorize = False
+        if not np.isscalar(include_z):
+            if np.all(include_z == include_z[0]):
+                include_z = include_z[0]
+            else:
+                vectorize = True
```
jorisvandenbossche, line 130: Alternatively, instead of using np.frompyfunc (to effectively apply the function on each geometry separately, right?), another option is to do what we did in geopandas: https://github.com/geopandas/geopandas/blob/f17b5350b8fa8acb8926e78ee011f876c7a03ff4/geopandas/_vectorized.py#L1099-L1117

Subsetting the input array in 2D and 3D geometries, and then transforming each of those subsets in a single go still using the natively vectorized transform function (without doing the vectorize on the python level with np.frompyfunc)

(might be useful to compare both approaches for performance)

jorisvandenbossche: Another option is also what we did in https://github.com/shapely/shapely/pull/1559, which is to infer `include_z` from the geometries (eg set it to True once there it at least one 3D geometry in the input), but then still use a single value for all geometries. 

For the `to_ragged_array` added in https://github.com/shapely/shapely/pull/1559 that is the only option (since the goal of that function is to return a single array). But of course here when it comes to transforming, you can easily have a transform function that cannot handle the NaNs very well (for example, in geopandas we ran into that with pyproj.Transformer.transform). So for the `shapely.transform` function, it probably _is_ useful to be able to have `include_z` per geometry.

shapely/coordinates.py
```diff
@@ -54,20 +148,117 @@ def transform(geometry, transformation, include_z=False):
...
+    try:
+        # First we try to apply func to x, y, z vectors.
+        return _transform(
+            geometry,
+            lambda coords: np.array(transformation(*coords.T)).T,
+            include_z=include_z,
+        )
+    except Exception:
+        # A func that assumes x, y, z are single values will likely raise a
+        # TypeError or a ValueError in which case we'll try again.
```
jorisvandenbossche, line 175: I am personally not a huge fan of such an automatic fallback. I know this exists in `shapely.ops.transform`. But if we would start from scratch (what we somewhat did for `shapely.transform`, and if we don't try to back `shapely.ops.transform` by `shapely.transform` with different defaults), I am not sure I would add this. If a user needs this, they can still write a small wrapper around the scalar function. 

(but I don't have a good sense for what `shapely.ops.transform` currently gets used, and with what kind of functions)

jorisvandenbossche (commented):
I added a few general remarks / thoughts from going through the PR, but see my comment at https://github.com/shapely/shapely/issues/1668#issuecomment-1366884171 for my preferred way forward before giving a detailed code review.

shapely/coordinates.py
idanmiara: > Alternatively, instead of using np.frompyfunc (to effectively apply the function on each geometry separately, right?), another option is to do what we did in geopandas: https://github.com/geopandas/geopandas/blob/f17b5350b8fa8acb8926e78ee011f876c7a03ff4/geopandas/_vectorized.py#L1099-L1117
> 
> Subsetting the input array in 2D and 3D geometries, and then transforming each of those subsets in a single go still using the natively vectorized transform function (without doing the vectorize on the python level with np.frompyfunc)
> 
> (might be useful to compare both approaches for performance)

I thought that it might be faster to do as you suggested, but then I thought that it wouldn't work for the old non-vectorized `rebuild` version, and the approach I chose works for all the cases without extra code, and also it would would be more complicated to implement and review, so I figured I will leave this idea for future PR.

shapely/coordinates.py
idanmiara: > I am personally not a huge fan of such an automatic fallback. I know this exists in `shapely.ops.transform`. But if we would start from scratch (what we somewhat did for `shapely.transform`, and if we don't try to back `shapely.ops.transform` by `shapely.transform` with different defaults), I am not sure I would add this. If a user needs this, they can still write a small wrapper around the scalar function.
> 
> (but I don't have a good sense for what `shapely.ops.transform` currently gets used, and with what kind of functions)

Well, I wouldn't use this fallback either, but I wanted to maintain full backwards compatibility. I also timed it figured that adding the `try` `except` wouldn't be a performance penalty for the first fast option, and if you ended up using the fallback function then I guess you don't care about performance anyway so that's also fine... 

Do we want eventually one function that has all the options with full backwards compatibility ? I think this way people can choose whichever combination of parameters they want as there is one function that supports them all.

If no, then I'm happy with removing this fallback, although I'm not convinced that there is any major drawback of keeping it.

idanmiara:
@jorisvandenbossche Thanks again for your review! I've fixed the issues as per your comments.
Please consider reviewing my solutions for https://github.com/shapely/shapely/issues/1693 and https://github.com/shapely/shapely/issues/1680 before I will split this PR into smaller chunks a you suggested.
Thanks!

shapely/coordinates.py
idanmiara: > For the `to_ragged_array` added in #1559 that is the only option (since the goal of that function is to return a single array). But of course here when it comes to transforming, you can easily have a transform function that cannot handle the NaNs very well (for example, in geopandas we ran into that with pyproj.Transformer.transform). So for the `shapely.transform` function, it probably _is_ useful to be able to have `include_z` per geometry.

Yes, exactly I thought that deducting a single `include_z` value for `include_z=None` could yield inconsistent results (`transform(func, [point2d, point3d], include_z=None)` could potentially yield different results then  `[transform(func, point2d, include_z=None), transform(func,, point3d, include_z=None)]`, which doesn't sound right.

idanmiara:
Hi @jorisvandenbossche,
I was going to split this commit, but then realized that I would have to rewrite all the tests if I remove the consolidated function.
In the end I ended up with the following:

1. I reverted the name change of `shapely.transform` so that would be still be the fastest option for whoever needed this option. so no changes to that function.
2. I named the consolidated function `shapely.transform_extra` as it has some extra parameters (but same default values as `shapely.transform`)
3. I added some documentation to all the functions, and I am not sure if we rather hide or expose them. I'm included to expose them as they each stand out by themselves, but I won't object to prefix them with `_` or removing them from `__all__`.
4. I was looking for what would be a name for the the functions that handle `interleave=False` and I ended up with planar as per https://stackoverflow.com/questions/66596023/name-for-the-opposite-of-interleaved-data-layout but if you have a better name I'll be happy to rename.
5. no further changes in this squash rebase besides renaming stuff and documentations.

If you still would want me to split this PR then please advise how do you want to test the remaining parts (as I think that the current tests would need to be heavily refactored if we are not going for the consolidated approach.

@brendan-ward I'd be happy for your input as well.

shapely/coordinates.py
```diff
@@ -54,20 +69,252 @@ def transform(geometry, transformation, include_z=False):
...
-            f"shape ({new_coordinates.shape})"
+            f"shape ({new_coordinates.shape}, but expected {coordinates.shape})"
         )
     geometry_arr = lib.set_coordinates(geometry_arr, new_coordinates)
     if geometry_arr.ndim == 0 and not isinstance(geometry, np.ndarray):
         return geometry_arr.item()
     return geometry_arr
 
 
+def transform_extra(
```
caspervdw, line 86: It could already came up in discussions with @jorisvandenbossche , but I couldn't find it:  what is the reason of adding a separate function for the two extra parameters? Wouldn't `transform_extra(geometry, transformation, interleaved=False, rebuild=False)` be equivalent to `transform(geometry, transformation)`?

shapely/coordinates.py
```diff
@@ -54,20 +69,252 @@ def transform(geometry, transformation, include_z=False):
...
+        transform_wrapper = np.frompyfunc(transform_wrapper, 3, 1)
+
+    return transform_wrapper(
+        geometry,
+        transformation,
+        include_z,
+    )
+
+
+def transform_planar(geometry, transformation, include_z: bool = False):
```
caspervdw, line 207: All these helper functions should be considered private, so prefixed with an underscore.

shapely/coordinates.py
caspervdw: I follow the reasoning and I am :+1: on implementing it this way. However, because it is an implementation that is not a "common practice", I think it the reasoning behind it should be annotated.

Also, the same implemtation occurs later in `transform_rebuild_planar`. Maybe you can factor this out in a helper function like

```
def _run_maybe_vectorized_transformation(transform, geometry, transformation, coords, include_z):
    ...
```

This would include the reasoning behind the `try ... except`

shapely/coordinates.py
```diff
@@ -54,20 +69,252 @@ def transform(geometry, transformation, include_z=False):
...
+    will get rebuilt from the transformation output
+
+    Parameters
+    ----------
+    geometry : Geometry or array_like
+    transformation : function
+        A function that transforms coordinates, in ``interleaved=True`` mode:
+        the function transforms a (N, 2) or (N, 3) ndarray of float64 to
+        another (N, 2) or (N, 3) ndarray of float64.
+        in ``interleaved=False`` mode: the function transform 2 or 3 N ndarray
```
caspervdw, line 112: The docs deserve some more attention.

caspervdw (commented):
Looks great, a few minor comments.

shapely/coordinates.py
idanmiara: > It could already came up in discussions with @jorisvandenbossche , but I couldn't find it: what is the reason of adding a separate function for the two extra parameters? Wouldn't `transform_extra(geometry, transformation, interleaved=False, rebuild=False)` be equivalent to `transform(geometry, transformation)`?

Hi @caspervdw thanks for your review!
`transform_extra(geometry, transformation, interleaved=True, rebuild=False)` is equivalent to `transform(geometry, transformation)` (I'm not a big fan of the keyword `interleaved`, so maybe if it confused you we should consider alternatives) 
Could you please clarify what are you suggesting? Are you suggesting to add the extra keywords to the original `transform` function?

idanmiara: My main reasoning behind not modifying `transform`, making `transform_planar` public and not adding a common function like `_run_maybe_vectorized_transformation` was performance, as I wanted to reduce the function calls and unnecessary checks for some common use cases.

shapely/coordinates.py
idanmiara: Note https://github.com/shapely/shapely/pull/1676#discussion_r1147679057

shapely/coordinates.py
caspervdw: Hi @idanmiara, thank you for your recent efforts on shapely!

Why are you worried about performance? Looking at the code, I don't expect any significant difference between the existing `transform` and the `transform_extra` (provided that `rebuild=False`). 

Either way, a comprehensible and maintainable API is also worth a great deal.  So yes I'd prefer a single top-level function `transform` that has all the options. But I may be convinced by some benchmarks :)

shapely/coordinates.py
caspervdw: Mainly, shorten it. I did an attempt:

```
transformation : function
        A function that transforms coordinates. If the default value of ``interleaved=True``
        is used, the function must transform a (N, 2) or (N, 3) ndarray
        of float64 to another (N, 2) or (N, 3) ndarray of float64.
        Unless ``rebuild=True``, the function may not change N.      
include_z : bool, optional, default False
        If True, include the third dimension in the coordinates array
        that is passed to the ``transformation`` function. If a
        geometry has no third dimension, the z-coordinates passed to the
        function will be NaN.
        If False, the third dimension will never be passed into the function.
        If None, the third dimension is only passed if present
interleaved: bool, default True
        If set to False, the transformation function should accept 2 or 3 separate
        one-dimensional arrays (x, y and optional x).
rebuild: bool, default False
        If set to True, the transformation function is allowed to change the number
        of coordinates per geometry.
```

shapely/coordinates.py
caspervdw: Additionally, the error would occur in the `transformation` function. So the try ... except (potentially) catches too much now.

Maybe a standard function wrapper?

```
def transpose_transformation(transformation):
    @functools.wraps(func)
    def wrapper(coords):
        try:
            return np.array(transformation(*coords.T).T
        except Exception:
            return np.array([transformation(*c) for c in coords])


# in the transform function do:
transformation = transpose_transformation(transformation) if not interleaved else transformation
```

shapely/coordinates.py
```diff
@@ -54,20 +69,252 @@ def transform(geometry, transformation, include_z=False):
...
+    transformation,
+    include_z: bool = False,
+):
+    """Returns a copy of a geometry array with a function applied to its coordinates.
+
+    Refer to `shapely.transform_extra` (``rebuild=True``, ``interleaved=False``) for full documentation.
+    """
+    if geometry.is_empty:
+        return geometry
+    if geometry.geom_type in ("Point", "LineString", "LinearRing", "Polygon"):
```
caspervdw, line 269: We have a `GeometryType` enum, which should be used here (in combination with `shapely.get_type_id`. If you care for performance ;)

shapely/coordinates.py
```diff
@@ -54,20 +69,252 @@ def transform(geometry, transformation, include_z=False):
...
+        except Exception:
+            # A func that assumes x, y, z are single values will likely raise a
+            # TypeError or a ValueError in which case we'll try again.
+            return transform_rebuild_single_part(
+                geometry,
+                lambda coords: [transformation(*c) for c in coords],
+                include_z=include_z,
+            )
+    elif (
+        geometry.geom_type.startswith("Multi")
```
caspervdw, line 286: Intuitively I don't like a string operations (`startswith`) on identifiers such as a geometry type. I would be much happier if there would be an ` in {...}` such as you did with the simple geometries.

Or just an `else`?

idanmiara:
Hi @caspervdw,
Thanks again for looking into this and for your thorough review!
I see your points.
I'm on a long holiday and don't have an access to a dev computer for the following month. I'd be happy to fix these points and continue the conversation when I'm back. 🙏

shapely/coordinates.py
```diff
@@ -54,20 +69,252 @@ def transform(geometry, transformation, include_z=False):
...
+
+def transform_rebuild_single_part(geometry, transformation, include_z: bool = False):
+    """Helper function for shapely.transform_rebuild, and shapely.transform_rebuild_planar
+    for a single part geometries
+    """
+    if geometry.geom_type in ("Point", "LineString", "LinearRing"):
+        return type(geometry)(
+            transformation(get_coordinates(geometry, include_z=include_z))
+        )
+    elif geometry.geom_type == "Polygon":
```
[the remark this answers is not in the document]
idanmiara: Although you are correct for some of this function uses, it's not true for all of them (`transform_rebuild`, `transform_rebuild_planar`) so if I split `transform_rebuild_single_part` I'd need to add a few more if statements in other places (especially given the merged `transform_planar` version), so I think it's more simple to leave it as is.

shapely/coordinates.py
idanmiara: Hi @caspervdw,
I think I've addressed all the issues, please review.
I left the extra 4 top level functions (for whoever want to choose a specific variations for performance reasons) I would run some benchmarks to justify this reasoning and we can discuss if it makes sense to hide them.

idanmiara:
Hi @caspervdw,
I think I've addressed all the issues, please review.
I've added the `_` prefix and hid the auxiliary functions, as the performance difference between direct use of the auxiliary function was negligible.

idanmiara:
Is anyone available to review this PR? Thanks!

caspervdw (changes requested):
Hi @idanmiara  I had the chance to have another look. As it is, I think this PR shouldn't be merged. The consolidation of the scalar and vectorized versions of transform resulted in the addition of some additional cases for which the implementation turns out the be pretty convoluted. I propose the following:

1. Seperately merge the "interleaved" and "include_z = None" for the vectorized transform function (https://github.com/shapely/shapely/pull/1849)
2. Leave out the "rebuild" parameter and instead implement a user-facing `transform_scalar(geometry: Geometry, transformation: Callable, include_z: Optional[bool], interleaved: True)`, also in a different PR.
3. Close this PR

So I advise against the "vectorized" version of the "rebuild" mode as it adds significant complexity and users are able to to that themselves when necessary.

caspervdw:
I do think having a consistent namespace is a good thing; we could just document the difference between the two like so:

```
def transform(...):
    """...

    For transformations that change the number of coordinates in a geometry,
    use transform_scalar.
```

```
def transform_scalar(...):
    """...

    Note: if the transformation accepts numpy arrays and does not change the
    number of coordinates, transform() is the more performant option.
```

idanmiara:
> I do think having a consistent namespace is a good thing; we could just document the difference

Ok, I'll rebase and continue when #1849 is merged.

jorisvandenbossche:
> instead implement a user-facing `transform_scalar(geometry: Geometry, transformation: Callable, include_z: Optional[bool], interleaved: True)`

How would `transform_scalar` and `transform` differ when passed a scalar Geometry? If the only difference is that `transform_scalar`, when passed an array of geometries, calls the transform on each individual geometry, then another option is to document how to do this manually to achieve the same result (either with a list comprehension, or like `np.vectorize(shapely.transform)(geometries)`)

idanmiara:
> > instead implement a user-facing `transform_scalar(geometry: Geometry, transformation: Callable, include_z: Optional[bool], interleaved: True)`
> 
> How would `transform_scalar` and `transform` differ when passed a scalar Geometry? If the only difference is that `transform_scalar`, when passed an array of geometries, calls the transform on each individual geometry, then another option is to document how to do this manually to achieve the same result (either with a list comprehension, or like `np.vectorize(shapely.transform)(geometries)`)

I think that what @caspervdw suggested is to name name the rebuild transform `transform_scalar` as it only operates on single geometries. That instead of adding another kwarg `rebuild` that also vectorizes rebuilding a geometry array, as I suggested in this PR.

jorisvandenbossche:
Ah, yes, I forgot about the rebuild aspect that re-creates the resulting geometry differently, allowing a change in size/dimension. Now, if that's the main difference in behaviour, "transform_scalar" might be not be the most clarifying name?

caspervdw:
Work is partially done in #1849 and pending in #1902 . Closing this one to avoid confusion. Thanks @idanmiara for you explorative work on this.

Issue #1664, opened 2022-12-13
ENH: C extension: use const GEOSGeometry pointers instead of plain GEOSGeometry pointers where appropriate


jorisvandenbossche:
_Originally reported at https://github.com/pygeos/pygeos/issues/227 by @brendan-ward_

Internal functions like `GeometryObject_FromGEOS` should accept `const` pointers instead of plain pointers, since those functions should not modify the GEOS Geometry.

Originally detected in #225 since we are getting `const` pointers back from GEOS functions but these were getting discarded on call to the PyGEOS C API.

PR #1678, opened 2022-12-18
CLN: Add consts to PyUFuncGenericFunction signatures

hoodmane:
When building with `-Wincompatible-function-pointer-types`, we get a bunch of errors like:
```C
error: incompatible function pointer types initializing 
'PyUFuncGenericFunction' (aka 'void (*)(char **, const long *, const long *, void *)') 
with an expression of type 
'void (*)(char **, npy_intp *, npy_intp *, void *)' (aka 'void (*)(char **, long *, long *, void *)') 
[-Wincompatible-function-pointer-types]
static PyUFuncGenericFunction Y_b_funcs[1] = {&Y_b_func};
                                              ^~~~~~~~~
```
This adds the consts needed to silence the error message.

### DIFF
--- a/src/ufuncs.c
+++ b/src/ufuncs.c
@@ -174,7 +174,7 @@ static void* is_ccw_data[1] = {GEOSGeom_isCCW_r};
 
 typedef char FuncGEOS_Y_b(void* context, void* a);
 static char Y_b_dtypes[2] = {NPY_OBJECT, NPY_BOOL};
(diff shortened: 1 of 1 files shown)


jorisvandenbossche (approved):
Yes, thanks a lot!

jorisvandenbossche:
(we have another `const` related issue about GeometryObject as well (https://github.com/shapely/shapely/issues/1664), but this is actually the easier part it seems)

PR #1679, opened 2022-12-18
Remove extra argument from GetNumInteriorRings

hoodmane:
GetNumInteriorRings is a supposed to be of type FuncGEOS_Y_i, which has signature
```C
int FuncGEOS_Y_i(void* context, void* a);
```
but when it's declared it takes an extra (unused) integer n argument. This fixes it.


jorisvandenbossche (approved):
Thanks for the catch!

Issue #1680, opened 2022-12-19
Merge shapely and pygeos tests folders (`shapely/tests`, `shapely/shapely/tests`) ?


idanmiara:
I find it confusing to have two root test folders:
`shapely/test` - originated from `shapely` <= 1.8
`shapely/shapely/tests` - originated from `pygeos`

Should we merge them both to either of the options?

PR #1681, opened 2022-12-19
TST: Move all remaining /tests to /shapely/tests

idanmiara:
Move `/tests` to `/shapely/tests` as noted in https://github.com/shapely/shapely/issues/962
Closes https://github.com/shapely/shapely/issues/1680
Moving the tests also means that all the tests will be covered by `pre-commit` (as of now only the tests under `shapely/test` are covered)


idanmiara:
Hi, 
This PR ready for review.
I'll squash when it's approved.
Thanks!

idanmiara:
@jorisvandenbossche please review

idanmiara:
@jorisvandenbossche can his be included in `2.0.1` ?

idanmiara:
@brendan-ward @mwtoews @caspervdw 
Please review. Thanks!

shapely/tests/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+    if geos_path:
+        os.add_dll_directory(geos_path + r"\bin")
+
+
+import numpy
+import pytest
+
+from shapely.geos import geos_version_string
+
+# Show some diagnostic information; handy for CI
```
brendan-ward, line 16: This should be consolidated into [pytest_report_header](https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_report_header) in `conftest.py`

shapely/tests/__init__.py
```diff
@@ -0,0 +1,24 @@
+import os
```
brendan-ward, line 1: I'm not clear on the role of this file; it seems like it should all be consolidated into `conftest.py`.  However, I don't have a Windows machine to test on, so I'm not sure of the effect of moving the DLL import to conftest instead of here.

shapely/tests/test_cga.py
```diff
@@ -34,8 +38,9 @@ def test_no_holes(self):
         assert polygon.exterior.is_ccw
 
     def test_holes(self):
-        polygon = Polygon([(0, 0), (0, 1), (1, 0)],
-                          [[(0.5, 0.25), (0.25, 0.5), (0.25, 0.25)]])
+        polygon = Polygon(
```
brendan-ward, line 41: Minor nit: consider surrounding this in `# fmt: off` / `# fmt: on` to prevent unwrapping; here it is helpful for the hole to be on a separate line.

shapely/tests/test_clip_by_rect.py
```diff
@@ -1,109 +1,123 @@
...
-    assert dump_wkt(geom2, rounding_precision=0) == wkt
-
-
-def polygon_overlapping_rectangle():
-    """Polygon overlapping rectangle"""
-    wkt = "POLYGON ((0 0, 0 30, 30 30, 30 0, 0 0), (10 10, 20 10, 20 20, 10 20, 10 10))"
-    geom1 = load_wkt(wkt)
-    geom2 = clip_by_rect(geom1, 5, 5, 15, 15)
-    assert dump_wkt(geom2, rounding_precision=0) == "POLYGON ((5 5, 5 15, 10 15, 10 10, 15 10, 15 5, 5 5))"
+"""
```
brendan-ward, line 1: Here a little signposting in the PR would be helpful.  If this large change is purely the result of running an autoformatter, then please say so; otherwise reviewers have to inspect more carefully and substantive changes may be hidden by the noise of purely formatting changes.

brendan-ward (commented):
Thanks for working on this @idanmiara ; it is great to have the tests consolidated into the same folder structure.

I only made it about 1/2 way through reviewing before running out of time, but the part I found challenging here was knowing where / if there were any substantive changes that needed more careful review that weren't purely the result of file renaming, autoformatting, and isort.  Are there any such changes here?

idanmiara:
> Thanks for working on this @idanmiara ; it is great to have the tests consolidated into the same folder structure.
> 
> I only made it about 1/2 way through reviewing before running out of time, but the part I found challenging here was knowing where / if there were any substantive changes that needed more careful review that weren't purely the result of file renaming, autoformatting, and isort. Are there any such changes here?

Thanks!
Nothing fancy in this PR.
Only rename and reformat with `pre-commit`.
only thing to consider that name clash:
`shapely/tests/test_predicates.py` -> `shapely/shapely/tests/test_predicates1.py` (as `test_predicates.py` was already exist)

shapely/tests/__init__.py
idanmiara: I'm thinking to leave this to another PR as it seems out of scope ?

shapely/tests/__init__.py
idanmiara: I'm thinking to leave this to another PR as it seems out of scope ?

shapely/tests/test_clip_by_rect.py
```diff
@@ -1,109 +1,123 @@
...
+
+import pytest
+
+from shapely.geos import geos_version
+from shapely.ops import clip_by_rect
+from shapely.wkt import dumps as dump_wkt
+from shapely.wkt import loads as load_wkt
+
+pytestmark = pytest.mark.skipif(geos_version < (3, 5, 0), reason="GEOS 3.5.0 required")
```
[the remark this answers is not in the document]
idanmiara: ok

shapely/tests/test_clip_by_rect.py
idanmiara: this is purely formatting, I'll check why it shows like that

shapely/tests/test_create_inconsistent_dimensionality.py
```diff
@@ -0,0 +1,214 @@
+"""
```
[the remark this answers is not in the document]
idanmiara: ok

shapely/tests/__init__.py
brendan-ward: Follow-on PRs are just fine for these changes, in order to keep this one focused purely on moving & formatting code.  They just jumped out as needing to be fixed at some point.  😄

shapely/tests/test_cga.py
idanmiara: ok

idanmiara:
@brendan-ward there were 3 files that `pre-commit` changed so much that when renamed in the same commit git treats them as deleted and added. I've split into two commits and I think now it looks good.
@jorisvandenbossche - please do not squash as the history of those files would be lost.

jorisvandenbossche:
@idanmiara sorry for the slow reply on this one. First on the original topic of the PR: yes, we eventually want to have all tests consolidated in `shapely/tests/`, and thus move the ones in `tests/`. And some of those tests have already been moved (see for example https://github.com/shapely/shapely/pull/1257). 

However, there are multiple aspects about those old tests: 1) they need to be moved, 2) they can use a clean-up (format with black/isort, i.e. pre-commit, migrate unittest to pytest, ..), and 3) some can use consolidation and de-duplication with other tests in `shapely/tests`. 
And I personally don't like to move them as-is to `shapely/tests` with only some formatting changes. That makes it a lot harder to see which files still need the other steps of clean-up, and which not. So I would personally prefer to either move those tests in smaller steps (multiple PRs) and directly clean-up properly, or either move them to some subdirectory inside `shapely/tests/` so it's clear which tests still need clean-up (and then when cleaning up, we can move them out of the subdirectory). One disadvantage of the latter is that they get moved twice.

shapely/tests/__init__.py
jorisvandenbossche: I don't think there is value in moving something we should just delete. I would prefer to clean it up at once, and I don't think there is anything in this file that is required.

idanmiara:
> or either move them to some subdirectory inside `shapely/tests/` so it's clear which tests still need clean-up (and then when cleaning up, we can move them out of the subdirectory). 

Moving to a subdirectory sounds like a good compromise so people can find them all and also they would be automatically included in `pre-commit` and also in the distribution.

> One disadvantage of the latter is that they get moved twice.

True, but I think that it's better than the alternative of not moving them for a long time so it's harder for people to find them.

shapely/tests/__init__.py
idanmiara: > I don't think there is value in moving something we should just delete. I would prefer to clean it up at once, and I don't think there is anything in this file that is required.

OK. 
What do you advise to do with the following code, delete or move?:
```
# Show some diagnostic information; handy for CI
print("Python version: " + sys.version.replace("\n", " "))
print("GEOS version: " + geos_version_string)
print("Numpy version: " + numpy.version.version)
```

idanmiara:
> @brendan-ward there were 3 files that `pre-commit` changed so much that when renamed in the same commit git treats them as deleted and added. I've split into two commits and I think now it looks good. You might want to look at them in commits view https://github.com/shapely/shapely/pull/1681/commits @jorisvandenbossche - please do not squash as the history of those files would be lost.

@jorisvandenbossche I've put these two files in https://github.com/shapely/shapely/pull/1743 so that we can rebase and squash this commit on top of it (so please review https://github.com/shapely/shapely/pull/1743 first)

brendan-ward:
> move them to some subdirectory inside shapely/tests/ so it's clear which tests still need clean-up

Very valid points @jorisvandenbossche and I like this solution.  While multiple moves are not ideal, having everything in the `shapely/tests` subtree means we don't have to remember to run tests against 2 different folders.  Longer term, I absolutely agree about needing to rework some of the tests - but I'd really like to see those in much smaller individual PRs, so a couple of bigger formatting and file move PRs to set up for that seems OK.

Maybe something like `shapely/tests/legacy` so that it is clear these are older tests in need of of update?

idanmiara:
> Maybe something like `shapely/tests/legacy` so that it is clear these are older tests in need of of update?

Sounds good to me. I'll rebase and implement once 
https://github.com/shapely/shapely/pull/1743 is merged.

idanmiara:
@brendan-ward @jorisvandenbossche 
I've squashed and rebased as there were too many conflicts to do it otherwise.

shapely/tests/legacy/test_clip_by_rect.py
```diff
@@ -1,123 +1,120 @@
-"""
```
idanmiara, line 1: I'm not sure why it shows this way. the only change in this file besides the rename is the removal of 
`pytestmark = pytest.mark.skipif(geos_version < (3, 5, 0), reason="GEOS 3.5.0 required")` and `from shapely.geos import geos_version`

shapely/tests/__init__.py
jorisvandenbossche: Can you keep this `__init__.py` in the legacy subdirectory? And the same for the conftest.py file? 

Both are not needed for the existing /shapely/tests/ to run, and as Brendan mentioned above, this content actually belongs elsewhere. So until that is done (which indeed doesn't need to be in this PR), let's leave it as is in the /legacy subdirectory

shapely/tests/legacy/test_create_inconsistent_dimensionality.py
```diff
@@ -187,7 +187,7 @@
 def test_create_from_geojson(geojson):
     # exact error depends on numpy version
     with pytest.raises((ValueError, TypeError)) as exc:
-        wkt = shape(geojson).wkt
+        wkt = shape(geojson).wkt  # NOQA
```
jorisvandenbossche, line 190: Why is this NOQA needed in this case?

idanmiara: because of the unused `wkt` assignment (we can also remove that variable, which I guess was put there for debugging purposes)

shapely/tests/__init__.py
idanmiara: resolved as per https://github.com/shapely/shapely/pull/1681#discussion_r1104118419

idanmiara:
@jorisvandenbossche I've rebased and I think I addressed all the issues.

shapely/tests/legacy/test_clip_by_rect.py
idanmiara: solved via rebasing and separated commits for move and edit

idanmiara:
@jorisvandenbossche I've addressed all the issue, then it seems that git picks up some changes as deletion and addition.
To workaround this issue I've rebased and recommitted 3 separate commits (moving files and some minor cleanups that don't confuse git, and 2 cleanup commits) this way git can track the changes correctly.
I advise not to squash in order to keep the file history intact.

tests/__init__.py
```diff
@@ -1,24 +0,0 @@
-import os
-import sys
-import unittest
-
-if os.name == "nt" and sys.version_info[1] >= 8:
```
brendan-ward, line 5: Were lines 5-8 and 22-24 not copied to to `shapely/tests/legacy/__init__.py` because they are no longer used?  Apologies if I missed the comment describing these changes.

idanmiara: > Were lines 5-8 and 22-24 not copied to to `shapely/tests/legacy/__init__.py` because they are no longer used? Apologies if I missed the comment describing these changes.

I recall so but I can't find the reference either. Should I restore them @jorisvandenbossche?

brendan-ward: From what I can tell, `shapely20_deprecated` is not used anywhere else, so I'm OK with dropping that here.

I don't have the ability to judge the impact of setting the GEOS DLL directory for Windows, but I'm assuming since all the Windows tests still pass that it isn't needed under a standard installation, and the operation of loading it isn't used anywhere else in the codebase.  So - I have no objections to removing this either.

Just wanted to be sure I was following the reason they we modified; all other changes here were expected and easy to review.  Thanks for the multiple rounds of refactoring & rebasing @idanmiara !

idanmiara: Yes, I don't think that section is needed either as there are many tests that passed on windows also before this change and all the tests pass also after this change...
I also tested it on windows myself

brendan-ward (approved):

idanmiara:
Thanks @brendan-ward

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks @idanmiara !

PR #1685, opened 2022-12-23
ENH: Add constrained_delaunay_triangles function

### LINKED ISSUE #772: Triangulation not working for non-convex polygon

Triangulation of a *Polygon* should not produce triangles outside the Polygon. While they have to be present for a *MultiPoint* with the same coords, it is not useful to have them for the Polygon.

Repro data in https://github.com/jpmaterial/test_data/tree/master/Shapely/772 

## Operating system

Windows 10

installed dependencies with conda, installed shapely 1.7a2 by git cloning from github and setup.py develop

rgangopadhya:
One of the items in https://github.com/shapely/shapely/issues/1570

Resolves #772


shapely/ops.py
```diff
@@ -164,6 +164,16 @@ def triangulate(geom, tolerance=0.0, edges=False):
     return [g for g in collection.geoms]
 
 
+def constrained_triangulate(geom):
+    """Creates the constrained Delaunay triangulation and returns a list of polygons
```
rgangopadhya, line 168: Should I just reference the docstring in `constructive`, like we're starting to do here? https://github.com/shapely/shapely/pull/1673

idanmiara: I think that should be the way to go. I've opened https://github.com/shapely/shapely/issues/1688 to discuss this further.

jorisvandenbossche: I don't think it is needed to add a version to `ops.py`. The existing functions in ops.py are (for now) kept for backwards compatibility, but other new functions have only been added as top-level functions (such as in `constructive.py`, which is then imported from there in the main `shapely.*` namespace)

shapely/constructive.py
```diff
@@ -436,6 +437,40 @@ def delaunay_triangles(geometry, tolerance=0.0, only_edges=False, **kwargs):
...
+
+    For non-polygonal inputs, returns an empty geometry collection.
+
+    Parameters
+    ----------
+    geometry : Geometry or array_like
+    **kwargs
+        For other keyword-only arguments, see the
+        `NumPy ufunc docs <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs>`_.
+
```
jorisvandenbossche, line 456: Can you add a "Returns" section? 

And I would also add a "See Also" section (see eg how this is used in `reverse` in this file) to point to `delaunay_triangles`, and also add this to the docstring of `delaunay_triangles` to point to `constrained_delaunay_triangles`

tests/test_constrained_delaunay.py
```diff
@@ -0,0 +1,39 @@
+import unittest
```
jorisvandenbossche, line 1: Can you move those tests to `shapely/tests/test_constructive.py`? 

(you will also notice that this file doesn't use `unittest`)

jorisvandenbossche (commented):
Thanks for the PR!

Added a few inline comments. For the docstring, it would also be nice if we could somewhat explain what the difference is between `delaunay_triangles` and `constrained_delaunay_triangles` (https://postgis.net/docs/ST_ConstrainedDelaunayTriangles.html gives an example of how the former can create edges that cross polygon boundaries)

jorisvandenbossche:
PostGIS actually has this as [`ST_TriangulatePolygon`](https://postgis.net/docs/ST_TriangulatePolygon.html), but I suppose that is mostly because [`ST_ConstrainedDelaunayTriangles`](https://postgis.net/docs/ST_ConstrainedDelaunayTriangles.html) already existed with SFCGAL instead of GEOS as backend. 

So for naming, while we have the general goal of using consistent names with PostGIS, in this case it might make sense to keep `constrained_delaunay_triangles`, to be consistent within shapely with the already existing `delaunay_triangles`. 
A "triangulate_polygon" name does make it more explicit that this version only works for polygons, though. 

cc @caspervdw

shapely/ops.py
rgangopadhya: Thanks, I wasn't sure what purpose having this in `ops.py` had :+1:

rgangopadhya:
Thanks for the quick review @jorisvandenbossche! I'll address your feedback in the next day or so. Regarding

>  A "triangulate_polygon" name does make it more explicit that this version only works for polygons, though.

I was playing with the idea of raising an exception on non-polygonal input as an additional way to communicate this, but opted for consistency with GEOS

tests/test_constrained_delaunay.py
rgangopadhya: Any particular reason not to use `unittest` here? Not a huge deal, but I do like the assert API provided through `unittest`. Will move back to similar-style asserts as in the other tests.

I'll also move `test_deluanay` over to `test_constructive` as well

rgangopadhya:
I believe I addressed all PR feedback @jorisvandenbossche, let me know if there is anything else I should consider. I was also thinking for consistency we could move `test_delaunay.py` over to `test_constructive`.

DennisVVV:
Are there any updates on this? Any idea when this will be available in Shapely? We are really looking forward to this.

rgangopadhya:
@jorisvandenbossche @caspervdw ?

ArthurDuysens:
I have made it work locally on windows, but I had to change the setup.py to accommodate for the lack of geos-config. What is the normal procedure for compiling it on Windows.

btw, I also tested it on Linux and it seemed to be working!

theroggy:
As it seemed ready to go at a time, it might be nice if it could be included in shapely 2.1?

jorisvandenbossche (approved):
Indeed, let's get this into 2.1 as `constrained_delaunay_triangles`.

Thanks @rgangopadhya for the PR! And apologies this has taken such a long time ..

rgangopadhya:
@jorisvandenbossche thanks! And thanks @theroggy for pushing this over the edge as well :) nice Friday surprise

Issue #1684, opened 2022-12-23
Feat: Vectorized orient function


kylebarron:
## Expected behavior and actual behavior.

It looks like the [`shapely.geometry.polygon.orient`](https://shapely.readthedocs.io/en/stable/manual.html#shapely.geometry.polygon.orient) function is not yet vectorized and only works with a single polygon object

## Steps to reproduce the problem.

```
import shapely
xmin = [0, 1]
ymin = [3, 4]
xmax = [5, 6]
ymax = [7, 8]
geoms = shapely.box(xmin, ymin, xmax, ymax, ccw=False)
shapely.geometry.polygon.orient(geoms)
# AttributeError: 'numpy.ndarray' object has no attribute 'exterior'
```

## Operating system

MacOS

## Shapely version and provenance

2.0.0 from pip

idanmiara:
Hi @kylebarron 
Please use the new vectorized function `shaply.ops.orient` added in https://github.com/shapely/shapely/pull/1690

kylebarron:
It looks like `shapely.ops.orient` is already vectorized as of 2.0, and additionally is already much faster than a python for loop (as we'd expect, so not sure if #1690 is necessary?):

```py
import shapely
xmin = [0, 1]
ymin = [3, 4]
xmax = [5, 6]
ymax = [7, 8]
geoms = shapely.box(xmin, ymin, xmax, ymax, ccw=False)
big_geoms = list(geoms) * 1000

%timeit shapely.ops.orient(big_geoms)
# 313 ns ± 7.58 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

%timeit [shapely.ops.orient(geom) for geom in big_geoms]
# 244 ms ± 4.66 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

I didn't know `shapely.ops.orient` existed; I'll move this to a doc request instead of a feature request. When I [search for `orient`](https://shapely.readthedocs.io/en/stable/search.html?q=orient) in the doc website, the only API doc result is for `shapely.geometry.polygon.orient`. Is there an API doc page in the website for everything in `shapely.ops`? Or is it laid out by intention, not by python module

idanmiara:
> It looks like `shapely.ops.orient` is already vectorized as of 2.0

Nope. It's just returns the original geom since it's not supported. just try your original test on it.

> Is there an API doc page in the website for everything in `shapely.ops`? Or is it laid out by intention

[...]

Issue #1366, opened 2022-04-15
ENH: vectorized function to force orientation of polygon rings


jorisvandenbossche:
Similar as https://github.com/pygeos/pygeos/issues/71

Shapely has a custom function to force the orientation of a geometry: `shapely.ops.orient` ([impl](https://github.com/shapely/shapely/blob/fc55006ecac8ad7835fe1618d6d391bfe076403c/shapely/geometry/polygon.py#L322-L335). But, this is currently a pure python function. Because this is not something that is supported by GEOS out of the box, this is also not supported by a vectorized ufunc that were merged from pygeos.

It would be good to add a vectorized ufunc version of `orient`. 

As reference, PostGIS has [`ST_ForcePolygonCCW`](https://postgis.net/docs/ST_ForcePolygonCCW.html) and related functions.

idanmiara:
Until a more optimized solution is proposed (ufunc, geos, etc...):
https://github.com/shapely/shapely/pull/1690 could be an alternative solution.

jorisvandenbossche:
Opened https://github.com/libgeos/geos/issues/779 to track this upstream in GEOS

idanmiara:
Please see new implementation for ``force_polygon_ccw` in https://github.com/shapely/shapely/pull/1690

jorisvandenbossche:
Closed by https://github.com/shapely/shapely/pull/2147

PR #1690, opened 2022-12-25
ENH: add vectorized fallback for `orient_polygons` for GEOS < 3.12

idanmiara:
Closes https://github.com/shapely/shapely/issues/1366
Closes https://github.com/shapely/shapely/issues/1684 
Closes (mostly) https://github.com/shapely/shapely/issues/1696


idanmiara:
I'm wondering if I could generalize the decorator to work on any object type, not just on Geometry.
For this I need to replace `isinstance(geoms, Geometry)` in the decorators with a function like `np.isscalar(geoms)` but that works also for objects. `np.isscalar` returns `False` for `Geometry` or other objects. Any ideas ?

jorisvandenbossche:
> I'm wondering if I could generalize the decorator to work on any object type, not just on Geometry.

What would be the goal / use case of that? 

In general, I am not sure if it is needed to already generalize this with a decorator, if we for now only have one place where we use this (and as long as we don't have plans to use it elsewhere as well).

---

Some other thoughts:

- If we add a vectorized version, we should probably directly add this as a top-level function as well, to be consistent with all other vectorized functions. 
- If we add it as a new top-level function, we should discuss the naming, though. We are now using `orient`, but we generally try to follow the PostGIS naming (unless there is a good reason to deviate), in which case it would be something like `force_ccw` or `force_polygon_ccw`.
- If we provide a vectorized version, it _would_ be nice to directly do this as a fast function in C (ufunc.c). I think that should be doable (without anything being added to GEOS upstream) and could mimic what we have for `force_2d` (that is also a custom implementation on our side, and forcing (c)cw would actually be simpler). If you would be interested in looking at this as well, I am happy to give some more pointers.

idanmiara:
> What would be the goal / use case of that?

No specific other use case at the moment, just thought it would be a cleaner solution. 

> In general, I am not sure if it is needed to already generalize this with a decorator, if we for now only have one place where we use this (and as long as we don't have plans to use it elsewhere as well).

I think that could be a nice, easy to implement vectorize method to other functions with a single geometry as the first argument. I could have used it for the old `transform` function, but ended up using something a bit more optimized for that use case. 

> If we add it as a new top-level function, we should discuss the naming, though. We are now using orient, but we generally try to follow the PostGIS naming (unless there is a good reason to deviate), in which case it would be something like force_ccw or force_polygon_ccw.

I don't have an opinion about the naming of the function. If you want to pick one I'd be happy to put as a top level function and put it also under the old name.module for backwards compatibility.

> If you would be interested in looking at this as well, I am happy to give some more pointers.

Thanks! Unfortunately, currently I don't have the time to go into this.

idanmiara:
> Some other thoughts:
> 
> * If we add a vectorized version, we should probably directly add this as a top-level function as well, to be consistent with all other vectorized functions.
> * If we add it as a new top-level function, we should discuss the naming, though. We are now using `orient`, but we generally try to follow the PostGIS naming (unless there is a good reason to deviate), in which case it would be something like `force_ccw` or `force_polygon_ccw`.

Done.
I could cleanup the commit history if you approve this.

idanmiara:
I've further improved `orient` and `force_ccw` using added new functions `reversed_conditioned` and using `is_ccw`.
since `reversed` and `is_ccw` are only available with GEOS >= 3.7 and the original `orient` did not have a GEOS minimum version, I've added a fallback implementation to both `reversed` and `is_ccw`, now all the tests for these functions also pass with GEOS < 3.7.

The new `orient`/`force_ccw` also support `LinearRing` (before it retuned the geometry unchanged), thus the relevant test was updated.

If you approve these additions but require smaller PRs for review I could split this PR.

jorisvandenbossche:
Thanks for further working on this!

> since `reversed` and `is_ccw` are only available with GEOS >= 3.7 and the original `orient` did not have a GEOS minimum version, I've added a fallback implementation to both `reversed` and `is_ccw`, now all the tests for these functions also pass with GEOS < 3.7.

In general, I think we shouldn't bother too much with custom code to support such an old GEOS version. It's fine that some features just raise an error that the GEOS version is too old, and we are probably going to require GEOS 3.7 as a minimum version in the near future anyway (https://github.com/shapely/shapely/issues/1000)

> I could cleanup the commit history if you approve this.

No need to clean up commit histories, we squash on merge anyway.

shapely/algorithms/cga.py
```diff
@@ -1,6 +1,29 @@
...
 import shapely
+from shapely import _is_ccw_geos_3_7
+from shapely.decorators import vectorize_geom
+from shapely.errors import ShapelyDeprecationWarning
+from shapely.geos import geos_version
+
+__all__ = [
+    "is_ccw",
+    "force_ccw",
+    "is_ccw_impl",
```
jorisvandenbossche, line 14: We don't want to expose this one top-level, I think

shapely/constructive.py
```diff
@@ -769,9 +772,62 @@ def remove_repeated_points(geometry, tolerance=0.0, **kwargs):
     return lib.remove_repeated_points(geometry, tolerance, **kwargs)
 
 
+def reverse_conditioned(geometry, condition=True, **kwargs):
```
jorisvandenbossche, line 775: What's the rationale of exposing this method publicly? Generally this seems something that a user can easily code themselves if they need some if/else logic around calling `reverse`?

shapely/algorithms/cga.py
idanmiara: OK, I'll remove this

idanmiara:
> Thanks for further working on this!
> 

Thanks for your review :)

> > since `reversed` and `is_ccw` are only available with GEOS >= 3.7 and the original `orient` did not have a GEOS minimum version, I've added a fallback implementation to both `reversed` and `is_ccw`, now all the tests for these functions also pass with GEOS < 3.7.
> 
> In general, I think we shouldn't bother too much with custom code to support such an old GEOS version. It's fine that some features just raise an error that the GEOS version is too old, and we are probably going to require GEOS 3.7 as a minimum version in the near future anyway (#1000)

OK, Would you suggest I'll remove the fallback implementations for `reverse` and `is_ccw` and mark `force_ccw` (and `orient`) for GEOS >= 3.7 ?

I think that till GEOS 3.7 is the minimum we could keep the `reverse` fallback/workaround and remove the weird `is_ccw_impl` workaround (I will make a separate PR, if you like this idea)

> > I could cleanup the commit history if you approve this.
> 
> No need to clean up commit histories, we squash on merge anyway.

Yes for simple PRs I agree that that's the way to go. 
If the PR is a bit more complex and the commit history makes sense I'd advise to rebase and keep the history.

idanmiara:
@jorisvandenbossche I think I've fixed everything as per your comments

idanmiara:
@jorisvandenbossche @kylebarron @brendan-ward I think this is ready for review.

idanmiara:
rebased to resolve conflicts with main

idanmiara:
@jorisvandenbossche @kylebarron @brendan-ward 
rebased to resolve conflicts with main
would you like to review?

idanmiara:
@jorisvandenbossche any further thoughts about this PR?

jorisvandenbossche:
In the meantime GEOS 3.12 added a GEOSOrientPolygons function, which was exposed as `orient_polygons` top-level ufunc in https://github.com/shapely/shapely/pull/2147. But so for now that ufunc requires GEOS > 3.12, and the implementation in this PR can still be used to provide a python fallback for older GEOS, so that `shapely.orient_polygons` can be available for all users regardless the GEOS version they have installed.

So I updated this PR to rename `force_ccw` to `orient_polygons` and use that in the top-level function for older GEOS. I simplified some of the helper functions to just that what is needed for this PR (e.g. using `np.frompyfunc` directly instead of the decorator) but reused the main logic, and added some more tests (my initial PR adding `shapely.orient_polygons` was not covering all cases that is needed for having a custom implementation).

Thanks @idanmiara for the original PR! (and again apologies for the unresponsiveness here)

Issue #1693, opened 2022-12-27
BUG: Debugging any shapely module fails due to relative imports and name clash causing circular import of the `enum` module


idanmiara:
## Expected behavior and actual behavior.

Expected behavior:

It should be possible to run shapely modules in debug mode in PyCharm without getting errors

Actual behavior:

```
C:\Python311\python.exe "C:/Program Files/JetBrains/PyCharm Community Edition 2022.3/plugins/python-ce/helpers/pydev/pydevd.py" --multiprocess --qt-support=auto --client 127.0.0.1 --port 52915 --file C:\dev\shapely\shapely\errors.py 
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2022.3\plugins\python-ce\helpers\pydev\pydevd.py", line 16, in <module>
    from _pydevd_bundle.pydevd_collect_try_except_info import collect_return_info
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2022.3\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_collect_try_except_info.py", line 12, in <module>
    from _pydevd_bundle.pydevd_constants import IS_PY38_OR_GREATER
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2022.3\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_constants.py", line 5, in <module>
    import platform
  File "C:\Python311\Lib\platform.py", line 117, in <module>
    import re
  File "C:\Python311\Lib\re\__init__.py", line 124, in <module>
    import enum
  File "C:\dev\shapely\shapely\enum.py", line 1, in <module>
    from enum import IntEnum
ImportError: cannot import name 'IntEnum' from partially initialized module 'enum' (most likely due to a circular import) (C:\dev\shapely\shapely\enum.py)

Process finished with exit code 1

```

Or running a doctest:

```
[...]

PR #1694, opened 2022-12-27
BUG: enum.py renamed to _enum.py to avoid name clash with the built in `enum` module

idanmiara:
Closes https://github.com/shapely/shapely/issues/1693


jorisvandenbossche:
Renaming sounds good to me, but I would suggest using `_enum.py` instead of `shapely_enum.py`

idanmiara:
> Renaming sounds good to me, but I would suggest using `_enum.py` instead of `shapely_enum.py`

oops, missed your comment. :)
Done.

jorisvandenbossche:
Thanks for the update. Small left-over issue: you accidentally committed `shapely/msvcp140.dll`, can you remove that again? 

(and https://github.com/shapely/shapely/pull/1703 is now also merged to prevent that in the future ;))

idanmiara:
> Thanks for the update. Small left-over issue: you accidentally committed `shapely/msvcp140.dll`, can you remove that again?
> 
> (and #1703 is now also merged to prevent that in the future ;))

Sure😄, now fixed and rebased. 
Thanks !

idanmiara:
Thanks @jorisvandenbossche

PR #1703, opened 2023-01-02
ENH:.gitignore - add *.dylib, *.dll to ignore binaries from MacOS and Windows

idanmiara:
`.gitignore` already includes `*.so`. 
This PR adds also ` *.dylib`, `*.dll` to ignore this binary file types on MacOS and Windows.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@ __pycache__/
 
 # C extensions
 *.so
+*.dylib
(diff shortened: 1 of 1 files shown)


jorisvandenbossche (approved):
Thanks!

idanmiara:
Thanks @jorisvandenbossche !

PR #1704, opened 2023-01-03
ENH: Allow getting matplotlib patch from polygon without plotting

sea-bass:
This PR stems from my use case to get a Matplotlib patch from a Shapely polygon, but manipulate it in [my own GUI](https://github.com/sea-bass/pyrobosim) instead of directly plotting it.

This adds a new `patch_from_polygon()` function that is internally called by `plot_polygon()` but can also be used standalone to return a patch without plotting. In the process I also did some slight cleanup.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -128,7 +128,7 @@ jobs:
           else
             pip install --upgrade Cython numpy==${{ matrix.numpy }} pytest pytest-cov coveralls;
           fi
-          if [ -n "${{ matrix.numpy }}" ]; then
+          if [ -n "${{ matrix.matplotlib }}" ]; then
             pip install matplotlib
           fi
           pip list
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,14 +40,12 @@ jobs:
           - python: "3.11"
             geos: 3.11.1
             numpy: 1.23.4
-            matplotlib: true
             doctest: true
             # TODO(shapely-2.0) temporary allow warnings
             # extra_pytest_args: "-W error"  # error on warnings
           # dev
           - python: "3.11"
             geos: main
-            matplotlib: true
             extra_pytest_args: "-W error" # error on warnings
           # enable two 32-bit windows builds:
           - os: windows-2019
@@ -129,9 +127,7 @@ jobs:
           else
             pip install --upgrade Cython numpy==${{ matrix.numpy }} pytest pytest-cov coveralls;
           fi
-          if [ -n "${{ matrix.numpy }}" ]; then
-            pip install matplotlib
-          fi
+          pip install matplotlib
           pip list
 
       - name: Set environment variables (Linux)
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,12 +40,14 @@ jobs:
           - python: "3.11"
             geos: 3.11.1
             numpy: 1.23.4
+            matplotlib: true
             doctest: true
             # TODO(shapely-2.0) temporary allow warnings
             # extra_pytest_args: "-W error"  # error on warnings
           # dev
           - python: "3.11"
             geos: main
+            matplotlib: true
             extra_pytest_args: "-W error" # error on warnings
           # enable two 32-bit windows builds:
           - os: windows-2019
@@ -127,7 +129,9 @@ jobs:
           else
             pip install --upgrade Cython numpy==${{ matrix.numpy }} pytest pytest-cov coveralls;
           fi
-          pip install matplotlib
+          if [ -n "${{ matrix.numpy }}" ]; then
+            pip install matplotlib
+          fi
           pip list
 
       - name: Set environment variables (Linux)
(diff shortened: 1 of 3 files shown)


jorisvandenbossche (commented):
That sounds reasonable, thanks for the PR!

Can you add a small test calling this new public function? (just a dummy test ensuring you can pass both a Polygon and MultiPolygon and get back a matplotlib PathPatch, is fine, it's mostly to assure we keep the function as is, and don't accidentally refactor it away)

shapely/plotting.py
```diff
@@ -5,28 +5,62 @@
...
+    ----------
+    polygon : shapely.Polygon or shapely.MultiPolygon
+    facecolor : matplotlib color specification
+        Color for the polygon fill.
+    edgecolor : matplotlib color specification
+        Color for the polygon boundary.
+    linewidth : float
+        The line width for the polygon boundary.
+    **kwargs
+        Additional keyword arguments passed to the matplotlib Patch.
```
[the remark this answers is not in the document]
sea-bass: Good idea -- that worked great.

sea-bass:
Interesting. The failures are all matplotlib import errors:

```
shapely/plotting.py:9: in <module>
    import matplotlib.pyplot as plt
E   ModuleNotFoundError: No module named 'matplotlib'
```

sea-bass:
The CI rules for matplotlib were kind of weird, in that `matplotlib` was only installed when a specific `numpy` version was provided.  Updated to always install it regardless of how numpy was installed.

shapely/tests/test_plotting.py
```diff
@@ -1,18 +1,41 @@
 import pytest
 from numpy.testing import assert_allclose
 
-import shapely
-from shapely import box, LineString, MultiLineString, Point
-from shapely.plotting import plot_line, plot_points, plot_polygon
+from shapely import box, get_coordinates, LineString, MultiLineString, Point
+from shapely.plotting import patch_from_polygon, plot_line, plot_points, plot_polygon
 
 pytest.importorskip("matplotlib")
```
sea-bass, line 7: I see... it's because I moved the imports to the top of `plotting.py` so even when matplotlib wasn't installed before, `pytest` would skip here.

sea-bass:
These last failures seem unrelated to the PR, as the last commits all passed and then I made a nonfunctional change.

Maybe worth seeing if someone broke on main and then re-running the jobs?

shapely/tests/test_plotting.py
mwtoews: I prefer to see this kept to skip, since matplotlib is not a requirement.

mwtoews:
CI failures are unrelated (https://github.com/libgeos/geos/issues/795)

shapely/tests/test_plotting.py
sea-bass: Sounds good. Do you want me to revert what I did in the GitHub actions config then? Because there are some jobs that straight up did not install matplotlib prior to this PR

mwtoews: yes, probably those too, thanks!

jorisvandenbossche: >  Because there are some jobs that straight up did not install matplotlib prior to this PR

That was intentional: since matplotlib is optional, we should have some builds without it, to ensure shapely properly works without it.

.github/workflows/tests.yml
```diff
@@ -128,9 +127,7 @@ jobs:
           else
             pip install --upgrade Cython numpy==${{ matrix.numpy }} pytest pytest-cov coveralls;
           fi
-          if [ -n "${{ matrix.numpy }}" ]; then
```
jorisvandenbossche, line 131: Whoops, I suppose the intention was that this check was `matrix.matplotlib` instead of `matrix.numpy` ..

sea-bass: I see! Will update accordingly a little later.

shapely/plotting.py
```diff
@@ -5,28 +5,53 @@
 exploration, debugging and illustration purposes.
 
 """
+
+import matplotlib.pyplot as plt
 import numpy as np
+from matplotlib import colors
+from matplotlib.patches import PathPatch
+from matplotlib.path import Path
```
mwtoews, line 13: these top-level imports will need to be nested to the functions below, since matplotlib is not a required dependency

sea-bass:
I think I reverted all the unintended things and it should be good to go!

sea-bass:
Seems like we got more unrelated failures...

All 3 of the `Py3.11 GEOS main` checks failed with the same failing test:

```
FAILED shapely/tests/test_set_operations.py::test_set_operation_array[difference-a7] - shapely.errors.GEOSException: IllegalArgumentException: Overlay input is mixed-dimension
```

idanmiara:
> Seems like we got more unrelated failures...
> 
> All 3 of the `Py3.11 GEOS main` checks failed with the same failing test:
> 
> ```
> FAILED shapely/tests/test_set_operations.py::test_set_operation_array[difference-a7] - 
> shapely.errors.GEOSException: IllegalArgumentException: Overlay input is mixed-dimension
> ```

Right, https://github.com/shapely/shapely/pull/1715

jorisvandenbossche (approved):
Thanks!

PR #1715, opened 2023-01-12
TST: skip failing test on GEOS master

idanmiara:
Skip failing test (`shapely.errors.GEOSException: IllegalArgumentException: Overlay input is mixed-dimension`) with GEOS master till that issue is resolved upstream

Related to: https://github.com/libgeos/geos/issues/797

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -166,6 +166,7 @@ jobs:
         # Enable this if we have failures on GEOS main
         # continue-on-error: ${{ matrix.geos == 'main' }}
         run: |
+          python -c "import shapely; print(f'GEOS version: {shapely.geos_version_string}')"
           pytest shapely/tests tests --cov --cov-report term-missing ${{ matrix.extra_pytest_args }}
 
       # Only run doctests on 1 runner (because of typographic differences in doctest results)
--- a/shapely/tests/test_set_operations.py
+++ b/shapely/tests/test_set_operations.py
@@ -46,13 +46,15 @@ non_polygon_types = [
 
 @pytest.mark.parametrize("a", all_types)
 @pytest.mark.parametrize("func", SET_OPERATIONS)
-def test_set_operation_array(a, func):
+def test_set_operation_array(request, a, func):
     if (
         func == shapely.difference
         and a == geometry_collection
         and shapely.geos_version >= (3, 12, 0)
     ):
-        pytest.xfail("https://github.com/libgeos/geos/issues/797")
+        request.node.add_marker(
+            pytest.mark.xfail(reason="https://github.com/libgeos/geos/issues/797")
+        )
     actual = func(a, point)
     assert isinstance(actual, Geometry)
 
--- a/shapely/tests/test_set_operations.py
+++ b/shapely/tests/test_set_operations.py
@@ -6,7 +6,7 @@ from shapely import Geometry, GeometryCollection, Polygon
 from shapely.errors import UnsupportedGEOSVersionError
 from shapely.testing import assert_geometries_equal
 
-from .common import all_types, empty, multi_polygon, point, polygon
+from .common import all_types, empty, geometry_collection, multi_polygon, point, polygon
 
 # fixed-precision operations raise GEOS exceptions on mixed dimension geometry collections
 all_single_types = [g for g in all_types if not shapely.get_type_id(g) == 7]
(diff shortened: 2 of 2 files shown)


mwtoews:
This is a recent issue with GEOS master, and should be fixed upstream

https://github.com/libgeos/geos/issues/797

idanmiara:
> This is a recent issue with GEOS master, and should be fixed upstream
> 
> [libgeos/geos#797](https://github.com/libgeos/geos/issues/797)

Yes, thanks, I thought to add a skip for this test when run against GEOS master till it fixed

idanmiara:
@mwtoews what do you think ?
Also - about https://github.com/libgeos/geos/issues/798 ?

mwtoews:
I'm waiting to see the outcome from the discussion, as it's not a straightforward bug.

Generally skips should be implemented for releases, as it's typical to see blips from development versions of libraries, which are usually ironed-out before they are released.

jorisvandenbossche:
@idanmiara in case we want to skip this test for the dev version, we already have the machinery to skip tests based on the version, so in this case it would be something like `@pytest.mark.skipif(shapely.geos_version >= (3, 12, 0), reason="...")` for skipping a full test, or something like you did in the diff. So I don't understand why something like https://github.com/libgeos/geos/issues/798 would be needed?

idanmiara:
> @idanmiara in case we want to skip this test for the dev version, we already have the machinery to skip tests based on the version, so in this case it would be something like `@pytest.mark.skipif(shapely.geos_version >= (3, 12, 0), reason="...")` for skipping a full test, or something like you did in the diff. So I don't understand why something like [libgeos/geos#798](https://github.com/libgeos/geos/issues/798) would be needed?

Hi,
I think it's not possible to skip a specific parameterized test (only difference on collection)
So either split the test, pytest skip it entirely or skip as I did. I don't mind modifying this if you advise.
libgeos/geos#798 would be required to make sure we only skip it on `3.12.0dev` and not on `3.12.0`

jorisvandenbossche:
> I think it's not possible to skip a specific parameterized test

You can do that, with `pytest.param(val, marks=pytest.mark.skipif(..))`, although that still depends on how the parameterization is written if that is easy to inject. 

Anyway, that's not really necessary here. We can also just call `pytest.skip()` inside the test function. And moreover, depending on the discussion on the GEOS side, we might actually want to check explicitly that it raises an error instead of skipping the test.

jorisvandenbossche:
> [libgeos/geos#798](https://github.com/libgeos/geos/issues/798) would be required to make sure we only skip it on `3.12.0dev` and not on `3.12.0`

I think in general we assume that whatever is in a development version can always become the final release, so we never made that distinction before. In case that this would have been a bug and we wanted to skip it right now, but it would get fixed on the GEOS side before the final release, we can just update the test then to remove the skip again. 
(and actually, if we want to get notified that this changes on the GEOS side, and the test would start passing again, we probably should use an xfail instead of skip)

idanmiara:
> > I think it's not possible to skip a specific parameterized test
> 
> You can do that, with `pytest.param(val, marks=pytest.mark.skipif(..))`, although that still depends on how the parameterization is written if that is easy to inject.
> 
> Anyway, that's not really necessary here. We can also just call `pytest.skip()` inside the test function. And moreover, depending on the discussion on the GEOS side, we might actually want to check explicitly that it raises an error instead of skipping the test.

Ok, Sounds good, so do you want me to modify this to call `pytest.skip()` inside the test function

jorisvandenbossche:
If the decision on the GEOS side is that this is not supported and the error is expected, then I think we can also change the test to reflect that (assert we get an error with pytest.raises(..):)

idanmiara:
> If the decision on the GEOS side is that this is not supported and the error is expected, then I think we can also change the test to reflect that (assert we get an error with pytest.raises(..):)

I changed to `xfail` on that case - I think that it serves the purpose the best. 
I wouldn't asserting failing a test that passed in previous version and might pass again in a future version (that suggests that passing on a future version is actually a bug) - I think expected failure would still be appropriate.

idanmiara:
@jorisvandenbossche  is this PR ready now ?

idanmiara:
@jorisvandenbossche @mwtoews What do you think of this proposed solution?
I find it annoying that all our other PRs are "failing" because of this unrelated issue.

jorisvandenbossche:
Yes, thanks, I think xfailing the test for now is good. We can always later change that in case there is a more definitive decision on the GEOS side. 

I just pushed a small commit to change the `pytest.xfail(..)` to `request.node.add_marker(pytest.mark.xfail(..))`. That's a bit more complicated, but ensures the test is still actually run and so pytest will report it in case the test would actually pass ("xpass"). For example in case a future GEOS version would fix this, we would then notice that.

jorisvandenbossche:
Thanks! I opened a PR to actually make the xfails strict by default, so to ensure what I said above is actually true ;) https://github.com/shapely/shapely/pull/1729

PR #1719, opened 2023-01-13
FIX: Copy geometries input to STRtree to prevent modification instead of adding write lock to array

### LINKED ISSUE #1714: Possible negative side-effect: STRtree makes passed geometry array immutable
#1422 included a change to mark the passed-in geometries as non-writeable, and included a [note](https://github.com/shapely/shapely/pull/1422#discussion_r896004573) that we should probably check downstream usages.

I've been running into an issue with this lately with Shapely 2.0 / GeoPandas 0.12.2 where I create the tree on a GeoDataFrame's geometry column, then want to update the contents of that column selectively.

```python
tree = shapely.STRtree(df.geometry.values)
left, right = tree.query(...other geoms..., predicate='intersects')

# construct some new geometries, perhaps via intersection of geometries indexed by left and right

# fails now:
df.loc[...some index..., 'geometry'] = new_geoms
```

I had previously been using this pattern frequently with pygeos without issue (the non-writeable flag was added to shapely2, not present in pygeos).  In cases where I'd do the above and then need to do another spatial index operation, I'd simply recreate the tree from the latest geometries (sidestepping the immutability issue) and query again.

Deleting the tree does not reset the non-writeable flag; it probably should.

This may be an edge case, but there are a couple work-arounds we may want to mention in the docs:
- make a copy of the geometry array as input to the tree if you plan on modifying it later
- delete the tree after querying to enable modifying the geometry array again (once we reset the flag on delete)

brendan-ward:
Resolves #1714 

Instead of adding a write lock to the geometry array input to construct the tree, we now take a copy of the input array, which allows the user to modify the input array later without modifying the geometries stored in the tree.

### DIFF
--- a/shapely/strtree.py
+++ b/shapely/strtree.py
@@ -45,6 +45,11 @@ class STRtree:
 
     Any mixture of geometry types may be stored in the tree.
 
+    The tree automatically adds a write lock to the input geometry array to
+    prevent modification of geometries indexed in the tree.  You can avoid this
+    by passing a copy of the geometry array as input or by deleting the tree
+    before modifying elements of the input geometry array.
+
     Note: the tree is more efficient for querying when there are fewer
     geometries that have overlapping bounding boxes and where there is greater
     similarity between the outer boundary of a geometry and its bounding box.
@@ -77,10 +82,15 @@ class STRtree:
         # Keep references to geoms
         self._geometries = np.asarray(geoms, dtype=np.object_)
         # prevent modification
+        self._prev_writeable = self._geometries.flags.writeable
         self._geometries.flags.writeable = False
 
         # initialize GEOS STRtree
         self._tree = lib.STRtree(self.geometries, node_capacity)
+
+    def __del__(self):
+        # release write lock on geometry array
+        self._geometries.flags.writeable = self._prev_writeable
 
     def __len__(self):
         return self._tree.count
--- a/shapely/strtree.py
+++ b/shapely/strtree.py
@@ -45,11 +45,6 @@ class STRtree:
 
     Any mixture of geometry types may be stored in the tree.
 
-    The tree automatically adds a write lock to the input geometry array to
-    prevent modification of geometries indexed in the tree.  You can avoid this
-    by passing a copy of the geometry array as input or by deleting the tree
-    before modifying elements of the input geometry array.
-
     Note: the tree is more efficient for querying when there are fewer
     geometries that have overlapping bounding boxes and where there is greater
     similarity between the outer boundary of a geometry and its bounding box.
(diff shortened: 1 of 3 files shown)


shapely/strtree.py
```diff
@@ -77,11 +82,16 @@ def __init__(
...
         # prevent modification
+        self._prev_writeable = self._geometries.flags.writeable
         self._geometries.flags.writeable = False
 
         # initialize GEOS STRtree
         self._tree = lib.STRtree(self.geometries, node_capacity)
 
+    def __del__(self):
+        # release write lock on geometry array
+        self._geometries.flags.writeable = self._prev_writeable
```
jorisvandenbossche, line 93: Should we put this in a try/except to be sure? I have seen other instances where this was done to prevent AttributeError in case the attribute was already deleted before this is called (not sure about the exact circumstances this happens, but in general one shouldn't rely on `__del__` always being called at the moment you expect)

brendan-ward: Just so I follow, do you mean something like this?

```python
def __del__(self):
    try:
        self._geometries.flags.writeable = self._prev_writeable
    except:
        pass
```

So that it doesn't break if either `self._geometries` or `self._prev_writeable` has already been deleted?

jorisvandenbossche: Yes

shapely/strtree.py
```diff
@@ -45,6 +45,11 @@ class STRtree:
 
     Any mixture of geometry types may be stored in the tree.
 
+    The tree automatically adds a write lock to the input geometry array to
+    prevent modification of geometries indexed in the tree.  You can avoid this
+    by passing a copy of the geometry array as input or by deleting the tree
```
jorisvandenbossche, line 50: Should we mention that this is actually a relatively cheap operation? (passing `arr.copy()`) 

Copying a numpy object array only creates a new array of pointers to the Python objects, but the actual geometries are not duplicated.

jorisvandenbossche: To illustrate:

```
In [8]: coords = np.random.randn(1_000_000, 2)

In [9]: arr = shapely.points(coords)

In [10]: %timeit shapely.points(coords)
176 ms ± 3.26 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [11]: %timeit arr.copy()
15.5 ms ± 248 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

So copying one million geometries takes 15ms (and here it are simple points, but it should take as long for complex geometries).

This is actually also faster than creating the tree:
```
In [16]: %timeit shapely.STRtree(arr)
253 ms ± 5.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [17]: %timeit shapely.STRtree(arr.copy())
275 ms ± 3.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

So should we maybe consider just copying ourselves under the hood? (instead of asking the user to do this)

brendan-ward: Thanks for the additional testing!  Given that it is fast enough and low-overhead memory wise, I agree that we can just do that automatically, and it sidesteps need to lock the array or ask the user to pass a copy / delete the tree.

shapely/tests/test_strtree.py
```diff
@@ -78,19 +72,6 @@ def test_init_with_invalid_geometry():
         STRtree(["Not a geometry"])
 
 
-def test_init_increases_refcount():
```
brendan-ward, line 81: The ref count tests and asserts no longer applicable now that we copy geometry array input to the tree

shapely/strtree.py
```diff
@@ -74,10 +74,9 @@ def __init__(
         geoms: Iterable[BaseGeometry],
         node_capacity: int = 10,
     ):
-        # Keep references to geoms
-        self._geometries = np.asarray(geoms, dtype=np.object_)
-        # prevent modification
-        self._geometries.flags.writeable = False
+        # Keep references to geoms in a copied array so that this array is not
+        # modified while the tree depends on it remaining the same
+        self._geometries = np.asarray(geoms, dtype=np.object_).copy()
```
jorisvandenbossche, line 79:
```suggestion
        self._geometries = np.array(geoms, dtype=np.object_, copy=True)
```

That would avoid doing a second copy in case the input `geoms` is not already a numpy array but eg a list

jorisvandenbossche (approved):
Looks good!

PR #1720, opened 2023-01-15
BUG: geometry constructors now support `Decimal` inputs

### LINKED ISSUE #1707: Shapely 2.0 doesn't accept Decimal Fields in the Polygon constructor.
Before the Shapely 2.0 update our code worked:

```
    def to_geo(self):
        p1 = (self.Longitude1, self.Latitude1)
        p2 = (self.Longitude2, self.Latitude2)
        p3 = (self.Longitude3, self.Latitude3)
        p4 = (self.Longitude4, self.Latitude4)
        p5 = (self.Longitude1, self.Latitude1)
        poly_coords = (p1, p2, p3, p4, p5)

        poly = Polygon(poly_coords)
        poly = poly.convex_hull

        return poly
```
now it returns a:
`ValueError: Inconsistent coordinate dimensionality`

I've run this also with the latest on master and gotten the same result. We are able to mitigate this by casting the Decimals to floats, but I'd rather not if this is truly a regression and not an intended change.

now it 

```
 # the following are latitude/longitude values normally on a Django model, but I've just changed them to the value for reproducibility. 
 from decimal import Decimal
 p1 = (Decimal('-76.572088'), Decimal('39.256796'))
 p2 = (Decimal('-76.569245'), Decimal('39.258841'))
 p3 = (Decimal('-76.567123'), Decimal('39.257046'))
 p4 = (Decimal('-76.571833'), Decimal('39.255359'))
 p5 = (Decimal('-76.572088'), Decimal('39.256796'))
 poly_coords = (p1, p2, p3, p4, p5)
 
 poly = Polygon(poly_coords)
 poly = poly.convex_hull
```

## Operating system

Ubuntu 22.04.1 LTS, but also recreated in a python docker image

shapely==2.0.0 installed from PyPI using pip

idanmiara:
Closes #1707

### DIFF
--- a/shapely/geometry/point.py
+++ b/shapely/geometry/point.py
@@ -76,6 +76,8 @@ class Point(BaseGeometry):
         if not np.issubdtype(coords.dtype, np.number):
             coords = [float(c) for c in coords]
         geom = shapely.points(coords)
+        if not isinstance(geom, Point):
+            raise ValueError("Invalid values passed to Point constructor")
         return geom
 
     # Coordinate getters and setters
--- a/shapely/geometry/point.py
+++ b/shapely/geometry/point.py
@@ -71,8 +71,7 @@ class Point(BaseGeometry):
 
         if coords.ndim > 1:
             raise ValueError(
-                "Invalid values passed to Point constructor, "
-                "only size-1 arrays are accepted"
+                f"Point() takes only scalar or 1-size vector arguments, got {args}"
             )
         if not np.issubdtype(coords.dtype, np.number):
             coords = [float(c) for c in coords]
--- a/shapely/geometry/point.py
+++ b/shapely/geometry/point.py
@@ -64,15 +64,19 @@ class Point(BaseGeometry):
             # Accept either (x, y) or [(x, y)]
             if not hasattr(coords, "__getitem__"):  # generators
                 coords = list(coords)
-
             coords = np.asarray(coords).squeeze()
         else:
             # 2 or 3 args
             coords = np.array(args).squeeze()
 
+        if coords.ndim > 1:
+            raise ValueError(
+                "Invalid values passed to Point constructor, "
+                "only size-1 arrays are accepted"
+            )
+        if not np.issubdtype(coords.dtype, np.number):
+            coords = [float(c) for c in coords]
         geom = shapely.points(coords)
-        if not isinstance(geom, Point):
-            raise ValueError("Invalid values passed to Point constructor")
         return geom
 
     # Coordinate getters and setters
--- a/shapely/tests/geometry/test_decimal.py
+++ b/shapely/tests/geometry/test_decimal.py
@@ -51,34 +51,29 @@ items3d = [
     [(90.0, 10, 3.0), (110.0, 10.0, 4.0), (100.0, 30.0, 5.0), (90.0, 10.0, 6.0)],
 ]
 
+all_geoms = [
+    [
+        Point(items[0][0]),
+        Point(*items[0][0]),
+        MultiPoint(items[0]),
+        LinearRing(items[0]),
+        LineString(items[0]),
+        MultiLineString(items),
+        Polygon(items[0]),
+        MultiPolygon(
+            [
+                Polygon(items[2]),
+                Polygon(items[0], holes=items[1:]),
+                Polygon(items[0], holes=items[2:]),
+            ]
+        ),
+        GeometryCollection([Point(items[0][0]), Polygon(items[0])]),
+    ]
+    for items in [items2d, to_decimal(items2d), items3d, to_decimal(items3d)]
+]
 
-@pytest.mark.parametrize("items", [items2d, items3d])
-def test_decimal(items):
-    assert Point(items[0][0]) == Point(to_decimal(items[0][0]))
-    assert Point(*items[0][0]) == Point(*to_decimal(items[0][0]))
-    assert MultiPoint(items[0]) == MultiPoint(to_decimal(items[0]))
-    assert LinearRing(items[0]) == LinearRing(to_decimal(items[0]))
-    assert LineString(items[0]) == LineString(to_decimal(items[0]))
-    assert MultiLineString(items) == MultiLineString(to_decimal(items))
-    assert Polygon(items[0]) == Polygon(to_decimal(items[0]))
-    assert Polygon(items[0], holes=items[1:]) == Polygon(
-        to_decimal(items[0]), holes=to_decimal(items[1:])
-    )
-    assert MultiPolygon(
-        [
-            Polygon(items[2]),
-            Polygon(items[0], holes=items[1:]),
-            Polygon(items[0], holes=items[2:]),
-        ]
-    ) == MultiPolygon(
-        [
-            Polygon(to_decimal(items[2])),
-            Polygon(to_decimal(items[0]), holes=to_decimal(items[1:])),
-            Polygon(to_decimal(items[0]), holes=to_decimal(items[2:])),
-        ]
-    )
-    assert GeometryCollection(
-        [Point(items[0][0]), Polygon(items[0])]
-    ) == GeometryCollection(
-        [Point(to_decimal(items[0][0])), Polygon(to_decimal(items[0]))]
-    )
+
+@pytest.mark.parametrize("geoms", list(zip(*all_geoms)))
+def test_decimal(geoms):
+    assert geoms[0] == geoms[1]
+    assert geoms[2] == geoms[3]
--- a/shapely/tests/geometry/test_decimal.py
+++ b/shapely/tests/geometry/test_decimal.py
@@ -13,42 +13,74 @@ from shapely import (
     Polygon,
 )
 
-
-def to_decimal(x):
-    if isinstance(x, (int, float)):
-        return Decimal(x)
-    if isinstance(x, tuple):
-        return tuple([to_decimal(x) for x in x])
-    if isinstance(x, list):
-        return list([to_decimal(x) for x in x])
-    return x
-
-
-def test_to_decimal():
-    assert to_decimal(1) == Decimal(1)
-    assert to_decimal([1]) == [Decimal(1)]
-    assert to_decimal([1, 2]) == [Decimal(1), Decimal(2)]
-    assert to_decimal((1, 2)) == (Decimal(1), Decimal(2))
-    assert to_decimal([(1, 2), [3], 4, [5, 6, 7]]) == [
-        (Decimal(1), Decimal(2)),
-        [Decimal(3)],
-        Decimal(4),
-        [Decimal(5), Decimal(6), Decimal(7)],
-    ]
-
-
 items2d = [
     [(0.0, 0.0), (70.0, 120.0), (140.0, 0.0), (0.0, 0.0)],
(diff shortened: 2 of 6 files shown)


shapely/tests/geometry/test_point.py
```diff
@@ -89,7 +89,7 @@ def test_from_invalid():
         Point(1, 2, 3, 4)
 
     # this worked in shapely 1.x, just ignoring the other coords
-    with pytest.raises(ValueError, match="Invalid values passed"):
+    with pytest.raises(TypeError, match="only size-1 arrays"):
```
jorisvandenbossche, line 92: The current error is not super informative, but this is also not an improvement (and also ValueError is better I think). I'll add a comment on the actual code about maybe catching this explicitly and raising a better error message.

shapely/geometry/point.py
```diff
@@ -64,12 +64,12 @@ def __new__(self, *args):
             # Accept either (x, y) or [(x, y)]
             if not hasattr(coords, "__getitem__"):  # generators
                 coords = list(coords)
-
             coords = np.asarray(coords).squeeze()
         else:
             # 2 or 3 args
             coords = np.array(args).squeeze()
 
+        coords = [float(c) for c in coords]
```
jorisvandenbossche, line 72: We could avoid doing this if we already have a numeric numpy array. So maybe something like:

```
if not np.issubdtype(coords.dtype, np.number):
    coords = [float(c) for c in coords]
```

and related to the other comment, we could maybe also first check explicitly for >1d array case to provide a better error message (untested):

```
if coords.ndim > 1:
    raise ValueError("...")
```

jorisvandenbossche (commented):
Thanks! Just a some minor points about the Point constructor

shapely/geometry/point.py
idanmiara: Done.

shapely/tests/geometry/test_decimal.py
```diff
@@ -0,0 +1,84 @@
...
+    [(60.0, 80.0, 1.0), (80.0, 80.0, 2.0), (70.0, 60.0, 3.0), (60.0, 80.0, 1.0)],
+    [(30.0, 10.0, 2.0), (50.0, 10.0, 3.0), (40.0, 30.0, 4.0), (30.0, 10.0, 2.0)],
+    [(90.0, 10, 3.0), (110.0, 10.0, 4.0), (100.0, 30.0, 5.0), (90.0, 10.0, 6.0)],
+]
+
+
+@pytest.mark.parametrize("items", [items2d, items3d])
+def test_decimal(items):
+    assert Point(items[0][0]) == Point(to_decimal(items[0][0]))
+    assert Point(*items[0][0]) == Point(*to_decimal(items[0][0]))
```
sgillies, line 58: @idanmiara every instance of `assert` in this function should be a different test. Please find a way to parameterize or add new `def` statements.

sgillies (changes requested):
I have a major change request with regards to the test and a couple other suggestions.

shapely/tests/geometry/test_decimal.py
```diff
@@ -0,0 +1,84 @@
...
+    [(0.0, 0.0, 1.0), (70.0, 120.0, 2.0), (140.0, 0.0, 3.0), (0.0, 0.0, 1.0)],
+    [(60.0, 80.0, 1.0), (80.0, 80.0, 2.0), (70.0, 60.0, 3.0), (60.0, 80.0, 1.0)],
+    [(30.0, 10.0, 2.0), (50.0, 10.0, 3.0), (40.0, 30.0, 4.0), (30.0, 10.0, 2.0)],
+    [(90.0, 10, 3.0), (110.0, 10.0, 4.0), (100.0, 30.0, 5.0), (90.0, 10.0, 6.0)],
+]
+
+
+@pytest.mark.parametrize("items", [items2d, items3d])
+def test_decimal(items):
+    assert Point(items[0][0]) == Point(to_decimal(items[0][0]))
```
sgillies, line 57: I don't agree with the use of to_decimal here. It's unnecessarily indirect. Why not just give instances of Decimal, explicitly?

shapely/geometry/point.py
```diff
@@ -64,15 +64,19 @@ def __new__(self, *args):
...
             coords = np.asarray(coords).squeeze()
         else:
             # 2 or 3 args
             coords = np.array(args).squeeze()
 
+        if coords.ndim > 1:
+            raise ValueError(
+                "Invalid values passed to Point constructor, "
+                "only size-1 arrays are accepted"
+            )
```
sgillies, line 76: It looks like this code isn't tested. We need a failing test to demonstrate that lines 72-76 are needed.

idanmiara: It is tested in the original test, see discussion about it above:
https://github.com/shapely/shapely/pull/1720#discussion_r1071623475

shapely/tests/geometry/test_decimal.py
idanmiara: OK, done.

shapely/tests/geometry/test_decimal.py
idanmiara: I'm not sure if you were referring to the use of the function `to_decimal` in general in this test or just to the `Point` test.
Anyway, I've parametrized the test as you requested above. 
If you'd prefer to remove `to_decimal` I could hardcode `Decimal` in copies of `items2d` and `items3d` and remove `to_decimal` - otherwise please explain what you meant please.

idanmiara:
> I have a major change request with regards to the test and a couple other suggestions.

No problem. Thanks for your review. Please advise if this is OK now.

shapely/geometry/point.py
```diff
@@ -64,15 +64,19 @@ def __new__(self, *args):
...
-
             coords = np.asarray(coords).squeeze()
         else:
             # 2 or 3 args
             coords = np.array(args).squeeze()
 
+        if coords.ndim > 1:
+            raise ValueError(
+                "Invalid values passed to Point constructor, "
+                "only size-1 arrays are accepted"
```
jorisvandenbossche, line 75: You can remove this second part, as that is exactly the confusing error message from passing an array to the Decimal(..) constructor (which happens if ndim > 1) that we want to avoid raising by checking this manually here. 

Maybe we can append the first part with something like ".. can only pass 2 or 3 coordinates".

idanmiara: `TypeError(f"Point() takes at most 3 arguments ({len(args)} given)")`  - is raised when more then 3 arguments are given
so maybe in a similar way: `ValueError(f"Point() takes only scalar or 1-size vector arguments, got {args}")`

shapely/tests/geometry/test_decimal.py
sgillies: Here's what I've come to understand about using convenience functions in tests. I used to do this somewhat often. But then a year or two later when I wanted to refactor code or update a feature, I would discover that the convenience function wasn't appropriate anymore and actually hindered the refactoring work. This is the awful truth about tests: they can easily make future changes hard. Shapely needs tests, but not too many, and not too fancy.

Please do remove to_decimal.

shapely/tests/geometry/test_decimal.py
```diff
@@ -0,0 +1,79 @@
...
+    [(30.0, 10.0), (50.0, 10.0), (40.0, 30.0), (30.0, 10.0)],
+    [(90.0, 10), (110.0, 10.0), (100.0, 30.0), (90.0, 10.0)],
+]
+
+items3d = [
+    [(0.0, 0.0, 1.0), (70.0, 120.0, 2.0), (140.0, 0.0, 3.0), (0.0, 0.0, 1.0)],
+    [(60.0, 80.0, 1.0), (80.0, 80.0, 2.0), (70.0, 60.0, 3.0), (60.0, 80.0, 1.0)],
+    [(30.0, 10.0, 2.0), (50.0, 10.0, 3.0), (40.0, 30.0, 4.0), (30.0, 10.0, 2.0)],
+    [(90.0, 10, 3.0), (110.0, 10.0, 4.0), (100.0, 30.0, 5.0), (90.0, 10.0, 6.0)],
+]
```
sgillies, line 52: @idanmiara I thought of another very important point!

Having multiple cases is good. However, all of these are basically the same case. Casting all of them to float will succeed because they are all well-formed numbers. We can reduce these lists to a single item and we would have equally good testing. There's a lot of redundancy here and that can bite us in the future if we change anything.

We only really need one 2D item and one 3D item to test the successful cases here.

To test weird edge cases we would want to check decimals that can't be exactly represented by floats (like by having hundreds of decimal places, etc). But that doesn't have to be done in this PR :sweat_smile:

shapely/tests/geometry/test_decimal.py
sgillies: This is important work @idanmiara and you've solved the problem well! I want to help you get the tests right.

idanmiara: @sgillies thanks, I see your point. while removing `to_decimal` I've added another test for mixed `decimal` and non-decimal inputs.

shapely/tests/geometry/test_decimal.py
```diff
@@ -0,0 +1,117 @@
...
+        items3d_mixed,
+        items3d_decimal,
+    ]
+]
+
+
+@pytest.mark.parametrize("geoms", list(zip(*all_geoms)))
+def test_decimal(geoms):
+    assert geoms[0] == geoms[1] == geoms[2]
+    assert geoms[3] == geoms[4] == geoms[5]
```
sgillies, line 117: Since we only changed the LinearRing, LineString, and Point constructors my preference would be to have just `test_linearring()`, `test_linestring()`, and `test_point()` within `test_decimal.py`. I think that would keep this test module easy to read and modify over time.

However, I don't insist on that. The work here is good and the changes I am thinking of could be done later.

Thanks @idanmiara !

sgillies (approved):

shapely/tests/geometry/test_decimal.py
idanmiara: Thanks for your review @sgillies !

> Since we only changed the LinearRing, LineString, and Point constructors...

I'm included to keep all these tests since implementations sometimes change, and we want to make sure this doesn't break in future refactors.

If you could find the time I'd be happy if you could review my other Shapely opened PRs
Thanks again!

shapely/geometry/point.py
```diff
@@ -64,12 +64,17 @@ def __new__(self, *args):
...
 
+        if coords.ndim > 1:
+            raise ValueError(
+                f"Point() takes only scalar or 1-size vector arguments, got {args}"
+            )
+        if not np.issubdtype(coords.dtype, np.number):
+            coords = [float(c) for c in coords]
         geom = shapely.points(coords)
         if not isinstance(geom, Point):
             raise ValueError("Invalid values passed to Point constructor")
```
jorisvandenbossche, line 80: @idanmiara I added this back. I know this is currently not needed (I don't think a test covers this), and I also can't directly think of a case where `shapely.points(..)` would not return a single Point that wouldn't already be catched by the `if coords.ndim > 1:` check. 
But, Point() really should ever only return a single Point, and `shapely.points()` doesn't guarantee that, so it feels safer to keep this anyway (our tests might not be complete).

jorisvandenbossche (approved):
Thanks!

Issue #987, opened 2020-09-17
parsing geometry type from WKB values returns bad values if Zs (and maybe for Ms and ZMs as well)


davidraleigh:
## Expected behavior and actual behavior.

when parsing binary from wkb, I would expect the 4 bytes following the endianness flag byte to be within the range described by the specification ([example](https://en.wikipedia.org/wiki/Well-known_text#Well-known_binary) from Wikipedia of values ranging from `0000` to `3024`). 

for example for a point with z values I should get 1001, but instead I'm getting 2147483649. for a multipolygon with z values I would expect 1006, but instead I get 2147483654

It looks like maybe Z isn't supported in geos for wkb?
https://github.com/libgeos/geos/blob/f43aa53d27b3889823365317fcf4ff45e60dc90f/include/geos/io/WKBConstants.h

## Steps to reproduce the problem.

```python
from shapely.wkb import dumps
from shapely.geometry import Point, MultiPolygon
point = Point(1, 2)
d_point = dumps(point)
endianness = "big"
if int.from_bytes(d_point[0:1], "big") == 1:
    endianness = "little"
self.assertEqual(1, int.from_bytes(point.wkb[1:5], endianness))

multi_polygon = MultiPolygon([(((0.0, 0.0), (0.0, 1.0), (1.0, 1.0), (1.0, 0.0)),
                               [((0.25, 0.25), (0.25, 0.5), (0.5, 0.5), (0.5, 0.25))])])
d_multi_polygon = dumps(multi_polygon)
self.assertEqual(6, int.from_bytes(d_multi_polygon[1:5], endianness))

point_z = Point(1, 2, 3)
d_point_z = dumps(point_z)
self.assertIsNotNone(d_point_z)
self.assertEqual(1001, int.from_bytes(d_point_z[1:5], endianness))

multi_polygon_z = MultiPolygon([(((0.0, 0.0, 0.0), (0.0, 1.0, 0.0), (1.0, 1.0, 0.0), (1.0, 0.0, 0.0)),
                               [((0.25, 0.25, 0.0), (0.25, 0.5, 0.0), (0.5, 0.5, 0.0), (0.5, 0.25, 0.0))])])
d_multi_polygon_z = dumps(multi_polygon_z)
self.assertEqual(1006, int.from_bytes(d_multi_polygon_z[1:5], endianness))
```

## Operating system

Mac OS X 10.15.3

## Shapely version and provenance

`Shapely==1.7.1`

[...]

Issue #984, opened 2020-09-16
API: exposing CoordinateSequence in Shapely 2.0


jorisvandenbossche:
One of the questions I run into while doing the Geometry classes refactor (subclassing PyGEOS geometry -> https://github.com/Toblerity/Shapely/pull/983), is what to do with CoordinateSequence.

So right now, if you access the `.coords` of a Shapely Geometry you get a `CoordinateSequence` object (at least for Point, LineString and LinearRing, the geometry types that are directly backed by a coordinate sequence).

Some examples usages that are possible now:

```python
>>> from shapely.geometry import LineString  
>>> geom = LineString([(0, 0), (0, 1), (1, 1)])
>>> print(geom) 
LINESTRING (0 0, 0 1, 1 1)

>>> geom.coords   
<shapely.coords.CoordinateSequence at 0x7f8663bd5438>

# slicing gives a list of tuples (this is done often in the tests)
>>> geom.coords[:]   
[(0.0, 0.0), (0.0, 1.0), (1.0, 1.0)]

# accessing single element gives (x, y) or (x, y, z) tuple
>>> geom.coords[0] 
(0.0, 0.0)

# iterating over coordinate tuples
>>> for x, y in geom.coords: 
...     print(x, y) 
0.0 0.0
0.0 1.0
1.0 1.0

# the xy attribute gives two 1D arrays (from the stdlib array module)
>>> geom.coords.xy
(array('d', [0.0, 0.0, 1.0]), array('d', [0.0, 1.0, 1.0]))

# convert to a 2D numpy array
>>> np.asarray(geom.coords) 
array([[0., 0.],
       [0., 1.],
       [1., 1.]])
```

Apart from the exact API (eg tuples and not numpy arrays), the current implementation also ensures you can *lazily* iterate over the coordinates or get a specific coordinate pair (instead of needing to convert all coordiantes of the full Geometry to a list/array first).

For Shapely 2.0, I think there are some different options (from most to least compatible with current shapely):

1) We also implement a `CoordinateSequence` extension type that maps to `GEOSCoordSequence` (like `pygeos.Geometry` wraps a `GEOSGeometry`), so that we can keep the exact same interface and the lazy conversion aspect.  
  However, this also gives quite some additional code complexity (implementing yet another extension type in C, although the code itself will be quite similar as the Geometry extension type).

[...]

PR #1722, opened 2023-01-16
ENH:base.py - add `coords_array` and `xy_array` properties

idanmiara:
This PR adds `coords_array` and `xy_array` properties to geometries. 
These are in addition to the existing `coords` and `xy` which are slower, and stay for backwards compatibility.

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -207,6 +207,18 @@ class BaseGeometry(shapely.Geometry):
         return CoordinateSequence(coords_array)
 
     @property
+    def coords_array(self):
+        """Access to geometry's full coordinates array [N, 2|3] (xy/xyz depending on coordinate dimensionality)"""
+        coords_array = shapely.get_coordinates(self, include_z=self.has_z)
+        return coords_array
+
+    @property
+    def xy_array(self):
+        """Access to geometry's xy coordinates array [N, 2] (regardless of coordinate dimensionality)"""
+        coords_array = shapely.get_coordinates(self)
+        return coords_array
+
+    @property
     def xy(self):
         """Separate arrays of X and Y coordinate values"""
         raise NotImplementedError
--- a/shapely/tests/geometry/test_linestring.py
+++ b/shapely/tests/geometry/test_linestring.py
@@ -1,5 +1,6 @@
 import numpy as np
 import pytest
+from numpy.testing import assert_array_equal
 
 import shapely
 from shapely import LinearRing, LineString, Point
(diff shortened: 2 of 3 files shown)


jorisvandenbossche:
For `coords` vs `coords_array`, that gives basically the same speed if you convert `geom.coords` to a numpy array:

```
In [1]: import shapely

In [2]: geom = shapely.LineString(np.random.randn(1000, 2))

In [3]: %timeit geom.coords_array
13.1 µs ± 805 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [4]: %timeit np.asarray(geom.coords)
13.2 µs ± 1.61 µs per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

So I think this is more about convenience than about speed. 

But then I am not sure that is worth it. I originally thought to just make this `.coords` attribute return directly the numpy array instead of this custom class, but that didn't make it in 2.0 (xref https://github.com/shapely/shapely/issues/984)

idanmiara:
> 

Thanks for this pointer. 
As you noted in #987, `.coords` doesn't return all the coordinates, and not implemented for polygons, multi*, and collections.
I don't see the point of having it besides for backwards compatibility.
Also I can't justify the overhead of converting to  `CoordinateSequence` and then just convert back to a numpy array - Thus the proposed solution.

jorisvandenbossche:
> `.coords` doesn't return all the coordinates, and not implemented for polygons, multi*, and collections.

It does return all the coordinates of the types it is implemented for AFAIK. It's indeed not implemented for the other types you mention, but this has been done intentionally (it's currently only implemented for geometry types that are backed by a single CoordinateSequence), so that would require some discussion if we actually want to change that (personally, I think a function like `shapely.get_coordinates` is a more flexible interface for that). 

> Also I can't justify the overhead of converting to `CoordinateSequence` and then just convert back to a numpy array

What do you mean exactly with "can't justify the overhead"? I think my example above shows that there is hardly any overhead?

idanmiara:
I can see a few reasons to add `.coords_array` as a separate property (and the same for `.xy` vs `.xy_array`):

1. backwards compatibility - with the intended use of `.coords_array` vs `.coords`- Some tests fail when I used `.coords_array` as `.coords`.
2. `.coords` was intentionally not implemented for multi part geometries - in some cases it might be appropriate to use `.coords_array` for multi part geometries.
3. speed - On my machine I get ~+30% overhead with your benchmark
4. convenience - much more convenient to use the property as is without casting it to a numpy array
5. advising users to use `np.array(geom.coords)` would imply that there is a reason to cast to `CoordinateSequence` and back to `np.array`, which I find confusing.

> personally , I think a function like `shapely.get_coordinates` is a more flexible interface for that

Many functions in `BaseGeometry` are just thin wrappers around some `shapely.function(self)` - It's just more convenient when you deal with a single geometry...

idanmiara:
@brendan-ward any thoughts about this PR?

caspervdw:
Closing because this PR was inactive and there is a general lack of enthusiasm for this addition. 

See https://github.com/shapely/shapely/pull/1448 for further discussion on the subject of closing inactive PRs

PR #1727, opened 2023-01-21
DOC: Improve docstring for prepared

### LINKED ISSUE #1726: Contains_xy not working with prepared polygon
Expected behavior:

prepare(polygon) should work with contains_xy according to documentation, but only works with an unprepared polygon. I'm trying to speed up code that is checking ~1,000,000 points to determine which lie within the polygon. Using an unprepared polygon this takes ~3.5 seconds.

Example code that is functioning properly, but very slowly:

```
x = np.linspace(36, 54, 1800), dtype='float32')
y = np.linspace(25, 31, 530), dtype='float32')

X, Y = np.meshgrid(x, y)
X_flat, Y_flat = X.flatten(order='F'), Y.flatten(order='F')

contains_xy(polygon, X_flat, Y_flat)
```

Shown below is a plot of the polygon in blue and the mesh grid points in black, it's clear the majority of them are within the polygon:

image

The code below returns all FALSE values for contains_xy:

```
x = np.linspace(36, 54, 1800), dtype='float32')
y = np.linspace(25, 31, 530), dtype='float32')

X, Y = np.meshgrid(x, y)
X_flat, Y_flat = X.flatten(order='F'), Y.flatten(order='F')

contains_xy(prepare(polygon), X_flat, Y_flat)
```

Windows 10 64 bit
Python version 3.11
Shapely version 2.0.0 installed from PyPI using pip

brendan-ward:
Closes #1726

### DIFF
--- a/shapely/creation.py
+++ b/shapely/creation.py
@@ -484,10 +484,12 @@ def prepare(geometry, **kwargs):
     This function does not recompute previously prepared geometries;
     it is efficient to call this function on an array that partially contains prepared geometries.
 
+    This function does not return any values; geometries are modified in place.
+
     Parameters
     ----------
     geometry : Geometry or array_like
-        Geometries are changed inplace
+        Geometries are changed in place
     **kwargs
         For other keyword-only arguments, see the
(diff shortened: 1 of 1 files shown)


jorisvandenbossche (approved):
Thanks!

PR #1735, opened 2023-01-23
TST: Cleanup STRtree xfails & skipifs

brendan-ward:
This address [#1729 comment](https://github.com/shapely/shapely/pull/1729#issue-1552080758) and cleans up xfails that were not actually necessary, removes xfails related to GEOS 3.5 (outdated), and removes version guards for < GEOS 3.6 since that is now the min version tested.


jorisvandenbossche (approved):
Ah, good, thanks!

PR #1687, opened 2022-12-24
Add proper type annotations for `array like` + usage examples

idanmiara:
Closes https://github.com/shapely/shapely/issues/1683
This PR is a proposed solution for adding proper type annotations for `array like` and also replace the current documentation statements like `Geometry or array_like` with 2 usage examples.

Issue #721, opened 2019-05-25
How about adding type hints?


GeorgySk:
As Shapely currently doesn't have [type hints](https://www.python.org/dev/peps/pep-0484/), results produced by type checkers are far from perfect. Some examples:
```python
from shapely.geometry import Polygon, LineString, Point

area = Polygon().area  # Inferred type: Any. Should be: float
geom_type = LineString().geom_type  # Inferred type: Union[str, List[str]]
# Should be LineString in this case
representative_point = Polygon().representative_point()
# Inferred type: Union[Polygon, Any]. Should be: Point
x = Point(1, 1).x  # Inferred type: Union[object, Tuple[object, object, object], Tuple[object, object]]
# Should be: float
```
and so on.

So, how about adding [stub files](https://mypy.readthedocs.io/en/latest/stubs.html) with the type hints, at least for the most basic functionality from the docs?

---
Shapely version: 1.6.4.post1, installed from conda.

sgillies:
I'm not opposed to adding stub files. However, since I'm not a mypy user (yet) and there is a lot of code to annotate, I am not going to take on this task. I would be happy to review a PR, especially if it had tests of the annotations.

snorfalorpagus:
I would love to see type annotations in Shapely.

mwtoews:
Annotations would be nice to have, perhaps for the 1.8 milestone (?)

One thing to consider planning this feature is whether to put annotations into stub files to maintain Python 2.7 compatibility for future versions of shapely. Or, make the next milestone (e.g. 1.8) release Python 3+ only, and maintain a separate branch for Python 2.7 bugs.

Note that Python 2.7 will not be maintained past 2020 ([about 7 months from now](https://pythonclock.org/)).

mwtoews:
Python 3.4 reached it's [end-of-life in March 2019](https://www.python.org/downloads/release/python-3410/), and #680 addressed that for Travis CI tests, so I think requiring Python 3.5+ for the next milestone (1.8?) makes sense.

But this means we'd need to maintain a 1.7 branch for Python 2.7 users (also <=3.4).

@sgillies is it time to break away from the Python 2.x era?

sgillies:
Let's revisit this after 2.0.

sgillies:
@mwtoews in https://github.com/shapely/shapely/discussions/1316 it looks like we welcome first time contributors to add type hints. I disagree and think that this is work that should be done by core developers who understand the API very well.

[...]

PR #1741, opened 2023-01-29
`base.py` - add type annoations

idanmiara:
Addition type annotations to `base.py` as per https://github.com/shapely/shapely/pull/1687#issuecomment-1407043373


jorisvandenbossche (commented):
A few inline comments, will post some general comments in the issue

shapely/geometry/base.py
```diff
@@ -308,33 +322,33 @@ def geom_type(self):
...
 
     @property
-    def area(self):
-        """Unitless area of the geometry (float)"""
+    def area(self) -> float:
+        """Unitless area of the geometry"""
         return float(shapely.area(self))
 
-    def distance(self, other):
+    def distance(self, other: MaybeGeometryArrayNLike) -> MaybeArrayN[float]:
```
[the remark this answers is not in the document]
idanmiara: > At the same time, while `MaybeGeometryArrayNLike` is shorter, I personally also don't find it super clear what this means. What's the "maybe"? (maybe a geometry, or maybe an array?) 

Also possible - `GeometryMaybeArrayNLike`, then you can have `Optional[GeometryMaybeArrayNLike]`.

> More descriptive would be something "GeometryOrGeometryArray"? But I do wonder if it's not clearer if we would use the more verbose but more explicit `Geometry | ArrayLike[Geometry]` in the code itself.

I agree that `None | Geometry | ArrayLike[Geometry]` is very readable, but this is Python >=3.10 syntax. 
I'm not sure that `Optional[Union[Geometry, ArrayLike[Geometry]]` is a readable.

Although I personally prefer `Optional[GeometryMaybeArrayNLike]` over the more verbose `Optional[Union[Geometry, ArrayLike[Geometry]]` - I don't have a strong opinion here. So any other wording is acceptable for me.

shapely/_typing.py
```diff
@@ -0,0 +1,121 @@
...
+        Polygon,
+    )
+    from shapely.geometry.base import BaseGeometry, BaseMultipartGeometry  # NOQA
+
+# todo: specify shapes once resolved: Numpy arrays with specific shapes: https://github.com/numpy/numpy/issues/16544
+
+NumpyArrayN = NumpyArray  # shape = (N,)
+NumpyArray2 = NumpyArray  # shape = (2,)
+NumpyArray3 = NumpyArray  # shape = (3,)
+NumpyArray4 = NumpyArray  # shape = (4,)
```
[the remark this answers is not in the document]
idanmiara: sure

shapely/_typing.py
```diff
@@ -0,0 +1,26 @@
...
+    NumpyArray = Union[np.ndarray, Sequence[T]]
+    ArrayLike = Union[NumpyArray[T], Sequence[T]]
+
+if TYPE_CHECKING:
+    from shapely import Geometry  # NOQA
+
+MaybeArray = Union[T, NumpyArray[T]]
+MaybeArrayLike = Union[T, ArrayLike[T]]
+
+GeometryMaybeArray = MaybeArray["Geometry"]
```
parched, line 25: Personally, I think this alias adds a small bit of obfuscation, without any more conciseness, so isn't really needed.

shapely/_typing.py
```diff
@@ -0,0 +1,26 @@
...
+    import numpy as np
+
+    # This is a fallback workaround for numpy < 1.20 (which is no longer supported since Jun 21, 2022)
+    NumpyArray = Union[np.ndarray, Sequence[T]]
+    ArrayLike = Union[NumpyArray[T], Sequence[T]]
+
+if TYPE_CHECKING:
+    from shapely import Geometry  # NOQA
+
+MaybeArray = Union[T, NumpyArray[T]]
```
parched, line 22: ScalarOrArray is more intuitive to me, but that might just be because I'm not familiar with the use of "maybe" in this codebase.

shapely/geometry/base.py
```diff
@@ -308,33 +321,33 @@ def geom_type(self):
...
 
     @property
-    def area(self):
-        """Unitless area of the geometry (float)"""
+    def area(self) -> float:
+        """Unitless area of the geometry"""
         return float(shapely.area(self))
 
-    def distance(self, other):
+    def distance(self, other: GeometryMaybeArrayLike) -> MaybeArray[float]:
```
parched, line 328: Would this be more precise?

```suggestion
    @typing.overload
    def distance(self, other: Geometry) -> float:
        ...
    @typing.overload
    def distance(self, other: ArrayLike[Geometry]) -> Array[float]:
        ...
    def distance(self, other: GeometryMaybeArrayLike) -> MaybeArray[float]:
```

shapely/_typing.py
idanmiara: Are you referring to `GeometryMaybeArray = MaybeArray["Geometry"]` ? Yes, I can see you point, I'll remove that.

shapely/_typing.py
idanmiara: That's a good name 👍

shapely/geometry/base.py
idanmiara: I'm not familiar with this syntax, is `def distance(self, other: GeometryMaybeArrayLike) -> MaybeArray[float]:` needed if you add the first two ?
But still, there are 27 functions like this in the unit alone, I'm afraid that this might add up to a lot of overhead. WDYT ?

shapely/geometry/base.py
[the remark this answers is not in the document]
idanmiara: @jorisvandenbossche @parched 
I've revised and minimalized `_typing` as per your suggestions so I'm marking this conversation as resolved.

shapely/geometry/base.py
parched: Yes, you need that last one which has the actual implementation. Personally, I think it's worth the overhead, but I can understand if the Shapely maintainers say it isn't. In that case, just typing it as `def distance(self, other: Geometry) -> float` would be my preferred option, leaving the array case as an advanced option, not supported by the typing. How often is the array case used? We don't use it in our project, but unsure about most of Shapely's users.

idanmiara: > How often is the array case used? We don't use it in our project, but unsure about most of Shapely's users.

Probably the more common use of the array inputs/outputs is in the top level functions which accept array also as the first geometry (instead of `self`) so I guess it makes more sense there. 
In that case also I don't think that the [generics solution](https://github.com/shapely/shapely/issues/721#issuecomment-1409410818) would apply because there can be 2 `ScalarOrArrayLike` inputs (like `shapely.intersection`) and only if both are `scalar`s than the output is `scalar`, and otherwise it's an `array`.
So for the top level functions for sure we need to type the arrays, but for `base` I don't have a strong opinion.

caspervdw: I think the object-oriented API of shapely is *mostly* used by people doing only scalar geometries. For these users, the typing doesn't really add value if you do (for instance) `distance(...) -> MaybeArray[float]`. A `cast` is then always necessary.

I am a bit torn on the solutions @parched suggested, but leaning towards the option of going all the way with the added `overload`. I'd like to hear the opinion of another maintainer on this @jorisvandenbossche  @sgillies . Would you accept added `@overload` decorated methods for a 100% correct typing?

idanmiara:
@caspervdw 
> As a generic remark; I am personally 👍 on the `from __future__ import annotations`, which would allow you to drop all the quotes around the types.
> 

Forward references were deferred from python 3.11 because it might cause problems, see:
https://lukasz.langa.pl/61df599c-d9d8-4938-868b-36b67fdb4448/
Thus I'm -1 for using `from __future__ import annotations`

shapely/geometry/base.py
idanmiara: >  Would you accept added `@overload` decorated methods for a 100% correct typing?

Yes.

caspervdw:
@idanmiara  I'd like to resume this subject of typing in shapely. Take for example the `Geometry.contains`. In this PR we have:

```
def contains(self, other: ScalarOrArrayLike["Geometry"]) -> ScalarOrArray[bool]:
```

Most people (@parched) work with scalar geometries. For these users, the problem here is that this will always need a `cast(bool, result)` to convince the typing that the return value is a `bool`. These users will actually find the following more useful:

```
def contains(self, other: "Geometry") -> bool:
    ...
```

Which, in my opinion, is also far more comprehensible. However, the more correct solution (as stated before would be using overloads:

```
@typing.overload
def contains(self, other: "Geometry") -> bool:
    ...

@typing.overload
def contains(self, other: ArrayLike["Geometry"]) -> Array[bool]:
    ...

def contains(self, other: ScalarOrArrayLike["Geometry"]) -> ScalarOrArray[bool]:
    ...  # actual code goes here
```

However, that is a lot of extra lines to cover and although it is good from a static type analysis point of view, it is harder to read. 

We have to decide on this to go forward on typing.  @idanmiara @jorisvandenbossche  @sgillies  could you give your opinion on this?

My personal perference is going with the scalar-only version (option 2) and postpone the arrays for later. Maybe we can experiment a bit with stub files, or something.

idanmiara:
Hi @caspervdw
Thanks for your input!
I agree about the scalar only version for the `Geometry` methods.

I think that we might consider the overload version for the shapely top level functions. I would assume that users who use vectorized input would use these mostly and not the `Geometry` methods wrappers.

sgillies:
@caspervdw I do not approve of how types are being figured out in a PR. I want to see an RFC or PEP type document explaining what the goal and scope of typing is and I want to see consensus on it before any work is considered. It feels to me like effort is being prematurely poured into this, and the effort is then being used to justify acceptance. I don't think that's how we should proceed.

idanmiara:
> @caspervdw I do not approve of how types are being figured out in a PR. I want to see an RFC or PEP type document explaining what the goal and scope of typing is and I want to see consensus on it before any work is considered. It feels to me like effort is being poured into this, and the effort is then being used to make a case for action. I don't think that's how we should proceed.

@caspervdw what did you mean by saying
> We have to decide on this to go forward on typing.
?

Did the maintainers have an internal discussion about it already? 
I don't mind writing an RFC if there is a consensus about including typing and we only need to decide the details. If you give me the the scope it could also help.

mwtoews:
Shapely RFCs go to https://github.com/shapely/shapely-rfc and see inspiration is from NumPy's [NEP 41](https://numpy.org/neps/nep-0041-improved-dtype-support.html) and [NEP 42](https://numpy.org/neps/nep-0042-new-dtypes.html).

sgillies:
@mwtoews Thanks! I acknowledge that I'm raising the bar for working on type hints, and I probably owe the project some work on templating or reviewing the existing template.

Maybe, maybe, with such a document, a sponsor might be convinced to pay Idan to do such work, and pay others to review it. We've seen other sponsored Shapely work, and more would be fantastic.

PR #1743, opened 2023-01-30
CLN: Reformat `/tests` using `pre-commit`

idanmiara:
Reformat `/tests` using `pre-commit` before https://github.com/shapely/shapely/pull/1681


jorisvandenbossche:
Maybe you could also run pre-commit just on all files here? The the reformatting is a single commit

brendan-ward (approved):
Thanks for pulling out formatting changes separately @idanmiara 

I am 👍 on the multiple steps outlined by @jorisvandenbossche in [#1681 comment](https://github.com/shapely/shapely/pull/1681#issuecomment-1409449891).  While there is certainly lots of work to do improving the tests, I personally am OK with a couple of large changesets that are purely about formatting (this one) and then moving (#1681) so that these follow a consistent formatting style and are located in the same subtree as the other tests.

jorisvandenbossche:
Thanks!

PR #1744, opened 2023-01-30
FIX: Fix read-only buffer error in `from_ragged_array`

kylebarron:
Previously calling `shapely.from_ragged_array` on a read-only buffer (such as a coordinates buffer owned by Arrow memory) you'd get an error that the array is read-only

```
File shapely/_geometry_helpers.pyx:96, in shapely._geometry_helpers.simple_geometries_1d()

File stringsource:660, in View.MemoryView.memoryview_cwrapper()

File stringsource:350, in View.MemoryView.memoryview.__cinit__()

ValueError: buffer source array is read-only
```

Since shapely is copying into GEOS memory anyways, it's fine for the buffer to be read-only and only needed to be marked as such with `const`

```py
cdef const double[:, :]
```

I have my editor set to automatically trim trailing whitespace on save. If it's important I can remove the unrelated changes.

### DIFF
--- a/shapely/tests/test_ragged_array.py
+++ b/shapely/tests/test_ragged_array.py
@@ -86,6 +86,17 @@ def test_include_z_default():
     assert coords.shape[1] == 2
 
 
+@pytest.mark.parametrize("geom", all_types)
+def test_read_only_arrays(geom):
+    # https://github.com/shapely/shapely/pull/1744
+    typ, coords, offsets = shapely.to_ragged_array([geom, geom])
+    coords.flags.writeable = False
+    for arr in offsets:
+        arr.flags.writeable = False
+    result = shapely.from_ragged_array(typ, coords, offsets)
+    assert_geometries_equal(result, [geom, geom])
+
+
 @pytest.mark.parametrize("geom", all_types_not_supported)
 def test_raise_geometry_type(geom):
     with pytest.raises(ValueError):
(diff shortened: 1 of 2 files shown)


kylebarron:
I'm not too familiar with the internals here; are there other places where this change makes sense too?

jorisvandenbossche:
Thanks, good catch! 
(that's what you get from not properly testing it anymore in your own code before releasing .. my python-geoarrow POC was still using pygeos, and I actually had an explicit `coords.copy()` in there to avoid this bug with a "# TODO copy is needed because of read-only memoryview bug in pygeos" comment ... ;))

> I'm not too familiar with the internals here; are there other places where this change makes sense too?

I pushed a small test that at least should cover multiple geometry types, and this seems sufficient for that (in the end all geometry types go through the simple_geometries_1d). The ufunc-based constructor functions (eg `points()`, or `polygons()` without passing indices) already work fine with read-only arrays.

jorisvandenbossche (approved):

shapely/tests/test_ragged_array.py
```diff
@@ -86,6 +86,17 @@ def test_include_z_default():
     assert coords.shape[1] == 2
 
 
+@pytest.mark.parametrize("geom", all_types)
+def test_read_only_arrays(geom):
+    # https://github.com/shapely/shapely/pull/1744
+    typ, coords, offsets = shapely.to_ragged_array([geom, geom])
+    coords.flags.writeable = False
+    for arr in offsets:
+        arr.flags.writeable = False
```
kylebarron, line 95: ah cool; I didn't know you could set a buffer as immutable so easily!

jorisvandenbossche: Yeah, in numpy nothing is ever _really_ immutable, as you can always easily change that flag to True :)

brendan-ward (approved):
Thanks @kylebarron @jorisvandenbossche

Issue #1648, opened 2022-12-03
ENH: add basic support for M values (parsing, checking for M, accessing values)


jorisvandenbossche:
With GEOS main (future GEOS 3.12), there is improved support for M values. Those can now be stored under the hood in the CoordinateSequence. The M value isn't used for much (quite similarly as Z values), but we could provide basic functionality to parse/create geometries with M values, check if a geometry has M values and access the M value. 

For example, that would support the workflow of parsing WKB/WKT with M and extracting that value. The parsing now works with GEOS main:

```
>>> geom = shapely.from_wkt("POINT M (1 2 3)")                                                                                                                                                                 
>>> geom                                                                                                                                                                                                       
<POINT M (1 2 3)>
```

But there is currently no way to access this value 3. 

Some possible places to add support for M values:

* Add a ``has_m`` function and class attribute (mimicking `has_z`). We should probably ask upstream to add a `GEOSHasM` to the C API.
* Add an `include_m` keyword in `get_coordinates`. Currently this function defaults to return 2D, but can return 3D if you specify `include_z=True`

idanmiara:
I was wondering how useful is the M coordinate, as I've never seen it actually used in the wild :)

> * Add a `has_m` function and class attribute (mimicking `has_z`). We should probably ask upstream to add a `GEOSHasM` to the C API.

As calling `has_z` and `has_m` in a row can be wasteful, It might be a good idea to also have a function `has_zm` that will either return Nx2 bool array (has z, has m, per geom) or  a N int8 array (first bit=has z, second bit=has m, per geom)

> * Add an `include_m` keyword in `get_coordinates`. Currently this function defaults to return 2D, but can return 3D if you specify `include_z=True`

In the same way, `include_z: int | bool` could be considered (2 bits), although have a confusing name, could be backwards compatible with `include_z: bool` without the addition of extra parameter.

Also we need to consider if there is code where `coord_dim=3` was considered equivalent to `has_z`. which now might not be the case with XYM geometries.

mwtoews:
[...]

PR #1752, opened 2023-02-04
BUG: fix handling of z coordinates in shapely.ops.substring

### LINKED ISSUE #1699: BUG: shapely.ops.substring issue with LineString Z in shapely 2.0.0
I used the `shapely.ops.substring` method in my project. There was no issue in shapely `1.8.4`. Today I upgraded to shapely `2.0.0` and I'm getting an error from calling this:
```
shapely.ops.substring(trajectory, dist1, dist2)
```
where the trajectory is a `LineString` type (LINESTRING Z), it contains Z coordinates.

```
setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (14,) + inhomogeneous part
```

It looks like only X, and Y of the `start_point` and `end_point` is used in the `vertex_list` for constructing the `LineString`

I'm using:
`Ubuntu 20.04`
`shapely 2.0.0`

songololo:
Closes #1699


mwtoews (approved):
Thanks, this does resolve the issue.

One future consideration is handling of coordinate dimensions (see #1648), since a three-item tuple may be either represent XYZ or XYM dimensions. But also, the next version of GEOS, 3.12, will have GEOSLineSubstring which I believe handles extended dimensions. While not yet implemented (#1570), I would expect the GEOS C implementation to be used over Shapely's own.

jorisvandenbossche (approved):
Thanks!

I merged the main branch to fix a conflict (the tests were moved)

PR #1776, opened 2023-02-27
TST: suppress warnings when creating geometry with NaNs in test suite (for GEOS main)

jorisvandenbossche:
The tests with GEOS main are failing (eg https://github.com/shapely/shapely/actions/runs/4282147357/jobs/7456054311) because of warnings generated when creating a geometry with NaNs as test data (we don't allow warnings in that test build). 

This is caused by a change in GEOS to already calculate the envelope upon geometry construction (and this calculation involves NaNs and raises floating point exceptions): https://github.com/libgeos/geos/issues/832

This change is probably there to stay in GEOS (it are valid warnings), and we can easily suppress those warnings in our own test suite.

### DIFF
--- a/shapely/tests/test_geometry.py
+++ b/shapely/tests/test_geometry.py
@@ -16,6 +16,7 @@ from shapely.tests.common import (
     empty_polygon,
     geometry_collection,
     geometry_collection_z,
+    ignore_invalid,
     line_string,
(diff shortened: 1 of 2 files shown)


brendan-ward (approved):
Thanks @jorisvandenbossche

PR #1786, opened 2023-03-15
Check for any `MultiPolygon`s when constructing a `MultiPolygon`

ianwilliamson:
Currently, the `MultiPolygon` constructor only validates the type of the first element of the `polygons` input. When a later element does not have the correct type, things silently fail.


jorisvandenbossche (commented):
Thanks for the PR! 
Could you add a test for it as well? (in `shapely/tests/geometry/test_multipolygon.py`, I think)

jorisvandenbossche:
Hey @ianwilliamson would you have time to add a small test for this?

jorisvandenbossche (approved):
Thanks @ianwilliamson for the PR, just pushed a small test

PR #1797, opened 2023-04-03
Consolidate point creation

caspervdw:
Small refactor in preparation of doing #1573 for points.

### DIFF
--- a/src/ufuncs.c
+++ b/src/ufuncs.c
@@ -435,7 +435,7 @@ static void Ydd_b_p_func(char** args, const npy_intp* dimensions, const npy_intp
 #if GEOS_SINCE_3_12_0
       ret = func(ctx, prepared_geom_tmp, in2, in3);
 #else
-      geom = create_point(ctx, in2, in3);
+      geom = create_point(ctx, in2, in3, NULL);
       if (geom == NULL) {
         if (destroy_prepared) {
           GEOSPreparedGeom_destroy_r(ctx, prepared_geom_tmp);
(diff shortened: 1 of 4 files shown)


mwtoews:
I realize this PR is a minor refactor, but is there is future scope to consider #1648 and further expand the functions signature to:
```
create_point(GEOSContextHandle_t ctx, double x, double y, double* z, double* m)
```
just checking now to ensure this refactor will be complementary with future work with XYZM and XYM ordinates.

caspervdw:
> I realize this PR is a minor refactor, but is there is future scope to consider #1648 and further expand the functions signature to:
> 
> ```
> create_point(GEOSContextHandle_t ctx, double x, double y, double* z, double* m)
> ```
> 
> just checking now to ensure this refactor will be complementary with future work with XYZM and XYM ordinates.

It is possible (and easy) to extend this to M coordinates (the signature you are proposing is fine); however I'd like to keep it out of scope of this PR. I would need to mock around with GEOS version conditionals.

src/ufuncs.c
```diff
@@ -2475,24 +2476,8 @@ static void points_func(char** args, const npy_intp* dimensions, const npy_intp*
...
-        errstate = PGERR_GEOS_EXCEPTION;
-        GEOSCoordSeq_destroy_r(ctx, coord_seq);
-        destroy_geom_arr(ctx, geom_arr, i - 1);
-        goto finish;
-      }
-    }
-    geom_arr[i] = GEOSGeom_createPoint_r(ctx, coord_seq);
-    // Note: coordinate sequence is owned by point; if point fails to construct, it will
-    // automatically clean up the coordinate sequence
+    geom_arr[i] = create_point(ctx, *(double*)ip1, *(double*)(ip1 + cs1),
```
brendan-ward, line 2479: This is the only part that is hard to follow without a bit more context because of the use of `qip1 + n*cs1` to step over latter dimensions, since that isn't immediately obvious from looking at the implementation of `SINGLE_COREDIM_LOOP_OUTER`.

Suggest adding a comment to explain this a bit more, for future reference.

brendan-ward (approved):
Thanks for working on this @caspervdw and apologies for the long delay in review.

Overall this looks good and the use of `NULL` to signal presence of a Z coordinate seems reasonable.  One minor suggestion re: adding a comment.

PR #1802, opened 2023-04-10
Set minimum versions: GEOS 3.7, NumPy 1.16, Python 3.8

mwtoews:
This PR upgrades the following versions for shapely-2.1 development:

- GEOS 3.7 (from GEOS 3.5) see [NEWS](https://github.com/libgeos/geos/blob/3.7/NEWS)
- NumPy 1.16 (from 1.14) see [NumPy 1.16.0 Release Notes](https://numpy.org/devdocs/release/1.16.0-notes.html)
- Python 3.8 (from 3.7) see [What’s New In Python 3.8](https://docs.python.org/3/whatsnew/3.8.html)

Most of the differences are due to dropping GEOS 3.5 and 3.6 logic.

One advantage in setting a minimum NumPy 1.16 is to undo https://github.com/pygeos/pygeos/pull/69/commits/6bece4f731ed283ae0b09fcf7460fd32a4388d05 (from https://github.com/pygeos/pygeos/pull/69), since gufunc signatures with fixed-size dimensions was added with that version.

There are no changes in this PR that would break Python 3.7 support, but this should be dropped simply because [PEP 537](https://peps.python.org/pep-0537/) specifies an EOL in June 2023. Breaking changes, like the walrus operator can happen later.

### Xref:

- https://github.com/shapely/shapely/discussions/1798
- https://github.com/shapely/shapely/issues/1000


jorisvandenbossche (approved):
Thanks a lot, looks good to me!

PR #1803, opened 2023-04-10
CI/BLD: update GEOS minor versions

mwtoews:
Update `maint-2.0` to use the latest GEOS minor versions for CI testing.

Also set the version for the binary wheels for the next maintenance release to use GEOS 3.11.2; see [NEWS.md](https://github.com/libgeos/geos/blob/3.11/NEWS.md) for changes.

### DIFF
--- a/ci/install_geos.sh
+++ b/ci/install_geos.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 # Build and install GEOS on a POSIX system, save cache for later
 #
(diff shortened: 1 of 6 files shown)


ci/install_geos.sh
```diff
@@ -36,13 +36,28 @@ build_geos(){
...
+    rm -rf build
     mkdir build
     cd build
-     # Use Ninja on windows, otherwise, use the platform's default
+    # Use Ninja on Windows, otherwise, use the platform's default
     if [ "$RUNNER_OS" = "Windows" ]; then
         export CMAKE_GENERATOR=Ninja
     fi
-    cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$GEOS_INSTALL -DCMAKE_INSTALL_LIBDIR=lib ..
+    # Avoid building tests, depends on version
```
jorisvandenbossche, line 46: Was there a specific reason for this change, or just to minimize the build time?

mwtoews: It is to align with  https://github.com/shapely/shapely/pull/1802/commits/7d54239a3b5656c2fd04098f8bde37ef4e6d8531 which minimizes build time

PR #1805, opened 2023-04-13
Add shapely.geometry.polygon.orient to __all__, bug #722

### LINKED ISSUE #722: 'orient' is not declared in __all__
When I import the [`orient`](https://shapely.readthedocs.io/en/latest/manual.html#shapely.geometry.polygon.orient) function as:
```python
```
PyCharm highlights it for me with the following message:
```
'orient' is not declared in __all__
```
image

I assume it was simply overlooked and should be added to `__all__`. (Or maybe it would be better to remove `__all__` completely as wild card imports are bad practice.)

---

quassy:
Fixes #722.

### DIFF
--- a/shapely/geometry/__init__.py
+++ b/shapely/geometry/__init__.py
@@ -9,13 +9,12 @@ from shapely.geometry.multilinestring import MultiLineString
 from shapely.geometry.multipoint import MultiPoint
 from shapely.geometry.multipolygon import MultiPolygon
 from shapely.geometry.point import Point
-from shapely.geometry.polygon import orient, LinearRing, Polygon
+from shapely.geometry.polygon import LinearRing, Polygon
 
 __all__ = [
     "box",
     "shape",
     "mapping",
-    "orient",
     "Point",
     "LineString",
     "Polygon",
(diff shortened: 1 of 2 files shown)


idanmiara:
related to https://github.com/shapely/shapely/pull/1690

quassy:
Any update on this?

mwtoews:
Apologies for the slowness! Happy to provide a review now.

Shouldn't `orient` be added to `__all__` in [`shapely/geometry/polygon.py`](https://github.com/shapely/shapely/blob/main/shapely/geometry/polygon.py#L13) instead of `shapely/geometry/__init__.py`?

quassy:
I have added it there, I think it belongs in both.

mwtoews:
Why does it need to be added to both modules? It's [documented](https://shapely.readthedocs.io/en/stable/manual.html#shapely.geometry.polygon.orient) as part of the `shapely.geometry.polygon` module, since it applies to Polygon geometries. Adding `orient` to multiple modules is a bit confusing.

quassy:
When it's documented like that I agree. 

For me it was a bit confusing that Polygon, LinearRing are both added to `shapely.geometry.polygon`, `shapely.geometry` and `shapely` (which is the documented location).

mwtoews:
True, Polygon and some other classes are in `__all__` for a few modules. But to resolve this issue, it just needs to be defined once in `shapely/geometry/polygon.py`. This PR only needs to roll-back changes to `shapely/geometry/__init__.py` and it should be good to go.

quassy:
Accidentally closed this PR by renaming the branch (thanks GitHub), so there is now a follow-up that should do exactly as required.

PR #1806, opened 2023-04-13
DOC: use more intersphinx links

mwtoews:
This PR leverages more intersphinx links, and resolves a missing productionlist token warning.

A simplified diff for the repetitive change (113 times) is:
```diff
- For other keyword-only arguments, see the NumPy ufunc docs.
+ See NumPy ufunc docs for other keyword arguments.
```
hopefully this new wording is OK, otherwise suggest a new wording.

### DIFF
--- a/docs/geometry.rst
+++ b/docs/geometry.rst
@@ -145,7 +145,7 @@ A format specification may also be used to control the format and precision.
     or hex-encoded as 0101000000cf6a813d263d65c0bdaab35a60ff32c0
 
 Shapely has a format specification inspired from Python's
-:ref:`formatspec`, described next.
+:ref:`python:formatspec`, described next.
 
 Semantic for format specification
 ---------------------------------
--- a/docs/geometry.rst
+++ b/docs/geometry.rst
@@ -145,8 +145,7 @@ A format specification may also be used to control the format and precision.
     or hex-encoded as 0101000000cf6a813d263d65c0bdaab35a60ff32c0
 
 Shapely has a format specification inspired from Python's
-`Format Specification Mini-Language
-<https://docs.python.org/3/library/string.html#formatspec>`_ described next.
+:ref:`formatspec`, described next.
 
 Semantic for format specification
 ---------------------------------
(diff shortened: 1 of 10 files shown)


docs/geometry.rst
```diff
@@ -145,15 +145,15 @@ A format specification may also be used to control the format and precision.
     or hex-encoded as 0101000000cf6a813d263d65c0bdaab35a60ff32c0
 
 Shapely has a format specification inspired from Python's
-`Format Specification Mini-Language
-<https://docs.python.org/3/library/string.html#formatspec>`_ described next.
+:ref:`formatspec`, described next.
```
jorisvandenbossche, line 148: When reading/editing this source, it might not be directly clear this is an external reference, so we could explicitly "namespace" it, like the following?

```suggestion
:ref:`python:formatspec`, described next.
```

shapely/_geometry.py
```diff
@@ -69,8 +69,7 @@ def get_type_id(geometry, **kwargs):
     ----------
     geometry : Geometry or array_like
     **kwargs
-        For other keyword-only arguments, see the
-        `NumPy ufunc docs <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs>`_.
+        See :ref:`NumPy ufunc docs <ufuncs.kwargs>` for other keyword arguments.
```
jorisvandenbossche, line 72: Personally, for _docstrings_ I am a bit on the fence about using intersphinx... The thing is that you also look at this in plain text when interactively checking the help in a terminal/notebook (and not just the online rendered html docs), and in that case the full link is something you can copy/paste, while the intersphinx label is basically useless.

mwtoews: A counterpoint is the first link from [a web search of "NumPy ufunc docs"](https://duckduckgo.com/?q=NumPy+ufunc+docs) is the same URL. Unless extensions are enabled, users may need to copy/paste something into a browser -- either a valid URL or phrase. But otherwise, your point is valid, as I often inspect plain-text docstrings too.

jorisvandenbossche: Yes, that's true, and this is also not a link that users will often have to check (the usage of one of those extra keyword arguments is far from general usage). So given that, also fine with your change.

mwtoews:
Backported to `maint-2.0` with 072f23c79f83f886ec6d72105a6c2b982ba22e15

PR #1808, opened 2023-04-20
ENH: Enable M and ZM in WKT and WKB outputs

mwtoews:
Working towards #1648 as a first step, to enable outputs with M or ZM dimensions. This requires GEOS 3.12+, which <s>is currently under development</s> was released 2023-06-27.

Earlier supported GEOS versions have always had limited support for reading these geometries:

- WKT will parse XYM geometries like `POINT M (1 2 4)` as XYZ, incorrectly interpreting M values as Z values
- WKT will parse XYZM geometries like `POINT ZM (1 2 3 4)` as XYZ, ignoring M values
- WKB will parse XYM geometries as XY, ignoring M values
- WKB will parse XYZM geometries as XYZ, ignoring M values

With GEOS 3.12+, the default `output_dimension` is adjusted from 3 to 4. Under the hood, this uses `GEOSWKBWriter_setOutputDimension_r` or `GEOSWKTWriter_setOutputDimension_r` to output XYZ, XYM or XYZM geometries.

I've flagged `test_repr_point_m_empty` is set to xfail, since `POINT M EMPTY` and `POINT ZM EMPTY` are incorrectly emitted as `POINT Z EMPTY`. This can be investigated/resolved later, as I'm aware there are workarounds in shapely to handle empty points.

Several of the tests were re-written to use more `struct` representations of binary strings, rather than raw binary strings. This was done to make them easier to read and modify.

### DIFF
--- a/shapely/tests/test_io.py
+++ b/shapely/tests/test_io.py
@@ -584,7 +584,9 @@ def test_repr_point_z_empty():
     assert repr(empty_point_z) == "<POINT Z EMPTY>"
 
 
-@pytest.mark.xfail(reason="TODO: fix WKT for empty M and ZM geometries", strict=True)
+@pytest.mark.xfail(
+    reason="TODO: fix WKT for empty M and ZM geometries; see GH-2004", strict=True
+)
 @pytest.mark.skipif(
     shapely.geos_version < (3, 12, 0),
     reason="M coordinates not supported with GEOS < 3.12",
@@ -951,10 +953,6 @@ def test_to_wkb_point_empty_zm(geom, expected):
     assert np.isnan(struct.unpack("<4d", actual[header_length:])).all()
 
 
-@pytest.mark.xfail(
-    shapely.geos_version < (3, 8, 0),
-    reason="GEOS<3.8 always outputs 3D empty points if output_dimension=3",
-)
 @pytest.mark.parametrize(
     "geom,expected",
     [
@@ -1092,7 +1090,7 @@ def test_pickle_z(geom):
 @pytest.mark.parametrize("geom", all_types_m)
 def test_pickle_m(geom):
     if shapely.get_type_id(geom) == shapely.GeometryType.LINEARRING:
-        pytest.xfail("TODO: M gets dropped for some reason")
+        pytest.xfail("TODO: M gets dropped for LinearRing types; see GH-2005")
     pickled = pickle.dumps(geom)
     actual = pickle.loads(pickled)
     assert_geometries_equal(actual, geom, tolerance=0)
@@ -1107,7 +1105,7 @@ def test_pickle_m(geom):
 @pytest.mark.parametrize("geom", all_types_zm)
 def test_pickle_zm(geom):
     if shapely.get_type_id(geom) == shapely.GeometryType.LINEARRING:
-        pytest.xfail("TODO: ZM gets dropped for some reason")
+        pytest.xfail("TODO: ZM gets dropped for LinearRing types; see GH-2005")
     pickled = pickle.dumps(geom)
     actual = pickle.loads(pickled)
     assert_geometries_equal(actual, geom, tolerance=0)
--- a/shapely/tests/test_io.py
+++ b/shapely/tests/test_io.py
@@ -21,29 +21,51 @@ from shapely.errors import UnsupportedGEOSVersionError
 from shapely.testing import assert_geometries_equal
 from shapely.tests.common import (
     all_types,
+    all_types_m,
     all_types_z,
+    all_types_zm,
     empty_point,
+    empty_point_m,
     empty_point_z,
+    empty_point_zm,
     equal_geometries_abnormally_yield_unequal,
     multi_point_empty,
+    multi_point_empty_m,
     multi_point_empty_z,
+    multi_point_empty_zm,
     point,
+    point_m,
     point_z,
+    point_zm,
     polygon_z,
 )
 
 EWKBZ = 0x80000000
+EWKBM = 0x40000000
+EWKBZM = EWKBZ | EWKBM
 ISOWKBZ = 1000
+ISOWKBM = 2000
+ISOWKBZM = ISOWKBZ + ISOWKBM
 POINT11_WKB = struct.pack("<BI2d", 1, 1, 1.0, 1.0)
 NAN = struct.pack("<d", float("nan"))
 POINT_NAN_WKB = struct.pack("<BI", 1, 1) + (NAN * 2)
 POINTZ_NAN_WKB = struct.pack("<BI", 1, 1 | EWKBZ) + (NAN * 3)
+POINTM_NAN_WKB = struct.pack("<BI", 1, 1 | EWKBM) + (NAN * 3)
+POINTZM_NAN_WKB = struct.pack("<BI", 1, 1 | EWKBZM) + (NAN * 4)
 MULTIPOINT_NAN_WKB = struct.pack("<BII", 1, 4, 1) + POINT_NAN_WKB
 MULTIPOINTZ_NAN_WKB = struct.pack("<BII", 1, 4 | EWKBZ, 1) + POINTZ_NAN_WKB
+MULTIPOINTM_NAN_WKB = struct.pack("<BII", 1, 4 | EWKBM, 1) + POINTM_NAN_WKB
+MULTIPOINTZM_NAN_WKB = struct.pack("<BII", 1, 4 | EWKBZM, 1) + POINTZM_NAN_WKB
 GEOMETRYCOLLECTION_NAN_WKB = struct.pack("<BII", 1, 7, 1) + POINT_NAN_WKB
 GEOMETRYCOLLECTIONZ_NAN_WKB = struct.pack("<BII", 1, 7 | EWKBZ, 1) + POINTZ_NAN_WKB
+GEOMETRYCOLLECTIONM_NAN_WKB = struct.pack("<BII", 1, 7 | EWKBM, 1) + POINTM_NAN_WKB
+GEOMETRYCOLLECTIONZM_NAN_WKB = struct.pack("<BII", 1, 7 | EWKBZM, 1) + POINTZM_NAN_WKB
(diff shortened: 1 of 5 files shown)


shapely/tests/test_io.py
```diff
@@ -579,6 +752,31 @@ def test_from_wkb_point_empty(wkb, expected_type, expected_dim):
...
+        (NESTED_COLLECTIONZM_NAN_WKB, 7, 4),
+    ],
+)
+def test_from_wkb_point_empty_m(wkb, expected_type, expected_dim):
+    geom = shapely.from_wkb(wkb)
+
+    assert shapely.is_empty(geom)
+    assert shapely.get_type_id(geom) == expected_type
+    assert shapely.get_coordinate_dimension(geom) == expected_dim
+
```
jorisvandenbossche, line 778: Maybe add a TODO here to add an `has_m`/`has_z` check once we have those?

jorisvandenbossche (commented):
Looks great!

Small comment: for `from_wkb`, you added test specific for empty points, but maybe we can also add a general test for from_wkb with a M / ZM geometry with actual coordinates? 
I think other geometries are being tested in a to_wkb/from_wkb roundtrip in `test_from_wkb_all_types`. Maybe we can add some M and ZM geometries to that parametrization?

shapely/tests/test_io.py
mwtoews: Coincidentally, `has_m` will be the next PR in the series I'm working on.

mwtoews:
The last commit expanded tests with `all_types_with_z` and `all_types_with_zm`, which highlighted many more "TODO" notes to investigate later, mostly related to empty geometry types. I'll create tickets for these after I get a sense of what they are, or if they are bugs with GEOS.

shapely/tests/test_io.py
```diff
@@ -790,6 +864,54 @@ def test_pickle(geom):
     assert_geometries_equal(pickle.loads(pickled), geom, tolerance=0)
 
 
+@pytest.mark.parametrize("geom", all_types_with_z)
+def test_pickle_z(geom):
+    if shapely.get_type_id(geom) == shapely.GeometryType.LINEARRING:
+        pytest.xfail("TODO: Z gets dropped for some reason")
```
jorisvandenbossche, line 870: This is probably because the custom code we have for LinearRings doesn't support more than 2D:

https://github.com/shapely/shapely/blob/60967815a3f4d40c4fcea719380e4287d2bc5e3b/shapely/geometry/polygon.py#L16-L22

We should probably add a helper function in C that converts a LineString to a LinearRing by just getting the CoordSequence from the line and put that in a LinearRing, so that it preserves the full coord seq and its dimensionality.

(doesn't need to happen for this PR)

shapely/tests/test_io.py
```diff
@@ -266,6 +269,55 @@ def test_from_wkb_all_types(geom, use_hex, byte_order):
...
+@pytest.mark.parametrize("use_hex", [False, True])
+@pytest.mark.parametrize("byte_order", [0, 1])
+def test_from_wkb_all_types_with_zm(geom, use_hex, byte_order):
+    if shapely.get_type_id(geom) == shapely.GeometryType.LINEARRING:
+        pytest.skip("Linearrings are not preserved in WKB")
+    wkb = shapely.to_wkb(geom, hex=use_hex, byte_order=byte_order)
+    actual = shapely.from_wkb(wkb)
+    assert_geometries_equal(actual, geom)
+    if actual.is_empty:
+        # TODO: Z gets enabled for some reason
```
jorisvandenbossche, line 314: enabled -> disabled?

jorisvandenbossche: So one example here (with a recent GEOS main for a few weeks ago, not the latest, so things might have changed):

```
In [1]: import shapely

In [2]: p = shapely.from_wkt("POINT ZM EMPTY")

In [3]: p
Out[3]: <POINT Z EMPTY>

In [4]: p.has_z
Out[4]: False

In [5]: shapely.get_coordinate_dimension(p)
Out[5]: 4

In [6]: shapely.to_wkt(p, output_dimension=4)
Out[6]: 'POINT Z EMPTY'

In [7]: shapely.to_wkt(p, output_dimension=3)
Out[7]: 'POINT Z EMPTY'

In [8]: shapely.to_wkt(p, output_dimension=2)
Out[8]: 'POINT Z EMPTY'
```

The overall dimensionality is correct, but `has_z` incorrectly gives False, and the WKT output incorrectly only has Z and not M. 
Also forcing to only show an output dimension up to 2 doesn't drop the "Z".

shapely/tests/test_io.py
mwtoews: Thanks for the tips, refactoring `get_coordinates` is somewhere down the track.

shapely/tests/test_io.py
mwtoews: I noticed that even with the current release `shapely.from_wkt("POINT Z EMPTY").has_z is True`, which I consider a bug. I'm considering splitting some of this PR into maintenance fixes for 2.0, as the scope is creeping out of enabling M coordinates.

mwtoews:
Note that this PR will be soon superseded by one or two PRs (when I get a chance to complete them). Thanks for the review @jorisvandenbossche , these helpful comments will be used for the next PRs.

Much has happened upstream in libgeos since this PR was started, which impact this PR:

- https://github.com/libgeos/geos/pull/908
- https://github.com/libgeos/geos/pull/915

mwtoews:
I've decided to update this PR rather than create a similar one. It has been updated to fit on `main`. There are 7 TODO notes to address later.

mwtoews:
@shapely/shapely-team anyone with time to review? It doesn't need to be too exhaustive, but I view this as a main shapely 2.1 feature, and am keen to progress with other related efforts.

shapely/tests/test_io.py
```diff
@@ -511,6 +618,45 @@ def test_to_wkb_z():
...
+
+    expected_wkb = struct.pack("<BI2d", 1, 1, 1.0, 2.0)
+    expected_wkb_m = struct.pack("<BI3d", 1, 1 | EWKBM, 1.0, 2.0, 4.0)
+    if shapely.geos_version < (3, 12, 0):
+        # previous behavior was to ignore M, treat as 2D
+        expected_wkb_m = expected_wkb
+
+    assert shapely.to_wkb(point, byte_order=1) == expected_wkb_m
+    assert shapely.to_wkb(point, output_dimension=2, byte_order=1) == expected_wkb
+    assert shapely.to_wkb(point, output_dimension=3, byte_order=1) == expected_wkb_m
```
jorisvandenbossche, line 637: So in this case (of M and not ZM), "M" is regarded as the third dimension? ( I would have naively expected this to drop the M values)

shapely/tests/test_io.py
```diff
@@ -669,6 +881,80 @@ def test_to_wkb_point_empty_z(geom, expected):
...
+    header_length = len(expected) - coordinate_length
+    assert len(actual) == header_length + coordinate_length
+    assert actual[:header_length] == expected[:header_length]
+    assert np.isnan(struct.unpack("<4d", actual[header_length:])).all()
+
+
+@pytest.mark.xfail(
+    shapely.geos_version < (3, 8, 0),
+    reason="GEOS<3.8 always outputs 3D empty points if output_dimension=3",
+)
```
jorisvandenbossche, line 957: Those lines probably were added back accidentally after updating with latest main?

shapely/tests/test_io.py
```diff
@@ -751,6 +1085,39 @@ def test_pickle_z(geom):
...
 
 
+@pytest.mark.skipif(
+    shapely.geos_version < (3, 12, 0),
+    reason="M coordinates not supported with GEOS < 3.12",
+)
+@pytest.mark.parametrize("geom", all_types_m)
+def test_pickle_m(geom):
+    if shapely.get_type_id(geom) == shapely.GeometryType.LINEARRING:
+        pytest.xfail("TODO: M gets dropped for some reason")
```
jorisvandenbossche, line 1095: Do you want to address that here, or open a follow-up issue? (a follow-up is fine for me as well , but we should track it to ensure it ideally gets done for 2.1)

src/pygeom.c
```diff
@@ -114,9 +118,11 @@ static PyObject* GeometryObject_ToWKT(GeometryObject* obj) {
   }
 
   GEOSWKTWriter_setRoundingPrecision_r(ctx, writer, precision);
+#if !GEOS_SINCE_3_12_0
   GEOSWKTWriter_setTrim_r(ctx, writer, trim);
   GEOSWKTWriter_setOutputDimension_r(ctx, writer, dimension);
```
jorisvandenbossche, line 123: Those have the correct defaults or are no longer needed for GEOS 3.12? Can you add a brief comment explaining this?

jorisvandenbossche (commented):
Looks good! Very nice, comprehensive but clear testing. Only added some small comments/questions

shapely/tests/test_io.py
```diff
@@ -487,6 +584,16 @@ def test_repr_point_z_empty():
     assert repr(empty_point_z) == "<POINT Z EMPTY>"
 
 
+@pytest.mark.xfail(reason="TODO: fix WKT for empty M and ZM geometries", strict=True)
```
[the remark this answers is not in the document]
mwtoews: This is a shapely issue #2004 (will also add to code comment for this PR)

shapely/tests/test_io.py
mwtoews: The first and second dims are always X and Y, but beyond that there are no guarantees. This is why I'm advocating renaming away from "3D" as it's ambiguous.

shapely/tests/test_io.py
mwtoews: Tracked with #2005, and comment adjusted for this PR

jorisvandenbossche (approved):

jorisvandenbossche:
@mwtoews thanks a lot!

jorisvandenbossche:
Failures that happened here on the dev builds are fixed in https://github.com/shapely/shapely/pull/2007

Issue #1814, opened 2023-05-04
Support `Feature` type in `__geo_interface__`


TomAugspurger:
Shapely doesn't currently support classes defining a `__geo_interface__` that returns an object with `{"type": "Feature"}`.

## Expected behavior and actual behavior.

Extract the `geometry` field and convert the feature to a Geometry

## Steps to reproduce the problem.

```python
import shapely.geometry


class Feature:
    @property
    def __geo_interface__(self):
        return {'type': "Feature", "geometry": {'type': "Point", "coordinates": [0, 0]}}

shapely.geometry.shape(Feature())
```

that raises

```pytb
---------------------------------------------------------------------------
GeometryTypeError                         Traceback (most recent call last)
Cell In[36], line 9
      5     @property
      6     def __geo_interface__(self):
      7         return {'type': "Feature", "geometry": {'type': "Point", "coordinates": [0, 0]}}
----> 9 shapely.geometry.shape(Feature())

File /srv/conda/envs/notebook/lib/python3.10/site-packages/shapely/geometry/geo.py:113, in shape(context)
    111     return GeometryCollection(geoms)
    112 else:
--> 113     raise GeometryTypeError(f"Unknown geometry type: {geom_type!r}")

GeometryTypeError: Unknown geometry type: 'feature'
```

## Shapely version and provenance

'2.0.1' from conda-forge.

PR #1815, opened 2023-05-04
Handle "Feature" type in `shape` (__geo_interface__)

TomAugspurger:
This adds a branch to `shape`'s `__geo_interface__` parsing to handle `Feature` type. I think this is the right thing to do, but I'm not an expert on GeoJSON or `__geo_interface__`.

Closes https://github.com/shapely/shapely/issues/1814


jorisvandenbossche:
https://gist.github.com/sgillies/2217756 indicates that `__geo_interface__` can also return a Feature type, so it seems the correct thing to do to support this in `shape(..)` then (cc@sgillies)

shapely/tests/geometry/test_feature.py
```diff
@@ -0,0 +1,14 @@
+import shapely.geometry
+
+
+def test_feature_from_geo_interface():
```
jorisvandenbossche, line 4: There is a ``shapely/tests/legacy/test_shape.py`` file with existing tests for `shape(..)`, you can move it there I think

jorisvandenbossche:
Thanks Tom!

PR #1820, opened 2023-05-18
TST: GEOS 3.12 uses parentheses for MultiPoint sub-members

mwtoews:
Upstream changes for GEOS 3.12 (current `main`) now use parentheses for MultiPoint sub-members. This change only broke one test. Other changes are expected for the doctests, which should be done when the version of GEOS is changed for doctests.

Xrefs:

- https://github.com/libgeos/geos/pull/903
- https://github.com/libgeos/geos/issues/902


mwtoews:
Backported to `maint-2.0` with 92b3cdbcfa04670ecd8106f129cc386f8d743a3a

PR #1822, opened 2023-05-23
Draft of a governance doc

sgillies:
Adapted from rasterio's, which is adapted from NumPy's.

### DIFF
--- a/governance.md
+++ b/governance.md
@@ -1,0 +1,323 @@
+Shapely project governance and decision-making
+===============================================
+
+The purpose of this document is to formalize the governance process used by the
+Shapely project in both ordinary and extraordinary situations, and to clarify
+how decisions are made and how the various elements of our community interact,
+including the relationship between open source collaborative development and
+work that may be funded by for-profit or non-profit entities.
+
+# Summary
+
+Shapely is a community-owned and community-run project. To the maximum extent
+possible, decisions about project direction are made by community consensus
+(but note that “consensus” here has a somewhat technical meaning that might not
+match everyone’s expectations – see below). Some members of the community
+additionally contribute by serving on the Shapely steering council, where they
+are responsible for facilitating the establishment of community consensus, for
+stewarding project resources, and – in extreme cases – for making project
+decisions if the normal community-based process breaks down.
+
+# The project
+
+The Shapely Project (The Project) is an open source software project. The goal
+of The Project is to develop open source software for manipulation and analysis
+of geometric objects in the cartesian plane in Python, and in particular the
+shapely Python package, along with related software. The Software developed by
+The Project is released under the BSD (or similar) open source license,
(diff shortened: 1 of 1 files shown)


caspervdw (approved):
Thanks @sgillies  for taking this initiative. A landmark in professionalizing Shapely!

I read through it and nothing stands out in particular. Do you have a specific section that needs extra attention?

One thought; when describing what `shapely` is, it might be good to mention GEOS, as shapely is largely an API around GEOS.

governance.md
```diff
@@ -42,6 +42,11 @@ Contributors work on behalf of and are responsible to the larger Project
 Community and we strive to keep the barrier between Contributors and Users as
 low as possible.
 
+Shapely relies on the [GEOS project](https://libgeos.org) and expresses, in
+Python, much of the GEOS API. The sets of contributors for these projects
+overlap, but they are organized separately and have different forms of
+governance.
```
sgillies, line 48: @caspervdw what do you think about this addition?

sgillies:
@caspervdw I was hoping that adapting Numpy's governance doc would be uncontroversial. And I find it to be both practical and philosophical.

As to professionalization: I don't want to take it too far! I'd like Shapely to not depart too far from its punk and DIY roots :smile:

governance.md
caspervdw: I think the core aim of this project should be expressing the GEOS API in Python. Maybe rephrase it slightly:

```
Shapely relies on the [GEOS project](https://libgeos.org) and aims to express its API
in Python. The sets of contributors for these projects
overlap, but they are organized separately and have different forms of
governance
```

sgillies: I think we mostly agree. The GEOS API has some things that are specific to PostGIS, like SRIDs, which we're not expressing. And the project has some Python geospatial interoperability features that have nothing to do with GEOS (`__geo_interface__` for example). To me. it doesn't feel quite right to say that expressing the API is the core aim of this project.

caspervdw:
> Unrelated to this PR: it looks like we have some flaky tests.

Yep the GEOS main branch often fails because we have lots of super strict tests. I think that’s a good thing but having the red X everywhere is not.
Maybe we should allow the failures against main(and manually check them)

caspervdw:
@sgillies  are you :+1: on merging this?

jorisvandenbossche (commented):
Thanks for getting this started! I am very supportive to describing our governance. And generally fine with adapting the NumPy governance documents.

One aspect I am wondering: do we want to say something about "commit rights" (who has those, how to get them, ...), or do we leave that detail to the practical day-to-day organization without being written out? (assuming this is not tied to being Steering Council member). Also the NumPy document doesn't mention this, and I can't directly find where this is described elsewhere for them (they do mention to have teams on their website, and list the "maintainers" members as one of the teams at https://numpy.org/teams/)

Issue #1732, opened 2023-01-23
REGR: equality (__eq__) now ignores the z dimension


jorisvandenbossche:
With shapely 2.0, we switched the implementation of `Geometry.__eq__` to use the "equals_exact" predicate from GEOS, instead of manually comparing the coordinates as we did before (`geom1.coords == geom2.coords`), in the assumption this is equivalent. But it seems that `GEOSEqualsExact` only checks the X and Y coordinates (https://github.com/libgeos/geos/issues/809), thus introducing a regression in comparing 3D geometries:

With shapely 1.8
```python
>>> Point(1, 2, 3) == Point(1, 2, 4)
False
```

With shapely 2.0:

```python
>>> Point(1, 2, 3) == Point(1, 2, 4)
True
```

idanmiara:
Hi @jorisvandenbossche Thanks for looking into this.
It would be useful to keep the current functionality (compare only xy) as a separate function, maybe `xy_equals(other)`.

jorisvandenbossche:
That already exists as `equals_exact` (the naming doesn't make this clear, and we need to update this in the docs that it actually _is_ only for x and y)

idanmiara:
> That already exists as `equals_exact` (the naming doesn't make this clear, and we need to update this in the docs that it actually _is_ only for x and y)

Thanks for the pointer! that reminds me of https://github.com/shapely/shapely/pull/1231 which I think almost ready (waiting for some final feedback there)

idanmiara:
@jorisvandenbossche as I understand from https://github.com/libgeos/geos/issues/809#issuecomment-1405242815 this issue would be solved properly only in future GEOS version, right?
If so - I think we need a quick fix for `2.0.1` as this is a major regression. Should I suggest a quick PR fix for `2.0.1` or are you working on it already ?

jorisvandenbossche:
Yes, we can't rely on that _right now_, so we still need to workaround for released versions of GEOS. I started on backporting https://github.com/libgeos/geos/pull/810 to our own code using the C API, and that could be used to fix this (although maybe a bit overkill for just this issue). Will open a PR shortly.

PR #1829, opened 2023-06-02
REGR: fix __eq__ to not ignore z values

jorisvandenbossche:
Closes https://github.com/shapely/shapely/issues/1732

Broken off from https://github.com/shapely/shapely/pull/1760, to just restore the previous behaviour (using the previous implementaiton) and already add the more extensive tests that I wrote for that PR. This way, the other PR can focus more on the API change.

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -201,6 +201,7 @@ class BaseGeometry(shapely.Geometry):
         if not isinstance(other, BaseGeometry):
             return NotImplemented
         # equal_nan=False is the default, but not yet available for older numpy
+        # TODO updated once we require numpy >= 1.19
         return type(other) == type(self) and np.array_equal(
             self.coords, other.coords  # , equal_nan=False
         )
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -200,7 +200,9 @@ class BaseGeometry(shapely.Geometry):
     def __eq__(self, other):
         if not isinstance(other, BaseGeometry):
             return NotImplemented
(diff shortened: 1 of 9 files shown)


shapely/geometry/base.py
```diff
@@ -197,6 +197,22 @@ def __sub__(self, other):
     def __xor__(self, other):
         return self.symmetric_difference(other)
 
+    def __eq__(self, other):
+        if not isinstance(other, BaseGeometry):
+            return NotImplemented
+        # equal_nan=False is the default, but not yet available for older numpy
```
mwtoews, line 203: Maybe put a note that it was introduced in version 1.19.0, which I imagine will be the minimum version in a few years.

shapely/geometry/base.py
```diff
@@ -917,6 +933,18 @@ def geoms(self):
     def __bool__(self):
         return self.is_empty is False
 
+    def __eq__(self, other):
+        if not isinstance(other, BaseGeometry):
+            return NotImplemented
+        return (
+            type(other) == type(self)
+            and len(self.geoms) == len(other.geoms)
+            and all(x == y for x, y in zip(self.geoms, other.geoms))
```
mwtoews, line 942: Use binary geometry aliases "a" and "b", instead of coordinate dimension names.
```suggestion
            and all(a == b for a, b in zip(self.geoms, other.geoms))
```

mwtoews (approved):
This looks good! I haven't had a chance to test this PR locally, but the changes and tests all look good. Only a few minor remarks to consider.

PR #1831, opened 2023-06-05
TST: update parallel_offset (offset_curve) tests for GEOS main (minimum quad_segs of 8)

jorisvandenbossche:
A recent commit on GEOS main (https://github.com/libgeos/geos/commit/68a5b4d4bfadee1c580fb6702597fe7cfbcc8118) changed to use a minimum quad_segs of 8 for OffsetCurve (this is only relevant for the default "round" join style, the "mitre" and "bevel" join styles still give more coarse results). 

This updates our tests to be compatible with both older as latest GEOS.

### DIFF
--- a/shapely/tests/legacy/test_parallel_offset.py
+++ b/shapely/tests/legacy/test_parallel_offset.py
@@ -30,21 +30,21 @@ class OperationsTestCase(unittest.TestCase):
 
         line2 = LineString([(0, 0), (5, 0), (5, -5)])
         assert_geometries_equal(
-            line2.parallel_offset(2, "left", resolution=1),
+            line2.parallel_offset(2, "left", join_style=3),
             LineString([(0, 2), (5, 2), (7, 0), (7, -5)]),
         )
         assert_geometries_equal(
-            line2.parallel_offset(2, "left", join_style=2, resolution=1),
+            line2.parallel_offset(2, "left", join_style=2),
             LineString([(0, 2), (7, 2), (7, -5)]),
         )
         # offset_curve alias
         assert_geometries_equal(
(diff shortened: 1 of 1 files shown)


shapely/tests/legacy/test_parallel_offset.py
```diff
@@ -30,21 +30,21 @@ def test_parallel_offset_linestring(self):
 
         line2 = LineString([(0, 0), (5, 0), (5, -5)])
         assert_geometries_equal(
-            line2.parallel_offset(2, "left", resolution=1),
+            line2.parallel_offset(2, "left", join_style=3),
             LineString([(0, 2), (5, 2), (7, 0), (7, -5)]),
```
jorisvandenbossche, line 34: With the default "round" join style, this is now using 8 points in the corner. But the "bevel" join style still gives this result, so updated the test to use that instead

shapely/tests/legacy/test_parallel_offset.py
```diff
@@ -30,21 +30,21 @@ def test_parallel_offset_linestring(self):
...
         )
         assert_geometries_equal(
-            line2.parallel_offset(2, "left", join_style=2, resolution=1),
+            line2.parallel_offset(2, "left", join_style=2),
```
jorisvandenbossche, line 37: The resolution keyword is ignored for a "mitre" join style, so just cleaning this up

jorisvandenbossche:
(there are still other failures on the builds with GEOS main, but the parallel offset test is fixed now)

caspervdw (approved):

mwtoews:
Backported to `maint-2.0` with 799912d7db86113fcd11346ca40db876d012aa16

PR #1841, opened 2023-06-26
TST: difference set operation works again for mixed dimensions on GEOS main

jorisvandenbossche:
Essentially reverting the xfail we added in https://github.com/shapely/shapely/pull/1715, as this was fixed upstream in GEOS: https://github.com/libgeos/geos/pull/923


mwtoews:
Backported to `maint-2.0` with eeb51ad4d80c902d5f0bcad209687f046019cb6e

PR #1842, opened 2023-06-26
TST: use numpy 0-d instead of 1-d arrays for testing non-mutation of array argument (numpy 1.25 compat)

jorisvandenbossche:
Those tests passing a numpy array of length 1 to the affine functions were added intentionally in https://github.com/shapely/shapely/pull/794. 
However, the main thing it is testing is that we don't mutate those values inplace, which can also be achieved by using a 0-d (scalar-like) array instead of a 1d array. 

The argument itself is being treated as a scalar by shapely (eg `shapely.affinity.rotate(geom, theta=np.array([10]))`, the `theta` is a scalar), but because of the numpy behaviour of treating a 1d array of length 1 as a scalar, this worked as well. Now that numpy deprecates this in 1.25, I think we can just follow numpy (and let users handle the numpy deprecation warning in a similar way as I am doing here), instead of adding some extra code to keep supporting this 1d array as argument. 

This fixes some failures on the CI dev build (where we turn warnings into errors).

### DIFF
--- a/shapely/tests/legacy/test_affinity.py
+++ b/shapely/tests/legacy/test_affinity.py
@@ -170,14 +170,14 @@ class TransformOpsTestCase(unittest.TestCase):
         ls = load_wkt("LINESTRING(240 400, 240 300, 300 300)")
         els = load_wkt("LINESTRING(220 320, 320 320, 320 380)")
         # check with degrees
-        theta = np.array([90.0])
+        theta = np.array(90.0)
         rls = affinity.rotate(ls, theta)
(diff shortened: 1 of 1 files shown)


mwtoews (approved):
Thanks, looks good! I wasn't aware of this change, which is [noted here](https://numpy.org/doc/stable/release/1.25.0-notes.html#deprecation).

mwtoews:
Backported to `maint-2.0` with f11509718abbfe210ae8b4c7a6ae91563258b6ea

PR #1843, opened 2023-06-26
FIX #1836 - polylabel runs into infinite loop with tiny polygon

### LINKED ISSUE #1836: polylabel runs into infinite loop with tiny polygon

**Expected:**
 `shapely.ops.polylabel` should not crash my machine! 

**Actual:**
`shapely.ops.polylabel` runs into an infinite loop, until the system runs out of memory, in rare cases with topologically questionable, but technically valid polygons.

```py
from shapely import wkt
from shapely.ops import polylabel


wkt_string = (
    "POLYGON ((536520.0679737709 5438764.374763639, 536520.0679737704 5438764.374763602, "
    "536520.0679737709 5438764.374763642, 536520.0679737709 5438764.374763639))"
)

polygon = wkt.loads(wkt_string)

# careful, this will eventually eat up your memory, if not interrupted manually!
polylabel(polygon)
```

I've tracked the problem down to this loop: https://github.com/shapely/shapely/blob/6cbae5d984060047d15b1b22e6ed9d912872508f/shapely/algorithms/polylabel.py#L110-L117

where in `y += cell_size` the tiny `cell_size` of `4.656612873077393e-10` is being added to the rather large `y` coordinate of `5438764.374763602`, causing the cell_size to loose its significance. Therefore, `y == y + cell_size` and `y` will never grow to the size of `maxy`.

## Operating system

Ubuntu 22.04.2 LTS based docker container `jupyter/base-notebook:notebook-6.5.3`.

 2.0.1 installed from conda-forge, but also reproducible with 1.8.5 from PyPi

djm93dev:
I believe this should fix issue #1836

### DIFF
--- a/shapely/algorithms/polylabel.py
+++ b/shapely/algorithms/polylabel.py
@@ -108,13 +108,9 @@ def polylabel(polygon, tolerance=1.0):
         best_cell = bbox_cell
 
     # build a regular square grid covering the polygon
-    x = minx
-    while x < maxx:
-        y = miny
-        while y < maxy:
+    for x in range(int(minx), int(maxx)):
+        for y in range(int(miny), int(maxy), int(cell_size)):
             heappush(cell_queue, Cell(x + h, y + h, h, polygon))
-            y += cell_size
-        x += cell_size
 
     # minimum priority queue
     while cell_queue:


jorisvandenbossche:
Thanks for the PR! And sorry for the slow reply .. Could you add a test for this? (I think you can take the example from the bug report for that, and add a test based on that in `test_polylabel.py`)

jorisvandenbossche:
And it also seems to actually change the result (failing test, didn't look further into how the result is different):

```
    def test_polylabel(self):
        """
        Finds pole of inaccessibility for a polygon with a tolerance of 10
    
        """
        polygon = LineString(
            [(0, 0), (50, 200), (100, 100), (20, 50), (-100, -20), (-150, -200)]
        ).buffer(100)
        label = polylabel(polygon, tolerance=10)
        expected = Point(59.35615556364569, 121.8391962974644)
>       assert expected.equals_exact(label, 1e-6)
E       assert False
E        +  where False = <bound method BaseGeometry.equals_exact of <POINT (59.356 121.839)>>(<POINT (57.906 108.751)>, 1e-06)
E        +    where <bound method BaseGeometry.equals_exact of <POINT (59.356 121.839)>> = <POINT (59.356 121.839)>.equals_exact

shapely/tests/legacy/test_polylabel.py:21: AssertionError
```

jorisvandenbossche:
As I mentioned on the issue, this is also being fixed by https://github.com/shapely/shapely/pull/1307 where we replace our custom polylabel implementation with a new function from GEOS. Therefore closing this PR. But @djm93dev thanks for the PR nonetheless!

PR #1847, opened 2023-07-01
Use Py3 integers where appropriate

eumiro:
Get rid of Python2 relics and use integers in mathematical operations.

### DIFF
--- a/docs/code/figures.py
+++ b/docs/code/figures.py
@@ -1,8 +1,8 @@
 from math import sqrt
 from shapely import affinity
 
-GM = (sqrt(5)-1.0)/2.0
-W = 8.0
+GM = (sqrt(5) - 1) / 2
+W = 8
 H = W*GM
 SIZE = (W, H)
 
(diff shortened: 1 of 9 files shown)


caspervdw (approved):

caspervdw:
Thanks @eumiro , I will merge when tests pass.

caspervdw:
@eumiro Tests are failing.

Can you explain to me why `(sqrt(5) - 1) / 2` is better than `(sqrt(5.0) - 1.0) / 2.0`? My idea is that a lot of superfluous int-to-float conversions are happening in the first version.

eumiro:
The test is failing because it expects `stroke-width="2.0"` where simple `2` should be actually correct. If you agree, I'd change the tests too.

About your `(sqrt(5) - 1) / 2` question: we're mathematically subtracting `1` and dividing by `2`. The `float` versions of these numbers may be a bit faster to calculate, but most probably not expressing what we're doing. There are also `60` minutes in an hour and not `60.0`.

caspervdw:
@eumiro I'll close this due to activity and absence of enthusiasm from maintainers. Thanks though for your contribution! In general we recommend raising an issue first to avoid wasted effort on a PR.

PR #1849, opened 2023-07-05
Port 'interleaved' and auto-z detection from PR 1676

caspervdw:
Port 'interleaved' and auto-z detection from PR #1676

### DIFF
--- a/shapely/coordinates.py
+++ b/shapely/coordinates.py
@@ -38,7 +38,7 @@ def transform(
         input geometry using ``has_z``, which may result in 2 calls to
         the transformation function. Note that this inference
         can be unreliable with empty geometries or NaN coordinates: for a
-        guaranteed result, it is recommended to specify the keyword).
+        guaranteed result, it is recommended to specify ``include_z`` explicitly.
     interleaved : bool, default True
         If set to False, the transformation function should accept 2 or 3 separate
         one-dimensional arrays (x, y and optional z) instead of a single
--- a/shapely/coordinates.py
+++ b/shapely/coordinates.py
@@ -29,16 +29,18 @@ def transform(
         A function that transforms a (N, 2) or (N, 3) ndarray of float64 to
         another (N, 2) or (N, 3) ndarray of float64.
         The function may not change N.
-    include_z : bool, optional, default False
-        If False, return 2D geometries. If True, include the third dimension
-        in the output (if a geometry has no third dimension, the z-coordinates
+    include_z : bool, optional
+        If False, return 2D geometries. If True, the data being passed to the
+        transformation function will include the third dimension
+        (if a geometry has no third dimension, the z-coordinates
         will be NaN). By default, will infer the dimensionality from the
         input geometries. Note that this inference can be unreliable with
         empty geometries (for a guaranteed result, it is recommended to
         specify the keyword).
-    interleaved: bool, default True
+    interleaved : bool, default True
         If set to False, the transformation function should accept 2 or 3 separate
-        one-dimensional arrays (x, y and optional z).
+        one-dimensional arrays (x, y and optional z) instead of a single
+        two-dimensional array.
 
     Examples
     --------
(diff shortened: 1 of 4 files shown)


shapely/coordinates.py
```diff
@@ -38,19 +52,51 @@ def transform(geometry, transformation, include_z=False):
...
+    <POINT Z (1 0 2)>
+
+    Using pyproj >= 2.1, the following example will reproject Shapely geometries
+    from EPSG 4326 to EPSG 32618:
+
+    >>> from pyproj import Transformer
+    >>> transformer = Transformer.from_crs(4326, 32618, always_xy=True)
+    >>> p = transform(Point(-75, 50), transformer.transform, interleaved=False)
+    >>> shapely.to_wkt(p, rounding_precision=2)
+    'POINT (500000 5538630.7)'
```
jorisvandenbossche, line 80: Is passing this rounding precision manually still needed? (I think with more recent GEOS versions, the default WKT repr tends to be improved, i.e. pruning unnecessary digits more by default)

jorisvandenbossche (commented):
One question for the `include_z` support is whether we want to support an array with mixed dimensions, see https://github.com/shapely/shapely/pull/1676/files#r1058565393. 

Currently we infer either True or False (so once there is one 3D geometry, we use 3D coordinates for all geometries, potentially introducing NaNs where you have 2D geometries)

In geopandas, we worked around this to preserve existing behaviour (https://github.com/geopandas/geopandas/pull/2641), by doing the transform in two steps (once on all 2D geoms, once on all 3D geoms). 
It might be nice to have this support in shapely itself (it might cover an important use case of why you would want a "scalar" transform, one call per geometry, i.e. the "rebuild" feature in https://github.com/shapely/shapely/pull/1676), but it's also easy to just keep this workaround in geopandas.

caspervdw:
@jorisvandenbossche  I updated the PR and additionally made the Z-autodetection per geometry (as per your suggestion).

I think it is good to merge; feel free to hit the button if you approve.

There is one breaking change here `include_z` now defaults to `None`. I think this has minor consequences, but I did write it down in the changelog. If you feel this is too large of a breaking change, I can revert the default back to `False`.

shapely/coordinates.py
```diff
@@ -20,11 +28,26 @@ def transform(geometry, transformation, include_z=False):
...
+    include_z : bool, optional, default False
+        If False, always return 2D geometries.
+        If True, the data being passed to the
+        transformation function will include the third dimension
+        (if a geometry has no third dimension, the z-coordinates
+        will be NaN). If None, will infer the dimensionality per
+        input geometry using ``has_z``, which may result in 2 calls to
+        the transformation function. Note that this inference
+        can be unreliable with empty geometries or NaN coordinates: for a
+        guaranteed result, it is recommended to specify the keyword).
```
caspervdw, line 41:
```suggestion
        guaranteed result, it is recommended to specify ``include_z`` explicitly.
```

caspervdw:
@jorisvandenbossche  I will merge this PR myself as soon as the CI passes - you already had a good look, it is well tested and doesn't break any existing function.

jorisvandenbossche:
Thanks for finishing up the PR! (and sorry for the unresponsiveness here)

shapely/coordinates.py
```diff
@@ -20,11 +28,26 @@ def transform(geometry, transformation, include_z=False):
...
+        guaranteed result, it is recommended to specify ``include_z`` explicitly.
+    interleaved : bool, default True
+        If set to False, the transformation function should accept 2 or 3 separate
+        one-dimensional arrays (x, y and optional z) instead of a single
+        two-dimensional array.
+
+    See Also
+    --------
+    has_z : Returns a copy of a geometry array with a function applied to its
+        coordinates.
```
jorisvandenbossche, line 50: This is a wrong copy-paste I assume, it's the explanation of `transform` itself, not `has_z`

idanmiara:
@caspervdw Thanks for your work!
@jorisvandenbossche Thanks for the review!

PR #1850, opened 2023-07-07
Make MultiPolygon accept polygons without holes

mdczaplicki:


mwtoews:
Thanks, this is an old bug!

A shapely 2 workaround is to use `shapely.multipolygons(coords2)`, which seems to work fine.

mdczaplicki:
That's what I'm using in the code. But it bothered me that I can't use the main API of the package.

mwtoews:
Agree @mdczaplicki that this should be fixed for main API. I've added a few commits to get it ready. I'll backport it for the next 2.0 micro release after this is reviewed and merged (the doc changes need to be moved to the 2.0.2 section).

jorisvandenbossche (approved):
Thanks, looking good!

mwtoews:
Backported to `maint-2.0` with 732370c93725136d8990bd26cc35e19ac8fdfd4f

Issue #1857, opened 2023-07-28
DOC: Enable pull request builds for documentation?


theroggy:
I would like to make some contributions to the shapely documentation... 

However, it would be a bit more convenient to do so if the builds for pull requests were activated. Would that be possible?

![image](https://github.com/shapely/shapely/assets/1565789/c2546a6e-a3fa-40eb-96c6-2b8bb70780b4)

mwtoews:
Almost done. @sgillies will need to address and error shown [here](https://readthedocs.org/projects/shapely/builds/):
> Could not send GitHub build status report for shapely. Make sure you have the correct GitHub repository permissions and your [GitHub account](https://readthedocs.org/accounts/social/connections/) is connected to Read the Docs.

mwtoews:
The PR builds are working fine, but they still don't appear in GitHub's list of checks. However, I don't know exactly where the magic button is.

theroggy:
It indeed isn't showing yet integrated in the pull request yet.

The following step in the [troubleshooter](https://docs.readthedocs.io/en/stable/guides/pull-requests.html) solved it for me in for another project:

> ### Build status is not being reported to your Git provider
> If opening a pull request does start a new build, but the build status is not being updated with your Git provider, then your  connected account may have out dated or insufficient permisisons.

> Make sure that you have granted access to the Read the Docs [OAuth App](https://github.com/settings/applications) for your personal or organization GitHub account. You can also try reconnecting your account with your Git provider.

The project in my case wasn't granted access to readthedocs via Oauth.

sgillies:
I've created a new webhook in the shapely settings and on RTD I can see a successful ping from GitHub.

theroggy:
Tried creating a new pull request... but the readthedocs build still doesn't show up, so there still needs to be some bump in the wire somewhere :-(...
https://github.com/shapely/shapely/pull/1864

sgillies:
@theroggy I see the build at https://readthedocs.org/projects/shapely/builds/ and at https://shapely--1864.org.readthedocs.build/en/1864/reference/shapely.intersects.html, but GitHub is not showing this as a PR check. I've enabled the RTD app for the shapely organization, so hopefully on your next push we'll see it in the list of PR checks.

theroggy:
Bingo! In my latest pull request the readthedocs build is shown as a check: #1865

Thanks!

PR #1858, opened 2023-07-30
Docs: pin Cython=0.29, geos=3.11, add pip for ReadTheDocs

mwtoews:
Since Cython 3.0 was released a few weeks ago, the ReadTheDocs builds have been failing:
```
$ python ./setup.py install --force
...
Error compiling Cython file:
------------------------------------------------------------
...
        self.last_error = <char *> malloc((1025) * sizeof(char))
        self.last_error[0] = 0
        self.last_warning = <char *> malloc((1025) * sizeof(char))
        self.last_warning[0] = 0
        GEOSContext_setErrorMessageHandler_r(
            self.handle, &geos_message_handler, self.last_error
                         ^
------------------------------------------------------------

shapely/_geos.pyx:30:25: Cannot assign type 'void (*)(const char *, void *) except *' to 'GEOSMessageHandler_r'
```
While this is a curious bug to fix for Cython 3.x, the short-term solution is to pin the docs/environment.yml for now. (This error doesn't occur in the main build, since `pyproject.toml` has `"Cython~=0.29"` in `build-system.requires`)

This also upgrades geos to 3.11 which is the same version used for the binary wheels. Furthermore, add pip for ReadTheDocs (which auto-adds a few pip deps).

Also, the RTD option described in #1857 is now toggled, so we should see the build in this PR.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -3,10 +3,11 @@ channels:
   - conda-forge
 dependencies:
   - python=3.10
-  - geos=3.10
+  - geos=3.11
   - numpy
-  - cython
+  - cython=0.29
   - sphinx-book-theme
   - sphinx-remove-toctrees
-  - numpydoc==1.1.*
+  - numpydoc=1.1
   - matplotlib
+  - pip


mwtoews:
> Also, the RTD option described in #1857 is now toggled, so we should see the build in this PR.

Not yet. I don't have sufficient permissions to enable RTD to send GitHub build status report back to this repo.

mwtoews:
Since `maint-2.0` was failing too, this is backported with 75aebf6c46402fb7ea04ea85a9b513e4eaf3cdd8

PR #1865, opened 2023-08-08
DOC: Clarification in doc of overlaps

theroggy:
Another pull request to test readthedocs integration

### DIFF
--- a/shapely/predicates.py
+++ b/shapely/predicates.py
@@ -781,9 +781,9 @@ def intersects(a, b, **kwargs):
 def overlaps(a, b, **kwargs):
     """Returns True if A and B spatially overlap.
 
-    A and B overlap if they have some but not all points in common, have the
-    same dimension, and the intersection of the interiors of the two geometries
-    has the same dimension as the geometries themselves.  That is, only polyons
+    A and B overlap if they have some but not all points/portion of space in common,
+    have the same dimension, and the intersection of the interiors of the two geometries
+    has the same dimension as the geometries themselves. That is, only polyons
     can overlap other polygons and only lines can overlap other lines.
 
     If either A or B are None, the output is always False.


shapely/predicates.py
```diff
@@ -862,10 +862,11 @@ def intersects(a, b, **kwargs):
...
 
-    A and B overlap if they have some but not all points in common, have the
-    same dimension, and the intersection of the interiors of the two geometries
-    has the same dimension as the geometries themselves.  That is, only polyons
-    can overlap other polygons and only lines can overlap other lines.
+    A and B overlap if they have some but not all points/space in common,
+    have the same dimension, and the intersection of the interiors of the two geometries
+    has the same dimension as the geometries themselves. That is, only polyons
+    can overlap other polygons and only lines can overlap other lines. If A covers or is
+    within B, overlaps won't be True.
```
sgillies, line 869: @theroggy can you format this block to 72 chars width? It's an archaic practice that I followed when I wrote the package and it really does help docs fit in the terminal when someone does a `help(overlaps)`.

Ideally we should set up a pre-commit hook for this 🤔

sgillies (changes requested):

shapely/predicates.py
theroggy: Sure. Note that there are very few docstrings now that seem to be limited to 72 characters at first sight. Most, not all seem to be restricted to 80 characters though.

Haven't used it yet, but ruff seems to supports having a different line length for "comments and docstring" lines:
https://docs.astral.sh/ruff/rules/doc-line-too-long/.

sgillies (approved):
Sorry about the review delay @theroggy ! I lost this one in my notifications.

shapely/predicates.py
```diff
@@ -874,11 +876,12 @@ def overlaps(a, b, **kwargs):
...
         Geometry or geometries to check.
     **kwargs
-        See :ref:`NumPy ufunc docs <ufuncs.kwargs>` for other keyword arguments.
+        See :ref:`NumPy ufunc docs <ufuncs.kwargs>` for other keyword
+        arguments.
 
     See Also
     --------
-    prepare : improve performance by preparing ``a`` (the first argument)
+    prepare : improve performance by preparing ``a`` (the 1st argument)
```
jorisvandenbossche, line 884:
```suggestion
    prepare : improve performance by preparing ``a`` (the first argument)
```

(going to revert this small change; at the moment we don't use that in shapely, and in general I think the recommendation for small numbers is to use the long version in written text. We can certainly discuss that but let's do in the context of the full docs)

jorisvandenbossche (approved):
Looks good!

Issue #1877, opened 2023-08-27
Multipolygon cannnot be created from an np.array of Polygons


JuriaanSioux:
## Expected behavior and actual behavior.

I would have expected the creation of `MultiPolygons` to work with a Numpy `array` of `Polygons`. However, this raises an exception.

The documentation does note that the polygons should be of type `Sequence` (and as far as I understand, a Numpy `array` does not meet this criteria),  but given everything else works with Numpy `arrays` as well, I think this one should work.
## Steps to reproduce the problem.

```python
import numpy as np
from shapely import Polygon, MultiPolygon

 # raises ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
MultiPolygon(np.array([Polygon(), Polygon()]))

# it does work if we use a list instead
MultiPolygon([Polygon(), Polygon()])
# or an array with a single element
MultiPolygon(np.array([Polygon()]))
```

## Possible fix

I think the fix is relatively straightforward. Change out the first comparison in the `__new__` method of `MultiPolygon' from 

```python
    def __new__(self, polygons=None):
        if not polygons:
```

to 

```python
    def __new__(self, polygons=None):
        if polygons is None:
```

The case of empty lists that are caught with the original implementation will be taken care of later under the `# Bail immediately if we have no input points.` section.

Let me know if you would like me to submit a pull-request for this.

## Operating system

Linux mint 21.2

## Shapely version and provenance

Shapely 2.0.1 on Python 3.10 installed via pip.

caspervdw:
Hi @JuriaanSioux , sorry for the late response, but I agree on that arrays of polygons should be accepted in the MultiPolygon constructor. Please feel free to open a PR and request a review!

jorisvandenbossche:
This was actually closed by https://github.com/shapely/shapely/pull/1880 (but it didn't automatically close this issue)

PR #1880, opened 2023-08-28
BUG: allow the creation of MultiPolygon from a numpy array of Polygons

JuriaanSioux:
… by replacing the check with a `is None` check so that it does not try to make a `bool` out of a `numpy array`. Also added tests to make sure that the `numpy array` creation is now allowed, and added a test for an empty `MultiPolygon` creation to ensure the original behavior is not broken. 

See [#1877](https://github.com/shapely/shapely/issues/1877).


caspervdw:
Looks good to me! Thanks @JuriaanSioux

caspervdw (approved):

jorisvandenbossche (approved):
Thanks!

PR #1884, opened 2023-09-05
FIX: setup.py didn't run geos-config properly on Windows/MSYS2

michalfapso:
Installing shapely in Windows/MSYS2 from source using
```
git clone https://github.com/shapely/shapely.git
cd shapely
pip install .
```
resulted in the `Could not find geos-config executable` error even though it was present in the system. My patch works on MSYS2 properly, I'm just not sure it's the way it should be done.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ def get_geos_config(option):
     cmd = os.environ.get("GEOS_CONFIG", "geos-config")
     try:
         stdout, stderr = subprocess.Popen(
-            [cmd, option], stdout=subprocess.PIPE, stderr=subprocess.PIPE
+            ['bash', '-c', cmd+' '+option], stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True
         ).communicate()
     except OSError:
         return


michalfapso:
I see it's failing on linux and macos, I'm just not sure how to write it properly. It actually doesn't work well even on Windows/MSYS2 when `GEOS_CONFIG` env variable contains spaces in the path. Do you have any ideas how to fix it properly?

caspervdw:
Thanks for your report @michalfapso 

The from-source installation is basically not supported on windows. The docs explicitly state that we don't have a recipe for that: https://shapely.readthedocs.io/en/stable/installation.html#installation-from-source-with-custom-geos-libary

It is not our aim to have the setup.py script deal with all different systems out there. That is the job of the binary distributions. For this specific case I'd recommend setting `GEOS_INCLUDE_PATH` and `GEOS_LIBRARY_PATH` manually, circumventing the whole need of geos-config. I hope that works out for you. I'll close this PR, please open an issue if you want to discuss this in greater detail.

michalfapso:
Thanks @caspervdw!

Setting those `GEOS_*` variables is indeed a nicer solution and it works well in Msys2.

The reason I needed to compile shapely from source is because `pip install shapely` fails because of the dependency on `oldest-supported-numpy` which unfortunately contains a bug which causes a compilation error on Msys2. It was fixed in newer numpy versions, so I had to download shapely from git and change the dependency to `numpy>=0.16` which allows it to use the newer numpy package already provided by Msys2.

Thanks again!

PR #1453, opened 2022-07-18
Require Python 3.7+, NumPy 1.14+

mwtoews:
For Shapely 2.0, it time to finally drop Python 3.6. NumPy 1.14+ is required for Python 3.7+ support.

As well removing an old version of Python, this will enable #1426 to move static metadata to `pyproject.toml`.

Closes #1205

Issue #1670, opened 2022-12-16
minimum_rotated_rectangle is incorrect in 2.0.0


sdoose:
## Expected behavior and actual behavior.

In shapely 2.0.0 the following code produces an incorrect
minimum_rotated_rectangle with area=20.4 when it should be 20.

In shapely <2.0.0 the correct minimum_rotated_rectangle with area=20 is found.

## Steps to reproduce the problem.
```python
"""
In shapely 2.0.0 the following code produces an incorrect
minimum_rotated_rectangle with area=20.4 and raises:

>       assert new_minimum_rotated_rectangle.area == true_minimum_rotated_rectangle.area
E       assert 20.400000000000006 == 20.0
E        +  where 20.400000000000006 = <POLYGON ((5.1 5.3, 1.5 6.5, -0.2 1.4, 3.4 0.2, 5.1 5.3))>.area
E        +  and   20.0 = <POLYGON ((1 1, 1 6, 5 6, 5 1, 1 1))>.area
bug_in_2_0_0.py:32: AssertionError
"""
import numpy as np
import matplotlib.pyplot as plt
from shapely.geometry import MultiPoint, Polygon

def test_minimum_rotated_rectangle():
    points = np.array(
        [[1, 1], [1, 5], [3, 6], [4, 2], [5, 5]]
    )
    pts = MultiPoint(points)
    new_minimum_rotated_rectangle = pts.minimum_rotated_rectangle

    true_minimum_rotated_rectangle = Polygon([[1, 1], [1, 6], [5, 6], [5, 1]])

    fig, ax = plt.subplots(nrows=1, ncols=1, figsize=(5, 5))
    ax.plot(*points.T, c="red", marker=".")
    ax.plot(*np.array(new_minimum_rotated_rectangle.boundary.coords).T,
            c="green", marker="o", alpha=0.2
            )
    ax.plot(*np.array(true_minimum_rotated_rectangle.boundary.coords).T,
            c="blue", marker="X", alpha=0.2
            )
    ax.set(xlim=(0, 10), ylim=(0, 10))
    plt.show()

    print(true_minimum_rotated_rectangle.area)
    print(true_minimum_rotated_rectangle.boundary)
    print(new_minimum_rotated_rectangle.area)
    print(new_minimum_rotated_rectangle.boundary)
    assert new_minimum_rotated_rectangle.area == true_minimum_rotated_rectangle.area
```

## Operating system

Ubuntu, MacOS, Windows

## Shapely version and provenance

[...]

PR #1885, opened 2023-09-08
Drop GEOS 3.7, add GEOS 3.12

caspervdw:
Follow-up from https://github.com/shapely/shapely/pull/1453

Ubuntu 18 is now EOL, so 3.7 can be dropped.


caspervdw:
I have been reading up on https://github.com/shapely/shapely/issues/1670 and https://github.com/shapely/shapely/pull/1708 considering the backwards incompatible change in `oriented_envelope` in GEOS 3.12

My personal opinion is that shapely shouldn't attempt to cover up backwards incompatible changes in GEOS. We should do nothing short of documenting that it happened. 

In this specific case we already had a solution in Shapely 1.x, so porting that back for GEOS 3.11 users could be done, however I think it is a very small user base that doesn't want to go forward to GEOS 3.12 so I think it is better to go forward and not fix anything at all.

shapely/constructive.py
```diff
@@ -971,16 +969,16 @@ def voronoi_polygons(
     --------
     >>> from shapely import LineString, MultiPoint, normalize, Point
```
mwtoews, line 970:
```suggestion
    >>> from shapely import LineString, MultiPoint, Point
```

mwtoews (approved):
Thanks for putting this together!

There is still one hidden failure with GEOS main from your upstream fixes:

    pytest shapely/tests/test_creation.py::test_points_nan_becomes_empty

Is it worth undoing f115e06adad5e5d494d6d9b847a65c47d042734c ?

caspervdw:
Thanks for the review @mwtoews . I’ll finish this up and then merge.

about the NaN points: I was planning to have a good look at the current situation when finishing https://github.com/shapely/shapely/pull/1811

PR #1886, opened 2023-09-12
CLN: Use exposed `contains_properly` for `PreparedGeometry`

### LINKED ISSUE #1875: contains_properly for PreparedGeometry does not accept array_like inputs.

For a prepared geometry, the contains_properly does not accept array_like inputs. This in contrast to all other methods.

This is related to the custom logic with STRtree for contains_properly that does take into account the fact that 'other' can be array-like.

```python
from shapely import Polygon

poly = Polygon(((0.0, 0.0), (0.0, 1.0), (1.0, 1.0), (1.0, 0.0)))

prep(poly).contains_properly(poly)         # this is ok, returns False

# this is not ok, raises: TypeError: Array should be one dimensional
# but you would expect it to return array([False])
prep(poly).contains_properly([poly]) 

# for example, the following returns array([True]) as you would expect
prep(poly).contains([poly])
```

## Tentative fix

I can make a pull-request to fix this issue if you want, but I am a bit hazy on where to add a test to catch this behavior. It seems like most tests for the prepared geometry have ended up in the legacy folder, and I am assuming we shouldn't be adding stuff there.

The fix would look something like this for the PreparedGeometry method:

```python
def contains_properly(self, other):
    """Returns True if the geometry properly contains the other, else False"""
    # TODO temporary hack until shapely exposes contains properly as predicate function
    from shapely import STRtree
    import numpy as np
    
    is_scalar = np.ndim(other) == 0
    other = np.atleast_1d(other)
    tree = STRtree(other)
    idx = tree.query(self.context, predicate="contains_properly")

    if is_scalar:
        return bool(len(idx))

    output = np.zeros_like(other, dtype=bool)
    output[idx] = True
    return output
```

## Operating system

Linux mint 21.2

Shapely 2.0.1 on Python 3.10 installed via pip.

JuriaanSioux:
Use the exposed geos function now, as it has been exposed now. Also resolves #1875.

### DIFF
--- a/shapely/prepared.py
+++ b/shapely/prepared.py
@@ -33,12 +33,7 @@ class PreparedGeometry:
 
     def contains_properly(self, other):
         """Returns True if the geometry properly contains the other, else False"""
-        # TODO temporary hack until shapely exposes contains properly as predicate function
-        from shapely import STRtree
-
-        tree = STRtree([other])
-        idx = tree.query(self.context, predicate="contains_properly")
-        return bool(len(idx))
+        return self.context.contains_properly(other)
 
     def covers(self, other):
         """Returns True if the geometry covers the other, else False"""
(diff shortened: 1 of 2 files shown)


jorisvandenbossche (approved):
Thanks!

jorisvandenbossche:
It seems `PreparedGeometry.contains_properly` actually wasn't covered at all in the tests, so updated an existing test to include it.

Issue #1852, opened 2023-07-11
Shapely 2.0.2 micro release?


mwtoews:
A micro release from `maint-2.0` could help resolve upstream GEOS 3.12 changes, e.g., see https://github.com/conda-forge/shapely-feedstock/pull/125

However, I suggest the binary wheels stick to GEOS 3.11.x. <s>I'll create a PR to bump the release to use [GEOS 3.11.2](https://github.com/libgeos/geos/releases/tag/3.11.2).</s> (not needed, it already is set to 3.11.2).

Here are the other pending changes: https://github.com/shapely/shapely/compare/2.0.1...maint-2.0

mwtoews:
Pretty much the only blocking issue is #1670 which currently fail the test suite with the latest release of GEOS 3.12.0

Is #1708 suitable for inclusion for `maint-2.0` / shapely 2.0.2?

musicinmybrain:
> Pretty much the only blocking issue is #1670 which currently fail the test suite with the latest release of GEOS 3.12.0

This matches what I see when I try packaging a snapshot of the `maint-2.0` branch, currently at eeb51ad4d80c902d5f0bcad209687f046019cb6e, for Fedora Rawhide. The only remaining test failures I see are:

```
=========================== short test summary info ============================
FAILED tests/test_constructive.py::test_oriented_envelope[geometry0-expected0]
FAILED tests/test_constructive.py::test_oriented_envelope[geometry2-expected2]
FAILED tests/test_constructive.py::test_minimum_rotated_rectangle[geometry0-expected0]
FAILED tests/test_constructive.py::test_minimum_rotated_rectangle[geometry2-expected2]
============ 4 failed, 3462 passed, 27 skipped, 3 xfailed in 6.65s =============
```

jorisvandenbossche:
I have a PR with a (final?) set of backports: https://github.com/shapely/shapely/pull/1904 (the eq regression, cython 3 fix, and linux aarch64 wheels update)

Then I will also make a PR with updating the release date etc.

jorisvandenbossche:
cc @shapely/shapely-team

caspervdw:
I am :+1: on doing the release after #1904 is in.

jorisvandenbossche:
I would like to include https://github.com/shapely/shapely/pull/1708 (the minimum_rotated_rectangle change) as well, given that this fixes a change in 2.0.x compared to 1.8

jorisvandenbossche:
[...]

PR #1893, opened 2023-09-22
Reenable Python 3.12 wheel building

djhoese:
I'm working on building Python 3.12 wheels for some of my packages that depend on shapely. It'd be great to have a shapely wheel for Python 3.12.

Side note: I think in pyproject.toml you're supposed to be able to just say "numpy" from 1.25 or 1.26 on. Of course I can't find it now, but I believe it is supposed to be bundled with backwards compatible APIs...I think.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -117,7 +117,7 @@ jobs:
         uses: pypa/cibuildwheel@v2.16.0
         env:
           CIBW_ARCHS: ${{ matrix.arch }}
-          CIBW_SKIP: cp36-* pp* *musllinux* *-manylinux_i686 cp312-*
+          CIBW_SKIP: cp36-* pp* *musllinux* *-manylinux_i686
           CIBW_ENVIRONMENT_LINUX:
             GEOS_VERSION=${{ env.GEOS_VERSION }}
             GEOS_INSTALL=/host${{ runner.temp }}/geos-${{ env.GEOS_VERSION }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,12 +38,14 @@ jobs:
             numpy: 1.24.4
             matplotlib: true
             doctest: true
+            extra_pytest_args: "-W error -W ignore:datetime.datetime.utcfromtimestamp\(\) is deprecated and scheduled for removal:DeprecationWarning" # error on warnings
           - python: "3.12"
             geos: 3.12.0
             numpy: 1.26.0
           # dev
           - python: "3.11"
             geos: main
+            extra_pytest_args: "-W error -W ignore:datetime.datetime.utcfromtimestamp\(\) is deprecated and scheduled for removal:DeprecationWarning" # error on warnings
           # enable two 32-bit windows builds:
           - os: windows-2019
             architecture: x86
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -41,7 +41,7 @@ jobs:
             # extra ignore for dateutil Python 3.12 warning fixed upstream (waiting on release 2.8.3+)
             extra_pytest_args: "-W error -W ignore:datetime.datetime.utcfromtimestamp:DeprecationWarning" # error on warnings
           # dev
-          - python: "3.11"
+          - python: "3.12"
             geos: main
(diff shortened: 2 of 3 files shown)


djhoese:
As a library maintainer, I'm also curious why you make the individual macos wheels and also the universal2 wheel?

caspervdw:
> As a library maintainer, I'm also curious why you make the individual macos wheels and also the universal2 wheel?

I am not sure actually. Looking at numpy, they *stopped* the universal2 wheel since version 1.22. https://github.com/numpy/numpy/issues/21233

Maybe we should too? Something for a different PR.

Either way, good to start testing and distribution for Python 3.12. Could you also append the test.yml?

djhoese:
I added it...I think, but I wasn't really sure what pattern to follow. Numpy 1.26.0 was just released so I set numpy to that. I also added a flag so Python 3.12 pre-releases can get used since 3.12 doesn't have a stable release out yet.

djhoese:
All the failures seem to be from this deprecation warning:

```
 >   EPOCH = datetime.datetime.utcfromtimestamp(0)
E   DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
```

djhoese:
Oh this is actually coming from dateutil which seems decently behind in stable releases supporting modern python versions. Maybe add this warning to the ignored warnings?

djhoese:
Ah it's already fixed, they just need a release: https://github.com/dateutil/dateutil/commit/f2293200747fb03d56c6c5997bfebeabe703576f

djhoese:
Ok, this is getting worse and worse. There was the deprecation warning about dateutil so I was going to filter it in setup.cfg, but pytest config `filterwarnings` are not the priority anymore so the `-W error` in the CI execution were still making it error out. So I removed the `-W error` and added `error` to the config...but now there is a pyproj warning that seems like something that should really be fixed...but also isn't specific to Python 3.12 from what I can tell?

mwtoews (approved):
Thanks @djhoese, this looks good! I'll be interested in looking at the macos and numpy questions in the next few months. Some of the cibuildwheel configuration is different than numpy as shapely covers a few earlier versions of Python.

mwtoews:
I've pushed some changes to use the "2023" matrix item for as the actual release year for Python 3.12, and also use the "dev" matrix item for Python 3.12.

djhoese:
> I've pushed some changes to use the "2023" matrix item for as the actual release year for Python 3.12, and also use the "dev" matrix item for Python 3.12.

Sounds good.

>  I'll be interested in looking at the macos and numpy questions in the next few months. Some of the cibuildwheel configuration is different than numpy as shapely covers a few earlier versions of Python.

This comes from a place of being a library maintainer myself, what's the reason for supporting older versions of Python?

mwtoews:
> This comes from a place of being a library maintainer myself and curiousity, what's the reason for supporting older versions of Python?

Fair question. The current status quo is to provide built wheels for the 4 to 5 most recent Python versions. Unifying around [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) and/or [SPEC 0](https://scientific-python.org/specs/spec-0000/) is an open discussion. There is now a good argument for adopting these policies, as NumPy is was brought as a requirement since Shapely 2.0. (Also, my personal observation is that folks are quicker to upgrade their Python environments now than they were, say, 5 years ago when there was an awkward mix of Python 2 and 3.)

djhoese:
>  (Also, my personal observation is that folks are quicker to upgrade their Python environments now than they were, say, 5 years ago when there was an awkward mix of Python 2 and 3.)

Ah ok. I thought you were saying much older versions (3.7 or older). Yeah I agree with this observation. I also like the excuse of "oh you want the new features in the library, you'll need to update your version of python, please use installation solution X, Y, or Z".

djhoese:
So the obvious question is, what is going to be the next released version and when do you expect to be able to make it?

mwtoews:
Appreciate your patience @djhoese as I'd like to see movement on #1852 too. Shapely 2.1 is not ready yet, but a micro release from `maint-2.0` should be ready. Xref #1898

PR #1898, opened 2023-10-01
Prepare for Python 3.12 wheels for 2.0.2

mwtoews:
* Bump pypa/cibuildwheel from 2.14.1 to 2.16.2
* Don't build manylinux_i686
* Add Python 3.12 classifier

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,8 +53,9 @@ jobs:
         include:
         - os: ubuntu-20.04
           arch: x86_64
-        - os: ubuntu-20.04
-          arch: i686
+        # Numpy no longer builds i686 packages
+        # - os: ubuntu-20.04
+        #   arch: i686
         # The aarch64 build has been transferred to Travis
         # - os: ubuntu-20.04
         #   arch: aarch64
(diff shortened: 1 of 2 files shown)


mwtoews:
Note that this would enable 6 Python versions (3.7, 3.8, 3.9, 3.10, 3.11, 3.12) for an upcoming Shapely 2.0.2 micro release (#1852). This is an unusually high number of Python releases.

.github/workflows/release.yml
```diff
@@ -53,8 +53,9 @@ jobs:
         include:
         - os: ubuntu-20.04
           arch: x86_64
-        - os: ubuntu-20.04
-          arch: i686
+        # Numpy no longer builds i686 packages
+        # - os: ubuntu-20.04
+        #   arch: i686
```
jorisvandenbossche, line 58: cc @caspervdw just to ensure you are aware (I think you originally added  this platform to the wheel building?)

For any other project I am involved in, we also don't distribute i686 wheels, so I am certainly fine with dropping them. But just to note that theoretically we can still provide them if there is a good use case (we don't theoretically _need_ numpy wheels (practically is something else ..), and certainly for users of this platform, building numpy from source with `pip install numpy` is typically easier than shapely, as we require an external C++ dependency)

jorisvandenbossche (approved):
Looks good, thanks!

.github/workflows/release.yml
djhoese: If someone goes to the work of building numpy from source on 32-bit platforms then these should really be done as a separate project so others have access to them. That said, if numpy dropped the building of them I have a feeling it was more work than it was worth (not just build times).

That said...if someone makes 32-bit wheels for numpy let me know.

caspervdw (approved):

mwtoews:
Just modified the minor version of cibuildwheel to [v2.16.2](https://github.com/pypa/cibuildwheel/releases/tag/v2.16.2) released 20 mins ago for the final Python 3.12.0 release.

.github/workflows/release.yml
mwtoews: FYI, NumPy's last built wheels for i686 were for [version 1.21.6](https://pypi.org/project/numpy/1.21.6/#files) from Apr 13, 2022. Fixing shapely to to use these older numpy's would need older cibuildwheel images to be configured to match.

PR #1902, opened 2023-10-03
Add transform_coordseq() and deprecate ops.transform()

caspervdw:
This proposal supersedes #1676 . I made a function specification to discuss about. The rationale here being:

- The consolidation of the current `shapely.transform` and `shapely.ops.transform` explored in #1676 by @idanmiara made clear that mixing the two results in the unnecessary addition of  cases for which the implementation turns out the be pretty hard. Specifically: "resizing" geometries in a geometry array of mixed types.
- Shapely should have a consistent API; the current two functions are confusing to most users.
- There should remain something that is backwards compatible with that `shapely.ops.transform`. We don't want a hard transition for existing users. 

To avoid the combinatorics, I opt for mostly a copy-paste of the existing `shapely.ops.transform`, with the `include_z` parameter added. Please review the docs I wrote carefully so that we are clear on the intended functionality. @jorisvandenbossche @idanmiara


shapely/coordinates.py
```diff
@@ -129,6 +142,127 @@ def transform(
...
+            return transformation(coords)
+        else:
+            # A transformation that assumes x, y, z are single values will likely raise a
+            # TypeError, in which case we'll try again.
+            try:
+                return zip(*transformation(*coords.T))
+            except TypeError:
+                # A func that assumes x, y, z are single values will likely raise a
+                # TypeError, in which case we'll try again.
+                return [transformation(*c) for c in coords]
```
jorisvandenbossche, line 235: Personally I don't really like this try/except attempt to "infer whether the function is array-wise or element-wise". 

We _could_ also consider providing that as two separate transform functions, if we think this is worth providing to the user, to make it a clear choice for the user (I think typically the user will always know / have control over what kind of function they have). 

Another reason making this mixture messy is that AFAIK the second one can actually never resize your coordinates? So the use case of "resize coordseq" and "scalar function" will also never overlap?

shapely/coordinates.py
```diff
@@ -129,6 +142,127 @@ def transform(
...
+    of coordinate pairs.
+
+    This function differs with `transform` in the following ways:
+
+    - The number of coordinates per Geometry is allowed to change.
+    - The input parameters to the transformation function can also be (scalar) floats.
+    - It only accepts scalar Geometry objects, not arrays.
+
+    We recommend only using this function only if 1) the number of coordinates in your Geometry
+    should change or 2) the transformation can't be made to accept numpy arrays.
```
jorisvandenbossche, line 161: I think another defining characteristic of this version is that the function is applied _per coordinate sequence_ (ring), instead of applied to all coordinates at once?

jorisvandenbossche (commented):
Some quick remarks on the general design. And thanks for looking into this!

shapely/coordinates.py
caspervdw: I also dislike this way inferring - I kept it only to make the migration of existing users of `ops.transform` easy.

> Another reason making this mixture messy is that AFAIK the second one can actually never resize your coordinates? So the use case of "resize coordseq" and "scalar function" will also never overlap?

That is well observed :) so in that case the "scalar function" version would better fit the existing `shapely.transform` function? I could think of an `call_with_scalars` kwarg for that function.

jorisvandenbossche: > so in that case the "scalar function" version would better fit the existing `shapely.transform` function? I could think of an `call_with_scalars` kwarg for that function.

I don't really like adding such a keyword, but at the same providing a third transform method is also not great. 

But a counterargument for a keyword to `transform` is that this might conflict/ be incompatible with other keywords (like interleaved)

Now, I assume we can leave that for another issue, and you are focusing this PR just on the new resize version?

jorisvandenbossche:
Seeing your latest doc updates and my comment about the "per coord seq" as main characteristic, I am wondering if a name like `transform_coordseq` (or a variant of this) would be a better name than `transform_resize`. 

For example, assume you have some custom simplification algorithm, and want to use this: you will make use of the fact that it can resize (eg reduce the number of coord pairs per sequence), but the more important thing is that you consider sequence per sequence, instead of all coordinates at once (without knowing which coordinate belongs to which geometry).

caspervdw:
> Seeing your latest doc updates and my comment about the "per coord seq" as main characteristic, I am wondering if a name like `transform_coordseq` (or a variant of this) would be a better name than `transform_resize`.

I gave this some thought and I really can't think of better alternatives then `transform_resize`. I think it is the best option and the docstring clarifies that is functions per coordseq.

Test failures seem unrelated BTW, something on MacOS with GEOS 3.8. I think we may safely ignore it.

shapely/coordinates.py
```diff
@@ -129,6 +141,116 @@ def transform(
...
+
+    The transformation function is applied per coordiante sequence. For polygons this
+    means: per ring. For collections this means: per element.
+
+    This function differs with `transform` in the following ways:
+
+    - It only accepts scalar Geometry objects, not arrays.
+    - The number of coordinates per coordinate sequence is allowed to change.
+
+    We recommend only using this function only if the number of coordinates in your
```
idanmiara, line 161: maybe state the reason? "Because this function is expected to be slower than `transform`, ...."

caspervdw:
I think this can be merged.   @jorisvandenbossche  can you confirm?

caspervdw:
@jorisvandenbossche  I am trying to spend a bit more time on open source again, shapely being one of the projects I want to resume working on. This PR is basically done, I dusted it off a bit and it should be ready for merge.

If you have time please check. And if there is something else that can use my attention, please let me know.

caspervdw:
I believe this PR has been reviewed well enough to warrant merging. I might even consider myself as an independent reviewer because I copied and edited @idanmiara 's work 2 years ago.

PR #1905, opened 2023-10-08
RLS/BLD: also build Python 3.12 wheels for linux aarch64

jorisvandenbossche:
We added those for the github actions based workflows, but not yet for CircleCI for Linux aarch64

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,9 +34,9 @@ workflows:
   wheel-build:
     jobs:
       - linux-aarch64-wheels:
-          filters:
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
I pushed this to upstream shapely as well (instead of my fork), so this builds on CircleCI (we don't build for forks). And it's passing there, with wheel for Python 3.12 now included: https://app.circleci.com/pipelines/github/shapely/shapely/152/workflows/9660e855-2c45-4371-9021-24d4479af692/jobs/101

PR #1907, opened 2023-10-08
Protect against Abort for wkt output with very large floats

### LINKED ISSUE #1903: shapely crashes on mac with "abort" on very large floats

If passed a very large float, `shapely` should maintain that representation, change to infinity, raise an exception, etc. Instead, I am getting a hard crash

Here is a minimalized example:
```python
import shapely
p = shapely.Point(1e+279, 0)
print(p)
```

On my Mac:

```shell
python shapely_thing.py
zsh: abort      poetry run python shapely_thing.py
```
## Operating system

Mac OS X 13.4.1

```python
>>> shapely.__version__
'2.0.1'
```

Installed from `PyPI` using `poetry`

## Other notes

1) It appears that the constructor runs fine (commenting out `print(p)` gets rid of the error)
2) If I replace `1e+279` with `np.inf` it also works fine (returns a point with Infinity as one of the coordinates)
3) If I replace `1e+279` with `1e+278` it works fine.
4) it works okay on linux (Ubuntu 20.04.4 LTS), the print statement gives this:

```shell
POINT POINT (1000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 0)```

### PR DESCRIPTION
**Pull request by caspervdw**
Closes #1903

### CODE REVIEW

**Comment by caspervdw** [comment]
Performance doesn't seem to suffer significantly:

this PR: 
```
[62.50%] ··· benchmarks.IOSuite.time_read_from_wkb                                                                                                                144±2ms
[75.00%] ··· benchmarks.IOSuite.time_read_from_wkt                                                                                                                936±2ms
[87.50%] ··· benchmarks.IOSuite.time_write_to_wkb                                                                                                                88.5±2ms
[100.00%] ··· benchmarks.IOSuite.time_write_to_wkt                                                                                                               449±0.8ms
```

vs. main:

```
[12.50%] ··· Running (benchmarks.IOSuite.time_read_from_wkb--)....
[62.50%] ··· benchmarks.IOSuite.time_read_from_wkb                                                                                                              143±0.9ms
[75.00%] ··· benchmarks.IOSuite.time_read_from_wkt                                                                                                                910±1ms
[87.50%] ··· benchmarks.IOSuite.time_write_to_wkb                                                                                                                87.5±1ms
[100.00%] ··· benchmarks.IOSuite.time_write_to_wkt                                                                                                               445±0.9ms
```

**Comment by caspervdw** [comment]
BTW, I just confirmed that the same issue is there for the Z coordinate, and we dont have a GetZMax GEOS function...

**Review by jorisvandenbossche — APPROVED** [review]
Looks good!

Do we want to include it for 2.0.2?

**Comment by jorisvandenbossche** [comment]
Seems we would need a lower threshold for Windows ..

**Comment by caspervdw** [comment]
> Do we want to include it for 2.0.2? 

Don’t think we should rush this for 2.0.2. It’s been there probably for a long time and it only just came up. I also think we should fix this for the Z coordinate.

> Seems we would need a lower threshold for Windows ..

I guess we need to approach the 128-size buffer mentioned in the GEOS issue theoretically.

**Comment by caspervdw** [comment]
The additional check on the Z coordinate does give ~ 7% performance regression:

```
· Discovering benchmarks
· Running 4 total benchmarks (1 commits * 1 environments * 4 benchmarks)
[ 0.00%] ·· Benchmarking existing-py_home_casper_code_shapely_.venv_bin_python
[12.50%] ··· Running (benchmarks.IOSuite.time_read_from_wkb--)....
[62.50%] ··· benchmarks.IOSuite.time_read_from_wkb                                                                                                              144±0.8ms
[75.00%] ··· benchmarks.IOSuite.time_read_from_wkt                                                                                                                912±3ms
[87.50%] ··· benchmarks.IOSuite.time_write_to_wkb                                                                                                                87.8±1ms
[100.00%] ··· benchmarks.IOSuite.time_write_to_wkt                                                                                                               475±0.9ms
```

**Comment by jorisvandenbossche** [comment]
> The additional check on the Z coordinate does give ~ 7% performance regression:

Should there somewhere be a shortcut for not looping over the whole coordseq if the geometry actually has no z coordinates? (in that case GEOSCoordSeq_getZ_r just always gives NaNs?)

**Comment by caspervdw** [comment]
> > The additional check on the Z coordinate does give ~ 7% performance regression:
> 
> Should there somewhere be a shortcut for not looping over the whole coordseq if the geometry actually has no z coordinates? (in that case GEOSCoordSeq_getZ_r just always gives NaNs?)

The reason I do this for all geometries, is that `GetCoordinateDimension` will return `2` if the first Z coordinate is a NaN - but then other coordinates could still contain out of bound coordinates resulting in a crash.
But we're lucky: GEOS 3.11 *also* does the same [link](https://github.com/libgeos/geos/blob/3.11/src/io/WKTWriter.cpp#L193) to decide whether to write Z coordinates in WKT.

GEOS 3.12 does `HasZ` when *not* removing empty dimensions (https://github.com/libgeos/geos/blob/35c10bfbc79a1af42dc427072a4a3f663288db33/src/io/WKTWriter.cpp#L217C39-L217C43) which is the default, which boils down to the same behaviour as GEOS 3.11.

So I'll add the `HasZ` and some tests to verify

For collections we probably still need to iterate, but for simple geometries not.

**Comment by caspervdw** [comment]
Performance regression is down to an acceptable levels again:

```
[62.50%] ··· benchmarks.IOSuite.time_read_from_wkb             144±0.7ms
[75.00%] ··· benchmarks.IOSuite.time_read_from_wkt               912±3ms
[87.50%] ··· benchmarks.IOSuite.time_write_to_wkb               88.2±1ms
[100.00%] ··· benchmarks.IOSuite.time_write_to_wkt              452±0.6ms
```

PR #1920, opened 2023-10-24
Disable check_to_wkt_trim_compatible for GEOS >= 3.13.0

mwtoews:
This is a follow-up to #1907 that considers upstream changes for upcoming GEOS 3.13.0 in https://github.com/libgeos/geos/pull/973

It also fixes a related regression with tests against GEOS main branch, where small non-zero coordinates are shown with scientific notation (with trim enabled) rather than misrepresenting them as zero.

### DIFF
--- a/shapely/tests/geometry/test_format.py
+++ b/shapely/tests/geometry/test_format.py
@@ -93,7 +93,12 @@ def test_format_polygon():
     )
 
     # 'g' format varies depending on GEOS version
-    if geos_version < (3, 10, 0):
+    if geos_version >= (3, 13, 0):
+        expected_2G = (
+            "POLYGON ((10 0, 7.07 -7.07, 6.12E-16 -10, -7.07 -7.07, "
+            "-10 -1.22E-15, -7.07 7.07, -1.84E-15 10, 7.07 7.07, 10 0))"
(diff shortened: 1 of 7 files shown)


caspervdw (approved):

shapely/tests/geometry/test_format.py
```diff
@@ -93,7 +93,12 @@ def test_format_polygon():
     )
 
     # 'g' format varies depending on GEOS version
-    if geos_version < (3, 10, 0):
+    if geos_version >= (3, 13, 0):
+        expected_2G = (
+            "POLYGON ((10 0, 7.07 -7.07, 6.12E-16 -10, -7.07 -7.07, "
+            "-10 -1.22E-15, -7.07 7.07, -1.84E-15 10, 7.07 7.07, 10 0))"
```
mwtoews, line 99: FYI, I'm a bit uneasy (and pedantic) about this one an and considering an upstream change to zap the near-zero values to zero, so this test might change once more before GEOS 3.13 is finalized.

mwtoews: xref https://github.com/libgeos/geos/pull/978

PR #1938, opened 2023-11-22
TST: Adjust tests and require_geos decorator implementation for docstring indent stripping in Python 3.13

### LINKED ISSUE #1937: Test failure on Python 3.13 because it strips indents from docstrings

In Python 3.13 (currently in alpha), as described in https://docs.python.org/3.13/whatsnew/3.13.html, “Compiler now strip indents from docstrings.” See https://github.com/python/cpython/issues/81283.

There doesn’t yet seem to be a working Python 3.13 wheel for `numpy` on PyPI, which makes this awkward to reproduce in a virtualenv, but we are already doing [early Python 3.13 testing](https://copr.fedorainfracloud.org/coprs/g/python/python3.13/) in Fedora Linux.

As reported in https://bugzilla.redhat.com/show_bug.cgi?id=2250678, some tests fail due to the new docstring handling.

```
_______________________ test_requires_geos_method[3.8.0] _______________________

version = '3.8.0', mocked_geos_version = '3.7.1', sphinx_doc_build = None

    @pytest.mark.parametrize("version", ["3.6.0", "3.8.0"])
    def test_requires_geos_method(version, mocked_geos_version, sphinx_doc_build):
        """The requires_geos decorator adjusts methods docstrings correctly"""
        wrapped = requires_geos(version)(SomeClass.func)
    
>       assert wrapped.__doc__ == expected_docstring.format(version=version, indent=" " * 8)
E       AssertionError: assert 'Docstring th...escription.\n' == 'Docstring th...on.\n        '
E           Docstring that will be mocked.
E         -         A multiline.
E         ? --------
E         + A multiline.
E           
E         -         .. note:: 'func' requires at least GEOS 3.8.0.
E         ? ------...
E         
E         ...Full output truncated (6 lines hidden), use '-vv' to show

tests/test_misc.py:123: AssertionError
=========================== short test summary info ============================
FAILED tests/test_misc.py::test_requires_geos_not_ok[3.7.2] - AssertionError:...
FAILED tests/test_misc.py::test_requires_geos_not_ok[3.8.0] - AssertionError:...
FAILED tests/test_misc.py::test_requires_geos_not_ok[3.8.1] - AssertionError:...
FAILED tests/test_misc.py::test_requires_geos_doc_build[3.6.0] - AssertionErr...
FAILED tests/test_misc.py::test_requires_geos_doc_build[3.8.0] - AssertionErr...
FAILED tests/test_misc.py::test_requires_geos_method[3.6.0] - AssertionError:...
FAILED tests/test_misc.py::test_requires_geos_method[3.8.0] - AssertionError:...
============ 7 failed, 3494 passed, 37 skipped, 3 xfailed in 5.74s =============
```

We expect that all tests should pass.

Compile Shapely and run the tests on Python 3.13. (The hard part is getting a proper installation of Numpy.)

## Operating system

Fedora Linux Rawhide – early testing of Python 3.13.

2.0.2, [`python-shapey`](https://src.fedoraproject.org/rpms/python-shapely)

-----

I intend to work on a PR for this.

musicinmybrain:
Fixes #1937.

I’m not thrilled with this, particularly with the unexplained `extra_indent`, but it does successfully match the behavior in Python 3.13.0a1.

For reference, here’s the current implementation of `_PyCompile_CleanDoc()`:

https://github.com/methane/cpython/blob/9c03215a3ee31462045b2c2ee162bdda30c54572/Python/compile.c#L7816

### DIFF
--- a/shapely/decorators.py
+++ b/shapely/decorators.py
@@ -39,7 +39,7 @@ class requires_geos:
             # Insert the message at the first double newline
             position = doc.find("\n\n") + 2
             # Figure out the indentation level
-            indent = 2
+            indent = 0
             while True:
                 if doc[position + indent] == " ":
                     indent += 1
--- a/shapely/tests/test_misc.py
+++ b/shapely/tests/test_misc.py
@@ -87,18 +87,18 @@ def expected_docstring(**kwds):
     doc = """Docstring that will be mocked.
 {indent}A multiline.
 
-{indent}{extra_indent}.. note:: 'func' requires at least GEOS {version}.
+{indent}.. note:: 'func' requires at least GEOS {version}.
 
 {indent}Some description.
-{indent}"""
+{indent}""".format(
+        **kwds
+    )
     if sys.version_info[:2] >= (3, 13):
         # There are subtle differences between inspect.cleandoc() and
         # _PyCompile_CleanDoc(). Most significantly, the latter does not remove
-        # leading or trailing blank lines. The extra_indent is required because
-        # the algorithm in shapely.decorators.requires_geos assumes a minimum
-        # indent of two spaces.
-        return cleandoc(doc.format(extra_indent="  ", **kwds)) + "\n"
-    return doc.format(extra_indent="", **kwds)
+        # leading or trailing blank lines.
+        return cleandoc(doc) + "\n"
+    return doc
 
 
 @pytest.mark.parametrize("version", ["3.10.0", "3.10.1", "3.9.2"])
--- a/shapely/tests/test_misc.py
+++ b/shapely/tests/test_misc.py
@@ -1,5 +1,6 @@
 import os
 import sys
+from inspect import cleandoc
 from itertools import chain
 from string import ascii_letters, digits
 from unittest import mock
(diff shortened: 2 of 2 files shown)


jorisvandenbossche:
@musicinmybrain Thanks for looking into this! We don't yet have CI for Python 3.13, but if you can test it on your end that this passes, I think we are certainly happy to already merge it.



> particularly with the unexplained `extra_indent`

That's maybe a bug to report to cpython? If I try with one of our docstrings with a `.. note::` using Python 3.11, that additional indent isn't present.

musicinmybrain:
I tried originally to form a simplified reproducer. For example, if I write `foo.py`:

```python
class SomeClass:
    def func(self):
        """Docstring that will be mocked.
        A multiline.

        .. note:: 'func' requires at least GEOS {version}.

        Some description.
        """
```

and then do

```
python3.13 -c 'from foo import SomeClass; print(repr(SomeClass.func.__doc__))'
```

I get what I would expect,

```
"Docstring that will be mocked.\nA multiline.\n\n.. note:: 'func' requires at least GEOS {version}.\n\nSome description.\n"
```

Now that I’m returning to this after a few days, I see exactly what’s happening. The docstring modification algorithm runs on the introspected (i.e. after byte-compilation) docstring, and it assumes the indentation level is at least `2`. Since the doc string has already had indents stripped, this isn’t a good assumption, but we end up with the note indented two spaces anyway.

https://github.com/shapely/shapely/blob/dba2704cdadbb5745edf9961f97aaf0c60826e42/shapely/decorators.py#L39-L50.

It seems like the choice of `2` is just a tiny performance optimization, and we could get rid of the “extra indent” without breaking anything else by simply starting at `0`. I’ll try it out.

musicinmybrain:
I began by force-pushing with an improved comment reflecting the above analysis.

musicinmybrain:
> It seems like the choice of `2` is just a tiny performance optimization, and we could get rid of the “extra indent” without breaking anything else by simply starting at `0`. I’ll try it out.

This approach appears to work as expected in my testing on Python 3.13 and 3.12. Updating this PR.

musicinmybrain:
If you wanted to find the indent more concisely without an explicit Python loop (in exchange for adding an otherwise-unnecessary temporary copy of part of the docstring), you could instead write:

```python
# Figure out the indentation level
indent = len(doc) - len(doc[position:].lstrip(" ")) - position
```

jorisvandenbossche:
Sorry for the slow follow-up, but rebased and still passing, so going to merge this now (and include in shapely 2.0.3). Thanks again!

Issue #1929, opened 2023-10-31
REGR: fallback oriented_envelope does not work with GeoSeries


martinfleis:
## Expected behavior and actual behavior.

When shapely 2.0.2 falls back to the `_oriented_envelope_min_area` algorithm when older GEOS is present, `shapely.oriented_envelope` fails when a GeoSeries is given.

## Steps to reproduce the problem.
```py
>>> import shapely
>>> import geopandas as gpd
>>> gs = gpd.GeoSeries([shapely.Point(0, 0).buffer(1), shapely.Point(90, 0).buffer(1)])
>>> shapely.oriented_envelope(gs)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/martin/miniforge3/envs/shapely_debug/lib/python3.12/site-packages/shapely/constructive.py", line 1034, in oriented_envelope
    return f(geometry, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/martin/miniforge3/envs/shapely_debug/lib/python3.12/site-packages/shapely/algorithms/_oriented_envelope.py", line 22, in _oriented_envelope_min_area
    if geometry.is_empty:
       ^^^^^^^^^^^^^^^^^
  File "/Users/martin/miniforge3/envs/shapely_debug/lib/python3.12/site-packages/pandas/core/generic.py", line 1519, in __nonzero__
    raise ValueError(
ValueError: The truth value of a Series is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
```

We need to pass a GeoSeries here but since it has a `geom_type` attribute, it is send to `geometry.is_empty` which fails.
https://github.com/shapely/shapely/blob/9540b77a939ba35d066d618183e67fc905ed9ec4/shapely/algorithms/_oriented_envelope.py#L20-L23

I suppose that replacing `not hasattr(geometry, "geom_type")` with some more generic check for array-like would fix it. Maybe check for `"len"`?

## Operating system

All of them

## Shapely version and provenance

Shapely 2.0.2 with GEOS < 3.12.

caspervdw:
Thanks @martinfleis for the report. It should be an easy fix.

PR #1939, opened 2023-11-26
REGR: ensure oriented_envelope (GEOS<3.12 code path) accepts array-like with ufunc protocol

jorisvandenbossche:
Closes https://github.com/shapely/shapely/issues/1929

In the end I thought to use `np.frompyfunc` and remove the manual iteration, instead of just fixing that single check for when to use the iteration code path. This ensures that it will work consistently with numpy in what it accepts (the kinds of array likes), and also that it will return the array like objects if those implement the numpy protocols (eg in shapely 2.0.1, calling this function on a geopandas.GeoSeries will also return a GeoSeries)

### DIFF
--- a/shapely/algorithms/_oriented_envelope.py
+++ b/shapely/algorithms/_oriented_envelope.py
@@ -17,8 +17,6 @@ def _oriented_envelope_min_area(geometry, **kwargs):
     """
     if geometry is None:
         return None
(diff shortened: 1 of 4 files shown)


jorisvandenbossche:
cc @martinfleis

brendan-ward (approved):
Thanks @jorisvandenbossche !

PR #1940, opened 2023-11-26
TST: move and consolidate Geometry __eq__ tests in geometry/test_equality.py

jorisvandenbossche:
The `test_eq` in `shapely/tests/test_geometry.py` is duplicated with `test_equality` in `shapely/tests/geometry/test_equality.py`

Deduplicating this, and also move the other equality (and hash) related tests in the former file to the latter, which already has more extensive tests on this topic.

### DIFF
--- a/shapely/tests/geometry/test_equality.py
+++ b/shapely/tests/geometry/test_equality.py
@@ -3,7 +3,11 @@ import pytest
 
 import shapely
 from shapely import LinearRing, LineString, MultiLineString, Point, Polygon
-from shapely.tests.common import all_types, all_types_z, ignore_invalid
+from shapely.tests.common import all_types, all_types_z, empty_point, ignore_invalid
+
+all_non_empty_types = np.array(all_types + all_types_z)[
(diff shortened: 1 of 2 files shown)


shapely/tests/geometry/test_equality.py
```diff
@@ -14,6 +18,13 @@ def test_equality(geom):
     assert not (geom != transformed)
 
 
+@pytest.mark.parametrize("geom", all_non_empty_types)
+def test_inequality(geom):
```
brendan-ward, line 22: Could this be combined with the hand-crafted `test_equality_false`, since both test inequality?

e. g., `[(geom, shapely.transform(geom, lambda x: x + 1)) for geom in all_non_empty_types] +` the set of test cases already there?

brendan-ward (commented):
Thanks @jorisvandenbossche !  One minor suggestion

shapely/tests/geometry/test_equality.py
jorisvandenbossche: That can indeed be combined!

Issue #1882, opened 2023-09-04
Add figures to all examples in docs


adishavit:
The Shapely docs are great.  
MIT would be even better if the example snippets had an accompanying figure for every example.  
Is there any reason this is not so?

jorisvandenbossche:
I fully agree that would be nice, but the main reason being that adding illustrations is also quite some work, and nobody started that for now. But contributions are certainly welcome!

PR #1941, opened 2023-11-27
ENH: Add parameters `method` and `keep_collapsed` to `make_valid` + improve doc

### LINKED ISSUE #1415: ENH: expose MakeValid parameters, introduced with GEOS 3.10.0
## Please note
Since this isn't a bug report, I'm not using the template, since it doesn't seem to fit.

## Feature request
Please expose the parameters to _MakeValid_, which have been introduced in _GEOS v3.10.0_ and solve most of the trouble we had with the "old" MakeValid.

## Background
In https://github.com/shapely/shapely/commit/1e5541a447901c27f7cb1884a416fb48a13e992c, _shapely_ got initial support for new _MakeValid_, which was introduced in https://github.com/libgeos/geos/blob/HEAD/NEWS.md#changes-in-380

In https://github.com/libgeos/geos/blob/HEAD/NEWS.md#changes-in-3100 there were added new parameters to _MakeValid_ with new validify enforcement approach from https://github.com/locationtech/jts/pull/704.

theroggy:
resolves #1415

reference #1882


shapely/constructive.py
```diff
@@ -525,8 +535,27 @@ def make_valid(geometry, **kwargs):
...
+
+        return lib.make_valid(geometry, **kwargs)
+
+    elif method == "structure":
+        if lib.geos_version < (3, 10, 0):
+            raise ValueError(
+                "The 'structure' method is only available in GEOS >= 3.10.0"
+            )
+
+        return lib.make_valid_with_params(geometry, 1, keep_collapsed, **kwargs)
```
brendan-ward, line 555:
```suggestion
        return lib.make_valid_with_params(geometry, np.intc(1), np.bool_(keep_collapsed), **kwargs)
```

brendan-ward (commented):
A couple suggestions to get the tests running.

shapely/constructive.py
```diff
@@ -508,12 +508,22 @@ def build_area(geometry, **kwargs):
...
     Parameters
     ----------
     geometry : Geometry or array_like
+    method:  {'linework', 'structure'}, default 'linework'
+        Algorithm to use when repairing geometry. The linework algorithm combines all
+        rings into a set of noded lines and then extracts valid polygons from that
+        linework. The structure algorithm first makes all rings valid, then merges
+        shells and subtracts holes from shells to generate valid result. It assumes that
+        holes and shells are correctly categorized.
+    keep_collapsed: bool, default True
```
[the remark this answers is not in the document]
theroggy: If I understand correctly, the linework algorithm doesn't support `keep_collapsed`=False based on this doc in Geos: https://libgeos.org/doxygen/geos__c_8h.html#a91442e5c832ac74d85f0fcc9731c769e

shapely/constructive.py
theroggy: Added casts

src/ufuncs.c
```diff
@@ -1555,6 +1555,91 @@ static void buffer_func(char** args, const npy_intp* dimensions, const npy_intp*
...
+  } else {
+    geom_arr[i] = GEOSMakeValidWithParams_r(ctx, in1, params);
+    if (geom_arr[i] == NULL) {
+      return PGERR_GEOS_EXCEPTION;
+    }
+  }
+  return PGERR_SUCCESS;
+}
+
+static char make_valid_with_params_dtypes[8] = {NPY_OBJECT, NPY_INT, NPY_BOOL,
```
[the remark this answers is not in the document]
theroggy: Corrected

shapely/constructive.py
```diff
@@ -525,8 +535,27 @@ def make_valid(geometry, **kwargs):
     False
     >>> make_valid(polygon)
     <MULTILINESTRING ((0 0, 1 1), (1 1, 1 2))>
+    >>> make_valid(polygon, method="structure", keep_collapsed=False)
+    <POLYGON EMPTY>
     """
-    return lib.make_valid(geometry, **kwargs)
+    if method == "linework":
```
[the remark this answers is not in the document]
theroggy: OK, done.

theroggy:
Odd that the runtime tests behave differently on windows vs linux. Thanks for the feedback! 

I made the proposed changes + added an extra check on GEOS version that should fix a compile error with older geos version.

So, ready for another try...

theroggy:
BTW... something I wanted to mention as I was a bit surprised by it... I don't very often work with lines, but apparently this line is valid: `LineString([(0, 0), (1, 1), (1, 2), (1, 1), (0, 0)])`... I didn't expect that.

brendan-ward:
> LineString([(0, 0), (1, 1), (1, 2), (1, 1), (0, 0)])
is valid as a `LineString` but not as a `LinearRing`.

shapely/constructive.py
```diff
@@ -525,8 +535,34 @@ def make_valid(geometry, **kwargs):
...
+    if not np.isscalar(keep_collapsed):
+        raise TypeError("keep_collapsed only accepts scalar values")
+
+    if method == "linework":
+        if keep_collapsed is False:
+            raise ValueError(
+                "The 'linework' method does not support 'keep_collapsed=False'"
+            )
+
+        return lib.make_valid(geometry, **kwargs)
```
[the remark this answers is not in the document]
theroggy: OK! I added a comment explaining this.

src/ufuncs.c
```diff
@@ -1555,6 +1555,95 @@ static void buffer_func(char** args, const npy_intp* dimensions, const npy_intp*
...
+  }
+
+  // allocate a temporary array to store output GEOSGeometry objects
+  geom_arr = malloc(sizeof(void*) * n);
+  CHECK_ALLOC(geom_arr);
+
+  GEOS_INIT_THREADS;
+
+  GEOSMakeValidParams* params = GEOSMakeValidParams_create_r(ctx);
+  if (params != 0) {
```
brendan-ward, line 1605:
```suggestion
  if (params != NULL) {
```

(it is a pointer to a struct instead of an integer return value, so I _think_ `NULL` is the right value here)

src/ufuncs.c
```diff
@@ -1555,6 +1555,95 @@ static void buffer_func(char** args, const npy_intp* dimensions, const npy_intp*
...
+      }
+      errstate = make_valid_with_params_inner(ctx, params, ip1, geom_arr, i);
+      if (errstate != PGERR_SUCCESS) {
+        destroy_geom_arr(ctx, geom_arr, i - 1);
+        break;
+      }
+    }
+  }
+
+  if (params != 0) {
```
brendan-ward, line 1631:
```suggestion
  if (params != NULL) {
```

brendan-ward (commented):
Thanks for working on this @theroggy !  Just a few more suggestions...

theroggy:
Suggestions applied...

theroggy:
@brendan-ward can you start the CI tests again on this PR.

shapely/constructive.py
```diff
@@ -508,12 +508,22 @@ def build_area(geometry, **kwargs):
...
 
 @multithreading_enabled
-def make_valid(geometry, **kwargs):
+def make_valid(geometry, method="linework", keep_collapsed=True, **kwargs):
     """Repairs invalid geometries.
 
     Parameters
     ----------
     geometry : Geometry or array_like
+    method:  {'linework', 'structure'}, default 'linework'
```
jorisvandenbossche, line 517:
```suggestion
    method :  {'linework', 'structure'}, default 'linework'
```

shapely/constructive.py
```diff
@@ -508,12 +508,22 @@ def build_area(geometry, **kwargs):
...
+        Algorithm to use when repairing geometry. The linework algorithm combines all
+        rings into a set of noded lines and then extracts valid polygons from that
+        linework. The structure algorithm first makes all rings valid, then merges
+        shells and subtracts holes from shells to generate valid result. It assumes that
+        holes and shells are correctly categorized. 'structure' requires GEOS >= 3.10.
+    keep_collapsed: bool, default True
```
jorisvandenbossche, line 523:
```suggestion
    keep_collapsed : bool, default True
```

shapely/constructive.py
```diff
@@ -508,12 +508,22 @@ def build_area(geometry, **kwargs):
...
 
     Parameters
     ----------
     geometry : Geometry or array_like
+    method:  {'linework', 'structure'}, default 'linework'
+        Algorithm to use when repairing geometry. The linework algorithm combines all
+        rings into a set of noded lines and then extracts valid polygons from that
+        linework. The structure algorithm first makes all rings valid, then merges
+        shells and subtracts holes from shells to generate valid result. It assumes that
+        holes and shells are correctly categorized. 'structure' requires GEOS >= 3.10.
```
jorisvandenbossche, line 522: What does "correctly categorized" mean here exactly?

jorisvandenbossche (commented):
Thanks a lot for working on this!

shapely/constructive.py
theroggy: @jorisvandenbossche Would you like me to try to rephrase to make it clearer, or do you have any suggestions?

jorisvandenbossche: Yes, I would prefer to rephrase it, or at least only write something that we ourselves at least understand what it means

jorisvandenbossche:
```suggestion
        holes and shells are correctly categorized in the original geometry. 'structure' requires GEOS >= 3.10.
```

Do you think that is correct? From reading the blogposts you referenced, my understanding of the structure algorithm is that it tries to keep the structure of the input geometry, i.e. what is interior/exterior in the original one is preserved. While the linework algorithm can introduce new holes where there were overlapping shells. 

If that's correct, the small clarification above might already help?

theroggy: I expanded the explanation based on another read of the blog posts and added an example image as well... Feedback welcome!

jorisvandenbossche: Very nice, thanks a lot!

shapely/constructive.py
```diff
@@ -508,12 +508,40 @@ def build_area(geometry, **kwargs):
...
+      geometries this produces rather complex results.
+    * the 'structure' algorithm tries to reason from the structure of the input to find
+      the 'correct' repair: exterior rings bound area, interior holes exclude area.
+      It first makes all rings valid, then shells are merged and holes are subtracted 
+      from the shells to generate valid result. It assumes that holes and shells are
+      correctly categorized in the input geometry.
+
+    Example:
+
+    |make_valid_methods|
```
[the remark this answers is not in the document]
jorisvandenbossche: And I _think_ you can also use `/_static/images/make_valid_methods.png` (i.e. absolute path, which means relative from the source root), then it doesn't depend on where this file lives (to ensure a correct relative path)

The `_static` subdirectory is not necessarily needed (that name is used by sphinx in the generated doc build, but we don't necessarily have to use that for the sources), it could also be just `/images`.

jorisvandenbossche (commented):
There is a small linting error to fix, see the CI build

shapely/constructive.py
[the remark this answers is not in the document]
theroggy: I had a look at how the images are inserted in manual.rst and that's a really nice system... so I use that now.

theroggy:
Feedback applied...

jorisvandenbossche (approved):
Thanks!

PR #1942, opened 2023-11-27
MAINT: update to latest GEOS point release in CI (GEOS 3.12.1 for wheels)

jorisvandenbossche:


jorisvandenbossche:
Opened an issue about the 3.9.5 failure on the GEOS side in case this is worth looking into (but just skipping here on our CI, for now): https://github.com/libgeos/geos/issues/1023

PR #1945, opened 2023-12-03
Fix incompatible pointer type passed to GEOSPolygonize_r

musicinmybrain:
Fixes the following warning by improving `const`-correctness:

```
  src/ufuncs.c: In function ‘polygonize_func’:
  src/ufuncs.c:2205:51: warning: passing argument 2 of ‘GEOSPolygonize_r’ from incompatible pointer type [-Wincompatible-pointer-types]
   2205 |     GEOSGeometry* ret_ptr = GEOSPolygonize_r(ctx, geoms, n_geoms);
        |                                                   ^~~~~
        |                                                   |
        |                                                   GEOSGeometry ** {aka struct GEOSGeom_t **}
  In file included from src/geos.h:15,
                   from src/ufuncs.c:16:
  /usr/include/geos_c.h:979:31: note: expected ‘const GEOSGeometry * const*’ {aka ‘const struct GEOSGeom_t * const*’} but argument is of type ‘GEOSGeometry **’ {aka ‘struct GEOSGeom_t **’}
    979 |     const GEOSGeometry *const geoms[],
        |     ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~
```

With Fedora Linux’s [PortingToModernC](https://fedoraproject.org/wiki/Changes/PortingToModernC) initiative, further described in [this mailing list message](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/GKCRXZESHSCCEKED2N5GNQ7GH32VSK2X/), this warning will become an error.

### DIFF
--- a/src/ufuncs.c
+++ b/src/ufuncs.c
@@ -2160,7 +2160,7 @@ static void polygonize_func(char** args, const npy_intp* dimensions, const npy_i
 
   GEOS_INIT;
 
-  GEOSGeometry** geoms = malloc(sizeof(void*) * dimensions[1]);
+  const GEOSGeometry** geoms = malloc(sizeof(void*) * dimensions[1]);
   if (geoms == NULL) {
     errstate = PGERR_NO_MALLOC;
     goto finish;


thesamesam:
FWIW, I still see:
```
src/ufuncs.c: In function ‘Ydd_b_p_func’:
src/ufuncs.c:436:23: warning: passing argument 2 of ‘func’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
  436 |       ret = func(ctx, prepared_geom_tmp, in2, in3);
      |                       ^~~~~~~~~~~~~~~~~
src/ufuncs.c:436:23: note: expected ‘void *’ but argument is of type ‘const GEOSPreparedGeometry *’ {aka ‘const struct GEOSPrepGeom_t *’}
```
but that's less of an issue.

src/ufuncs.c
```diff
@@ -2160,7 +2160,7 @@ static void polygonize_func(char** args, const npy_intp* dimensions, const npy_i
 
   GEOS_INIT;
 
-  GEOSGeometry** geoms = malloc(sizeof(void*) * dimensions[1]);
+  const GEOSGeometry** geoms = malloc(sizeof(void*) * dimensions[1]);
```
jorisvandenbossche, line 2163: Is it "correct" to mark `geoms` as const if we assign into `geoms` in the loop below to fill the array?

musicinmybrain: Yes. The type is changed from “pointer to pointer to `GEOSGeometry`” to “pointer to pointer to `const GEOSGeometry`”, or conceptually, from “array of pointers to `GEOSGeometry`” to “array of pointers to `const GEOSGeometry`.” We only assign pointers; we don’t modify the `GEOSGeometry` they reference, so this is fine.

We couldn’t change it to `const GEOSGeometry* const*`, because that *would* make the pointers `const`.

jorisvandenbossche: Ah, I was reading it as const array instead of array of const Geometry*. Thanks for the clarification!

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

PR #1948, opened 2023-12-14
DOC: fix plot_points color typo

nobkd:
`markeredgecolor` was duplicated and marker fill color (`markerfacecolor`) was not represented


jorisvandenbossche (approved):
Thanks!

PR #1953, opened 2023-12-18
Optimize GeoArrow/Arrow IO

paleolimbot:
Getting closer! This now works in both directions.

Currently requires a development version of nanoarrow for Python:

```
pip install "https://github.com/paleolimbot/arrow-nanoarrow/archive/3041a02594b9d8955fb578924befbb2adb7e53d4.zip#egg=nanoarrow&subdirectory=python"
```

```python
import pyogrio
import pyarrow as pa
import geoarrow.pyarrow as ga
import numpy as np
from shapely.geoarrow import to_pyarrow, from_arrow, infer_pyarrow_type, Encoding

# http://geoarrow.org/data
# ! curl -L https://github.com/geoarrow/geoarrow-data/releases/download/v0.1.0/ns-water-water_line-wkb.arrow -o ns-water-water_line-wkb.arrow
# ! curl -L https://github.com/geoarrow/geoarrow-data/releases/download/v0.1.0/ns-water-water_line.fgb.zip -o file.zip
# ! unzip -d . file.zip
# ! rm file.zip
df = pyogrio.read_dataframe("ns-water-water_line.fgb")

%timeit pa.array(df.geometry.to_wkb())
#> 1.13 s ± 5.77 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

from shapely.io import to_ragged_array
%timeit to_ragged_array(df.geometry)
#> 450 ms ± 15.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit to_pyarrow(df.geometry, Encoding.WKB)
#> 190 ms ± 570 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit to_pyarrow(df.geometry, Encoding.GEOARROW)
#> 220 ms ± 591 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit to_pyarrow(df.geometry, Encoding.GEOARROW_INTERLEAVED)
#> 217 ms ± 3.03 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit infer_pyarrow_type(df.geometry)
#> 45 ms ± 790 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

chunked = to_pyarrow(df.geometry)
%timeit from_arrow(chunked)
# 407 ms ± 10 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/shapely/geoarrow.py
+++ b/shapely/geoarrow.py
@@ -1,7 +1,6 @@
-
 import numpy as np
 
-from shapely._geoarrow import ArrayReader, GeoArrowGEOSException
+from shapely._geoarrow import ArrayReader, GeoArrowGEOSException  # NOQA
 
 
 def from_arrow(arrays, schema):
--- a/shapely/geoarrow.py
+++ b/shapely/geoarrow.py
@@ -1,7 +1,18 @@
 
-from shapely._geoarrow import ArrayReader
+import numpy as np
+
+from shapely._geoarrow import ArrayReader, GeoArrowGEOSException
 
 
 def from_arrow(arrays, schema):
     schema = schema.__arrow_c_schema__()
     reader = ArrayReader(schema)
+    geom_arrays = []
+    for array in arrays:
+        _, array = array.__arrow_c_array__()
+        geom_arrays.append(reader.read(array))
+
+    if geom_arrays:
+        return np.concatenate(geom_arrays)
+    else:
+        return np.array([], dtype=object)
--- a/shapely/geoarrow.py
+++ b/shapely/geoarrow.py
@@ -91,15 +91,15 @@ def to_pyarrow(obj, schema_to):
     return pa.chunked_array(chunks_pyarrow, type=schema_to)
 
 
-def from_arrow(arrays, schema):
+def from_arrow(arrays, schema, n=None):
     schema = schema.__arrow_c_schema__()
     reader = ArrayReader(schema)
-    geom_arrays = []
+
+    if n is not None:
+        reader.reserve(n)
+
     for array in arrays:
         _, array = array.__arrow_c_array__()
-        geom_arrays.append(reader.read(array))
+        reader.read(array)
 
-    if geom_arrays:
-        return np.concatenate(geom_arrays)
-    else:
-        return np.array([], dtype=object)
+    return reader.finish()
--- a/shapely/geoarrow.py
+++ b/shapely/geoarrow.py
@@ -11,7 +11,7 @@ def to_pyarrow(obj, schema_to):
 
     chunks_pyarrow = []
     for holder in builder.finish():
-        array = pa.Array._import_from_c(holder._addr(), schema_to.__arrow_c_schema__())
+        array = pa.Array._import_from_c(holder._addr(), schema_to)
         chunks_pyarrow.append(array)
 
     return pa.chunked_array(chunks_pyarrow, type=schema_to)
--- a/shapely/geoarrow.py
+++ b/shapely/geoarrow.py
@@ -18,84 +18,51 @@ class Encoding(enum.Enum):
     GEOARROW_INTERLEAVED = SchemaCalculator.ENCODING_GEOARROW_INTERLEAVED
 
 
-def type_pyarrow(encoding, geometry_type=None, dimensions=None):
-    import pyarrow as pa
+class GeoArrowTypeExporter:
 
-    encoding = Encoding(encoding).value
-
-    if geometry_type == GeometryType.POINT:
-        wkb_type = 1
-    elif geometry_type in (GeometryType.LINESTRING, GeometryType.LINEARRING):
-        wkb_type = 2
-    elif geometry_type == GeometryType.POLYGON:
-        wkb_type = 3
-    elif geometry_type == GeometryType.MULTIPOINT:
-        wkb_type = 4
-    elif geometry_type == GeometryType.MULTILINESTRING:
-        wkb_type = 5
-    elif geometry_type == GeometryType.MULTIPOLYGON:
-        wkb_type = 6
-    elif geometry_type == GeometryType.GEOMETRYCOLLECTION:
-        wkb_type = 7
-    else:
+    def __init__(self, encoding, geometry_type=None, dimensions=None):
+        encoding = Encoding(encoding).value
         wkb_type = 0
 
-    if dimensions is not None:
-        dimensions = dimensions.lower()
+        if geometry_type == GeometryType.POINT:
+            wkb_type = 1
+        elif geometry_type in (GeometryType.LINESTRING, GeometryType.LINEARRING):
+            wkb_type = 2
+        elif geometry_type == GeometryType.POLYGON:
+            wkb_type = 3
+        elif geometry_type == GeometryType.MULTIPOINT:
+            wkb_type = 4
+        elif geometry_type == GeometryType.MULTILINESTRING:
+            wkb_type = 5
+        elif geometry_type == GeometryType.MULTIPOLYGON:
+            wkb_type = 6
+        elif geometry_type == GeometryType.GEOMETRYCOLLECTION:
+            wkb_type = 7
+        elif geometry_type is not None:
+            raise ValueError(f"Unknown value for geometry_type: '{geometry_type}'")
 
-    if dimensions == "xy" or dimensions is None:
-        wkb_type += 0
-    elif dimensions == "xyz":
-        wkb_type += 1000
-    elif dimensions == "xym":
-        wkb_type += 2000
-    elif dimensions == "xyzm":
-        wkb_type += 3000
+        if dimensions is not None:
+            dimensions = dimensions.lower()
 
-    capsule = SchemaCalculator.from_wkb_type(encoding, wkb_type)
-    return pa.DataType._import_from_c_capsule(capsule)
+        if dimensions == "xy" or dimensions is None:
+            wkb_type += 0
+        elif dimensions == "xyz":
+            wkb_type += 1000
+        elif dimensions == "xym":
+            wkb_type += 2000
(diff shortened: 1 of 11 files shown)


jorisvandenbossche:
Just for clarity of the scope of the custom code _for shapely_: all the .c/.h files are currently vendored (from https://github.com/geoarrow/geoarrow-c-geos and https://github.com/geoarrow/geoarrow-c), and essentially only `_geoarrow.pyx`/`geoarrow.py` is new code here?

shapely/geoarrow.py
```diff
@@ -0,0 +1,17 @@
+import numpy as np
+
+from shapely._geoarrow import ArrayReader, GeoArrowGEOSException  # NOQA
+
+
+def from_arrow(arrays, schema):
```
kylebarron, line 6: Can we let `arrays` be either a single array or multiple? In the single case we'd check for `__arrow_c_array__` on the input directly?

In the future, this could check for `__arrow_c_stream__` for a chunked array? 

`schema` could also be optional here, in which case you just assert that the schemas of each array chunk are the same? Or, really, there's no reason to disallow arrays with differing schemas, right? So it seems preferable to load the schema from `__arrow_c_array__` anyways, unless instantiating `ArrayReader` is expensive.

paleolimbot: You're right on all counts! I didn't get very far in the implementation before hitting a segfault...probably the final version will use `nanoarrow.carray_stream()` to sanitize the input as an array stream, since that's basically what the array reader is designed to read (which would include anything that implements `__arrow_c_array__`).

paleolimbot: This show now work! Now the signature is just `from_arrow(array_or_stream)` 🙂

shapely/geoarrow.py
```diff
@@ -0,0 +1,163 @@
...
+    ArrayBuilder,
+    ArrayReader,
+    GeoArrowGEOSException,
+)  # NOQA
+
+
+class Encoding(enum.Enum):
+    WKB = SchemaCalculator.ENCODING_WKB
+    WKT = SchemaCalculator.ENCODING_WKT
+    GEOARROW = SchemaCalculator.ENCODING_GEOARROW
```
kylebarron, line 17: Does this default to separated encoding? Since GEOS is interleaved I have been defaulting to interleaved encoding for all my Python APIs (And I need interleaved for lonboard). But not sure which we should suggest.

kylebarron:
I think `to_pyarrow` but `from_arrow` might be a little confusing for users. Do you expect a `to_nanoarrow` as well? Should we have `to_arrow` with a potential keyword argument in the future to switch arrow impl?

shapely/geoarrow.py
kylebarron: Presumably this needs a non-table stream, right? I.e. it'll error on struct input that doesn't conform to points?

shapely/geoarrow.py
paleolimbot: This PR only just got to the point where we can test! I haven't been able to observe meaningful differences between the two in terms of timing, although for things like small arrays of very huge linestrings I wonder if it would pop through. The real benefit would be if GEOS made this type of export available internally, since then the overhead of all the individual C API calls wouldn't overwhelm the timing.

I think that the performance characteristics (e.g., write to file, filter, take, calculate bounding box) of the separated coordinate type (particularly for points) is better, so the default here is separated. I need to benchmark that to be sure, though.

For a library that lonboard that prefers interleaved values, you can always request them.

```
exporter = to_arrow(some_array)
// Not sure if pa.data_type() exists or if it handles generic __arrow_c_schema__ imports yet...
type = pa.data_type(exporter)
if type.geometry_type != ga.GeometryType.GEOMETRY:
  pa.array(exporter, type.with_coord_type(ga.CoordType.INTERLEAVED))
else:
  pa.array(exporter, type)
```

shapely/geoarrow.py
paleolimbot: I think it will error for anything that isn't an extension type at the moment, although this is a deeper geoarrow-c (current) limitation. It would probably be fine to accept some non-extension storage types, but this gets hairy since multipoint/linestring share a storage type. But it should eventually (probably) accept a binary (WKB) and text (WKT) array.

kylebarron:
> Good call! I was thinking that`to_arrow()` should return a `GeoArrowExporter`. Then, it's the caller's responsibility to invoke the dependency they had in mind.

That extra level of indirection seems like it would be annoying for most users. I'd be +1 on conditionally importing either pyarrow or nanoarrow and returning those directly.

shapely/geoarrow.py
```diff
@@ -0,0 +1,163 @@
...
+        builder = ArrayBuilder(requested_schema)
+        builder.append(self._geometries)
+        chunks = builder.finish(ensure_non_empty=True)
+
+        if len(chunks) == 1:
+            return requested_schema, chunks[0]
+        else:
+            raise ValueError(
+                f"Can't export geometries to single chunk ({len(chunks)} required)"
+            )
```
kylebarron, line 96: It's a little weird to me that we have both `__arrow_c_array__` and `__arrow_c_stream__` on the same class. I'd have usually expected that there would be a single underlying data representation in a class, and that would inform which PyCapsule API is defined on that class.

Maybe it makes sense to have both `ArrayBuilder` and `ChunkedArrayBuilder`? Maybe there are some occasions where a user wants to have all geometries in a single array. I.e. does `pyarrow.Table.from_pandas` always create a table with a single chunk? It looks like it does. In that case, exporting a non-chunked array of GeoArrow data would be important, so that the geometries could be appended on the table. And we'd presumably never want to raise an exception here.

(Ideally, I'd love `pyarrow.Table.from_pandas` to optionally export a chunked Table, but alas)

kylebarron:
> Something like lonboard might want to instantiate the `GeoArrowExporter` directly to avoid a situation where the user has neither `nanoarrow` nor `pyarrow` installed

Lonboard depends on pyarrow directly. It would be _lovely_ to remove this dependency (especially as that's the biggest blocker to getting lonboard to work in Wasm in jupyterlite/pyodide), but we rely heavily on `pyarrow.Table.from_pandas`, which I don't really want to reimplement myself

shapely/geoarrow.py
paleolimbot: Functionally they are the same unless the export is WKB *and* there's more than 2GB of WKB in the array. In that case, `__arrow_c_array__` will error and `__arrow_c_stream__` won't. In nanoarrow for Python, I try to almost always use `__arrow_c_stream__` (and fall back to `__arrow_c_array__` to generate a length-one stream).

I think it is fine to have both (nanoarrow's `Array` class does this too)...if the caller needs or prefers contiguous memory, they can call `__arrow_c_array__` (if they don't, they should probably call `__arrow_c_stream__`).

kylebarron: > Functionally they are the same

Even in the cases where there's a single chunk under the hood, I disagree that they're functionally the same because usually there are distinct APIs for contiguous and chunked arrays. 

So far in my experience with the PyCapsule APIs, it's been expected that you can infer the storage format of the producer based on the dunder method that exists on the object. Is this not why `pyarrow.Array` and `pyarrow.RecordBatch` have `__arrow_c_array__` implemented but not `__arrow_c_stream__`, and vice versa for `pyarrow.ChunkedArray` and `pyarrow.Table`? 

I very often use the presence of the dunder methods to decide on the most efficient way to operate on data, and I'd much prefer community consensus around whether both APIs should ever exist on a single class.

paleolimbot: It's pretty easy to separate and I don't mind doing it, although I think that you will almost always want to call the chunked version because it's not that hard to end up with >2GB of WKB. Calling the Array version is not great for the generic case because a downstream library will get errors when converting big (mixed) arrays.

jorisvandenbossche:
> but we rely heavily on `pyarrow.Table.from_pandas`, which I don't really want to reimplement myself in Rust

We should maybe reimplement that in nanoarrow ..

jorisvandenbossche:
@paleolimbot apologies for the delay in response here. And thanks for exploring this!

Some general concerns / thoughts:

- Shapely is a volunteer-only project with limited maintainer capacity (we are already too slow in reviewing and are not always able to keep up with new GEOS features), and this PR is adding a _lot_ of code. I know part of it is vendored (but I don't know if that code is already used elsewhere as well?), but we still need to understand and know that (not much documented) vendored code well enough to review and maintain the shapely-specific code. I am a bit afraid about the current scope of code addition here.
- Some parts of what is implemented here feels somewhat duplicative with existing functionality of shapely. For example, the type inference could maybe also be done with a shapely alternative based on `np.unique(shapely.get_type_id(arr))` ? While this might mostly be part of the vendored code, from shapely's point of view it might still be duplicative additions
- Further thinking about the scope, the main thing we want here is implement the GEOS-specific aspects about producing/consuming Arrow data (i.e. those parts where you need to build against GEOS). So it might also be an option to focus on that, and leave it at capsules, not bothering with pyarrow or geoarrow-pyarrow or nanoarrow or ..?
- Should we advocate for moving some core functionality of this into GEOS itself? (that might also be useful for GDAL?)
- I did have some preliminary cython code that just optimizes `to_ragged_array` (https://github.com/pygeos/pygeos/pull/93). From a quick test using your water_line test data, that's actually still a bit faster than this PR (but of course not yet that feature-full, like missing values support). Have to profile a bit in more detail to understand this better.

Oreilles:
Considering that it is possible with your work to call `from_arrow` and parse EWKB data while preserving the SRID in the newly created geometries, do you think it would be possible to also write EWKB (with SRID included) with the function `to_pyarrow` ? It currently isn't the case, meaning a round-trip with these two functions would result in a loss of information.

PR #1956, opened 2023-12-18
DOC: fix typo in buffer docs (single_side -> single_sided)

llimllib:


jorisvandenbossche (approved):
Good catch, thanks for the PR!

Issue #1954, opened 2023-12-18
Very slow MultiPoint initialization


jobh:
## Expected behavior and actual behavior.

I expect that initializing a large MultiPoint geometry is roughly as expensive as initializing a LineString from the same coordinates. Instead, it is ~2000x slower.

## Steps to reproduce the problem.

(using ipython syntax for statement timing):

```
import numpy as np
import shapely

coords = np.random.uniform(size=(1000,2))

%timeit shapely.LineString(coords)
# --> 3.39 µs

%timeit shapely.MultiPoint(coords)
# --> 6.64 ms
```

Most of that extra time is in the non-vectorized check for empty geometries, which can be bypassed:
```
%timeit shapely.multipoints(coords)
# --> 119 µs
```

This is still a factor of ~50x slower. I haven't investigated why. But vectorizing that empty-geometry check seems like low-hanging fruit for a ~50x speedup.

## Operating system

Ubuntu

## Shapely version and provenance

Shapely v2.0.1 from conda-forge. I did check the latest main, it seems like the non-vectorized error check is still present.

JamesParrott:
The non-vectorised check does indeed look like it could be improved (if only by iterating over points, not over a range https://github.com/shapely/shapely/blob/9c852a664c7b545cc3d914b6c1c2b4b215fcd4d1/shapely/geometry/multipoint.py#L53 ).  

But users can choose multipoints or MultiPoint, if the speedup is important to them.  


As to the why, MultiPoint is typed for any sequence, whereas multipoints expects an array like argument.  It (multipoints) is also wrapped in a multithreading decorator, and calls out to create_collection which is implemented in a C extension.

jobh:
Thanks for the explanation @JamesParrott!

As you say, users can choose the faster variant, as I did. I guess there is no easy vectorization fix if the sequence can be heterogeneous.

Maybe it would help to put a pointer to this option in the documentation or docstring? Feel free to close the issue if desired, I opened it only because I was so surprised by the almost 2000x performance difference compared to LineString initialization.

JamesParrott:
The docs talk about ufuncs at the start.  But even so, if multipoints is a ufunc, yes indeed the docs could be much clearer.  I would add something at the start of the Geometry / API section here:

https://shapely.readthedocs.io/en/stable/geometry.html

Otherwise, Shapely already forces users to install Numpy.  The Shapely docs should prioritise the parts of the API that make the most of Numpy.

[...]

PR #1961, opened 2024-01-03
PERF: improve MultiPoint(..) constructor from numpy array of coordinate values

jorisvandenbossche:
Addresses #1954


jobh:
Looks good!

Just a heads-up, there is a semantic change because an all-NaN point is considered an empty geometry:
```
coords = np.array([[np.nan, np.nan]])
shapely.multipoints(coords)
# <MULTIPOINT EMPTY>
shapely.MultiPoint(coords)  # before PR
# EmptyPartError: Can't create MultiPoint with empty component
```

jorisvandenbossche:
Good point! We should try to keep that consistent. The `points()` constructor now has a `handle_nan` keyword that can be set to disallow NaNs. 

Now, that would also disallow single NaNs for just x or y, which currently is allowed:

```
In [7]: shapely.MultiPoint([(0, 1), (2, np.nan)])
Out[7]: <MULTIPOINT ((0 1), (2 NaN))>
```

so if we want to retain the actual current behaviour of MultiPoint(..), we could also first create points, and then just use the vectorized method to check if there are any empties in there.

sgillies (approved):
Thanks @jorisvandenbossche !

PR #1965, opened 2024-01-15
Add geoms property to BaseGeometry

### LINKED ISSUE #1881: Add `.geoms` to all shapes
Multi part geometries have a `.geoms` sequence of all the parts.  
I suggest adding a `.geoms` to the single part geometries too.  
It will return just the shape itself but make usability much more consistent without the need to differentiate.  
Today, I actually have a wrapper to allow mw to work consistently.

It should be almost trivial to add, be efficient (no copies) and would not break existing code.  
WDYT?

cookiedan42:
Added geoms property to BaseGeometry
Added tests for Simple Geometries to test .geom method

Resolves #1881

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -106,6 +106,10 @@ class BaseGeometry(shapely.Geometry):
         return shapely.from_wkt("GEOMETRYCOLLECTION EMPTY")
 
     @property
+    def geoms(self):
+        return GeometrySequence(self)
+
+    @property
     def _ndim(self):
         return shapely.get_coordinate_dimension(self)
 
--- a/shapely/tests/geometry/test_geometry_base.py
+++ b/shapely/tests/geometry/test_geometry_base.py
@@ -262,5 +262,3 @@ def test_geoms(geom):
 
     with pytest.raises(IndexError):  # index out of range
         result[1]
-    
-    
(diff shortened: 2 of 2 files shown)


cookiedan42:
failing tests on seem to be tied to using ` if hasattr(geom, "geoms")` to identify MultiGeometries and decompose them, adding the .geoms attribute to BaseGeometry leads to recursion limit in the tests' helper function

shapely/geometry/base.py
```diff
@@ -105,6 +105,10 @@ def __new__(self):
         )
         return shapely.from_wkt("GEOMETRYCOLLECTION EMPTY")
 
+    @property
+    def geoms(self):
+        return GeometrySequence(self)
```
sgillies, line 110: @cookiedan42 here's what I suggest to solve the recursion issue. Do the following for the base class:

```suggestion
        raise NotImplementedError
```

And then for single-part `Point`, etc, classes, do this:

```python
@property
def geoms(self):
    return [self]
```

`GeometrySequence` is nothing more than a Python sequence protocol backed by a GEOS multi-part geometry, but as you notice, introduces some complication.

sgillies (changes requested):
@cookiedan42 sorry for the delay in review! I have a suggestion that should unblock this work.

sgillies:
@cookiedan42 I enabled the workflow. There are some linting issues and also this change really causes trouble for `test_set_operations.py` which assumes that recursively walking the parts of geoms is closed at some level. Until now I didn't really get the infinite recursion implications of this feature. It suggests to me that we may want to consider, e.g.,  `Point`'s,  lack of a `geoms` property as a safety feature and not a bug.

PR #1969, opened 2024-01-22
ENH: add coverage_simplify exposing topological simplification of coverages

### LINKED ISSUE #2184: Expose TPVWSimplifier
Does shapely already expose the GEOS implementation of [Topology-Preserving Visvalingam-Whyatt Line Simplification](https://libgeos.org/doxygen/classgeos_1_1coverage_1_1TPVWSimplifier.html) through its Python API? (I couldn’t find it in the docs, but I might have looked at the wrong place). If it’s not exposed yet, would a Python wrapper be in scope for shapely?

martinfleis:
xref #1570

GEOS supports this simplification for collections of polygons only. Here, I have currently opted for returning other geometries untouched but the correct behaviour is up to a discussion. When checking for correct geom types, nesting og GeometryCollections is not supported and a GeometryCollection of a GeometryCollections will not be simplified no matter the contents of the inner collection. 

If we did not do this check, when points or line strings are passed, shapely returns (empty) GEOS Exception. When passing GeometryCollection with points or line strings, it kills the kernel, so I suppose that the checks I have included now are quite a feasible way forward.

I am mirroring the API of PostGIS, not that of GEOS. So the keyword is `simplify_boundary` rather than `preserve_boundary` which seems more intuitive.

resolves #2184


martinfleis:
I'll fish those cases out and report to GEOS, good point. 

I actually wrote more tests but apparently did not push the commit. Will do once I'll get to laptop.

theroggy:
I suppose that the input coverage needs to be entirely topologically sound? Eg. that if two polygons touch not in a common point but a point touches a vector of another polygon between two points, this will not be treated as a common boundary?

If this is the case, I think it will be useful to explicitly mention this in the doc to avoid misunderstandings...

jorisvandenbossche:
We also need to decide how to treat the input geometry argument. AFAIK in the current state of the PR, you just pass each input geometry to the GEOS function, and so this assumes that the user already collected their set of polygons into a collection. 

Alternatively, we can also do that for the user. Similarly as [`ST_CoverageSimplify`](https://postgis.net/docs/ST_CoverageSimplify.html) is a window function (thus taking a set of geometries to simplify together), and similarly as for `union_all` we also collect the input geometries in a collection first before passing to GEOSUnaryUnion.

martinfleis:
I have changed the behaviour to follow the logic of `union_all` or [`ST_CoverageSimplify`](https://postgis.net/docs/ST_CoverageSimplify.html) to make the function operate on the whole array and return a collection. The axis keyword is not yet tested but please have a look if this makes more sense API-wise.

shapely/constructive.py
```diff
@@ -993,6 +994,63 @@ def simplify(geometry, tolerance, preserve_topology=True, **kwargs):
...
+
+    The function allows simplification of all edges including the outer boundaries of
+    the coverage or simplification of only the inner (shared) edges.
+
+    If there are other geometry types than Polygons or MultiPolygons present, the
+    resulting GeometryCollection will not undergo simplification and geometries are
+    returned collected but unchanged.
+
+    If the geometry is polygonal but does not form a valid coverage due to overlaps,
+    it will be simplified but it may result in invalid topology.
```
sgillies, line 1020: @martinfleis Lovely docstring, thank you!

shapely/constructive.py
```diff
@@ -993,6 +994,63 @@ def simplify(geometry, tolerance, preserve_topology=True, **kwargs):
...
+        The degree of simplification roughly equal to the square root of the area
+        of triangles that will be removed.
+    simplify_boundary : bool, optional
+        By default (True), simplifies both internal edges of the coverage as well
+        as its boundary. If set to False, only simplifies internal edges.
+    axis : int, optional
+        Axis along which the operation is performed. The default (None)
+        performs the operation over all axes, returning a scalar value.
+        Axis may be negative, in which case it counts from the last to the
+        first axis.
```
sgillies, line 1035: `axis` is new to me. It comes from Pandas, right? I think boundaries between Pandas/GeoPandas and Shapely are a good thing and that we should resist blurring the projects if we can. What does `axis` really mean to Shapely, given that inputs can only be a scalar geometry or a 1-D array of geometries? Can't we infer axis from the type of input?

martinfleis: This comes directly from `shapely.union_all` and other predicate _all functions which have the same notion of using axis to define what should be treated as a single group. So it is not new.

> What does axis really mean to Shapely, given that inputs can only be a scalar geometry or a 1-D array of geometries?

That is not correct as far as I'm aware. While some operations will not work on a N-D array (STRtree notably), most of shapely operations are perfectly capable of handling N-D arrays of geometries and there's a clear use case for that in the vector data cube space for dealing with time-dependent geometries.

If you think this should change, we should do that across all functions that use it, so I'd keep it for a separate discussion and PR.

sgillies: Yeah, since the low-level operations were written like ufuncs, I shouldn't be surprised that they would work on N-D arrays.

Even, so why would we want to ravel in one case, and rollaxis in the other? What does this mean to a GIS programmer? If `lib.coverage_simplify()` can accept N-D arrays, can the caller not ravel or rollaxis or neither (perhaps?) as they need before calling the method?

In matters of software design, I tend to be anti-convenience. Instead of accepting an `axis` argument and then modifying input based on the value, callers should just modify their inputs in their own applications. Then we have less code, fewer special cases and branches to test. I would make an exception for situations where GEOS C API methods must be called, because how is a user going to do that? But here the user already has numpy installed and can easily reshape their inputs.

sgillies (changes requested):
I need some clarification on what `axis` really stands for in this method and I'd like a discussion about whether this is a real feature or just a convenience.

martinfleis:
I mean, I am not attached to the `axis` keyword. I even said above

> The axis keyword is not yet tested but please have a look if this makes more sense API-wise.

The reason I included it in this way is that we need to collect individual geometries in some way before passing them to GEOS, otherwise we expect user to do that. The latter was originally implemented here but as pointed out by @jorisvandenbossche above, it may not be what user expects from this function.

When collecting geometries, shapely has a precedent of using the axis keyword. I simply followed that for the sake of consistency. Coverage simplification then works as aggregation, since it returns a GeometryCollection, following practically the same pattern as `union_all` and other functions. 

I fully understand your concern but I am not sure if this function is different enough to deviate from this pattern. But in the end, that is not my call to make.

I can remove the axis and follow the route currently implemented for `axis=None`, I don't mind that, if you think that it does not belong here. But then you should maybe have a discussion if it belongs elsewhere as well.

jorisvandenbossche:
> `axis` is new to me. It comes from Pandas, right? I think boundaries between Pandas/GeoPandas and Shapely are a good thing and that we should resist blurring the projects if we can. 

To be clear, the `axis` keyword does not come from pandas (I don't think we actually even use it in geopandas in any of the shapely methods we call), but it comes from numpy. 
See https://numpy.org/doc/stable/glossary.html#term-axis and https://numpy.org/doc/stable/glossary.html#term-along-an-axis for operations that can be performed _along_ an axis. Such operations have an `axis` keyword, and so for example all reduction functions or generalized ufuncs in numpy have that keyword.

Given that the core GEOS functions we now wrap are implemented as numpy ufuncs, I think providing the `axis` keyword in reducing ufuncs is very natural (for people familiar with numpy, and we can of course do a better job explaining this keyword in our docs for people less familiar).

> Even, so why would we want to ravel in one case, and rollaxis in the other? What does this mean to a GIS programmer? If `lib.coverage_simplify()` can accept N-D arrays, can the caller not ravel or rollaxis or neither (perhaps?) as they need before calling the method?

The `ravel` or `rollaxis` call is an implementation detail that currently ensures the function behaves as expected for a numpy reduction: the default of `axis=None` means reducing over all values (producing a single value), specifying an axis means reducing along that axis (producing an array with the specified axis removed from the result' shape).

Can you the user do this themselves? Potentially yes, depending on what signature we would provide instead. 
But would it be expected that you need to do this yourself? I would say no, for someone who expects an interface consistent with numpy (and I would personally also not call this just "convenience", using `rollaxis` is quite advanced, and my guess is the many numpy users will know the `axis` keyword, but will never have used `rollaxis` directly)

> What does this mean to a GIS programmer?

To be fair, I don't think this specific capability is _that_ useful for typical GIS applications. I think it is quite normal that a user would want to simplify different groups of geometries together, but it is probably not that likely that those different groups are then exactly the same size such that you actually can put them in a multidimensional numpy array and make use of reducing over an axis. 
(hypothetical example: you have 5 groups of 10 polygons where you want to simplify per group; in that case you could put those geometries in a (5, 10) array, and call `coverage_simplify` with `axis=1` to get the 5 simplified geometry collections (one per group). But so if those groups are not exactly the same size, you cannot make use of this feature and you will have to manually loop through the groups anyway)

But while this specific case doesn't seem that useful, there are several other aspects about the numpy ufunc capabilities that _are_ very useful, such as the broadcasting behaviour (allowing you to do pairwise operations). And when we essentially provide standard numpy functions with those features, I would personally go for being consistent in following numpy's design entirely, and not only provide that subset of features that we think will be useful.

---

That said, we still need to make a choice about what _kind_ of function we want to provide, and there are different options which all have some pros and cons (and not all of those options have an `axis` keyword when following numpy):

- A plain simple element-wise ufunc (i.e. shape in = shape out): this is what Martin did in the first version of this PR. This means that it is up to the user to collect the polygons in geometry collections, and there would be no `axis` keyword.
- A reduction-like function: this is the current state of the PR, where we collect the polygons in a collection under the hood. This returns a single scalar result by default (but can be changed with the `axis` keyword to reduce only over a specific axis)
- A generalized ufunc with 1 dimension that gets reduced (this could actually be fully implemented in ufuncs.c in that case). This would be similar to `polygonize` that has a ufunc signature of `(d)->()`, which means that it reduces over the last dimension.
- A final option could also be a function that does both the collection and explosion for the user, i.e a function that has "shape in = shape out" but is not elementwise (in the meaning of processing every element independently). It would simplify _all_ geometries in the input array together, but preserve the shape of the input array (and essentially hide the fact that the core GEOS function is implemented in terms of GeometryCollections and not Polygons)

The first option is the closest to the pure GEOS function, the last option is closest to how PostGIS exposes this as `[ST_CoverageSimplify](https://postgis.net/docs/ST_CoverageSimplify.html)` (and I think the most user friendly or "convenient"). 

Note that there is prior art in "hiding" the fact that certain GEOS functions work on a GeometryCollection (instead of on an array of geometries, but that might be because the collection is easier in the C API). The `unary_union` function that shapely has exposed for a long time, has always collected the input geometries in a collection first, instead of requiring the user to do that (and same for eg `polygonize_full`)

martinfleis:
I have just tested that doing the option four, meaning collect-simplify-explode, is possible as the order of geometries in the collection seems to be preserved throughout the simplification procedure. That is how I would like this to behave in GeoPandas but within shapely, it feels inconsistent as other reductions like polygonize return a collection of polygons, not an array of individual geometries. But truth is, that a user does not need to know the implementation details in GEOS and we can use the API that more mirrors normal `simplify` than polygonize or union, even though that is how it practically works.

So I'm happy to update the PR to follow the first option. However, with a N-D array (e.g. time x geom used in vector data cubes), we'd need to slice and loop over 1D arrays, compared to the current state of the PR.

jorisvandenbossche:
> within shapely, it feels inconsistent as other reductions like polygonize return a collection of polygons, not an array of individual geometries. 

Yes, but we could also argue that `coverage_simplify` is not a reduction. Just like `simplify` is also not a reduction. Taking option 4 would make `simplify` and `coverage_simplify` more similar in usage, with the only difference that `simplify` works purely element-by-element where `coverage_simplify` considers the full input as context to simplify each element.

(for `polygonize` there is also no 1-to-1 mapping between input and output, so that's a bit of a different situation and would neither fit in a reduction or window function if it returned the exploded array, so I think here making it a reduction makes more sense)

martinfleis:
Alright. I'll try to update the PR over the weekend.

jorisvandenbossche:
(maybe wait a bit to give Sean time to respond as well)

jorisvandenbossche:
@sgillies do you have time to take another look at the discussion above?

sgillies:
@jorisvandenbossche @martinfleis thank you for the discussion. I've finally got my head around the current implementation. It makes sense to me. I'm also attracted to the 4th option.

Do either of you know if GEOS coverage simplification *always* conserves polygons? If pieces of a coverage are ever simplified away to nothing, then we could not implement the fourth option, right?

martinfleis:
The VW algorithm reduced a polygon to a triangle at most, so there's always a polygon as an output. I will verify that again but so far I believe that we can implement the fourth option.

jorisvandenbossche:
That's indeed what the doc comment in the C API guarantees (https://github.com/libgeos/geos/blob/2b5721bc96d64c3d768bd0e91dba6c3bbca354a9/capi/geos_c.h.in#L3960):

> Geometries never disappear, but they may be simplified down to just
> a triangle. Also, some invalid geoms (such as Polygons which have too
> few non-repeated points) will be returned unchanged.

sgillies:
Let's do the fourth option, then, and agree that this isn't a reduction operation?

BTW, I'm happy with the 5 instances in set_operations.py.

martinfleis:
I have changed the behaviour to follow the fourth model but I don't think it is quite right implementation-wise. Right now, it works in principle but does not work like a ufunc, so casting it on a GeoSeries does not preserve GeoSeries output etc... And I think that this logic should happen in C but I think I've hit the limits of my C and wasn't able to make it work. 

So marking this as a draft but I'll probably need to pass this on to someone more capable in C. It got more complicated than anticipated.

jorisvandenbossche:
@martinfleis I finally took the time to properly re-read our discussion earlier to understand again what we decided, and your last comment and updated implementation. And I think your implementation is actually just fine, and it's only a matter of updating the tests to follow this new behaviour (4th option).

> Right now, it works in principle but does not work like a ufunc, so casting it on a GeoSeries does not preserve GeoSeries output etc... And I think that this logic should happen in C but I think I've hit the limits of my C and wasn't able to make it work.


I think this is a bit the consequence of choosing the 4th option, because that option does not fit in the numpy ufunc model (it is not a simple element-wise ufunc, but also not a generalized ufunc with eg a reducing or reused axis). It is a function that preserves the shape, but looks at all the elements together, and so not a ufunc. 

I suppose we could still make it work that it behaves like a ufunc in the sense of preserving the original object (I think that would then rather use the `__array_function__` protocol instead of `__array_ufunc__`, see https://numpy.org/doc/stable/user/basics.interoperability.html#operating-on-foreign-objects-without-converting) 
But this seems like less critical, and something I could take a look at.

martinfleis:
I am getting a strange bare `GEOSException` when passing a scalar Polygon. Everything else is fine. I thought that it may be some GEOS issue but using command line `geosop` is fine. No clue where that comes from.

jorisvandenbossche (approved):
Thanks for the update! Looking good!

shapely/tests/test_constructive.py
```diff
@@ -1097,6 +1097,87 @@ def test_concave_hull_kwargs():
     assert shapely.get_num_coordinates(result4) < shapely.get_num_coordinates(result3)
 
 
+@pytest.mark.skipif(shapely.geos_version < (3, 12, 0), reason="GEOS < 3.12")
+@pytest.mark.parametrize("geometry", all_types)
+def test_coverage_simplify_scalars(geometry):
+    actual = shapely.coverage_simplify(geometry, 0.0)
+    assert isinstance(actual, Geometry)
+    # in some cases, a collection may be returned
+    assert shapely.get_type_id(actual) in (shapely.get_type_id(geometry), 7)
```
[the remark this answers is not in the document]
martinfleis: No, not anymore. That was relevant for the earlier implementation.

martinfleis:
I don't think the CI failure is related.

jorisvandenbossche:
Indeed unrelated. Comparing with the previous succeeding build, I see that `coverage` upgraded from 7.6.10 to 7.6.11 (which would be strange to be related), and for the rest we install GEOS and numpy dev versions. Will see if I can reproduce with latest numpy

jorisvandenbossche:
With latest numpy I can't reproduce it, but will try later with latest GEOS (there was a buffer related change a few days ago https://github.com/libgeos/geos/pull/1233)

jorisvandenbossche:
Opened https://github.com/libgeos/geos/issues/1236

jorisvandenbossche:
Thanks @martinfleis!

Issue #1972, opened 2024-01-24
Compatibility with NumPy 2.0


jorisvandenbossche:
NumPy 2.0 is coming soon, and we should ensure shapely is fully compatible to ensure a smooth transition of our users. Overview issue on numpy's side: 

* https://github.com/numpy/numpy/issues/24300

Quick summary of the most important aspects for us:

- NumPy 2.0 as a major release changes the C ABI, so any package that builds against the NumPy C API like shapely will have to be rebuilt with numpy 2.0 to be able to run with numpy 2.0. That also means that our current released packages (built with older versions of numpy) will break (fail to import) with numpy 2.0, and so we should ideally add a `numpy<2` pin.  
  This happens at the moment when installing shapely from a wheel in an environment with numpy nightly.  
* Since numpy 1.25, we don't need `oldest-supported-numpy` anymore (https://numpy.org/devdocs/release/1.25.0-notes.html#compiling-against-the-numpy-c-api-is-now-backwards-compatible-by-default and https://github.com/scipy/oldest-supported-numpy#deprecation-notice-for-numpy-20), so for newer Python versions (where the minimum supported version of numpy is <=1.25), we can start compiling with the latest numpy.  
  This also helps with the point above, because once numpy 2.0 is released, we can therefore _built_ against numpy 2.0 and have a resulting wheel that can _run_ with both numpy 1.x and 2.0. See https://numpy.org/devdocs/dev/depending_on_numpy.html#numpy-2-0-specific-advice for more details.
* Additionally to C ABI changes, there are also a whole set of changes in the Python API, and so we should test with numpy nightly to ensure we are compatible on that front.

We don't yet test with numpy nightly 

A to do list based on my current understanding:

- [x] Start testing with numpy nightly (we currently only use `--pre` in the dev build, so we only get a dev numpy when they do a beta or RC release). This will ensure that shapely at least builds and runs fine with the upcoming numpy. -> https://github.com/shapely/shapely/pull/1973
- [x] We should ideally do another release of shapely that adds a `numpy<2` pin, so that when numpy 2.0 is out and someone does `pip install shapely`, you don't get a broken env. Therefore, for the maintenance 2.0.x branch:
  - [x] Add a `numpy<2` pin to the requirements in the maintenance branch -> https://github.com/shapely/shapely/pull/1975
  - [x] Make another 2.0.x release  
- [x] For the main development branch, we don't need a runtime pin but we need to update the build dependencies for packaging:
[...]

PR #1973, opened 2024-01-26
TST: test against numpy nightly

jorisvandenbossche:
First part of https://github.com/shapely/shapely/issues/1972, to start with testing running numpy 2.0.dev

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -123,7 +123,7 @@ jobs:
           pip install --upgrade wheel setuptools
           if [ -z "${{ matrix.numpy }}" ]; then
             pip install --upgrade --pre Cython pytest pytest-cov coveralls;
-            pip install --upgrade --pre --only-binary :all: -i https://pypi.anaconda.org/scipy-wheels-nightly/simple numpy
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -122,7 +122,8 @@ jobs:
           python -m pip install --disable-pip-version-check --upgrade pip
           pip install --upgrade wheel setuptools
           if [ -z "${{ matrix.numpy }}" ]; then
-            pip install --upgrade --pre Cython numpy pytest pytest-cov coveralls;
+            pip install --upgrade --pre Cython pytest pytest-cov coveralls;
+            pip install --upgrade --pre --only-binary :all: -i https://pypi.anaconda.org/scipy-wheels-nightly/simple numpy
```
martinfleis, line 126:
```suggestion
            pip install --upgrade --pre --only-binary :all: -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple numpy
```

Scipy channel is no longer in use, the latest numpy there is from 2023-06-16.

jorisvandenbossche: I copied it from their docs ... but the page I landed on from googling is the one from 1.24 ;)

jorisvandenbossche:
Seems to be all good!

caspervdw:
Big relief that numpy 2.* “just works” for shapely!

PR #1974, opened 2024-01-26
BLD: update build-system requires to use latest numpy instead of oldest-supported-numpy + build nightly wheels with numpy nightly

jorisvandenbossche:
Part of https://github.com/shapely/shapely/issues/1972

And see https://numpy.org/devdocs/dev/depending_on_numpy.html#numpy-2-0-specific-advice for a direct link to the NumPy guidelines on updating build dependencies for NumPy 2.0.


jorisvandenbossche:
I verified locally that the wheels built here installed in an env with numpy nightly works (while installing the latest released wheel fails to import)

PR #1975, opened 2024-01-26
COMPAT: Add a numpy<2 pin to install requirements for 2.0.x

jorisvandenbossche:
See https://github.com/shapely/shapely/issues/1972 for the details.

Current Shapely wheels for 2.0.x (and for an upcoming 2.0.3 release) are not compatible with future numpy 2.0 packages (if you install shapely now in an environment with nightly numpy, it fails to import), therefore it is recommended to add this upper pin now for releases. 
We will keep the more flexible pin on the development branch, assuming by the time we release shapely 2.1, numpy 2.0 will be out and we can build shapely 2.0 to be compatible with all numpy versions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ classifiers = [
 ]
 requires-python = ">=3.7"
 dependencies = [
-    "numpy>=1.14,<2'",
+    "numpy>=1.14,<2",
 ]
 
 [project.optional-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ classifiers = [
 ]
 requires-python = ">=3.7"
 dependencies = [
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -38,7 +38,7 @@ classifiers = [
 ]
 requires-python = ">=3.7"
 dependencies = [
-    "numpy>=1.14",
+    "numpy>=1.14,<2'",
```
mwtoews, line 41:
```suggestion
    "numpy>=1.14,<2",
```

mwtoews:
It seems there is a bit of work needed for `maint-2.0` to work with GEOS main, I'll investigate ....

jorisvandenbossche:
Did some additional fixes in https://github.com/shapely/shapely/pull/1977, will update this PR now

Issue #1963, opened 2024-01-04
API: need for a element-wise binary coverage_union method?


jorisvandenbossche:
We have a binary `coverage_union` function (with an API that matches the binary `union` function), and a `coverage_union_all` reduction function (matching `union_all` / `unary_union`).

As brought up in https://github.com/geopandas/geopandas/issues/3117, the binary version doesn't seem to be working correctly, though. When this was added in https://github.com/pygeos/pygeos/pull/142, it seems we only tested this with scalar geometry inputs, and not with arrays. 

Consider the following two arrays of polygons:

```python
>>> import shapely
>>> a = shapely.box([0, 1], [0, 1], [1, 2], [1, 2])
>>> b = shapely.box([1, 2], [0, 1], [2, 3], [1, 2])
>>> a
array([<POLYGON ((1 0, 1 1, 0 1, 0 0, 1 0))>,
       <POLYGON ((2 1, 2 2, 1 2, 1 1, 2 1))>], dtype=object)
>>> b
array([<POLYGON ((2 0, 2 1, 1 1, 1 0, 2 0))>,
       <POLYGON ((3 1, 3 2, 2 2, 2 1, 3 1))>], dtype=object)
```

Passing the first geometry of each works fine:

```python
>>> shapely.coverage_union(a[0], b[0])
<POLYGON ((0 0, 0 1, 1 1, 2 1, 2 0, 1 0, 0 0))>
```

But passing arrays still reduces all inputs to a single geometry, instead of working element-wise:

```python
>>> shapely.coverage_union(a, b)
<POLYGON ((0 0, 0 1, 1 1, 1 2, 2 2, 3 2, 3 1, 2 1, 2 0, 1 0, 0 0))>
```

The `coverage_union` function is implemented in terms of the reducing one (which is what is implemented in C):

```python
def coverage_union(a, b, **kwargs):
    return coverage_union_all([a, b], **kwargs)
```

However, `coverage_union_all` by default reduces over all axes, while here we want to preserve the first axis. So the above should be:

```python
>>> shapely.coverage_union_all([a, b], axis=0)
array([<POLYGON ((0 0, 0 1, 1 1, 2 1, 2 0, 1 0, 0 0))>,
       <POLYGON ((1 1, 1 2, 2 2, 3 2, 3 1, 2 1, 1 1))>], dtype=object)
```

This now worked correctly in a binary, element-wise manner.

---

[...]

PR #1982, opened 2024-02-06
ENH: add disjoint_subset_union and disjoint_subset_union_all

martinfleis:
Exposing `GEOSDisjointSubsetUnion` as `disjoint_subset_union` and `disjoint_subset_union_all`, matching the API of union and coverage_union (assuming coverage_unoin will be fixed as discussed in #1963).

Tests are currently only reusing the tooling for other set operations. Not sure if there's a need to go beyond that.

xref #1510


shapely/set_operations.py
```diff
@@ -501,3 +503,105 @@ def coverage_union_all(geometries, axis=None, **kwargs):
...
+    <POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))>
+    """
+    if isinstance(a, Geometry | None) and isinstance(b, Geometry | None):
+        pass
+    elif isinstance(a, Geometry | None):
+        a = np.full_like(b, a)
+    elif isinstance(b, Geometry | None):
+        b = np.full_like(a, b)
+    elif len(a) != len(b):
+        raise ValueError("Arrays a and b must have the same length")
```
martinfleis, line 551: Since `disjoint_subset_union` operates on a GeometryCollection, we need to roll-out custom broadcasting here.

shapely/set_operations.py
```diff
@@ -501,3 +503,110 @@ def coverage_union_all(geometries, axis=None, **kwargs):
     # create_collection acts on the inner axis
     collections = lib.create_collection(geometries, GeometryType.GEOMETRYCOLLECTION)
     return lib.coverage_union(collections, **kwargs)
+
+
+@multithreading_enabled
```
jorisvandenbossche, line 508:
```suggestion
@requires_geos("3.12.0")
@multithreading_enabled
```

(and same for the `_all` version below)

shapely/set_operations.py
```diff
@@ -501,3 +503,110 @@ def coverage_union_all(geometries, axis=None, **kwargs):
...
+    elif isinstance(b, Geometry) or b is None:
+        b = np.full_like(a, b)
+    elif len(a) != len(b):
+        raise ValueError("Arrays a and b must have the same length")
+    return disjoint_subset_union_all([a, b], axis=0, **kwargs)
+
+
+@multithreading_enabled
+def disjoint_subset_union_all(geometries, axis=None, **kwargs):
+    """Returns the union of multiple polygons of a geometry collection.
```
jorisvandenbossche, line 560:
```suggestion
    """Returns the union of multiple polygons.
```

The "of a geometry collection" might be a bit confusing, given we create the collection under the hood?

shapely/tests/test_set_operations.py
```diff
@@ -66,6 +73,8 @@ def test_set_operation_array(a, func):
 @pytest.mark.parametrize("func", SET_OPERATIONS)
 @pytest.mark.parametrize("grid_size", [0, 1])
 def test_set_operations_prec_not_supported(func, grid_size):
+    if "grid_size" not in inspect.signature(func).parameters:
```
jorisvandenbossche, line 76: I would maybe hardcode `if func is shapely.disjoint_subset_union`, to avoid that this might accidentally skip tests if something went wrong in this check

shapely/set_operations.py
```diff
@@ -501,3 +503,110 @@ def coverage_union_all(geometries, axis=None, **kwargs):
...
+    if (isinstance(a, Geometry) or a is None) and (
+        isinstance(b, Geometry) or b is None
+    ):
+        pass
+    elif isinstance(a, Geometry) or a is None:
+        a = np.full_like(b, a)
+    elif isinstance(b, Geometry) or b is None:
+        b = np.full_like(a, b)
+    elif len(a) != len(b):
+        raise ValueError("Arrays a and b must have the same length")
```
jorisvandenbossche, line 554: The logic looks good here (to mimic broadcasting), just wondering if this is already covered by the standard tests we have? (did you have to add this to get the tests passing?)

jorisvandenbossche: Ah, and it might be that this can be simplified as `a, b = np.broadcast_arrays(a, b)`

jorisvandenbossche (commented):
Thanks a lot, looks good! A few minor comments

shapely/set_operations.py
martinfleis: I tried `broadcast_array` before writing this and it does not work as we need it.  Some of the tests fail with `TypeError: One of the arguments is of incorrect type. Please provide only Geometry objects.`

And yes, I had to include this to make the tests green, so it is properly covering by existing tests.

jorisvandenbossche (approved):
Thanks for the updates!

shapely/set_operations.py
```diff
@@ -506,12 +506,10 @@ def disjoint_subset_union(a, b, **kwargs):
     :func:`union`. As such, it is recommeded to use ``disjoint_subset_union`` with
     GeometryCollections rather than individual geometries.
 
-    Requires GEOS >= 3.12.0
```
jorisvandenbossche: Forgot that this gets actually added by the decorator, so removing it here to not have it duplicated.

jorisvandenbossche:
The windows failure is unrelated and is being fixed by https://github.com/shapely/shapely/pull/2127, I think

jorisvandenbossche:
Thanks @martinfleis!

PR #1985, opened 2024-02-12
DOC: fix version shown on readthedocs (take 2)

jorisvandenbossche:
https://github.com/shapely/shapely/pull/1914 is no longer working to get the tags (I suppose because the checkout depth is by default only 50)


jorisvandenbossche:
That seems to be working again!

PR #1986, opened 2024-02-12
MAINT: backports for 2.0.3

jorisvandenbossche:


mwtoews (approved):
Thanks! It will be nice to see the correct version in the docs.

shapely/constructive.py
```diff
@@ -1028,7 +1028,7 @@ def oriented_envelope(geometry, **kwargs):
     <POLYGON EMPTY>
     """
     if lib.geos_version < (3, 12, 0):
-        f = _oriented_envelope_min_area
+        f = _oriented_envelope_min_area_vectorized
```
Lash-L, line 1031: Sorry to revive an old PR for this - but ever since you pointed out ufuncs to me in a previous PR I have been using them a ton, I noticed I got a minimal difference in running the min_rotate_rectangle function with ufuncs, so I started diving in, found this, and found the corresponding GEOS PR.

I don't fully have a grasp, but I had a few things - hopefully these are constructive to the project and not just me wasting time for you

1) Shouldn't this be geos_version < 3,10, 0 as that's what it was backported to? Or since there are theoretically versions of 3,10,0 out there without the proper functionality, you require 3,12,0?
2) If the expectation is that anything below 3,10,0/ 3,12,0 does not properly work and should instead use the in shapely approach - Should _oriented_envelope_geos have a requires_geos("3.10.0") instead of 3.6.0? Or is it rather the fact that the issue was only introduced in 3.10.0 and this should instead be if geos_version >= 3.10.0 and <= 3.12.0?
3) There is a rather large timing difference with the three:

```
calling min rotated rectangle for each element in the list:
Point: 0.8094756999926176 seconds
Lines: 14.574645699991379 seconds
Polygons: 16.986191800009692 seconds
calling min rotated rectangle with a list:
Point: 1.3272610999993049 seconds
Lines: 17.333137600013288 seconds
Polygons: 15.052347099990584 seconds
calling _Oriented envelope:
Point: 0.12419709999812767 seconds
Lines: 0.3855291999934707 seconds
Polygons: 0.2791209999995772 seconds
```

How large of a problem are taking results of _oriented_envelope_geos() directly vs using the _oriented_envelope_min_area_vectorized()?

jorisvandenbossche: Apologies for the slow reply

> 1. Shouldn't this be geos_version < 3,10, 0 as that's what it was backported to? Or since there are theoretically versions of 3,10,0 out there without the proper functionality, you require 3,12,0?

Note that this backport PR is for Shapely, not for GEOS. And shapely 2.0.x can be installed and work with any version of GEOS >= 3.5

> 2\. Should _oriented_envelope_geos have a requires_geos("3.10.0") instead of 3.6.0?

Ah, that `@requires_geos("3.6.0")` is probably a left-over, we can remove that entirely, given this will never be called with an older GEOS version

> 3\. There is a rather large timing difference with the three

Yes, we know that it is a lot faster when using GEOS. The problem is that GEOS was giving wrong results for GEOS < 3.12, and therefore when you use shapely with an older version of GEOS, we will fall back to a slower python implementation included in shapely. 

See https://github.com/shapely/shapely/issues/1670 for context.

Issue #1742, opened 2023-01-30
2.0.x releases


jorisvandenbossche:
I released 2.0.1 today with a first batch of regression/bug fixes (https://github.com/shapely/shapely/releases/tag/2.0.1). 

There are still a few regressions/bugs to be tackled for 2.0.x (see current list in the 2.0.2 milestone: https://github.com/shapely/shapely/milestone/55), so I would propose to do a 2.0.2 shortly once a few of those are fixed.

idanmiara:
Are we considering small enhancements in minor version updates? 
If so, suggesting to consider marking the following for `2.0.1` (or otherwise - for `2.1`): 
https://github.com/shapely/shapely/pull/1231
https://github.com/shapely/shapely/pull/1681

jorisvandenbossche:
I am planning to do another 2.0.x release (2.0.3), mainly to ensure we have wheels that are "compatible" with the upcoming numpy 2.0 release (i.e. pin to <2 to avoid build compatibility issues), see https://github.com/shapely/shapely/issues/1972 for all details.

In addition, this also allows us to release wheels with the latest GEOS bug-fix release of the 3.11.x cycle (3.11.3; I think the idea is that GEOS 3.12 would be kept for shapely 2.1)

cc @shapely/shapely-team

sgillies:
Great. Thanks @jorisvandenbossche !

mwtoews:
Awesome, thanks! The one main thing that comes to mind is that GitHub recently introduced [native macOS Sillicon runners for open source projects](https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/). Some refactoring will need to enable it, and should resolve this warning:
> [Build arm64 wheels on macos-11](https://github.com/shapely/shapely/actions/runs/7874085014)
While arm64 wheels can be built on x86_64, they cannot be tested. The ability to test the arm64 wheels will be added in a future release of cibuildwheel, once Apple Silicon CI runners are widely available. To silence this warning, set `CIBW_TEST_SKIP: "*-macosx_arm64"`.

I'll see if I can tackle this one...

jorisvandenbossche:
Yes, it would be nice to make use of that new feature. However, specifically for 2.0.x, I wouldn't consider that as critical, given that we already have wheels for MacOS arm64

jorisvandenbossche:
[...]

PR #1990, opened 2024-02-12
RLS/BLD: use native Apple Silicon macOS for arm64 wheels; remove universal2

mwtoews:
Use native Apple Silicon macOS runners ([described here](https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/)) to build arm64 wheels.

Also, remove `universal2` wheels, since I think it's fine to have separate `x86_64` builds for macOS.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -72,12 +72,9 @@ jobs:
         - os: macos-11
           arch: x86_64
           cmake_osx_architectures: x86_64
-        - os: macos-11
+        - os: macos-14
           arch: arm64
           cmake_osx_architectures: arm64
-        - os: macos-11
-          arch: universal2
-          cmake_osx_architectures: "x86_64;arm64"
 
     steps:
       - name: Checkout source
@@ -119,6 +116,7 @@ jobs:
         env:
           CIBW_ARCHS: ${{ matrix.arch }}
           CIBW_SKIP: cp36-* pp* *musllinux* *-manylinux_i686
+          CIBW_TEST_SKIP: "cp38-macosx_arm64"
           CIBW_ENVIRONMENT_LINUX:
             GEOS_VERSION=${{ env.GEOS_VERSION }}
             GEOS_INSTALL=/host${{ runner.temp }}/geos-${{ env.GEOS_VERSION }}


mwtoews:
Up to @jorisvandenbossche to decide about these changes for the upcoming maintenance release version 2.0.3 described in #1742

I'm working on a separate PR for `main`.

jorisvandenbossche:
It's still showing the same warning / not actually running tests:

> Warning: While cibuildwheel can build CPython 3.8 universal2/arm64 wheels, we cannot test the arm64 part of them, even when running on an Apple Silicon machine. This is because we use the x86_64 installer of CPython 3.8. See the discussion in https://github.com/pypa/cibuildwheel/pull/1169 for the details. To silence this warning, set `CIBW_TEST_SKIP: "cp38-macosx_*:arm64"`.

jorisvandenbossche:
(I think it is also fine to directly target this PR to main, can still backport the eventual commit to 2.0.x if we want)

mwtoews:
switched to `main` and attempted to silence the issue with cibuildwheel and CPython 3.8 universal2/arm64 wheels

mwtoews:
Warnings all cleared up now, ready for final review.

I should elaborate more on why to remove `universal2`. They are nearly twice the size, since they contain both `x86_64` and `arm64`. E.g. with the shapely 2.0.2 [files](https://pypi.org/project/shapely/2.0.2/#files):
```
$ file universal2/shapely/_geos.cpython-39-darwin.so 
universal2/shapely/_geos.cpython-39-darwin.so: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit x86_64 bundle, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL>] [arm64:Mach-O 64-bit arm64 bundle, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL>]
```
which is cool, but there are individual wheels for each architecture, so pip should be able to fetch the correct built wheel for the host platform. But also be warned that I'm not a macOS user, so I don't have any direct experience with this stuff.

martinfleis:
Do you want me to test something here? I have access to both M1 and Intel Mac.

mwtoews:
@martinfleis My main curiosity is with `pip install --dry-run --force shapely==2.0.2`, which binary wheel is selected? `universal2` or `arm64`?

martinfleis:
That would pull `arm64`.

jorisvandenbossche:
> I should elaborate more on why to remove `universal2`. They are nearly twice the size, since they contain both `x86_64` and `arm64`

Yes, certainly on board with that. In most other projects where I am involved in, we also dropped the universal wheels the last years (eg pandas, and also numpy did the same I see)

jorisvandenbossche (approved):
Thanks, looks good!

PR #1997, opened 2024-02-26
Upload nightly wheels to Scientific Python nightly channel

jorisvandenbossche:
See https://github.com/scientific-python/upload-nightly-action/issues/63


jorisvandenbossche:
We are only going to see if this worked or not on main, so going to merge this to verify (happy to do any follow-ups!)

PR #1998, opened 2024-02-26
Fix uploading of nightly wheels

jorisvandenbossche:
Follow-up on https://github.com/shapely/shapely/pull/1997, as apparently the action only works on ubuntu, so we need to run it once after all wheels builds succeeded


jorisvandenbossche:
This is working now, and the first set of wheels is uploaded! 
To make this more usable, we will have to add a 2.1.0.dev0 tag, I think, because right now, the version is 2.0.1+78 (after 2.0.1, we branched for maint-2.0)

PR #1999, opened 2024-02-26
DOC: document how to install nightly wheels

jorisvandenbossche:
Follow-up on https://github.com/shapely/shapely/pull/1997 / https://github.com/shapely/shapely/pull/1998.

I still need to update the tags to make this actually work. Currently, those wheels have versions like 2.0.1+78 (after 2.0.1, we branched for maint-2.0).
To make this more usable (so the wheels have a version number that is considered higher than the released version 2.0.3, and gets preferred by pip when specifying the index), I think we will have to add a `2.1.0.dev0` tag to the main branch (e.g. on the commit just after we branched maint-2.0)

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -25,6 +25,15 @@ Shapely is available on the conda-forge channel. Install as follows::
 
     $ conda install shapely --channel conda-forge
 
+Installation of the development version using nightly wheels
+------------------------------------------------------------
+
+If you want to test the latest development version of Shapely, the easiest way
+to get this version is by installing it from the Scientific Python index of
+nightly wheel packages::
+
+    python -m pip install --pre --upgrade --extra-index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple shapely
+
 
 Installation from source with custom GEOS libary
 ------------------------------------------------


jorisvandenbossche:
> I think we will have to add a `2.1.0.dev0` tag to the main branch (e.g. on the commit just after we branched maint-2.0)

cc @shapely/shapely-team for the above

jorisvandenbossche:
I added the dev tag to https://github.com/shapely/shapely/commit/eeef37fcf398afeb6342d37b63037dc398fa1a5c, which was the first commit on main not included in the maint-2.0 branch:

```
git tag 2.1.0.dev0 eeef37fcf398afeb6342d37b63037dc398fa1a5c
git push upstream --tags
```

jorisvandenbossche:
I re-triggered the wheel build and uploads on the last commit on main, which uploaded nightly wheels with 2.1.0.dev0 versions, and now the instructions in this PR to install that nightly wheel work fine

mwtoews:
Neat! I see it all at https://anaconda.org/scientific-python-nightly-wheels/shapely

Running this command gets shapely-2.0.1+78:

    pip install --upgrade -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple shapely

And to get shapely-2.1.0.dev0+64 the `--pre` flag is needed:

    pip install --upgrade --pre -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple shapely

mwtoews (approved):
Looks great!

jorisvandenbossche:
I removed the files with the 2.0.1+.. version from the index, to avoid confusion (otherwise they would also be removed automatically in some time)

The `--pre` flag is indeed needed to ensure pip prefers the dev version from the extra index over the released version on the default PyPI index.

PR #2003, opened 2024-03-04
Add shapely.geometry.polygon.orient to __all__

### LINKED ISSUE #722: 'orient' is not declared in __all__

quassy:
Add orient to `__all__` in polygon.py, supersedes #1805.

Fixes #722 where linting warnings like this appear:

> 'orient' is not declared in `__all__`

### DIFF
--- a/shapely/geometry/polygon.py
+++ b/shapely/geometry/polygon.py
@@ -10,7 +10,7 @@ from shapely.geometry.base import BaseGeometry
 from shapely.geometry.linestring import LineString
 from shapely.geometry.point import Point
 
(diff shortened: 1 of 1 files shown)


mwtoews:
Thanks!

(Note that the errors are unrelated)

PR #2006, opened 2024-03-05
BLD: update circleci image for arm wheel building

jorisvandenbossche:
See https://discuss.circleci.com/t/linux-image-deprecations-and-eol-for-2024/50177


jorisvandenbossche:
The wheel build is green. There are some failures on the dev CI build (I suppose rather related to numpy dev than GEOS dev)

PR #2007, opened 2024-03-05
COMPAT: fix assert_geometries_equal to work with numpy 2.0 copy changes

jorisvandenbossche:
Recent change in numpy main with different semantics for the `copy` keyword, where `copy=False` will now guarantee (instead of try) to not make a copy, and thus raise an error when it cannot create the array without allocating data.


mwtoews (approved):
Thanks! I see the upstream change https://github.com/numpy/numpy/commit/acf49fe0627f40a1dc00645b999c13fd6220ab30 and various connected issues.

PR #2008, opened 2024-03-06
ENH: add has_m

mwtoews:
Add a basic `has_m` unary predicate function, available with GEOS >= 3.12.0.

#### Other changes to note
* Update some related docs for `get_coordinate_dimension`
* Add `has_z` to `UNARY_PREDICATES` for shapely/tests/test_predicates.py

### DIFF
--- a/shapely/tests/test_io.py
+++ b/shapely/tests/test_io.py
@@ -1028,7 +1028,7 @@ def test_from_wkb_point_empty_m(wkb, expected_type):
     assert shapely.get_type_id(geom) == expected_type
     assert shapely.get_coordinate_dimension(geom) == 3
     assert not shapely.has_z(geom)
-    # TODO: assert shapely.has_m(geom)
+    assert shapely.has_m(geom)
 
 
 @pytest.mark.skipif(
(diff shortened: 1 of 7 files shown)


shapely/tests/test_io.py
```diff
@@ -1095,7 +1095,8 @@ def test_pickle_m(geom):
     actual = pickle.loads(pickled)
     assert_geometries_equal(actual, geom, tolerance=0)
     assert not actual.has_z
-    # TODO: assert actual.has_m
+    if not actual.is_empty:  # GEOSHasM with EMPTY geometries is inconsistent
```
jorisvandenbossche, line 1098: There is an upstream issue about this? 
Or is for the case of multi empty geometries (https://github.com/libgeos/geos/issues/888)?

shapely/tests/test_predicates.py
```diff
@@ -218,17 +227,47 @@ def test_dwithin():
...
+    assert not shapely.has_z(geometry)
+    assert shapely.has_m(geometry)
+
+
+@pytest.mark.skipif(
+    shapely.geos_version < (3, 12, 0),
+    reason="M coordinates not supported with GEOS < 3.12",
+)
+@pytest.mark.parametrize("geometry", all_types_zm)
+def test_has_m_all_types_zm(geometry):
```
jorisvandenbossche, line 266:
```suggestion
def test_has_z_has_m_all_types_zm(geometry):
```

jorisvandenbossche (approved):
Looks good!

shapely/tests/test_io.py
mwtoews: That's the main issue, and there are other GEOS-version-specific issues related to GEOSHasZ too like https://github.com/libgeos/geos/pull/887. These will hopefully get less complicated when support for older GEOS versions drops off. There are a few other "EMPTY geometries is inconsistent" comments that I've added to `main` with the same theme.

Issue #2004, opened 2024-03-04
TODO: fix WKT for empty M and ZM geometries


mwtoews:
On shapely `main` with #1808 to enable M and ZM in WKT outputs, one "TODO" item needs to be resolved for `test_repr_point_m_empty`. To demonstrate the issue:
```python
import shapely

assert shapely.geos_version[0:2] >= (3, 12), shapely.geos_version_string

empty_wkt = ["POINT Z EMPTY", "POINT M EMPTY", "POINT ZM EMPTY"]
empties = shapely.from_wkt(empty_wkt)

# This is one issue with repr
repr(empties)
# array([<POINT Z EMPTY>, <POINT Z EMPTY>, <POINT Z EMPTY>], dtype=object)
# but expected same as input wkt

# Another curious related issue
shapely.to_wkt(empties)
# array(['POINT Z EMPTY', None, None], dtype=object)

# But these really are different geometries
shapely.has_z(empties)
# array([ True, False,  True])
```

This is a shapely issue, and not with GEOS (3.12+). See `wkt_empty_3d_geometry` in `src/geos.c` for a possible resolution. Probably also rename away from "3D" too, since the third dimension is not always Z.

jorisvandenbossche:
> See `wkt_empty_3d_geometry` in `src/geos.c` for a possible resolution.

Should we just skip using that helper in case of GEOS 3.12+? Assuming that GEOS can now handle those cases correctly? (AFAIK that helper was added to correct some issues on the GEOS side, but it was also added in a time we didn't consider M values, so it's not surprising that it doesn't work correctly)

mwtoews:
Agree on skipping helper for GEOS 3.12+. I have a PR ready after #2012 is merged.

PR #2012, opened 2024-03-10
FIX: to_wkt with empty Z geometry arrays

mwtoews:
This fixes an issue while using `to_wkt` as a ufunc on an array with GEOS >= 3.9.0 with empty geometries. The issue (discovered with GH-2004) is that the WKT writer finishes sometimes too early while processing geometries.

For instance, using shapely 2.0.2 with GEOS 3.12.1 via conda-forge:
```python
import shapely

empty_geoms = shapely.from_wkt(["POINT Z EMPTY", "POINT Z EMPTY"])
list(shapely.to_wkt(empty_geoms))
# ['POINT Z EMPTY', None]
```

It also seems that there could be a missing `Py_XDECREF(*out);`, but I'd appreciate someone with Python C API expertise to weight in.

This can be backported to `maint-2.0` too.

### DIFF
--- a/shapely/tests/test_io.py
+++ b/shapely/tests/test_io.py
@@ -415,6 +415,15 @@ def test_to_wkt_none():
     assert shapely.to_wkt(None) is None
 
 
+def test_to_wkt_array_with_empty_z():
+    # See GH-2004
+    empty_wkt = ["POINT Z EMPTY", None, "POLYGON Z EMPTY"]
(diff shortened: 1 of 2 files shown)


jorisvandenbossche (approved):
Thanks! This looks correct to me now

mwtoews:
Backported to `maint-2.0` with 5976300d197570b6df1f27d1a9c4764836d1b02e

Issue #1994, opened 2024-02-16
`interleaved` parameter is missing from `shepely.transform` in the released versions but not in the main branch and documentation


zergar:
## Please note

If you are reporting an installation or module import issue, please note that
this project only accepts reports about problems with packages downloaded from
the Python Package Index. Conda users should take issues to one of the
following trackers:

- https://github.com/ContinuumIO/anaconda-issues/issues
- https://github.com/conda-forge/shapely-feedstock

### Expected
I can use the `interleaved` parameter in the `shapely.transform` method from the `coordiantes.py`  file that is described in the Documentation (https://shapely.readthedocs.io/en/latest/reference/shapely.transform.html) as well as the main branch.

### Actual

The `interleaved` parameter is missing in the `shapely.transform` method from the `coordinates.py` file in the maint-2.0 branch (https://github.com/shapely/shapely/blob/maint-2.0/shapely/coordinates.py#L8)

## Steps to reproduce the problem.

Download the files from the last release (https://github.com/shapely/shapely/releases/tag/2.0.3) or 

## Operating system

any

## Shapely version and provenance

2.0.2 and 2.0.3

jorisvandenbossche:
You are looking at the "latest" version of the docs, while you want the "stable" docs for the latest released version ("latest" is a bit a confusing name for the development version ..)

See https://shapely.readthedocs.io/en/stable/reference/shapely.transform.html

The `interleaved` keyword will be included in a future shapely 2.1.0 release

jorisvandenbossche:
> ("latest" is a bit a confusing name for the development version ..)

Although I see that our latest docs say they are for "shapely 2.0.1". That makes it even more confusing, and something we should fix to make it clear this is for the dev version.

zergar:
Ah ok, thanks. I guess i was confused by the website stating "shapely 2.0.1" in the top left.

Could you maybe also mark the parameter as a future one in the docs? This would clarify it even more.

jorisvandenbossche:
The latest docs (https://shapely.readthedocs.io/en/latest/) now say "2.1.0.dev", so that should be clearer.

> Could you maybe also mark the parameter as a future one in the docs? This would clarify it even more.

Yes, that's a good idea. We should more consistently add "versions added" markers for new functions or keywords.

PR #2013, opened 2024-03-10
DOC: add versionadded note to new functions or keywords for 2.1

jorisvandenbossche:
xref https://github.com/shapely/shapely/issues/1994#issuecomment-1987266626

With the upcoming 2.0 release, I think it will be nice to start documenting more explicitly when keywords or entire functions are new (I think it will be normal to require any >= 2, and when doing that you can only use features that already existed in 2.0).

Sphinx has a `versionadded` directive for this, and so this PR is adding that to the relevant places for 2.1.

If we like this, we should also try to think about doing this directly in PRs adding the new features.


jorisvandenbossche:
Example how this looks in the built docs:

image

mwtoews (approved):
Thanks, looks good! And going forward, I'll include this note for new enhancements.

PR #2014, opened 2024-03-11
FIX: fix WKT for empty M and ZM geometries

### LINKED ISSUE #2004: TODO: fix WKT for empty M and ZM geometries

mwtoews:
This uses `wkt_empty_3d_geometry` only for GEOS >= 3.9.0 and < 3.12.0, otherwise it is not defined or used.

Closes #2004

### DIFF
--- a/shapely/tests/test_io.py
+++ b/shapely/tests/test_io.py
@@ -593,9 +593,6 @@ def test_repr_point_z_empty():
     assert repr(empty_point_z) == "<POINT Z EMPTY>"
 
 
-@pytest.mark.xfail(
-    reason="TODO: fix WKT for empty M and ZM geometries; see GH-2004", strict=True
-)
 @pytest.mark.skipif(
     shapely.geos_version < (3, 12, 0),
     reason="M coordinates not supported with GEOS < 3.12",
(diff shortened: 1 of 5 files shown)


jorisvandenbossche (approved):
Thanks!

PR #2016, opened 2024-03-11
BLD: use geos INSTALL_NAME_DIR for MacOS instead of specifying custom LDFLAGS with rpath

jorisvandenbossche:
_If_ this works, it might be a bit cleaner (from https://github.com/isciences/exactextract/pull/87)


mwtoews (approved):
Neat! See [INSTALL_NAME_DIR](https://cmake.org/cmake/help/latest/prop_tgt/INSTALL_NAME_DIR.html) for description of this Apple platform variable.

PR #2019, opened 2024-03-12
ENH: get_m and point.m property

mwtoews:
Adds `get_m` and a `m` point property, available with GEOS 3.12 or later. For example:

```python
import shapely

assert shapely.geos_version[0:2] >= (3, 12), shapely.geos_version_string

pt = shapely.from_wkt("POINT M (1 2 4)")
assert pt.m == shapely.get_m(pt) == 4.0
```
And similar to the `z` property:
```python
pt = shapely.geometry.Point(1, 2)
shapely.get_m(pt)  # nan
pt.m
# raises: DimensionError: This point has no m coordinate.
```

Other checks to `shapely/tests/geometry/test_point.py` will be added after reworking `Point()`.

### DIFF
--- a/shapely/geometry/point.py
+++ b/shapely/geometry/point.py
@@ -3,6 +3,7 @@
 import numpy as np
 
 import shapely
+from shapely.decorators import requires_geos
 from shapely.errors import DimensionError
 from shapely.geometry.base import BaseGeometry
 
@@ -12,7 +13,7 @@ __all__ = ["Point"]
 class Point(BaseGeometry):
     """
     A geometry type that represents a single coordinate with
-    x,y and possibly z values.
+    x, y and possibly z and/or m values.
 
     A point is a zero-dimensional feature and has zero length and zero area.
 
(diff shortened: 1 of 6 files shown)


shapely/geometry/point.py
```diff
@@ -99,6 +100,14 @@ def z(self):
             raise DimensionError("This point has no z coordinate.")
         return shapely.get_z(self)
 
+    @property
+    @requires_geos("3.12.0")
```
jorisvandenbossche, line 104: I think you can leave out the decorator here, because calling `shapely.has_m` will already raise an error for that

jorisvandenbossche (approved):

shapely/geometry/point.py
mwtoews: Sure. Also I've added a "versionadded" note for the property that shows up in the [docs](https://shapely--2019.org.readthedocs.build/en/2019/reference/shapely.Point.html#shapely.Point.m).

mwtoews:
@jorisvandenbossche do you have any idea why "Create a merge commit" is no longer enabled for this repository? This is usually my "go to" method, otherwise I can manually re-write this PR with one commit.

PR #2030, opened 2024-03-26
DOC: update user manual for functions that ~moved from shapely.ops to shapely

theroggy:
From shapely 2.0, many functions that were in `shapely.ops` before, became available directly in the `shapely` namespace, I suppose in the spirit of [Flat is better than nested](https://peps.python.org/pep-0423/#avoid-deep-nesting). Some functions were also renamed to be more consistent with GEOS/PostGIS.

However, the user manual wasn't updated to reflect this. This PR tries to incorporate these changes.

I expected this to be a mainly cosmetic change, but while making the changes I encountered many more differences between the functions than I was expecting, so to be discussed how to deal with this:
- line-like input versus explicit iterables of Linestrings: line oriented functions in `shapely.ops` support line-like input, but the equivalent functions in `shapely` don't . I changed the doc to reflect that, but maybe this is not wanted? Relevant functions:
    - shapely.ops.polygonize` vs `shapely.polygonize`
    - `shapely.ops.polygonize_full` vs `shapely.polygonize_full`
- `shapely.line_merge` is a vectorized function and expects a (arraylike of) MultiLineStrings as input. Hence, many merge operations can be done in one call. `shapely.ops.linemerge` expects an iterable of line-line elements as input, and will merge those, so only one merge operation can be done per call.
- some functions return slightly different results, e.g.:
    - `shapely.shortest_line` returns a LineString vs `shapely.ops.nearest_points` returns a Tuple of 2 Points


docs/manual.rst
```diff
@@ -2362,37 +2347,37 @@ Voronoi diagram from a collection points, or the vertices of any geometry.
...
+  >>> from shapely import shortest_line
   >>> triangle = Polygon([(0, 0), (1, 0), (0.5, 1), (0, 0)])
   >>> square = Polygon([(0, 2), (1, 2), (1, 3), (0, 3), (0, 2)])
-  >>> list(nearest_points(triangle, square))
-  [<POINT (0.5 1)>, <POINT (0.5 2)>]
+  >>> shortest_line(triangle, square)
+  <LINESTRING (0.5 1, 0.5 2)>
 
 
 Note that the nearest points may not be existing vertices in the geometries.
```
sgillies, line 2372:
```suggestion
Note that the shortest line may not connect to vertices of the input geometries.
```

docs/manual.rst
```diff
@@ -2362,37 +2347,37 @@ Voronoi diagram from a collection points, or the vertices of any geometry.
...
+of geometries.
 
-.. function:: shapely.ops.nearest_points(geom1, geom2)
+.. function:: shapely.shortest_line(geom1, geom2)
 
-   Returns a tuple of the nearest points in the input geometries. The points are
+   Returns a tuple of shortest line between the input geometries. The points are
    returned in the same order as the input geometries.
 
    `New in version 1.4.0`.
```
sgillies, line 2361:
```suggestion
   `New in version 2.0`.
```

docs/manual.rst
```diff
@@ -2443,22 +2428,27 @@ paths between two linear geometries.
...
-   It may be convenient to snap the splitter with low tolerance to the geometry. For example in the case of splitting a line by a point, the point must be exactly on the line, for the line to be correctly split.
-   When splitting a line by a polygon, the boundary of the polygon is used for the operation.
-   When splitting a line by another line, a ValueError is raised if the two overlap at some segment.
+   It may be convenient to snap the splitter with low tolerance to the geometry. For
+   example in the case of splitting a line by a point, the point must be exactly on the
+   line, for the line to be correctly split. When splitting a line by a polygon, the
+   boundary of the polygon is used for the operation. When splitting a line by another
+   line, a ValueError is raised if the two overlap at some segment.
 
    `New in version 1.6.0`
```
sgillies, line 2453:
```suggestion
```

1.6 is ancient history, I think we can safely remove this. Perhaps we remove all such references to releases before 2.0?

sgillies (approved):
@theroggy thank you for doing this! I made a couple of suggestions inline.

I'd love to see the text wrapped at 79 chars like the original manual, but that's not important.

sgillies (approved):
Thanks again!

sgillies:
@theroggy Correct, the 1.x API (which still exists in some form in shapely.ops) had functions that accepted coordinate sequences and implicitly converted them to geometries, but the 2.x API doesn't promote this anymore and it's good to remove this usage from the manual.

Issue #1995, opened 2024-02-21
get_coordinates(point) is 4x faster than (point.x, point.y)


Zaczero:
At this moment, this is not a bug report nor a feature request. I primarily want to highlight the fact that `get_coordinates(point)` is approximately 4x faster than using `point.x, point.y`.

I believe this matters substantially as the `point.x, point.y` pattern is very popular for retrieving Python float numbers for a given point. It is also quite often present in hot loops.

A simple naive search on GitHub for "point.x, point.y" [yields](https://github.com/search?q=%22point.x%2C+point.y%22+language%3APython&type=code&l=Python) almost 7000 results, many of which are in for loops.

### Benchmark

```txt
Python 3.12.1
shapely 2.0.3
```

```sh
python -m timeit -s "from shapely import Point" -s "a = Point(12,34)" "a.x, a.y"
# 20000 loops, best of 5: 11.3 usec per loop
```

```sh
python -m timeit -s "from shapely import Point, get_coordinates" -s "a = Point(12,34)" "get_coordinates(a)[0]"
# 100000 loops, best of 5: 2.98 usec per loop
```

Lash-L:
Yeah this has been on my mind for a while. We use shapely a lot for work, and when we do our optimization passes, we always go through and just do x = point.x and then use x whenever we need the value. I wasn't aware of get_coordinates() 

I do struggle to understand why a calculation needs to be done at all for x and y and why they aren't just stored. Perhaps i'm missing something basic, but operations aren't done in place that I can think of, so I don't fully get why you would need a calculation - but I also haven't dove into the GEOS library to see.

It definitely should at the very least be a note in the docs as it can be a significant slowdown to codebases.

martinfleis:
Given what shapely is doing here is calling respective GEOS functions here, the question boils down to why `GEOSGeomGetX_r`, `GEOSGeomGetY_r` is slower than `GEOSGeom_getCoordSeq_r` within GEOS itself. 

Also note that the comparison above is not precise as  doing `a.x, a.y` is calling `get_x` first and then `get_y`, while `get_coordinates` does only a single pass. I see two actionable items:

1. Follow up with GEOS in https://github.com/libgeos/geos and ask the same question there.
2. Use `get_coordinates` internally in `Point.x` and `Point.y`. At the same time, we should try to steer users towards `get_coordinates` or `get_x` when they need coordinates and do the loop as the performance boost of the vectorized operation will be much higher than the internal refactoring.

[...]

PR #2032, opened 2024-04-11
ENH: Optimize z property of point

Lash-L:
Slightly based off of conversation here: https://github.com/shapely/shapely/issues/1995

I noticed that getting .z was incredibly slow compared to get_coordinates.

This refactors .z to be significantly faster.

My test:
```
start = time.time()
for _ in range(100000):
    p = geo.Point(random.random(), random.random(), random.random())
    z = p.z
    f = p.z
    g = p.z
print(time.time() - start)
```

edit: Removed old comparisions

Nightly comparision:
Old: 4.941876173019409
New:3.2745301723480225


mwtoews:
With upcoming development for shapely 2.1 with GEOS 3.12+ the third coordinate (i.e. `coords[2]`) with XYM geometries is the M coordinate (although not at the moment, as I'm currently still planning `get_coodinates()`. So this behavior may change soon.

Lash-L:
> With upcoming development for shapely 2.1 with GEOS 3.12+ the third coordinate (i.e. `coords[2]`) with XYM geometries is the M coordinate (although not at the moment, as I'm currently still planning `get_coodinates()`. So this behavior may change soon.

Interesting. well I think i actually just found an approach that is even faster if you are open to it? 

Basically it still uses get_z, but it skips the has_z. as get_z does that for us by returning nan. It's a bit faster. Getting 3.1s 

I also realized my numbers for get_z above were incorrect. it is in the 5s, not the 10. Not sure what happened there.

mwtoews:
Intriguing that `has_z` is slow, could be something to investigate further.

One surprise that I don't see from the existing tests is that newer GEOS versions allow NaN as valid coordinate values, i.e.:
```
In [1]: import shapely

In [2]: geom = shapely.from_wkt("POINT Z (1 2 NaN)")

In [3]: geom.has_z
Out[3]: True

In [4]: geom.z
Out[4]: nan

In [5]: shapely.get_coordinates(geom)
Out[5]: array([[1., 2.]])
```
which means this geometry should continue to allow access to `.z` (returning `nan`). Not sure why this failure didn't pop up.

Lash-L:
I wouldn't say has_z is super slow - but it def has some slowdown and some redundant things are happening if I am understanding things correctly.

This is in the case of getting z when we do have a z coord - I can do this same thing with when we don't have a z coord if oyu're interested.

But in the case that we do have a z coord, half of our time is spent in the has_z function.

Old method: 12791345
New method: 7440167 
About a 40% speed up
image



> One surprise that I don't see from the existing tests is that newer GEOS versions allow NaN as valid coordinate values,

That seems counter to what I see in the GEOS code(pulled from main iirc)
image

I could be misunderstanding something(still trying to get a grasp of GEOS and my cpp is rusty), but it seems to still check if it is nan.

Although I guess if hasz is true, the nan check doesn't matter.

I'm going to try to learn how to build a newer version of GEOS locally for shapely so I can play with things

jorisvandenbossche:
> I could be misunderstanding something(still trying to get a grasp of GEOS and my cpp is rusty), but it seems to still check if it is nan.
> 
> Although I guess if hasz is true, the nan check doesn't matter.

Yeah, depending on how the Point is created, it "knows" its dimension, and it has stored whether it has a z coord in `m_hasz`. It's only when that information is not available that it falls back to checking the third coordinate being NaN or not. 

That means that with recent GEOS, you can explicitly create a 3D POINTZ (`m_hasz` will be set) with a NaN as z coord.

shapely/geometry/point.py
```diff
@@ -95,9 +95,10 @@ def y(self):
     @property
     def z(self):
         """Return z coordinate."""
-        if not shapely.has_z(self):
+        z = shapely.get_z(self)
+        if np.isnan(z):
```
jorisvandenbossche, line 99:
```suggestion
        if np.isnan(z) and not shapely.has_z(self):
```

This would ensure we keep returning NaN if it actually has a Z coord that is NaN, while for the common case avoiding the `has_z` check

Lash-L: That makes sense to me. Will be faster for all use cases except for those where nan was passed in as a variable to the Point. I think that is a pretty good trade off.

PR #2035, opened 2024-04-12
PERF: optimize coordinate access for scalar Point

jorisvandenbossche:
Illustration of what it would entail to optimize specifically the `Point.x` et al attributes for scalar points.

```python
In [1]: import shapely

In [2]: p = shapely.Point(12, 34)

In [3]: %timeit p.x
3.15 µs ± 3.46 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)    # <-- main
236 ns ± 0.164 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)  # <-- PR
```

The reason this is so much faster is mostly from avoiding the ufunc machinery overhead.

While this is relatively little code for just those `x`/`y`/`z` attributes, the question is of course then "could we also speed up this other attribute .. ?", and we definitely don't want to end up reimplementing all ufuncs for scalar geometries.

### DIFF
--- a/shapely/_geometry_helpers.pyx
+++ b/shapely/_geometry_helpers.pyx
@@ -24,6 +24,7 @@ from shapely._geos cimport (
     GEOSGeom_createPolygon_r,
     GEOSGeom_destroy_r,
     GEOSGeometry,
+    GEOSGeomGetX_r,
     GEOSGeomTypeId_r,
     GEOSGetExteriorRing_r,
     GEOSGetGeometryN_r,
@@ -392,6 +393,21 @@ def collections_1d(object geometries, object indices, int geometry_type = 7, obj
             geom_idx_1 += collection_size[coll_idx]
 
     return out
+
+
+def point_x(point):
+    cdef GEOSGeometry *geom = NULL
+    cdef double result
+
+    with get_geos_handle() as geos_handle:
+        if PyGEOS_GetGEOSGeometry(<PyObject *>point, &geom) == 0:
+            raise TypeError("One of the arguments is of incorrect type. "
+                            "Please provide only Geometry objects.")
+
+        if GEOSGeomGetX_r(geos_handle, geom, &result) == 0:
+            return  # GEOSException is raised by get_geos_handle
+
+    return result
 
 
 def _geom_factory(uintptr_t g):
--- a/shapely/_geos.pxd
+++ b/shapely/_geos.pxd
@@ -44,6 +44,7 @@ cdef extern from "geos_c.h":
     int GEOSCoordSeq_setZ_r(GEOSContextHandle_t handle, GEOSCoordSequence* s, unsigned int idx, double val) nogil
     int GEOSCoordSeq_getSize_r(GEOSContextHandle_t handle, GEOSCoordSequence* s, unsigned int* size) nogil
 
+    int GEOSGeomGetX_r(GEOSContextHandle_t handle, const GEOSGeometry *g, double *x)
 
 cdef class get_geos_handle:
     cdef GEOSContextHandle_t handle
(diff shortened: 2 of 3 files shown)


theroggy:
> While this is relatively little code for just those `x`/`y`/`z` attributes, the question is of course then "could we also speed up this other attribute .. ?", and we definitely don't want to end up reimplementing all ufuncs for scalar geometries.

I definitely understand this...

Then again, there are cases where a vectorized approach is not feasable, e.g. when implementing more complicated algorithms using many basic spatial operations on a single geometry. If a shortcut that bypasses the ufunc machinery if the input is a single element would give a big performance difference...

caspervdw:
I am in doubt whether it is forth the effort to optimise scalar calls in shapely. We might shave of a significant part of the scalar call, but still, in most situations where the shapely call is actual a bottleneck, you’ll have many of such calls and you can just use the vectorised shapely functions.

Still there is lots of code out there built with shapely 1.0, and also lots of users out there just use the OO interface without wanting to think much about vectorisation. For those use cases, special casing some implementations would be acceptable.

> While this is relatively little code for just those x/y/z attributes, the question is of course then "could we also speed up this other attribute .. ?", and we definitely don't want to end up reimplementing all ufuncs for scalar geometries.

Why not? It potentially is a lot of code, sure, but it’s not complex. Maybe we can share part of the ufunc code with the scalar calls.

theroggy:
Algorithms like [centerline](https://github.com/pygeoops/pygeoops/blob/main/pygeoops/_centerline.py) are not really feasible to implement without many calls on scalar functions I think. Could be implemented in cython of course, but...

In an ideal world the optimized scalar path would be implemented in c I suppose so it can be tackled centrally, at least for all functions that don't need custom code in ufunc.c? But I don't know that code at all, so possibly/probably this is not possible?

caspervdw:
I did a small study on [this branch](https://github.com/caspervdw/Shapely/tree/casper-ufunc-scalar) and identified the performance bottlenecks:

| **Function** | **Time (ns)** | **Overhead Removed** | **Cumulative Savings** | **What Was Eliminated** |
|--------------|---------------|---------------------|------------------------|------------------------|
 `shapely.get_x` | **2,250** |  |  |  |
| `shapely.lib.get_x` | **1,120** | **-1,130ns** | **1,130ns** | Python wrapper + `@multithreading_enabled` decorator |
| `shapely.lib.get_x_1` | **345** | **-775ns** | **1,905ns** | NumPy ufunc machinery |
| `shapely.lib.get_x_2` | **322** | **-23ns** | **1,928ns** | for loop |
| `shapely.lib.get_x_4` | **217** | **-105ns** | **2,033ns** | Kept the GIL during the call |
 `shapely.lib.get_x_5` | **122** | **-75ns** | **2,128ns** | Used global GEOS context (but released GIL again)|
| `shapely.lib.get_x_3` | **60.3** | **-62ns** | **2,190ns** | Optimizations 4 and 5 together|

## Optimization Breakdown

1. **Remove Python wrapper** (-1,130ns): Eliminate decorator overhead and function call
2. **Bypass NumPy ufunc** (-775ns): Skip array machinery for scalar operations  
3. **Use global GEOS context** (-200ns): Avoid repeated init/cleanup
4. **Keep GIL held** (-62ns): Eliminate overhead of releasing/reacquiring the GIL
## Proposal for going forward

- The `multithreading_enabled` decorator can be circumvented safely because it does nothing for scalars. This is an easy win and already saves about a microsecond. 
- The ufunc machinery can be avoided by implementing a scalar version from C or Cython. I'd opt for reusing code (either the scalar reusing the vectorized function, like I did on my branch, or both functions using a shared base). In any case, this is major work. It would remove almost another microsecond.
- Using the global GEOS context and keeping the GIL for scalar (or small) arrays.

caspervdw:
Closing this now #2339 is merged

PR #2050, opened 2024-05-13
ENH: Add polygonhull_simplify exposing PolygonHullSimplifier

aloi214:
GEOS introduces a PolygonHullSimplifier in v3.11. This PR adds a shapely binding.

### DIFF
--- a/shapely/constructive.py
+++ b/shapely/constructive.py
@@ -30,6 +30,7 @@ __all__ = [
     "remove_repeated_points",
     "reverse",
     "simplify",
+    "polygonhull_simplify",
     "snap",
     "voronoi_polygons",
     "oriented_envelope",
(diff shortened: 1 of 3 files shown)


aloi214:
@jorisvandenbossche I hope I’m not interrupting you at this time. It's my first PR and I want to know what do I need to do with this PR. :smile:

shapely/constructive.py
```diff
@@ -926,6 +927,61 @@ def simplify(geometry, tolerance, preserve_topology=True, **kwargs):
         return lib.simplify(geometry, tolerance, **kwargs)
 
 
+@multithreading_enabled
+def polygonhull_simplify(
```
martinfleis, line 931: I suppose that given the aim to match the PostGIS naming, this should be called `simplify_polygon_hull` after [`ST_SimplifyPolygonHull`](https://postgis.net/docs/ST_SimplifyPolygonHull.html).

shapely/constructive.py
```diff
@@ -926,6 +927,61 @@ def simplify(geometry, tolerance, preserve_topology=True, **kwargs):
...
+    >>> polygonhull_simplify(Polygon([(0,0), (0, 4), (1,3), (3,4), (10, 4), (10, 0), (0, 0)]), 0.1, "area", False)
+    <POLYGON ((0 0, 10 0, 10 4, 3 4, 1 3, 0 0))>
+    >>> polygonhull_simplify(Polygon([(0,0), (0, 4), (1,3), (3,4), (10, 4), (10, 0), (0, 0)]), 0.1, "area", True)
+    <POLYGON ((0 0, 0 4, 10 4, 10 0, 0 0))>
+    """
+    if lib.geos_version < (3, 11, 0):
+        raise UnsupportedGEOSVersionError(
+            "Polygonhull Simplifier require GEOS >= 3.11.0, "
+            f"found {lib.geos_version_string}"
+        )
```
martinfleis, line 968: You shall replace this with the `@requires_geos("3.11.0")` decorator.

shapely/constructive.py
```diff
@@ -926,6 +927,61 @@ def simplify(geometry, tolerance, preserve_topology=True, **kwargs):
...
+    vertices or the input geometry's area retained retained in the result.
+
+    Parameters
+    ----------
+    geometry : Geometry or array_like
+    parameter : float or array_like
+        Larger values produce less concave results. A value of 1 produces
+        the convex hull; a value of 0 produces the original geometry.
+    parameter_mode : str
+        vertice - Fraction of input vertices retained, area - Ratio of simplified hull area to input area
```
martinfleis, line 945: I think that it should be "vertex", not "vertice".

aloi214: Thanks for your opinion, I'll adjust the name.

shapely/constructive.py
```diff
@@ -926,6 +927,56 @@ def simplify(geometry, tolerance, preserve_topology=True, **kwargs):
...
+    """Computes a boundary-respecting hull of a polygonal geometry, with hull
+    shape determined by a target parameter specifying the fraction of the input
+    vertices or the input geometry's area retained retained in the result.
+
+    Parameters
+    ----------
+    geometry : Geometry or array_like
+    parameter : float or array_like
+        Larger values produce less concave results. A value of 1 produces
+        the convex hull; a value of 0 produces the original geometry.
```
martinfleis, line 944: This is actually correct only for area mode. In the vertex mode, it is the vice versa. 0 is a convex_hull, while 1 is original geometry (as in PostGIS). That can be a bit confusing. I would recommend keeping both the same, probably following the vertex mode and PostGIS.

martinfleis: Also, it is convex hull only if `is_outer=True`, otherwise it is the largest triangle within the geom (I think).

shapely/constructive.py
```diff
@@ -926,6 +927,56 @@ def simplify(geometry, tolerance, preserve_topology=True, **kwargs):
...
+        parameter_mode = 2
+    elif parameter_mode == "vertex":
+        parameter_mode = 1
+    else:
+        raise ValueError("The 'parameter_mode' only support 'area' or 'vertex'")
+    return lib.simplify_polygon_hull(
+        geometry,
+        np.bool_(is_outer),
+        np.intc(parameter_mode),
+        np.double(parameter),
```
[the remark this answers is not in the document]
aloi214: The reason I casting numpy dtypes is that the running tests in CI/CD workflow failed in ubuntu 3.11/3.12 and macos 3.11/3.12 as below. So I cast parameters to numpy dtypes.
`E           TypeError: ufunc 'simplify_polygon_hull' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''`

shapely/constructive.py
```diff
@@ -926,6 +927,56 @@ def simplify(geometry, tolerance, preserve_topology=True, **kwargs):
...
+    shape determined by a target parameter specifying the fraction of the input
+    vertices or the input geometry's area retained retained in the result.
+
+    Parameters
+    ----------
+    geometry : Geometry or array_like
+    parameter : float or array_like
+        Larger values produce less concave results. A value of 1 produces
+        the convex hull; a value of 0 produces the original geometry.
+    parameter_mode : str
```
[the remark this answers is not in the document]
aloi214: Sure, I think `parameter` can be replaced by `tolerance` to be consistent with `simplify`.

shapely/constructive.py
aloi214: Yes, it's easy to misunderstand. The explanation should be:

Two parameters are supported:
- Vertext Mode: the tolerance means Vertex Number fraction. the fraction of the input vertices retained in the result. Value 1 produces the original geometry. Smaller values produce less concave results. For outer hulls, value 0 produces the convex hull (with triangles for any holes). For inner hulls, value 0 produces a triangle (if no holes are present). The value should be in the range [0,1].
- Area Mode: the tolerance means Area Delta ratio. the ratio of the change in area to the input area. Value 0 produces the original geometry. Larger values produce less concave results. The value must be 0 or greater.

aloi214: Parameter is in range [0, 1] for vertex mode and in range [0, ∞) for area mode. 
```python
a = shapely.Polygon([[0,0], [100,0], [200, 500], [100, 10], [0, 10]])
b = shapely.simplify_polygon_hull(a, 1, "area", True)
c =shapely.simplify_polygon_hull(a, 10, "area", True)
d = shapely.simplify_polygon_hull(a, 99, "area", True)
print(a.area, b.area, c.area, d.area) # 1500.0 1500.0 1500.0 26000.0
```
I can keep both same by limit the tolerance in the range [0, 1] like
```python
    if parameter_mode == "area":
        parameter_mode = 2
        tolerance = 1-tolerance if tolerance < 1 else 0
    elif parameter_mode == "vertex":
        parameter_mode = 1
```
However, it will limit result in area mode as the area delta ratio couldn't be larger than 1 in this case. It's hard to keep both same.😟

tomkom:
Just checking, is this planned for a release? I would love to have the ability to compute inner hulls and bring shapely to parity to PostGIS in this function, missing in shapely to date.

PR #2053, opened 2024-05-14
ENH: add copy=None to CoordinateSequence.__array__

mwtoews:
Recent CI failures have picked up a recent NumPy 2.0 change, noted in the migration guide under ["Adapting to changes in the copy keyword"](https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword).

A simple fix appears to be to simply add `copy=None` to the `__array__` signature. This value is not used by the method.

---

[Here is an example failure](https://github.com/shapely/shapely/actions/runs/9070802200/job/24923285344#step:16:178):
```
________________________ TestCoords.test_data_promotion ________________________

self = <shapely.tests.geometry.test_coords.TestCoords object at 0x7f6bbc72fb90>

    def test_data_promotion(self):
        coords = np.array([[12, 34], [56, 78]], dtype=np.float32)
>       processed_coords = np.array(LineString(coords).coords)
E       DeprecationWarning: __array__ implementation doesn't accept a copy keyword, so passing copy=False failed. __array__ must implement 'dtype' and 'copy' keyword arguments.

shapely/tests/geometry/test_coords.py:16: DeprecationWarning
```

Note that test failures with macos in this PR are unrelated.

### DIFF
--- a/shapely/coords.py
+++ b/shapely/coords.py
@@ -46,8 +46,13 @@ class CoordinateSequence:
         else:
             raise TypeError("key must be an index or slice")
 
-    def __array__(self, dtype=None):
-        return self._coords
+    def __array__(self, dtype=None, copy=None):
+        if copy is False:
+            raise ValueError("`copy=False` isn't supported. A copy is always created.")
+        elif copy is True:
+            return self._coords.copy()
+        else:
+            return self._coords
 
     @property
     def xy(self):
--- a/shapely/tests/geometry/test_coords.py
+++ b/shapely/tests/geometry/test_coords.py
@@ -2,6 +2,7 @@ import numpy as np
 import pytest
 
 from shapely import LineString
+from shapely.tests.common import line_string, line_string_z, point, point_z
 
 
 class TestCoords:
@@ -84,3 +85,18 @@ class TestXY:
         assert list(x) == [0.0, 1.0]
         assert len(y) == 2
         assert list(y) == [0.0, 1.0]
+
+
+@pytest.mark.parametrize("geom", [point, point_z, line_string, line_string_z])
+def test_coords_array_copy(geom):
+    """Test CoordinateSequence.__array__ method."""
+    coord_seq = geom.coords
+    assert np.array(coord_seq) is not np.array(coord_seq)
+    assert np.array(coord_seq, copy=True) is not np.array(coord_seq, copy=True)
+
+    # Behaviour of copy=False is different between NumPy 1.x and 2.x
+    if int(np.version.short_version.split(".", 1)[0]) >= 2:
+        with pytest.raises(ValueError, match="A copy is always created"):
+            np.array(coord_seq, copy=False)
+    else:
+        assert np.array(coord_seq, copy=False) is np.array(coord_seq, copy=False)


jorisvandenbossche:
Thanks! I had noted this as a TODO item in https://github.com/shapely/shapely/issues/1972, but didn't yet get to it (and didn't expect our CI to fail for it, but so this is a good reminder ;)) 

Ideally we would honor the `copy` keyword _if_ passed. I think in our case, this just means raising an error if `copy is False`, because the coordinates we return are always a copy? 
But this is less critical, and could also be done later in a follow-up.

mwtoews:
I had a play with various copy=True/False/None versions, and it all seems minor impact, mostly because CoordinateSequence objects aren't really used as separate objects. Each `.coords` property access creates a new sequence object by default.

A simple implementation would be:
```python
    def __array__(self, dtype=None, copy=None):
        if copy:
            return self._coords.copy()
        return self._coords
```
then tested (somewhere):
```python
def test_coords_array():
    coord_seq = point.coords
    assert np.array(coord_seq) is not np.array(coord_seq)
    assert np.array(coord_seq, copy=False) is np.array(coord_seq, copy=False)
    assert np.array(coord_seq, copy=True) is not np.array(coord_seq, copy=True)
```

jorisvandenbossche:
Hmm, yes, it depends on how you interpret the CoordinateSequence object. Is that just a wrapper around an already materialized numpy array `self._coords` (how it is implemented now in practice) or is it a logical abstraction around the _geometry's_ coord seq (how it was actually also implemented in shapely 1.x)? 

Because in the former case, you can indeed say that `np.array(geom.coords, copy=False)` is fine because getting the `.coords` object in `geom.coords` itself has already made a copy, and so the subsequent conversion to numpy does not require a copy. 
But in the latter case you could argue that `np.array(geom.coords, copy=False)` should raise an error because it can never the the `geom.coords` without involving a copy of the actual values (like, `np.array(geom.coords, copy=False) is np.array(geom.coords, copy=False)` will never be true).

Now, since this CoordinateSequence class is a separate object from the Geometry objects, it might be that the former interpretation is the only viable one.

mwtoews:
From [The `__array__()` method](https://numpy.org/devdocs/user/basics.interoperability.html#the-array-method) docs "where copy=False should raise an exception if a copy is needed". And the [basic dispatch example](https://numpy.org/devdocs/user/basics.dispatch.html#basics-dispatch) raises ValueError for this case. This PR is now updated, renamed as an enhancement.

mwtoews:
Any further review/comment @jorisvandenbossche ?

shapely/coords.py
```diff
@@ -46,8 +46,13 @@ def __getitem__(self, key):
         else:
             raise TypeError("key must be an index or slice")
 
-    def __array__(self, dtype=None):
-        return self._coords
+    def __array__(self, dtype=None, copy=None):
+        if copy is False:
+            raise ValueError("`copy=False` isn't supported. A copy is always created.")
+        elif copy is True:
+            return self._coords.copy()
```
jorisvandenbossche, line 53: On thing I am still wondering: if we say that a copy is always created (in the error message above), then maybe we should not actually do the explicit copy here?

Of course, if one saved the `.coords` separately, one might still expect this actual copy ..

```
coords = geom.coords
coords_arr1 = np.asarray(coords, copy=True)
coords_arr2 = np.asarray(coords, copy=True)
```

i.e. one would expect in the case above that those two numpy arrays are not views of each other .. So I assume the above explicit copy is fine. We are just a bit inconsistent about the copy=False vs copy=True

jorisvandenbossche (approved):
Sorry, this fell off my radar. Let's get this in to fix the CI failures

PR #2054, opened 2024-05-14
CI: use macos-13 for tests to run x64 python

jorisvandenbossche:
Installing python with a x64 architecture stopped working with `macos-latest` (https://github.com/actions/setup-python/issues/855#issuecomment-2096792205), so as a quick workaround specifically asking for macos-13 runner instead (longer term we should check how to run with latest macos, but that requires changing the set up of the `architecture` matrix entry a bit)


mwtoews (approved):
This was the advice I read too, and it does solve this particular issue.

PR #2071, opened 2024-06-24
BLD: replace pkg_resources, prepend numpy include dirs

mwtoews:
This PR replaces setuptools' deprecated [`pkg_resources`](https://setuptools.pypa.io/en/latest/pkg_resources.html) module with a simple function to compare versions as tuples.

Expressions like this should be fine: `assert version_tuple("3.8") < version_tuple("3.13.0dev")`

Other changes:

- Replace `subprocess.Popen` with simpler `subprocess.run`
- Prepend `numpy.get_include()` rather than append, otherwise system libs (e.g. [`/usr/include/python3.10`](https://packages.ubuntu.com/jammy/amd64/python3-numpy/filelist)) will take precedence over locally/venv installed libs. This is important while developing for new numpy 2.0 libs.
- The `tests` subdir was moved, so remove from MANIFEST.in

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,7 +4,6 @@ exclude MANIFEST.in
 include CITATION.cff CHANGES.txt CREDITS.txt LICENSE.txt README.rst
 include pyproject.toml versioneer.py
 recursive-include src *.c *.h
-recursive-include tests *.py
 recursive-include shapely *.pxd *.pyx
 recursive-exclude shapely *.c
 include docs/*.rst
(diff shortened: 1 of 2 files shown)


jorisvandenbossche (approved):
Thanks for the cleanup, looks good to me!

Is the numpy insert instead of append of include dirs something they document/recommend on their side?

mwtoews:
I don't see any specific numpy docs around this, but locally on an ubuntu system I see this difference. Before this change it will always compile with numpy 1.x if installed on the system rather than numpy 2.x installed via pip. This threw me in a loop for a while when I kept seeing [this message](https://numpy.org/doc/stable/user/troubleshooting-importerror.html#downstream-importerror-or-attributeerror).

PR #2072, opened 2024-06-24
CI: Update macos-11 image to macos-13

mwtoews:
According to the 2024 GitHub Actions macOS runner image roadmap, macos-11 is removed. Macos-12 will be deprecated next month, so it could make sense to jump to macos-13.


jorisvandenbossche (approved):
Thanks!

PR #2078, opened 2024-06-27
DOC: LineStrings with zero length are invalid

void-rooster:
The validity test is meaningful for more than Polygons and MultiPolygons, as LineStrings with zero length can be created and are invalid. This PR corrects the manual.


mwtoews (approved):
Thanks for the improvements!

PR #2084, opened 2024-07-10
Drop Python 3.8 support, require Python 3.9+

EwoutH:
With the migration to Shapely 2.0 successfully completed and Python 3.13 soon coming in, this seems like a good moment to drop Python 3.8 support. It's almost [end-of-life](https://endoflife.date/python) and [SPEC 0](https://scientific-python.org/specs/spec-0000/) already recommended dropping Python 3.8 over a year ago.

So this PR bumps the required Python version to 3.9, removes 3.8 from the CI and test files.

Of course this will only affect future releases from the main branch, existing releases and releases from maintenance branches (like `maint-2.0`) stay compatible with Python 3.8 and can continued to be used.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -11,7 +11,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: 3.12
+          python-version: 3.11
 
       - name: Run black, flake8 and isort
         uses: pre-commit/action@v3.0.1
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -11,7 +11,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: 3.8
+          python-version: 3.12
 
       - name: Run black, flake8 and isort
         uses: pre-commit/action@v3.0.1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -119,7 +119,6 @@ jobs:
         env:
           CIBW_ARCHS: ${{ matrix.arch }}
           CIBW_SKIP: cp36-* pp* *musllinux* *-manylinux_i686
-          CIBW_TEST_SKIP: "cp38-macosx_arm64"
           CIBW_ENVIRONMENT_LINUX:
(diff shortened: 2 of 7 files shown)


EwoutH:
The linting error is a bit weird. It are multiple cases of flake8 throwing a `E231 missing whitespace after ','` (or after `':'`) [error](https://www.flake8rules.com/rules/E231.html). But all those cases are in the middle of strings, not actual code.

What I find strange, is why is it showing up _now_.

EwoutH:
Couldn't pinpoint what's happening exactly, but on Python 3.11, the flake8 linting job passes without problem, so let's use that for this PR. Probably somehow related to https://github.com/PyCQA/flake8/issues/1845.

The Windows error seems unrelated, please re-run that job.

EwoutH:
@mwtoews, since you did the previous two version bumps (in #1453 and #1802), could you review this PR?

Maybe we also want to update NumPy and GEOS minimum version, but that can also be done in another PR.

adamjstewart:
> What I find strange, is why is it showing up _now_.

Each new Python release contains subtle changes to the parser. Python 3.12 specifically contains changes to [f-string parsing](https://docs.python.org/3/whatsnew/3.12.html#pep-701-syntactic-formalization-of-f-strings), and may include other parser changes. This would mean that what was once a string is now parsed as a string containing an expression. The linter would then have access to this new information and provide more detailed feedback.

My suggestion would be to always use the latest Python release for linting and correct any new linter errors that pop up. Of course, you can do this in a different PR if you want to keep this one focused.

mwtoews:
Hi @EwoutH, apologies for the delay. We'll drop at least Python 3.8, but perhaps more. I'm not sure where a discussion should be held (there could already be an issue for it, I can't recall). Essentially, I'd like to see adoption for SPEC 0 but retain compatibility with older/supported GEOS versions. I'll add a review here soon.

EwoutH:
Would you like me to keep this PR open or can I close it?

EwoutH:
Closing. If a new PR is appreciated, let me know.

mwtoews:
I appreciate the effort and interest @EwoutH , particularly since development efforts in this project are slow moving. I'm going to start a version policy discussion sometimes this week to see what minimum Python versions should be supported.

PR #2089, opened 2024-07-13
ENH: add musllinux_x86_64 binary wheels

### LINKED ISSUE #1996: musllinux wheel missing

There should be a wheel for `musllinux`, to speed up CI build times significantly when using Alpine-based Python base images.

pip install shapely => Builds from source instead of using a pre-built wheel for musllinux

## Operating system

Alpine Linux, x86_64 

Until at least 2.0.3

mwtoews:
Closes #1996

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,6 +11,7 @@ jobs:
     environment:
       CIBUILDWHEEL: 1
       CIBW_BUILD: "cp*_aarch64"
+      CIBW_SKIP: "pp* *i686 cp37-musl* cp38-musl*"
       CIBW_MANYLINUX_AARCH64_IMAGE: ghcr.io/shapely/manylinux2014_aarch64_geos:3.11.4
       CIBW_MUSLLINUX_AARCH64_IMAGE: ghcr.io/shapely/musllinux_1_2_aarch64_geos:3.11.4
       CIBW_ENVIRONMENT: "PIP_PRE=1"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
       CIBUILDWHEEL: 1
       CIBW_BUILD: "cp*_aarch64"
       CIBW_MANYLINUX_AARCH64_IMAGE: ghcr.io/shapely/manylinux2014_aarch64_geos:3.11.4
-      CIBW_MUSLLINUX_AARCH64_IMAGE: ghcr.io/shapely/musllinux_1_1_x86_64_geos:3.11.4
+      CIBW_MUSLLINUX_AARCH64_IMAGE: ghcr.io/shapely/musllinux_1_2_aarch64_geos:3.11.4
       CIBW_ENVIRONMENT: "PIP_PRE=1"
       CIBW_TEST_REQUIRES: "pytest"
       CIBW_TEST_COMMAND: "pytest --pyargs shapely.tests"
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
Quickly looking at the logs, when it starts with the musllinux docker image, I see:

```
Running before_all...
  
      + /opt/python/cp38-cp38/bin/python -c 'import sys, json, os; json.dump(os.environ.copy(), sys.stdout)'
      + sh -c ./ci/install_geos.sh
  /project /project
  Using cached install /host/home/runner/work/_temp/geos-3.11.4
  /project
```

So it's not rebuilding GEOS in the musllinux image, which seems problematic. We probably need some logic in `install_geos.sh` to avoid doing that (include the architecture in the path?)

mwtoews:
The primary issue that I see that libgeos is built (and cached) using the github runner image, while cibuildwheel uses a different image. These images have different libraries, and are incompatible with each other.

---

I'm having some good success using [custom linux images](https://cibuildwheel.pypa.io/en/stable/options/#linux-image) that I've [published as packages](https://github.com/mwtoews?tab=packages). For example:
```Dockerfile
# docker build -t ghcr.io/shapely/musllinux_1_1_x86_64_geos:3.12.2 --build-arg GEOS_VERSION=3.12.2 musllinux_1_1_x86_64
# docker run -it --rm ghcr.io/shapely/musllinux_1_1_x86_64_geos:3.12.2

FROM quay.io/pypa/musllinux_1_1_x86_64:latest

ARG GEOS_VERSION=local
ENV GEOS_VERSION=${GEOS_VERSION}

RUN curl -OL --retry 5 https://download.osgeo.org/geos/geos-$GEOS_VERSION.tar.bz2
RUN tar xfj geos-$GEOS_VERSION.tar.bz2 && rm geos-$GEOS_VERSION.tar.bz2
RUN cmake -DCMAKE_BUILD_TYPE=Release -S geos-$GEOS_VERSION -B build
RUN cmake --build build -j 4
RUN cmake --install build
RUN cd build && ctest --output-on-failure
RUN rm -rf build geos-$GEOS_VERSION
```
I also have success with aarch64 via emulation, so have updated the circleCI configuration.

I'm able to push these images into GitHub packages/containers [here](https://github.com/orgs/shapely/packages) although they are currently "private", as I don't have necessary permissions to make them public. @jorisvandenbossche (or @sgillies) are you able to review and enable? The Dockerfiles for the images can be put into a subrepo too, which would be linked to the packages. But again, I don't have the credentials to create new repos.

mwtoews:
Good news, the linux binary wheels look good! I'm using my personal containers at `ghcr.io/mwtoews` for now, but I'd prefer to see `ghcr.io/shapely` be used for this PR when ready.

Bad news, some unrelated issue appears to have broken the macos-14 builds.

sgillies:
@mwtoews @jorisvandenbossche I'll make the images public and then tick the "public" option here if that's okay with you.

Untitled

jorisvandenbossche:
@mwtoews will try to take a closer look tomorrow, but just a quick note that in the pyogrio wheel building workflow, we build and cache custom images in github actions using `docker/build-push-action`, without the need to manually push some images somewhere.

mwtoews:
Thanks for pointing me to pyogrio's approach @jorisvandenbossche ! I see a similar approach could be used, which is much simpler with only libgeos to bundle.

jorisvandenbossche:
Yes, indeed for pyogrio the dockerfiles are quite complicated because it needs to build GDAL, as you showed above a customized manylinux dockerfile for adding GEOS can be simpler.

mwtoews:
Superseded by #2107

PR #2092, opened 2024-07-20
RLS/CI: upgrade GEOS versions to latest minor, add more to CI matrix

mwtoews:
This PR upgrades a few components for CI testing and the GEOS version used in binary wheels. It is similar to #2086 but for the `main` branch.

- <s>Add a "2024" CI matrix entry with numpy-2.0.0</s> (it's not so easy to insert another matrix entry for GEOS=3.12.2, so wait on this)
- Add test with macos-14 for an apple silicon runner
- Remove extra ignore for dateutil Python 3.12 warning in CI
- Fix properly catching ERRORLEVEL on Windows, but don't fail if tests fail (they are disabled for POSIX)
- Fix doctests to work with GEOS 3.12.2
- <s>Change GEOS install to use [`CMAKE_PLATFORM_NO_VERSIONED_SONAME=ON`](https://cmake.org/cmake/help/latest/variable/CMAKE_PLATFORM_NO_VERSIONED_SONAME.html), which will remove the soversion number from the internal library included with the binary wheel.</s>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ jobs:
     # https://circleci.com/product/features/resource-classes/
     resource_class: arm.medium
     environment:
-      GEOS_VERSION: 3.12.1
+      GEOS_VERSION: 3.12.2
       CIBUILDWHEEL: 1
       CIBW_BUILD: "cp*-manylinux_aarch64"
       CIBW_ENVIRONMENT_PASS_LINUX: "GEOS_VERSION GEOS_INSTALL GEOS_CONFIG LD_LIBRARY_PATH"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,7 +50,7 @@ jobs:
     name: Build ${{ matrix.arch }} wheels on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     env:
-      GEOS_VERSION: "3.12.1"
+      GEOS_VERSION: "3.12.2"
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-13, windows-2019]
         architecture: [x64]
-        geos: [3.8.4, 3.9.5, 3.10.6, 3.11.3, 3.12.1, main]
+        geos: [3.8.4, 3.9.5, 3.10.6, 3.11.4, 3.12.2, main]
         include:
           # 2019
           - python: 3.8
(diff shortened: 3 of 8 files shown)


jorisvandenbossche:
Changes look good!

> * Change GEOS install to use [`CMAKE_PLATFORM_NO_VERSIONED_SONAME=ON`](https://cmake.org/cmake/help/latest/variable/CMAKE_PLATFORM_NO_VERSIONED_SONAME.html), which will remove the soversion number from the internal library included with the binary wheel.

Is that considered better practice for .so's included in wheels?

mwtoews:
> Is that considered better practice for .so's included in wheels?

I'm not sure. From what I've seen, this option is selected for Android targets. And I suppose cibuildwheel is embedding these libraries in a binary wheel, so perhaps soname version numbers are unnecessary. Here are the macOS file differences:
```diff
- shapely/.dylibs/libgeos.3.11.3.dylib
+ shapely/.dylibs/libgeos.dylib
- shapely/.dylibs/libgeos_c.1.17.3.dylib
+ shapely/.dylibs/libgeos_c.dylib
```
and for Linux:
```diff
- shapely.libs/libgeos-f99bd675.so.3.11.3
+ shapely.libs/libgeos-f2240383.so
- shapely.libs/libgeos_c-da48124d.so.1.17.3
+ shapely.libs/libgeos_c-91b6a8c5.so
```
(the hash name is done by auditwheel to create a unique library name).
 
The counter-argument for keeping the version numbers is that we can clearly see GEOS 3.11.3 in the file name, which is perhaps convenient. I can be convinced to keep the former behaviour, since it hasn't made any trouble.

jorisvandenbossche:
> The counter-argument for keeping the version numbers is that we can clearly see GEOS 3.11.3 in the file name, which is perhaps convenient. I can be convinced to keep the former behaviour, since it hasn't made any trouble.

Yeah, and given this has been working fine, I would be inclined to not change it .. (unless there would be a good reason to do so)

mwtoews:
Let's keep the soversion filenames for now, as there is no compelling reason to do so.

This option would make sense in other scenarios, where symlinks are created to shared libs, e.g. libgeos.so -> libgeos.so.3.12.2. In this scenario, the symlinks get converted to regular file copies with wheel unpack/pack, creating larger whl sizes. But this scenario isn't currently happening.

---

Unrelated to this PR, a new error appears likely due to recent numpy-2.1dev changes:
```
shapely/creation.py:433: in multipoints
    return lib.create_collection(geometries, typ, out=out, **kwargs)
E   TypeError: ufunc 'create_collection' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

jorisvandenbossche (approved):
Thanks!

PR #2093, opened 2024-07-20
BUG: fix base geometry "project" to return Python float

mwtoews:
Similar to #2074, this PR restores [`line.project(point)`](https://shapely.readthedocs.io/en/stable/manual.html#object.project) to return a Python float, and not a NumPy float64 scalar.  It similarly changes `line_locate_point` as well.

This is more obvious with NumPy 2.0:
```pycon
>>> import shapely
>>> line = shapely.from_wkt("LINESTRING (0 0, 1 0)")
>>> point = shapely.from_wkt("POINT(0 1)")
>>> line.project(point)
np.float64(0.0)
>>> line.line_locate_point(point)
np.float64(0.0)
```
this PR changes the functions to return Python float(0.0) instead (or `0.0`).

Tests are improved to check other scalar results, and test `project` and `interpolate` function aliases too.

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -834,7 +834,9 @@ class BaseGeometry(shapely.Geometry):
 
         Alias of `project`.
         """
-        return shapely.line_locate_point(self, other, normalized=normalized)
+        return _maybe_unpack(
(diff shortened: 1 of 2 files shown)


jorisvandenbossche (approved):
Thanks, looks good!

PR #2095, opened 2024-07-22
MAINT: replace black, flake8 and isort by ruff

theroggy:
Same approach as used in geopandas and pyogrio.

Didn't enable pydocstyle yet to keep the PR somewhat managable.


shapely/tests/geometry/test_equality.py
```diff
@@ -232,16 +235,16 @@ def test_comparison_not_supported():
...
-        geom1 < geom2
+        assert geom1 < geom2
 
     with pytest.raises(TypeError, match="not supported between instances"):
-        geom1 >= geom2
+        assert geom1 >= geom2
 
     with pytest.raises(TypeError, match="not supported between instances"):
-        geom1 <= geom2
+        assert geom1 <= geom2
```
mwtoews, line 247: Plenty of good fixes like this in here.

shapely/tests/test_set_operations.py
```diff
@@ -334,8 +335,8 @@ def test_coverage_union_non_polygon_inputs(geom_1, geom_2):
         def effective_geom_types(geom):
             if hasattr(geom, "geoms") and not geom.is_empty:
                 gts = set()
-                for geom in geom.geoms:
-                    gts |= effective_geom_types(geom)
+                for part in geom.geoms:
+                    gts |= effective_geom_types(part)
```
mwtoews, line 339: More good catches!

mwtoews (approved):
Thanks for PR! I've recently been switching other projects to ruff too, but you have beat me to it!

shapely/tests/geometry/test_equality.py
jorisvandenbossche: Why is this actually needed? We should never get to that `assert`, as the code is meant to raise an error (I don't think I ever add explicit asserts when using `with pytest.raises` contexts)

shapely/tests/test_ragged_array.py
```diff
@@ -317,8 +318,10 @@ def test_multilinestrings():
 def test_multipolygons():
     arr = shapely.from_wkt(
         [
-            "MULTIPOLYGON (((35 10, 45 45, 15 40, 10 20, 35 10), (20 30, 35 35, 30 20, 20 30)))",
-            "MULTIPOLYGON (((40 40, 20 45, 45 30, 40 40)),((20 35, 10 30, 10 10, 30 5, 45 20, 20 35),(30 20, 20 15, 20 25, 30 20)))",
+            "MULTIPOLYGON (((35 10, 45 45, 15 40, 10 20, 35 10), "
+            "(20 30, 35 35, 30 20, 20 30)))",
+            "MULTIPOLYGON (((40 40, 20 45, 45 30, 40 40)),"
+            "((20 35, 10 30, 10 10, 30 5, 45 20, 20 35),(30 20, 20 15, 20 25, 30 20)))",
             "MULTIPOLYGON EMPTY",
```
jorisvandenbossche, line 325: For cases like this a `  # noqa: E501` might be nicer?

jorisvandenbossche (commented):
Thanks for doing this!

shapely/tests/geometry/test_equality.py
theroggy: Not really needed in this specific instance, but in most cases this is a bug, so ruff gives a linting error on it...

shapely/tests/test_ragged_array.py
theroggy: Same to me... I'll change it...

jorisvandenbossche (approved):
Thanks!

PR #2097, opened 2024-07-27
DOC: Replace "pygeos.get_parts" by "shapely.get_parts" in docstring of `from_geojson`

hamdanal:


jorisvandenbossche (approved):
Good catch, thanks for the PR!

Issue #2098, opened 2024-07-28
ufunc 'create_collection' not supported for the input types


mwtoews:
This is a recent issue identified using nightly numpy binary wheels with the main branch of shapely.

The error looks like this:
```
=========================== test session starts ============================
platform linux -- Python 3.10.12, pytest-8.2.2, pluggy-1.5.0
benchmark: 4.0.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /home/mtoews/src/shapely/shapely
configfile: setup.cfg
plugins: syrupy-4.6.1, anyio-4.4.0, dotenv-0.5.2, flaky-3.8.1, hypothesis-6.108.5, cov-5.0.0, xdist-3.6.1, benchmark-4.0.0
collected 9 items / 1 error                                                

================================== ERRORS ==================================
__________ ERROR collecting shapely/tests/geometry/test_coords.py __________
shapely/tests/geometry/test_coords.py:5: in <module>
    from shapely.tests.common import line_string, line_string_z, point, point_z
shapely/tests/common.py:26: in <module>
    multi_point = shapely.MultiPoint([(0, 0), (1, 2)])
shapely/geometry/multipoint.py:62: in __new__
    return shapely.multipoints(subs)
shapely/decorators.py:77: in wrapped
    return func(*args, **kwargs)
shapely/creation.py:433: in multipoints
    return lib.create_collection(geometries, typ, out=out, **kwargs)
E   TypeError: ufunc 'create_collection' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
========================= short test summary info ==========================
ERROR shapely/tests/geometry/test_coords.py - TypeError: ufunc 'create_collection' not supported for the input types,...
!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!
============================= 1 error in 0.41s =============================
```
I've bisected the issue to https://github.com/numpy/numpy/commit/f3b669d391cfe660b064af052973b9bff9372dd8 from https://github.com/numpy/numpy/pull/26905 (ping @seberg) which is related to [NEP 50](https://numpy.org/neps/nep-0050-scalar-promotion.html).

mwtoews:
Here is a short reproducer:
```pycon
>>> import shapely
[...]

PR #2099, opened 2024-07-29
Fix compatibility with numpy 2.1 dev to cast GeometryType to int as ufunc input

jorisvandenbossche:
Fixes https://github.com/shapely/shapely/issues/2098


mwtoews (approved):
Agree this is the correct cast type, and is consistent with other ufunc enums.

After merge, this should be cherry-picked into `maint-2.0` to resolve CI failures too.

EwoutH:
It's a bit weird that NumPy 2.1 breaks stuff right, as a minor release?

EwoutH:
If all else is not feasible, could we set NumPy to <2.1 on the 2.0 branch and sprint a release out?

https://github.com/shapely/shapely/blob/2a839056e91ec7d35e108404d0bfa16d9f3f8038/pyproject.toml#L46

orthorhombic:
NumPy just released 2.1.0, and I noticed this impacting my code. It would be nice to get a release out.

@EwoutH - Impressive that you guessed "NumPy could release 2.1 at any moment" <4 hours before they released it.

EwoutH:
What’s the strangest is that @jorisvandenbossche added this to the [2.0.6](https://github.com/shapely/shapely/milestone/58) milestone over two weeks ago, but 2.0.6 was never released.

axelande:
Will 2.0.6 be released or what is missing from that being released?

jorisvandenbossche:
> What’s the strangest is that @jorisvandenbossche added this to the [2.0.6](https://github.com/shapely/shapely/milestone/58) milestone over two weeks ago, but 2.0.6 was never released.

That's because milestones are used for future releases ;)

We were planning to cut a 2.0.6 release if needed (if numpy 2.1 was released before a next shapely release), which is the case now, so will try to do that the coming hours.

jorisvandenbossche:
Shapely 2.0.6 is released (it's on pypi, conda-forge can still take a few hours)

PR #2105, opened 2024-08-11
TST/CI: enable testing Python 3.13, NumPy 2.1, GEOS 3.13

mwtoews:
Follows #2086 to finally enable testing of Python 3.13 with NumPy 2.1. Currently, both these are available as release candidates. Also note that NumPy 2.0.x does not have any binary wheels out for Python 3.13.

I'm not sure when GEOS 3.13 is ready, but this would ideally be put under the "2024" item too when it's out.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,10 +39,10 @@ jobs:
             matplotlib: true
             doctest: true
             extra_pytest_args: "-W error"  # error on warnings
-          # 2024 -- wait until Python 3.13 with binary wheels for numpy are available
-          # - python: "3.13"
-          #   geos: 3.13.0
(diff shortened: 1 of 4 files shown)


EwoutH:
Looks like it needs pyproj wheels for 3.13, which is blocked on https://github.com/pyproj4/pyproj/issues/1425.

mwtoews:
This PR now adjusts a few tests for GEOS 3.13.0beta1; see [changes](https://github.com/libgeos/geos/blob/3.13.0beta1/NEWS.md).

The more significant change is with RelateNG, which is more tolerant with invalid geometries.

Changes to STRTree are also more tolerant with invalid geometries.

jorisvandenbossche (approved):
Looks good, thanks!

PR #2106, opened 2024-08-12
ENH: add function minimum_clearance_line

theroggy:
ref #1570

### DIFF
--- a/shapely/constructive.py
+++ b/shapely/constructive.py
@@ -23,6 +23,7 @@ __all__ = [
     "build_area",
     "make_valid",
     "normalize",
+    "minimum_clearance_line",
     "node",
     "point_on_surface",
     "polygonize",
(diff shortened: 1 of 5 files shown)


shapely/constructive.py
```diff
@@ -595,6 +596,35 @@ def make_valid(geometry, method="linework", keep_collapsed=True, **kwargs):
         raise ValueError(f"Unknown method: {method}")
 
 
+@multithreading_enabled
+def minimum_clearance_line(geometry, **kwargs):
+    """Return a LineString whose endpoints define the minimum clearance.
+
```
jorisvandenbossche, line 602: Could you add an extended description paragraph explaining a bit more what "minimum clearance" exactly is? 
In the GEOS code itself I don't find much more than what you wrote above, but something about distance one of the vertices can be moved to become invalid.

jorisvandenbossche (commented):
Thanks @theroggy!

For the API, I was checking PostGIS and they seem to expose two versions: one that returns just the distance ([`ST_MinimumClearance`](https://postgis.net/docs/ST_MinimumClearance.html), which might often be what one wants), and one that returns the line ([`ST_MinimumClearanceLine`](https://postgis.net/docs/ST_MinimumClearanceLine.html)).

So we need to decide whether we want to do that here as well or not.

theroggy:
> Thanks @theroggy!
> 
> For the API, I was checking PostGIS and they seem to expose two versions: one that returns just the distance ([`ST_MinimumClearance`](https://postgis.net/docs/ST_MinimumClearance.html), which might often be what one wants), and one that returns the line ([`ST_MinimumClearanceLine`](https://postgis.net/docs/ST_MinimumClearanceLine.html)).
> 
> So we need to decide whether we want to do that here as well or not.

The version that just returns the distance is already available in shapely: https://shapely.readthedocs.io/en/stable/reference/shapely.minimum_clearance.html#shapely.minimum_clearance

jorisvandenbossche:
Whoops, sorry ;) Was going through it a bit too quickly, missed that the function being added is already called `.._line` and that we already have the basic form!

shapely/constructive.py
jorisvandenbossche:
```suggestion

    A geometry's "minimum clearance" is the smallest distance by which
    a vertex of the geometry could be moved to produce an invalid geometry.

```

shapely/constructive.py
```diff
@@ -595,6 +596,35 @@ def make_valid(geometry, method="linework", keep_collapsed=True, **kwargs):
...
+    If the geometry has no minimum clearance, an empty LineString will be returned.
+
```
jorisvandenbossche, line 604:
```suggestion

   .. versionadded:: 2.1.0
```

jorisvandenbossche (approved):
Thanks!

PR #2107, opened 2024-08-16
ENH: use docker to build wheels for linux

### LINKED ISSUE #1996: musllinux wheel missing

mwtoews:
This a rewrite of parts of the cibuildwheel procedure for Linux, using local docker containers. Parts of it are inspired from pyogrio.

Some notes:

- Upgrade a few components, such as upload/download artifacts, cibuildwheel (for cp313)
- Windows and macOS aspects remain the same, but are separated from Linux
- For macOS there are recent failures due to some clang default changes. These are side-stepped with `CFLAGS="-ferror-limit=0 -Wno-error=incompatible-function-pointer-types"`
- Adds musllinux images from pypa (same versions used by numpy)
- Use QEMU to build aarch64 wheels - so I suppose CircleCI could be removed

This PR currently targets `maint-2.0`.

Supersedes #2089
Closes #1996


.github/workflows/release.yml
```diff
@@ -37,43 +41,121 @@ jobs:
...
+  build_wheels_linux:
+    name: Build wheels on Linux with ${{ matrix.baseimage }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: x86_64
+            baseimage: manylinux2014_x86_64
+            basetag: 2024.08.12-1
```
jorisvandenbossche, line 60: Do you anticipate this to be different across images? Otherwise we could also (or at least for now) add this as top-level matrix entry to not have to repeat it here?

.github/workflows/release.yml
```diff
@@ -37,43 +41,121 @@ jobs:
...
+          buildkitd-flags: --debug
+
+      - name: Build Docker image with GEOS
+        # using build-push-action (without push) to make use of cache arguments
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: ci/Dockerfile
+          platforms: linux/${{ matrix.arch }}
+          tags: ${{ matrix.baseimage }}_geos:${{ env.GEOS_VERSION }}
```
jorisvandenbossche, line 101: Including the GEOS_VERSION here is a way to ensure it actually gets rebuilt when we update that number?

.github/workflows/release.yml
```diff
@@ -37,43 +41,121 @@ jobs:
...
+        env:
+          CIBW_ARCHS: ${{ matrix.arch }}
+          CIBW_BUILD: ${{ matrix.cibw_build }}
+          CIBW_SKIP: pp* *i686 cp37-musl* cp38-musl*
+          CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014_x86_64_geos:${{ env.GEOS_VERSION }}
+          CIBW_MUSLLINUX_X86_64_IMAGE: musllinux_1_1_x86_64_geos:${{ env.GEOS_VERSION }}
+          CIBW_MANYLINUX_AARCH64_IMAGE: manylinux2014_aarch64_geos:${{ env.GEOS_VERSION }}
+          CIBW_MUSLLINUX_AARCH64_IMAGE: musllinux_1_2_aarch64_geos:${{ env.GEOS_VERSION }}
+          CIBW_ENVIRONMENT_LINUX:
+            PIP_PRE=1
```
jorisvandenbossche, line 128: I know this was copied from below, but not sure if this is still needed?

jorisvandenbossche (commented):
@mwtoews thanks for working on this! 

I do wonder if it would be better to directly target this to the main branch instead (it's a somewhat bigger change to the build setup, and it also grows the matrix of wheel builds considerably, and on the maint-2.0 branch we maintain more Python versions which increases it quadratic)

jorisvandenbossche:
FWIW, in my effort to get a 2.0.6 release out (and without having noticed this PR, although this one would have been more difficult to move forward that quickly), I essentially already committed (and backported) two aspects: update cibuildwheel to get Python 3.13 wheels (https://github.com/shapely/shapely/pull/2103) and the `-Wno-error=incompatible-function-pointer-types` for the macOS failures (https://github.com/shapely/shapely/pull/2114, I am not sure if the `-ferror-limit=0` that you added additionally is still needed)

.github/workflows/release.yml
mwtoews: This could probably be a top-level item, as all images appear to have the same tag. It's better to tag than it is to have `latest` for stability and hopefully caching (I'm still observing how image caching works in GHA).

.github/workflows/release.yml
mwtoews: It's perhaps a legacy of how I was handling this in #2089. It can be `latest` as it's a local image (same as pyogrio).

jorisvandenbossche: Yes, but I remember from pyogrio that sometimes I just need to do a dummy edit to the dockerfile (or update the base image's tag) just to ensure we rebuild the docker image to get an updated GDAL version. SO it _might_ be that including the geos version here would help with that (i.e. to avoid it still uses the cached image when updating the GEOS version, as that would then not actively test it with the update version)

mwtoews:
Sure, this can target `main` instead, as the higher number of changes more out-of-scope for maintenance. The growing number of binary wheel combinations is concerning, so I have a plan to start a discussion this week to adopt [SPEC 0](https://scientific-python.org/specs/spec-0000/) to limit the number of Python versions.

.github/workflows/release.yml
mwtoews: moved to `env.basetag`

.github/workflows/release.yml
mwtoews: ok, keep this tag for now. I'm still exploring the caching for these actions, as the qemu matrix items take over an hour to build the images.

mwtoews:
This PR is now updated to target `main` which now has cp39-cp313 or five versions per os/arch combo (eight), so there are 40 binary wheels.

mwtoews:
@jorisvandenbossche does the gha cache mechanism work only after it's merged (and not in a PR?) For instance, in a recent [pyogrio run](https://github.com/geopandas/pyogrio/actions/runs/10623640468/job/29450459921)  I see messages like "importing cache manifest from gha" / "CACHED" which is encouraging. But I didn't see any caching messages after a trivial commit in this PR, and this was after a successful commit (2951dfdea3eaf952126fe9be56a33452540374d0) that took nearly two hours to complete. This PR is also ready for review.

mwtoews:
Wait, I was partially wrong, caching seemed to work for musllinux_1_2_aarch64, but for some reason not for manylinux2014_aarch64. The image initially took 1h15 min to build, then the wheels with testing took 30 mins. So this means a "normal" run where a cached image is used should take 30 mins. I can disable testing for aarch64 if this is too long.

.github/workflows/release.yml
```diff
@@ -119,12 +184,6 @@ jobs:
         env:
           CIBW_ARCHS: ${{ matrix.arch }}
           CIBW_SKIP: pp* *musllinux* *-manylinux_i686
```
jorisvandenbossche, line 186: Given this is now only mac+win, the musllinux and manylinux skips can probably be removed

.github/workflows/release.yml
```diff
@@ -119,12 +184,6 @@ jobs:
...
-          CIBW_TEST_SKIP: "cp38-macosx_arm64"
-          CIBW_ENVIRONMENT_LINUX:
-            GEOS_VERSION=${{ env.GEOS_VERSION }}
-            GEOS_INSTALL=/host${{ runner.temp }}/geos-${{ env.GEOS_VERSION }}
-            GEOS_CONFIG=/host${{ runner.temp }}/geos-${{ env.GEOS_VERSION }}/bin/geos-config
-            LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/host${{ runner.temp }}/geos-${{ env.GEOS_VERSION }}/lib
```
jorisvandenbossche, line 127: Setting those env variables apparently don't seem to be needed in practice for linux?

pyproject.toml
```diff
@@ -71,6 +71,12 @@ include = ["shapely", "shapely.*"]
 [tool.setuptools.package-data]
 "shapely" = ["*.pxd"]
 
+[tool.cibuildwheel]
+skip = ["pp*", "*_i686", "*_ppc64le", "*_s390x"]
```
jorisvandenbossche, line 75: Not too important, but just wondering to what extent it is actually needed to list eg `"*_i686` given we are running with explicit many/musllinux docker images (I would not expect that it tries to build such wheels by default in a manylinux2014_x86-64 image)

.github/workflows/release.yml
jorisvandenbossche: (that's maybe because in the docker image we are just installing GEOS in the expected system default location?)

jorisvandenbossche (commented):
Looking good!

jorisvandenbossche:
Could you push a temp trivial commit once more to have another check how caching works. Given only 1 out of 4 builds used the cache, I am wondering if we need to do something to make the cache specific per job in the matrix.

jorisvandenbossche:
Maybe check https://github.com/docker/build-push-action/issues/546, they talk about a similarly sounding image and are using `scope` in the cache-to/from

jorisvandenbossche:
Otherwise a local cache (with plain github actions caching) might also work: https://docs.docker.com/build/ci/github-actions/cache/#local-cache

jorisvandenbossche:
> I can't say from the docs whether this is needed, but might be something to try

Based on https://docs.docker.com/build/cache/backends/gha/#scope (from https://stackoverflow.com/questions/76227392/github-build-push-action-and-gha-cache-should-i-set-the-github-ref-name-in-the), it might actually indeed be needed to set the scope to a unique name for each matrix entry

.github/workflows/release.yml
mwtoews: These are not needed in an image since it's installed to a "default" path  `/usr/local` which is automatically discovered.

pyproject.toml
mwtoews: I think it makes it more explicit to not build these. It changes the output from (e.g.) a command `cibuildwheel --print-build-identifiers`

mwtoews:
Thanks for finding the `scope` detail, it seems to describe what is needed here. I'll wait for this one commit to complete before adding another commit to see if it works...

mwtoews:
Caching is now working as expected! I've tested skipping tests on aarch64 wheels, and the times are 20 mins without tests compared to 30 mins with tests. I might re-enable tests, as it's not significantly different.

And this PR also removes CircleCI, assuming the new workflows here are a suitable replacement.

jorisvandenbossche:
> I've tested skipping tests on aarch64 wheels, and the times are 20 mins without tests compared to 30 mins with tests. I might re-enable tests, as it's not significantly different.

One thing I am wondering is whether we should maybe only run those slower builds on main or when explicitly triggered (eg with a specific slug in the PR title or commit message). Since they are still by far the slowest jobs in the full CI run now ..

mwtoews:
Perhaps I'm too quick to dismiss CircleCI. It has [wheel-build runtimes of 5-10 mins](https://app.circleci.com/pipelines/github/shapely/shapely) on native ARM hardware. I'll explore their [docs](https://circleci.com/docs/) to see if a similar image-building-caching workflow can be implemented.

jorisvandenbossche:
Even without caching, it is faster, so I don't think that is necessarily a big deal. (and maybe by the end of the year we will have native arm images on github actions .. https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/#get-started-using-arm-hosted-runners-today)

mwtoews:
Great to hear that native arm64 runners will be available before the end of the year! I think we should keep the existing and unmodified CircleCI for now, which only builds manylinux, and commented-out the GHA matrix entries for aarch64 wheels. Adding musllinux_1_2_aarch64 can wait.

jorisvandenbossche (approved):
Can you add a note to CHANGES.txt that there are now musllinux wheels?

jorisvandenbossche:
Thanks @mwtoews!

PR #2112, opened 2024-08-19
TST: adjust for GEOS 3.13.0beta1

mwtoews:
See [changes](https://github.com/libgeos/geos/blob/3.13.0beta1/NEWS.md) for GEOS 3.13.0beta1.

The more significant change is with RelateNG, which is more tolerant with invalid geometries.

Changes to STRTree are also more tolerant with invalid geometries.


mwtoews:
This is better suited to add into #2112 2105

Issue #2081, opened 2024-07-03
Fails to build with GCC 14


sebastic:
As reported in [Debian Bug #1075423](https://bugs.debian.org/1075423) shapely fails to build with GCC 14:
```
src/ufuncs.c: In function ‘polygonize_func’:
src/ufuncs.c:2205:51: error: passing argument 2 of ‘GEOSPolygonize_r’ from incompatible pointer type [-Wincompatible-pointer-types]
 2205 |     GEOSGeometry* ret_ptr = GEOSPolygonize_r(ctx, geoms, n_geoms);
      |                                                   ^~~~~
      |                                                   |
      |                                                   GEOSGeometry ** {aka struct GEOSGeom_t **}
```

See also: https://gcc.gnu.org/gcc-14/porting_to.html#incompatible-pointer-types

sebastic:
gcc-14 is the default compiler in Debian unstable now, the use of `-Wno-error=incompatible-pointer-types` is required until a proper fix for this issue is available.

jorisvandenbossche:
The problem is that the second argument of `GEOSPolygonize_r` is defined as `const GEOSGeometry * const geoms[]`, and not `GEOSGeometry** geoms` (as it is done in some other functions taking an array of geometries, and how we create and pass it in our code).

Not directly sure about the best C way to resolve this.


https://github.com/shapely/shapely/blob/9a8bf73ecf2ef51d684b5866ab5718220b402640/src/ufuncs.c#L2275-L2299

musicinmybrain:
This should already be fixed in `main` by https://github.com/shapely/shapely/pull/1945, which was merged on 2024-12-05 as 05455886750680728dc751dc5888cd02086d908e but never cherry-picked over to `maint-2.0`.

jorisvandenbossche:
Good catch, thanks @musicinmybrain. We should indeed ideally have backported that.

jorisvandenbossche:
> Good catch, thanks @musicinmybrain. We should indeed ideally have backported that.

I did that now for the case we would have to do another 2.0.x bug-fix release, we would at least not forget to include it

jorisvandenbossche:
Given that this is fixed, at least for the upcoming 2.1 release, closing this issue.

Issue #1983, opened 2024-02-08
2.0.2: build fails with `[-Wincompatible-pointer-types]` error


kloczek:
With new glibc is possible to build with `-Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS` which turns some warnings into errors.
<details>

```console
+ cd shapely-2.0.2
+ ASMFLAGS='-m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -O2 -g -grecord-gcc-switches -pipe -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -fdata-sections -ffunction-sections -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -flto=auto -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS -Wall -Werror=format-security'
+ CFLAGS='-m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -O2 -g -grecord-gcc-switches -pipe -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -fdata-sections -ffunction-sections -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -flto=auto -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS -Wall -Werror=format-security'
[...]

PR #2114, opened 2024-08-19
BLD: suppress 'incompatible-function-pointer-types' error for clang>=16

jorisvandenbossche:
Starting with clang 16 ([announcement](https://discourse.llvm.org/t/clang-16-notice-of-potentially-breaking-changes/65562)), it started to default `-Wincompatible-function-pointer-types` to be on. This causes a bunch of hard errors in the wheel build for MacOS arm64 on the macos-14 image, like:

```
         src/ufuncs.c:213:47: error: incompatible function pointer types initializing 'PyUFuncGenericFunction' (aka 'void (*)(char **, long *, long *, void *)') with an expression of type 'void (*)(char **, const npy_intp *, const npy_intp *, void *)' (aka 'void (*)(char **, const long *, const long *, void *)') [-Wincompatible-function-pointer-types]
        static PyUFuncGenericFunction Y_b_funcs[1] = {&Y_b_func};
                                                      ^~~~~~~~~
```

In this case it seems to be the issue of `const long *` vs `long *`.

This could probably be fixed "properly", but for the time being (given this has been done like this for a long time and didn't give any problems, and to get CI green right now), this PR is suppressing those errors by passing `-Wno-error=incompatible-function-pointer-types`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -130,6 +130,7 @@ jobs:
             GEOS_CONFIG=${{ runner.temp }}/geos-${{ env.GEOS_VERSION }}/bin/geos-config
             MACOSX_DEPLOYMENT_TARGET=10.9
             CMAKE_OSX_ARCHITECTURES='${{ matrix.cmake_osx_architectures }}'
+            CFLAGS='-Wno-error=incompatible-function-pointer-types'
           CIBW_ENVIRONMENT_WINDOWS:
             GEOS_INSTALL='${{ runner.temp }}\geos-${{ env.GEOS_VERSION }}'
             GEOS_LIBRARY_PATH='${{ runner.temp }}\geos-${{ env.GEOS_VERSION }}\lib'


jorisvandenbossche:
Given this is required to get the wheels for macos arm64 building, and I would like to do a release today (to fix the numpy 2.1 compat issue), I am going to merge this myself quickly. But any additional review is certainly still welcome, we can fix this better later if needed.

jorisvandenbossche:
@mwtoews I am only now taking a look at https://github.com/shapely/shapely/pull/2107, and I see you had the same idea of passing `-Wno-error=incompatible-function-pointer-types`

mwtoews:
Thanks for pushing this along, and indeed it seems that `-ferror-limit=0` is not needed as I had in the other PR. Suppressing the warning/error for maint-2.0 branch is fine.

And agree that a better fix for the main branch is to resolve the warnings, xref #1983 and #2081.

musicinmybrain:
This should already be fixed in `main` by https://github.com/shapely/shapely/pull/1945, which was merged on 2024-12-05 as 05455886750680728dc751dc5888cd02086d908e but never cherry-picked over to `maint-2.0`.

musicinmybrain:
Ah, you’re right! I didn’t read closely enough. That commit should fix #2081, but not this issue.

Issue #2111, opened 2024-08-19
Serialization of `POINT EMPTY` to GeoJSON should not raise


Oreilles:
This issue a copy of https://github.com/pygeos/pygeos/issues/464, I'm creating both as I don't know how the two projects are linked.

Serialization of empty points to GeoJSON raise `ValueError: GeoJSON output of empty points is currently unsupported`.

This tracks back to an [old issue](https://trac.osgeo.org/geos/ticket/1139) in GEOS that was closed 3 years ago by [this commit](https://github.com/libgeos/geos/commit/a51060c15ac2d8813e60de33bbc0ce0d07cc0d3b):

In shapely, these lines were responsible for checking is the geometry being serialized was an empty point, and raising an error if it was:
https://github.com/shapely/shapely/blob/775f7792ecd945fa30b3f98b46d99f996e275f00/src/ufuncs.c#L3562

This check should be removed, as GEOS can now perfectly handle empty points i/o for GeoJSON.

jorisvandenbossche:
> This issue a copy of [pygeos/pygeos#464](https://github.com/pygeos/pygeos/issues/464), I'm creating both as I don't know how the two projects are linked.

We should more clearly archive the `pygeos` repo, but so all code of pygeos has been merged into shapely, and all development happens here.

PR #2118, opened 2024-08-21
BUG: fix to_geojson with empty point

Oreilles:
Fixes https://github.com/shapely/shapely/issues/2111

Serialization of empty points in GEOS has been fixed in [3.10.2](https://github.com/libgeos/geos/blob/7cb33c988bb0115f4fe107785570ae5786ed2401/NEWS#L73), so I updated the C code and tests to handle that.

### DIFF
--- a/shapely/tests/test_io.py
+++ b/shapely/tests/test_io.py
@@ -1254,9 +1254,7 @@ def test_to_geojson_exceptions():
         shapely.to_geojson(1)
 
 
-@pytest.mark.skipif(
-    (3, 10, 2) <= shapely.geos_version < (3, 10, 0), reason="3.10.2 <= GEOS < 3.10"
-)
+@pytest.mark.skipif(shapely.geos_version < (3, 10, 0), reason="GEOS < 3.10")
 @pytest.mark.parametrize(
     "geom",
     [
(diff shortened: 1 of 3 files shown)


Oreilles:
```
      /home/runner/work/shapely/shapely/geosinstall/geos-3.13.0beta1/include/geos_c.h:67:28: error: invalid suffix "beta1" on integer constant
         67 | #define GEOS_VERSION_PATCH 0beta1
            |                            ^~~~~~
      src/ufuncs.c:3562:12: note: in expansion of macro ‘GEOS_VERSION_PATCH’
       3562 |       #if (GEOS_VERSION_PATCH < 2)
```

That is unfortunate. Is there a way to use a compiler directive to ensure the minor version is lower than 2 even in the case it not a number ? Or maybe we can just remove this check alltogether, and just expect developpers to specifically avoid `3.10.0` and `3.10.1`, any prior/later version being fine.

mwtoews:
There is a long-ish workaround [here](https://github.com/shapely/shapely/blob/cf4aebf74c73316e7e599c21068cfb9d156ca313/src/lib.c#L64-L71) to extract the integer part from `GEOS_VERSION_PATCH`.

Oreilles:
After a bit of fiddling I'm starting to think that what I'm trying to do is impossible using only macros (which is probably why `int geos_version_patch_int = GEOS_VERSION_PATCH_STR[0] - '0'` is written as an instruction and not a macro).

So I'm not really sure how to handle that...

shapely/tests/test_io.py
```diff
@@ -1260,9 +1260,13 @@ def test_to_geojson_exceptions():
...
-    # Pending GEOS ticket: https://trac.osgeo.org/geos/ticket/1139
-    with pytest.raises(ValueError):
-        assert shapely.to_geojson(geom)
+    if shapely.geos_version < (3, 10, 2):
+        # Pending GEOS ticket: https://trac.osgeo.org/geos/ticket/1139
+        with pytest.raises(ValueError):
+            assert shapely.to_geojson(geom)
+        return
+    # Else, just confirm that the operation succeed
+    shapely.to_geojson(geom)
```
jorisvandenbossche, line 1269: Could we test that it roundtrips correctly?

jorisvandenbossche:
Personally, I think it is fine to not care too much about the patch version, and simply assume that if someone is using GEOS 3.10, they are using the latest patch release of that series (that's also what we do in our tests, in the end)

shapely/tests/test_io.py
Oreilles: Sure, just changed it.

Oreilles:
Alright 👍

jorisvandenbossche:
And can you maybe also add a brief sentence in CHANGES.txt ?

Oreilles:
All done

jorisvandenbossche:
Thanks @Oreilles!

Issue #2119, opened 2024-08-21
Version policy


mwtoews:
Discussions with regards to **Python version** support seems to pop up regularly. My recommendation is to adopt Python scientific community’s [SPEC 0](https://scientific-python.org/specs/spec-0000/) to determine the supported versions that should be supported in `main`. This time-based policy also has recommended schedules for supporting NumPy and matplotlib versions too.

The motivation for adopting SPEC 0 for Python version is to reduce maintenance burden, and have fewer combinations of binary wheels. Too many Python version combinations chew up CI time and PyPI disk space (particularly when more combinations of binary wheels are considered, e.g. #1996).

---

Related in this discussion is the minimum supported **GEOS version**, most recently discussed with #1000. Refer to [GEOS RFC 11 - GEOS Version and EOL Policy](https://libgeos.org/project/rfcs/rfc11/). Personally, I think it's fine that shapely to follow this GEOS version policy, as shapely has already done the work to support several GEOS versions. Furthermore, it's conceivable that users may need the behavior of a supported GEOS version. Supporting several GEOS versions does add a bit more maintenance burden and CI time, but does not add extra PyPI disk space since only one GEOS version is added to binary wheels.

---

And lastly, upon community agreement of version policies, how should this be visible to contributors and maintainers? In a root document (e.g. [POLICY.md](https://github.com/esp8266/Arduino/blob/ccea72823ac50290bc05c67350d2be6626e65547/POLICY.md)) or somewhere in the regular docs (e.g. [for JAX](https://jax.readthedocs.io/en/latest/deprecation.html))? There are a few other policy items that should be added the document, e.g. #1924.

mwtoews:
Oh, and I should mention that the implementations of adopting my recommendations would set the minimum Python version to 3.10 and minimum GEOS version to 3.9. This means dropping support for Python 3.8 and 3.9, also GEOS 3.8.

The versions that actually make it to the [2.1.x](https://github.com/shapely/shapely/milestone/52) series depend on when we release it (#2065).

martinfleis:
+1 on adoption of SPEC0. It seems to be the way forward for a lot of projects in the ecosystem so it only makes sense to be in sync. Given the sparse release cycle of shapely, it may mean some larger jumps in supported versions between releases but I don't think that is an issue.

I also agree on following GEOS' RFC.

> how should this be visible to contributors and maintainers

[...]

PR #2124, opened 2024-08-25
MAINT: Drop GEOS 3.8

mwtoews:
GEOS 3.8.4 reached it's EOL on 2023/11/12, thus it should be dropped for shapely's main branch.

The CI test matrix is substituted with GEOS 3.9.0 for now, as the number of combinations need to also include Python versions too (see #2119).

Many of the tests are simplified to remove conditionals.

<s>The C lib also contains some important corrections, which I may bring as a backport for maint-2.0.</s> (Update: the corrections only apply to main, thus should not be backported). There were some partially removed GEOS 3.7 features from #1885 that need refining, e.g. the def for `GEOS_SINCE_3_8_0` was removed, but there were still active blocks that used that def, now handled in this PR.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,11 +14,11 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-13, windows-2019]
         architecture: [x64]
-        geos: [3.9.0, 3.9.5, 3.10.6, 3.11.4, 3.12.2, 3.13.0beta1, main]
+        geos: [3.9.5, 3.10.6, 3.11.4, 3.12.2, 3.13.0beta1, main]
         include:
           # 2019
           - python: 3.8
-            geos: 3.9.0  # this older version is here temporarily to allow matrix to work
+            geos: 3.9.5  # this is here temporarily to allow matrix to work
             numpy: 1.17.3
           # 2020
           - python: 3.9
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,7 +18,7 @@ jobs:
         include:
           # 2019
           - python: 3.8
-            geos: 3.9.0
+            geos: 3.9.0  # this older version is here temporarily to allow matrix to work
             numpy: 1.17.3
           # 2020
           - python: 3.9
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,11 +14,11 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-13, windows-2019]
         architecture: [x64]
-        geos: [3.8.4, 3.9.5, 3.10.6, 3.11.4, 3.12.2, 3.13.0beta1, main]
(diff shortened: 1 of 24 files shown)


shapely/io.py
```diff
@@ -45,10 +45,7 @@ def to_wkt(
 
     The following limitations apply to WKT serialization:
 
-    - for GEOS <= 3.8 a multipoint with an empty sub-geometry will raise an exception
-    - for GEOS <= 3.8 empty geometries are always serialized to 2D
-    - for GEOS >= 3.9 only simple empty geometries can be 3D, collections are still
-      always 2D
+    - only simple empty geometries can be 3D, collections are always 2D
```
jorisvandenbossche, line 48:
```suggestion
    - only simple empty geometries can be 3D, empty collections are always 2D
```

(just to be a bit more explicit)

.github/workflows/tests.yml
```diff
@@ -14,11 +14,11 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-13, windows-2019]
         architecture: [x64]
-        geos: [3.8.4, 3.9.5, 3.10.6, 3.11.4, 3.12.2, 3.13.0beta1, main]
+        geos: [3.9.0, 3.9.5, 3.10.6, 3.11.4, 3.12.2, 3.13.0beta1, main]
```
jorisvandenbossche, line 17: Is it needed to test this 4-year old version that had bug fixes applied afterwards? (as now we need to add special cases to our tests for failures that have already been fixed in GEOS also for 3.9)

I am personally fine with our general practice of only testing the latest GEOS releases

jorisvandenbossche (commented):
Thanks for this clean-up work!

.github/workflows/tests.yml
mwtoews: Generally yes, this matrix should only have the latest GEOS versions. But this is temporarily a placeholder to make the GHA matrix work for the oldest Python version. This will be removed when the older Python version is removed from the matrix. But also, libgeos 3.9.0 is [still in oldstable bullseye debian](https://packages.debian.org/bullseye/libgeos-dev).

jorisvandenbossche: > But this is temporarily a placeholder to make the GHA matrix work for the oldest Python version

Yes, but for that purpose we could also just use 3.9.5 for that? (so testing that version for both Python versions)

jorisvandenbossche: > But also, libgeos 3.9.0 is [still in oldstable bullseye debian](https://packages.debian.org/bullseye/libgeos-dev).

This is for https://github.com/shapely/shapely/issues/2119, but if we start looking at those kind things, one could also say that GEOS 3.8 is still in Ubuntu 20.04 LTS ..

mwtoews: As I thought, repeating GEOS 3.9.5 skips over Py3.8 testing. The matrix needs unique numbers to work, hence why I picked one. However, this isn't a big deal as I expect this to be gone soon when the Python versions are assessed.

jorisvandenbossche (approved):

PR #2127, opened 2024-08-30
MAINT: drop Python 3.8 support

jorisvandenbossche:
A first step following https://github.com/shapely/shapely/issues/2119 (after also dropping GEOS 3.8 in https://github.com/shapely/shapely/pull/2124), dropping at least Python 3.8.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ classifiers = [
 ]
 requires-python = ">=3.9"
 dependencies = [
-    "numpy>=1.19",
+    "numpy>=1.20",
 ]
 
 [project.optional-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,9 +39,9 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
     "Topic :: Scientific/Engineering :: GIS",
 ]
-requires-python = ">=3.8"
+requires-python = ">=3.9"
 dependencies = [
-    "numpy>=1.16",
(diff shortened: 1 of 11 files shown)


pyproject.toml
mwtoews, line 1: Also modify the classifiers in this file, removing 3.8 and adding 3.12 and 3.13

jorisvandenbossche:
> I think setting NumPy >= 1.20 would make more sense to pair with Python >=3.9. This is because NumPy 1.19.0-1.19.2 support Python 3.6-3.8.

I was just following what we were testing (numpy 1.19.5 with python 3.9). Can also make it explicitly `>=1.19.5` (or 1.19.3), but don't mind using 1.20 either (and then update the tests to use that version)

mwtoews:
The [oldest-supported-numpy](https://github.com/scipy/oldest-supported-numpy/blob/main/setup.cfg) is complex, but is most commonly 1.19.3 for for Python 3.9.

I think 1.20 is a cleaner version increment, since it keeps a simpler major-minor increment (rather than a mid-way major-minor-micro increment). See also [1.20.0 release notes](https://numpy.org/doc/stable/release/1.20.0-notes.html) if there is anything compelling. But I'm also not too bothered by this detail, as `>= 1.19.3` is fine too.

The windows-2019 test is failing due to numpy==1.16.2, which has no cp39 binary wheels.

mwtoews (approved):
Looks good, thanks!

PR #2128, opened 2024-08-30
MAINT: remove Travis CI configuration and badge

mwtoews:
Shapely's Travis CI status is [not active](https://app.travis-ci.com/github/shapely/shapely/not-active), and this has been the case for two years.

This can be backported to `maint-2.0`.


jorisvandenbossche:
It was still testing on platforms we don't test elsewhere ... But yes, given they just disabled our repo, even though I asked for OSS credits (and got them at some point), and then they point to https://app.travis-ci.com/organizations/shapely/plan to check the balance or change the settings, but that pages simply gives a "There was an error, please try again.". Let's get rid of it..

PR #2129, opened 2024-08-30
MAINT: clean-up numpy build dependency in pyproject.toml

jorisvandenbossche:
Now that we only support Python >= 3.9, we can remove the oldest-supported-numpy entirely. 

And also added a rough upper pin on numpy<3, because a numpy 3 would probably again mean ABI breaking changes that would require a re-build of packages.


mwtoews (approved):
Thanks, looks good. Here are the [build-time dependency docs](https://numpy.org/devdocs/dev/depending_on_numpy.html#adding-a-dependency-on-numpy) for reference.

PR #2104, opened 2024-08-07
MAINT: apply pydocstyle

theroggy:
Apply pydocstyle and fix all linting errors because of this.

Changes to the documentation are "minimal": only changes needed to conform to the pydocstyle checks.

PR #2131, opened 2024-09-01
MAINT: apply pydocstyle followup

theroggy:
Apply final round of feedback on PR #2104 (after it was merged).

### DIFF
--- a/shapely/predicates.py
+++ b/shapely/predicates.py
@@ -596,9 +596,8 @@ def contains(a, b, **kwargs):
 
 @multithreading_enabled
 def contains_properly(a, b, **kwargs):
-    """Return True if geometry B is completely inside geometry A.
-
-    They should have no common boundary points.
+    """Return True if geometry B is completely inside geometry A, with no common
+    boundary points.
 
     A contains B properly if B intersects the interior of A but not the
     boundary (or exterior). This means that a geometry A does not
@@ -642,7 +641,7 @@ def contains_properly(a, b, **kwargs):
     >>> contains_properly(area1, area3)
     True
 
-    """
+    """  # noqa: D205
     return lib.contains_properly(a, b, **kwargs)
 
 
@@ -1011,7 +1010,8 @@ def within(a, b, **kwargs):
 
 @multithreading_enabled
 def equals_exact(a, b, tolerance=0.0, normalize=False, **kwargs):
-    """Return True if the geometries are equivalent within a given tolerance.
+    """Return True if the geometries are structurally equivalent within a given
+    tolerance.
 
     This method uses exact coordinate equality, which requires coordinates
     to be equal (within specified tolerance) and in the same order for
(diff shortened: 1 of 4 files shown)


shapely/predicates.py
```diff
@@ -596,9 +596,8 @@ def contains(a, b, **kwargs):
 
 @multithreading_enabled
 def contains_properly(a, b, **kwargs):
-    """Return True if geometry B is completely inside geometry A.
-
-    They should have no common boundary points.
+    """Return True if geometry B is completely inside geometry A, with no common
```
sgillies, line 599: @theroggy thanks for looking after the docstrings!

Have you ever seen an open source tool for checking the *prose* of a docstring? The actual writing style?

I've aspired to never begin with "Return ..." or "Returns ..." because returning (on success) is in the very nature of a Python method and we have a numpy "Returns" section specifically for that. I must say that I don't always live up to my aspirations and so a tool would be handy.

I also prefer a first line that is as terse as possible, as per https://numpydoc.readthedocs.io/en/latest/format.html#short-summary.

sgillies (approved):

shapely/predicates.py
jorisvandenbossche: > I also prefer a first line that is as terse as possible, as per https://numpydoc.readthedocs.io/en/latest/format.html#short-summary.

FWIW, the change was because my explicit ask to make it longer again ..

I fully agree that ideally the first line is as short as possible and to generally follow the rule of a single line summary. However, I also think that summary should still be as clear as possible, and so I commented on the previous PR that IMO in this case "practicality beats purity" and asked to expand that first sentence, to make it more explicit what the exact difference is of "contains properly" (i.e. it's the part of "no common boundary points" that differentiates it from `contains`)

theroggy: > Have you ever seen an open source tool for checking the _prose_ of a docstring? The actual writing style?

Ruff, via the [pydocstyle rules](https://docs.astral.sh/ruff/rules/#pydocstyle-d) does check the writing style recommendations that are in [PEP 257](https://peps.python.org/pep-0257/) e.g.:
- The first line/summary of the docustring should be one line: [blank-line-after-summary](https://docs.astral.sh/ruff/rules/blank-line-after-summary/)
- The summary should be in an "imperative mood". So e.g. "Returns..." is not allowed, it should be "Return...": [non-imperative-mood](https://docs.astral.sh/ruff/rules/non-imperative-mood/)
- The summary should not start with "This": [docstring-starts-with-this](https://docs.astral.sh/ruff/rules/docstring-starts-with-this/)
- ...
 
> I've aspired to never begin with "Return ..." or "Returns ..." because returning (on success) is in the very nature of a Python method and we have a numpy "Returns" section specifically for that. I must say that I don't always live up to my aspirations and so a tool would be handy.

I do understand your line of thinking, but """Return the pathname of the KOS root directory.""" is actually the example in PEP 257 for a [one line docstring](https://peps.python.org/pep-0257/#one-line-docstrings)... so at least the authors of PEP 257 were OK with that. Hence, I doubt that there will be linting tools that explicitly check for this out of the box...

jorisvandenbossche (approved):

mwtoews (approved):
I've checked and verified the fixes in the rendered docs. Thanks all for the improvements!

PR #2132, opened 2024-09-03
BLD: refactor setup.py, consolidate define_macros, cleanup warnings

mwtoews:
This refactors the setuptools-based build with a few changes:

- Rename `get_geos_paths()` to `get_ext_options()`, which also embeds NumPy include dir.
- Also use `define_macros` (aka `-D`) for `GEOS_USE_ONLY_R_API`, `NPY_NO_DEPRECATED_API`, and `NPY_TARGET_VERSION`
- Remove custom `build_ext` cmdclass, which used to embed NumPy include dir. This references a PR from 8 years ago as a workaround for some issue that is probably historic by now (although I don't fully understand what the issue was)
- Use ruff check / format on setup.py to bring it to the same standard as other Python sources
- Resolve a few basic C-level warnings; there are still other unresolved warnings
- Remove macros from source files, as they are moved to `define_macros` in setup.py
- Add `-v` for verbose builds in CI to see compilation commands and warnings

I'd think it's fine for the 2.1 release series to keep with setuptools-based build, but future 2.2 development should aim to have a [meson-python build](https://mesonbuild.com/meson-python/) (I have some [work-in-progress here](https://github.com/mwtoews/shapely/tree/meson-python)).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -167,9 +167,9 @@ jobs:
         # ensure we build against numpy nightly (numpy 2.0 compat)
         run: |
           if [ -z "${{ matrix.numpy }}" ]; then
-            pip install -e . --no-build-isolation
+            pip install -v -e . --no-build-isolation
           else
-            pip install -e .
+            pip install -v -e .
           fi
 
       - name: Overview of the Python environment (pip list)
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,5 @@
-import builtins
+"""Setuptools build."""
+
 import logging
 import os
 import subprocess
@@ -6,7 +7,6 @@ import sys
 from pathlib import Path
 
 from setuptools import Extension, setup
-from setuptools.command.build_ext import build_ext as _build_ext
 
 # ensure the current directory is on sys.path so versioneer can be imported
 # when pip uses PEP 517/518 build rules.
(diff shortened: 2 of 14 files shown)


sgillies:
@mwtoews I've not yet used meson, but it seems like a good idea for us.

mwtoews:
@sgillies it seems many projects that have compiled objects have shifted away from setuptools to python-meson. The primary devs are from scipy (thus this has had the longest support), more recently pandas 2.1, scikit-learn 1.4, matplotlib 3.9, etc.

My suggestion to wait until the next dev cycle is to not stall the upcoming 2.1 release any further. I'd rather see a full dev cycle to work out any issues with a new build backend.

jorisvandenbossche (approved):
Nice clean-up!

jorisvandenbossche:
> but future 2.2 development should aim to have a [meson-python build](https://mesonbuild.com/meson-python/) (I have some [work-in-progress here](https://github.com/mwtoews/shapely/tree/meson-python)).

I had been thinking about using [scikit-build-core](https://github.com/scikit-build/scikit-build-core) (or planning to at some point experiment with it), but meson-python would certainly be a good option as well. 
Any of those two will probably be am improvement (although I would say that our use of setuptools is also not too bad)

Issue #2137, opened 2024-09-08
Support for Python 3.13 freethreading build


greglucas:
I am starting to build Cartopy to support Python 3.13t and Shapely is one of our dependencies that we interact with through Cython and the C-layer. I couldn't find any issues discussing the current status of Shapely's support for free threading or if there are plans to work on this at all, so I figured I would open up an issue and draft PR with some simple updates to enable the free threading build to discuss this.

Is libgeos thread safe? https://libgeos.org/project/rfcs/rfc03/ I am not sure what the status of that RFC for the C-API is.

A link to some good discussion for how to approach various thread unsafe C-API usages: https://py-free-threading.github.io/porting/

caspervdw:
There has been some news on this subject, but I am not sure if this makes shapely completely "freethreading" ready.

In #2343 we eliminated the global GEOS context that was used (only) in the geometry deallocator. Instead we started using a threadlocal GEOS context. This threadlocal GEOS context is now used BTW for every GEOS operation, not only the geometry deallocator.

Stuff that needs to happen still (based on https://py-free-threading.github.io/porting/):

- [ ] Define and document thread-safety guarantees
- [ ] Be explicit about that we don't support mutating shared ndarrays (like numpy) see discussion in #2338
- [ ] Go through all mutable global state and see if it is needed to make threadlocals (#2343 did the most part of it)
- [ ] Check if we have mutable global caches (I think not)
- [ ] Check if the Geometry classes are thread-safe (do they have mutable state?)
- [x] Declare free-threaded support in the C-extension (already done in #2138)
- [ ] Review usage of the `Py_BEGIN_ALLOW_THREADS` and `Py_END_ALLOW_THREADS` macros

PR #2138, opened 2024-09-08
MNT: Add Python 3.13t free threading support to the extensions

greglucas:
This declares the package as supporting nogil Python releases. I have tested it locally and it works as expected. Without this I get a warning that shapely doesn't support the nogil free threading, without this it runs as expected and passes the tests.

Some links for information on what I am updating here.

Update to the module definition:
https://py-free-threading.github.io/porting/#__tabbed_1_1

CI testing:
Added installation from the deadsnakes setup-python version, which I think will only work on the ubuntu runner:
https://py-free-threading.github.io/ci/

cibuildwheel support to build the free threaded wheels:
https://cibuildwheel.pypa.io/en/stable/options/#free-threaded-support

xref: #2137

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -124,7 +124,9 @@ jobs:
           # once Cython has an official release that supports free threading we can remove this
           CIBW_BUILD_FRONTEND: "pip; args: --no-build-isolation"
           CIBW_BEFORE_BUILD: |
-            pip install --pre cython &&
+            # For free threading support in Cython install a pre-release version of Cython
+            PRE_FLAG="$(python -c"import sysconfig; print('--pre' if sysconfig.get_config_var('Py_GIL_DISABLED') else '')")" &&
+            pip install $PRE_FLAG cython &&
             pip install numpy setuptools wheel
           CIBW_BUILD: ${{ matrix.cibw_build }}
           CIBW_MANYLINUX_X86_64_IMAGE: ${{ matrix.baseimage }}_geos:${{ env.GEOS_VERSION }}
@@ -208,10 +210,14 @@ jobs:
           # once Cython and numpy have official releases that support free threading we can remove this
           CIBW_BUILD_FRONTEND: "pip; args: --no-build-isolation"
           CIBW_BEFORE_BUILD: |
-            pip install --pre cython &&
+            # For free threading support in Cython install a pre-release version of Cython
+            PRE_FLAG="$(python -c"import sysconfig; print('--pre' if sysconfig.get_config_var('Py_GIL_DISABLED') else '')")" &&
+            pip install $PRE_FLAG cython &&
             pip install numpy setuptools wheel
           CIBW_BEFORE_BUILD_WINDOWS:
-            pip install --pre cython &&
+            # For free threading support in Cython install a pre-release version of Cython
+            PRE_FLAG="$(python -c"import sysconfig; print('--pre' if sysconfig.get_config_var('Py_GIL_DISABLED') else '')")" &&
+            pip install $PRE_FLAG cython &&
             pip install numpy setuptools wheel &&
             pip install delvewheel
           CIBW_REPAIR_WHEEL_COMMAND_WINDOWS: delvewheel repair --add-path ${{ runner.temp }}\geos-${{ env.GEOS_VERSION }}\bin -w {dest_dir} {wheel}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -125,8 +125,7 @@ jobs:
           CIBW_BUILD_FRONTEND: "pip; args: --no-build-isolation"
           CIBW_BEFORE_BUILD: |
             # For free threading support in Cython install a pre-release version of Cython
-            PRE_FLAG="$(python -c"import sysconfig; print('--pre' if sysconfig.get_config_var('Py_GIL_DISABLED') else '')")" &&
-            pip install $PRE_FLAG cython &&
+            python -c "import sysconfig, subprocess; subprocess.check_call(['pip', 'install', '--pre', 'cython'] if sysconfig.get_config_var('Py_GIL_DISABLED') else ['pip', 'install', 'cython'])" &&
             pip install numpy setuptools wheel
           CIBW_BUILD: ${{ matrix.cibw_build }}
           CIBW_MANYLINUX_X86_64_IMAGE: ${{ matrix.baseimage }}_geos:${{ env.GEOS_VERSION }}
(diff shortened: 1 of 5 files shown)


mwtoews:
The CIBW config should be moved centrally to pyproject.toml, and skip Windows. This is because [numpy files](https://pypi.org/project/numpy/2.1.1/#files) have `cp313t` for most platforms except Windows (I don't know why this is).
```toml
[tool.cibuildwheel]
# Enable free-threaded support
free-threaded-support = true

# Skip building free-threaded compatible wheels on Windows
free-threaded-support = true
skip = "*t-win*"
```

greglucas:
I think the wheel builds will still fail because they aren't getting the `cython>=3.1.0a0` from the nightlies (the current pypi cython release doesn't work). I couldn't fully track down how numpy is handling their cython dependency to make the nightly releases unfortunately. In the testing action here, I forced the install to use the nightly cython and numpy releases which is why that one works.

shapely/_geometry_helpers.pyx
```diff
@@ -1,3 +1,4 @@
+# cython: freethreading_compatible=True
```
jorisvandenbossche, line 1: I think we should also be able to specify that in setup.py for all cython files?

jorisvandenbossche:
@greglucas thanks for looking into this!

> I think the wheel builds will still fail because they aren't getting the `cython>=3.1.0a0` from the nightlies (the current pypi cython release doesn't work). I couldn't fully track down how numpy is handling their cython dependency to make the nightly releases unfortunately.

You can probably reuse something similar as what I cleaned up in https://github.com/shapely/shapely/pull/2075 for when we were building against numpy nightly

shapely/_geometry_helpers.pyx
greglucas: Yes, good call. I have added that directive to the Cythonize call.

greglucas:
> You can probably reuse something similar as what I cleaned up in https://github.com/shapely/shapely/pull/2075 for when we were building against numpy nightly

Thanks for the link! I copied that over to this PR now.

rgommers:
> The CIBW config should be moved centrally to pyproject.toml, and skip Windows. This is because [numpy files](https://pypi.org/project/numpy/2.1.1/#files) have `cp313t` for most platforms except Windows (I don't know why this is).

Note that NumPy 2.1.3 is out now and has Windows wheels.

.github/workflows/release.yml
```diff
@@ -119,6 +119,11 @@ jobs:
         uses: pypa/cibuildwheel@v2.21.3
         env:
           CIBW_ARCHS: ${{ matrix.arch }}
+          # TEMP don't use automated/isolated build environment, but manually
+          # install build dependencies so we can build with cython 3.1.0a0
+          # once Cython and numpy have official releases that support free threading we can remove this
+          CIBW_BUILD_FRONTEND: "pip; args: --no-build-isolation"
+          CIBW_BEFORE_BUILD: pip install --pre --extra-index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple numpy setuptools cython wheel
```
jorisvandenbossche, line 126: Wondering if there is a way to limit this to cython or limit this to the free-threading build (so we still build all the other binaries with released versions)

jorisvandenbossche: Looking at the numpy and pandas setup, they essentially create one entry in the workflow's matrix per wheel that is being built (instead of letting cibuildwheel generate multiple wheels for the different python versions per platform), and so that way they can control this a bit easier ..

.github/workflows/tests.yml
```diff
@@ -96,7 +100,7 @@ jobs:
         if: ${{ matrix.geos == 'main' }}
 
       - name: Set up Python ${{ matrix.python }}
-        uses: actions/setup-python@v5
+        uses: Quansight-Labs/setup-python@b9ab292c751a42bcd2bb465b7fa202ea2c3f5796  # v5.3.1
```
greglucas, line 103: https://github.com/Quansight-Labs/free-threaded-compatibility/issues/107

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-    "Cython",
+    "Cython>=3.1.0a0",
```
greglucas, line 3: If we are OK with putting this in the build system area to a high pin, we can automatically get pip to install the required Cython for us without needing to update any of the cibuildwheel BEFORE_BUILD steps.

Note that we can still be built with `Cython<3.1`, for example if you install with `pip install --no-build-isolation .` and have a lower Cython installed in your environment.

greglucas (commented):
I just reworked this a bit to try and simplify some of the things and noted them in this review here.

.github/workflows/release.yml
rgommers: I think building a free-threaded wheel with Cython 3.1.0a0 is perfectly fine, because that's the minimum version with any kind of support.

I wouldn't add `>=3.1.0a0` in `pyproject.toml` though, that will give problems for distros and is unnecessary for anything but free-threading. Plus there may be as-yet-undetected regressions, it's an alpha release after all.

greglucas: I thought most distros built with `no-build-isolation` generally so they could bring in their own versions of things. Although I guess they still have to parse that section to find out what the minimum requirement should be on their build dependencies... :+1: I reverted that change now.

rgommers: Yes, the parsing indeed, mostly distros have some tooling for that (e.g., Grayskull for conda-forge).

jorisvandenbossche: But so then it's still the question how to get cython 3.1.0a0 for just the free-threaded wheel build?

I suppose one other option would be to do some if/else logic in the CIBW_BEFORE_BUILD section. I remember now that I did something like that in https://github.com/apache/arrow/pull/44193/files#diff-1b07285b8a4a94b9f2e56a12d1baf7b7b2f186abc44e5b93f644fd1771f7dc10

greglucas: :+1: Thanks for that link. I think we can put that logic in pretty easily, I added it as a one-liner because I wasn't entirely sure how the line continuations would work for the multi-line if block.

.github/workflows/release.yml
```diff
@@ -196,7 +205,21 @@ jobs:
...
+          CIBW_BUILD_FRONTEND: "pip; args: --no-build-isolation"
+          CIBW_BEFORE_BUILD: |
+            # For free threading support in Cython install a pre-release version of Cython
+            PRE_FLAG="$(python -c"import sysconfig; print('--pre' if sysconfig.get_config_var('Py_GIL_DISABLED') else '')")" &&
+            pip install $PRE_FLAG cython &&
+            pip install numpy setuptools wheel
+          CIBW_BEFORE_BUILD_WINDOWS:
+            # For free threading support in Cython install a pre-release version of Cython
+            PRE_FLAG="$(python -c"import sysconfig; print('--pre' if sysconfig.get_config_var('Py_GIL_DISABLED') else '')")" &&
+            pip install $PRE_FLAG cython &&
```
greglucas, line 220: This is problematic in powershell apparently. I tried to figure this out briefly on my own fork, but couldn't get the right Windows incantation here with several different attempts. Does anyone else have Windows knowledge or know what I need to do here to get this to work under Windows?

```
PRE_FLAG="$(python -c"import sysconfig; print('--pre' if sysconfig.get_config_var('Py_GIL_DISABLED') else '')")" && pip install $PRE_FLAG cython && pip install numpy setuptools wheel && pip install delvewheel
  'PRE_FLAG' is not recognized as an internal or external command,
  operable program or batch file.
```
https://github.com/shapely/shapely/actions/runs/11945885851/job/33427753116?pr=2138#step:7:458

rgommers: I'm terrible at the various Windows shells, so I'd change to `bash` or try to steal the invocation from another project. E.g., from NumPy:

https://github.com/numpy/numpy/blob/59647e846b6c25f0e44905936693d9a49c1e0ac3/tools/wheels/cibw_before_build.sh#L59

jorisvandenbossche: Or with an if/else instead of trying to pass the flag to the single pip call?

greglucas: I could not get the Windows shell to cooperate. I fell back to an if/else within a Python subprocess to call the respective pip install command instead (also used this in the linux BEFORE_BUILD steps for consistency).

jorisvandenbossche: Given this is only temporary until cython does a release, let's just go with this since it is working!

jorisvandenbossche (approved):
Thanks a lot @greglucas!

PR #2141, opened 2024-09-13
CI: enable autofixing for ruff linter + add a few generic pre-commit hooks for EOL and trailing whitespace

mwtoews:
This PR adds a few useful pre-commit hooks to check EOLs and trailing whitespace.

<s>It also moves the CI job to [pre-commit.ci](https://pre-commit.ci/) (in progress, requested)</s>. The GHA is renamed from "Lint" to "pre-commit".


jorisvandenbossche:
> It also moves the CI job to [pre-commit.ci](https://pre-commit.ci/) (in progress, requested)

Is there a good reason to move? (the github actions workflow we have now is tiny and works fine, I think)

I know that pre-commit.ci can push to PRs to fix issues, but personally I am -1 on letting a bot push commits to a PR automatically

sgillies:
I'd prefer to not add a new service.

mwtoews:
Personally I think pre-commit.ci is a quick an nifty way to offload CPU time away from GHA, but it's true that it's a tiny GHA job that can be kept.

I've re-added the GHA again, but renamed it away from "Lint" to "pre-commit. It's pinned to '3.x' for the latest stable Python, and I've removed the CITATION.cff check stuff, as this was moved as a pre-commit.

.pre-commit-config.yaml
```diff
@@ -1,7 +1,22 @@
-files: 'shapely/'
+ci:
+  autoupdate_schedule: quarterly
```
jorisvandenbossche, line 2:
```suggestion
  autoupdate_schedule: quarterly
  autofix_prs: false
```

That still uses the pre-commit.ci to regularly update the versions here, without the pushing to other PRs. That would personally be my preference.

.pre-commit-config.yaml
```diff
@@ -1,7 +1,22 @@
...
+    - id: trailing-whitespace
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: "v0.5.2"
+    rev: v0.5.2
     hooks:
+      # Run the linter.
+      - id: ruff
+        args: [ --fix ]
+      # Run the formatter.
       - id: ruff-format
```
jorisvandenbossche, line 18: And thanks for fixing this! I was planning to open a PR as well, as the comments about non-sorted imports without actually fixing them were becoming annoying ;)

.pre-commit-config.yaml
mwtoews: Mind you, these configurations are dormant for now, unless we later decide to switch service to pre-commit.ci

jorisvandenbossche: Ah yes, but then should we then either just enable it, or otherwise remove the configuration? (or at least add a comment that it is not used?)

mwtoews: Let's remove the unused configuration

PR #2146, opened 2024-09-13
PERF: prepare single geometry input to contains_xy and intersects_xy

### LINKED ISSUE #1734: ENH: prepare input polygon automatically for contains_xy
As noted in [#1726 comment](https://github.com/shapely/shapely/issues/1726#issuecomment-1399434147), there is a major speedup to preparing the input polygon.

In this case, we can detect if there is just one polygon as input, and prepare it automatically.

jorisvandenbossche:
Closes #1734

See https://github.com/shapely/shapely/issues/1726#issuecomment-1399434147 for example of performance impact

Still need to update the docstring, if we want to do this.

### DIFF
--- a/shapely/predicates.py
+++ b/shapely/predicates.py
@@ -1261,6 +1261,8 @@ def contains_xy(geom, x, y=None, **kwargs):
     if y is None:
         coords = np.asarray(x)
         x, y = coords[:, 0], coords[:, 1]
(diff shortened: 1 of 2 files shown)


brendan-ward (approved):
The speedup suggests that this is well worth it, and the overhead of preparing geometries is pretty low (unless done repeatedly).

PR #2147, opened 2024-09-13
ENH: add orient_polygons

jorisvandenbossche:
This was an old branch I had lying around, and exposes `GEOSOrientPolygons` (was new in GEOS 3.12). At the moment directly as `shapely.orient_polygons()` ufunc, however there are some other API options.

PostGIS exposes this as [`ST_ForcePolygonCCW`](https://postgis.net/docs/ST_ForcePolygonCCW.html) and [`ST_ForcePolygonCW`](https://postgis.net/docs/ST_ForcePolygonCW.html).  
However, they also have [`ST_IsPolygonCCW`](https://postgis.net/docs/ST_IsPolygonCCW.html) and [`ST_IsPolygonCW`](https://postgis.net/docs/ST_IsPolygonCW.html), while we have [`shapely.is_ccw`](https://shapely.readthedocs.io/en/stable/reference/shapely.is_ccw.html).

We also have `shapely.force_2d` and `force_3d`, so in that line a `shapely.force_ccw` or `force_polygon_ccw` would certainly fit as well. Of course, if we want to provide both CCW and CW options, that requires two functions instead of one for `orient_polygons`

We already have a non-vectorized `shapely.ops.orient()` function, but if we go with the "orient" terminology, I would prefer the more explicit `orient_polygons` over `orient` for the top-level vectorized function.

Some related discussion in https://github.com/shapely/shapely/pull/1690 (once we decide in this PR, I can revive that older PR to provide a fallback implementation for GEOS < 3.12)


sgillies:
I find the PostGIS use of "Force" unfortunate and I prefer a less violent name. Numpy's `asarray()`, for example, returns what you ask for, copying only as necessary. There's no use of force 😄

jorisvandenbossche:
Them maybe let's keep it as `orient_polygons` as I did here. Then we don't follow PostGIS in this case but take the naming consistent with GEOS. 
(and I think @theroggy also mentioned his preference for `orient_polygons` in the geopandas meeting>)

theroggy (approved):
Looks fine to me.

Maybe just add a note in CHANGES.txt?

PR #2152, opened 2024-09-16
TST: add some extra test cases for from_wkt with on_invalid=fix

theroggy:
Add some more test cases as "documentation" on how different situations are handled by the "fix" logic of GEOS.

Followup on #2094


shapely/tests/test_io.py
```diff
@@ -190,14 +190,28 @@ def test_from_wkt_on_invalid(wkt, on_invalid, error, message):
...
+        ),
     ],
 )
 def test_from_wkt_on_invalid_fix(wkt, expected_wkt):
     """Tests for on_invalid="fix".
 
     Geometries that cannot be fixed are returned as None.
     """
+    # Make sure the wkt needs fixing
+    assert shapely.from_wkt(wkt, on_invalid="ignore") is None
```
sgillies, line 212: @theroggy Let's not use `from_wkt()` like this to pre-assert. If there were a bug involving `on_invalid="ignore"` we could be misled.

sgillies (changes requested):
Made a request inline.

shapely/tests/test_io.py
theroggy: I was actually trying to avoid being misled if the "fix" functionality of GEOS would be expanded to support more cases. 

But just a classic potato-patato case... so no strong opinion, so I'll remove it.

sgillies (approved):
Thanks @theroggy !

shapely/tests/test_io.py
```diff
@@ -190,14 +190,25 @@ def test_from_wkt_on_invalid(wkt, on_invalid, error, message):
...
+            None,
+        ),
     ],
 )
 def test_from_wkt_on_invalid_fix(wkt, expected_wkt):
     """Tests for on_invalid="fix".
 
     Geometries that cannot be fixed are returned as None.
     """
+    # Test
```
jorisvandenbossche, line 211:
```suggestion
```

mwtoews (approved):
Looks good, although it doesn't test `from_wkb` (anymore?), so perhaps edit the PR title.

theroggy:
@mwtoews True. Fixed.

jorisvandenbossche (approved):
Thanks!

PR #2153, opened 2024-09-17
Split up new test to eliminate conditions

sgillies:
Conditions in a test signal to me that we should split up the test. We have more tests, but they are simpler, and won't have any bugs of their own.

### DIFF
--- a/shapely/tests/test_io.py
+++ b/shapely/tests/test_io.py
@@ -142,41 +142,60 @@ def test_from_wkt():
 
 
 def test_from_wkt_none():
-    # None propagates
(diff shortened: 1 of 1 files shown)


theroggy:
Just to clarify, I won't loose any sleep on this being changed ;-).

Even though I agree in general, I wrote this test as it is/was to have a clear overview in the parameters how each input WKT is treated depending on the different possible values of `on_invalid`.
When there is a new relevant case is to be added to the test in the future it is also very clear and easy to add this for the different cases.

With the test split up you need to scroll through the different tests and it is quite a bit more difficult to get this overview.

PR #2155, opened 2024-09-18
DOC: apply doc8 for rst files + add to pre-commit

theroggy:
Structural solution for comment that line length in manual should be limited to 79 character: https://github.com/shapely/shapely/pull/2030#pullrequestreview-2290631366

[doc8](https://github.com/PyCQA/doc8) is a style checker for sphinx (or other) rst documentation.


pyproject.toml
```diff
@@ -79,9 +79,14 @@ test-command = "pytest --pyargs shapely.tests"
...
 
+[tool.doc8]
+include = ["docs"]
+ignore = ["D000", "D004"]
+allow-long-titles = true
+
 [tool.ruff]
 line-length = 88
-extend-exclude = ["docs/*", "benchmarks/*", "shapely/_version.py"]
+extend-exclude = ["docs/*", "benchmarks/*", "shapely/_version.py", "versioneer.py"]
```
mwtoews, line 89: I've taken this suggested edit for #2141 (also removing `files: 'shapely/'` from the pre-commit config).

pyproject.toml
```diff
@@ -79,9 +79,14 @@ test-command = "pytest --pyargs shapely.tests"
 source = ["shapely"]
 omit = ["shapely/tests/*"]
 
+[tool.doc8]
+include = ["docs"]
+ignore = ["D000", "D004"]
+allow-long-titles = true
```
mwtoews, line 85: Also add `max-line-length = 79` to enforce this rule.

mwtoews (approved):
This pre-commit is a bit slow, but nothing too tedious. And if it is too slow, we could comment it out most of the time, enabling it briefly before releases.

Rebase to latest main, as there were some related pre-commit changes merged.

pyproject.toml
theroggy: This is the default, so was already enforced... but "explicit is better than implicit" so I added it...

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks @theroggy!

PR #2156, opened 2024-09-19
ENH: add coverage validation function

jorisvandenbossche:
Exposing the new `GEOSCoverageIsValid_r`, which will be useful in combination with https://github.com/shapely/shapely/pull/1969.

This only exposes the boolean validation part (valid or not valid), and not the ability to get the invalid edges (I would expose that as a separate function, PostGIS has [ST_CoverageInvalidEdges](https://postgis.net/docs/ST_CoverageInvalidEdges.html) for that).

Still need to add tests and expand the docstring.

Two main questions:
- Naming: do we strictly follow GEOS (`coverage_is_valid()`), or do we make it a more natural order (`is_valid_coverage()`) ?
  - The latter has the benefit that it matches all our other `is_...` predicates (although those all work element-wise and not on a full array like this one), the former has the benefit that it keeps all `coverage_...` functions with a consistent prefix (e.g. with tab completion you can then easily see all coverage related functions) 
- Exact reduction API: the ufunc itself is written as a generalized ufunc that removes the last axis. But on the python side we can make this work a bit differently if we want. For example, I think the best default is to let it work on the full array, also if it is higher dimensional (>1D). Question is if we only support that (i.e. always ravel under the hood, which is what I did now), or still support an `axis` keyword to optionally verify along a certain axis (like `unary_union` can optionally union along a certain axis). Consistency with a similar function like `unary_union` is maybe the best option.


jorisvandenbossche:
cc @martinfleis @theroggy

martinfleis:
Considering the vector data cube application for time-varying geometries which can be stored in N-D arrays, I can image that we might want to do coverage simplification and hence also coverage validation per time period, hence axis keyword might be useful. But if that will not be exposed in coverage simplification, we would have to loop over time dimension anyway in xvec so it does not matter too much.

I generally think that if you have N-D array of geoms you are mostly interested in applying these operations along the axis given the additional dimension is likely going to be time so it would be useful to have that exposed directly here. But it would have to propagate through all relevant functions. In coverage simplification of, all geoms belonging to the same individual time coordinate might form a valid coverage but when considering whole array together, it is unlikely going to be a single coverage. (writing all this, I am not sure if our decision in #1969 was the best but I'll try to get to that in coming days).

sgillies:
@jorisvandenbossche My understanding is that `Coverage` in `GEOSCoverageIsValid` is an attempt to preserve some sense of the C++ `coverage` namespace. But since shapely is flat these days, I think `is_valid_coverage` reads better than `coverage_is_valid`. The latter would be better written as `coverage.is_valid` in my opinion, using an actual namespace and not a pseudo-namespace.

jorisvandenbossche:
> I think `is_valid_coverage` reads better

Yes, agreed, and that is also the reason why I choose `is_valid_coverage` in the current version of this PR. However, after a night's sleep, I think that I now have a slight preference for `coverage_is_valid` .. ;)

We will also want to add a `coverage_invalid_edges()`, I think, and at that point it seems nice to have all those coverage related functions start with `coverage_..`. Also, that makes it a bit more distinct from the other `is_..` functions which might actually be a good thing (because `is_valid_coverage`/`coverage_is_valid` is not a simple element-wise predicate like all other `is_..` functions).

Now, no strong opinion, so in the end either way is fine

theroggy:
> However, after a night's sleep, I think that I now have a slight preference for `coverage_is_valid` .. ;)
> 
> Now, no strong opinion, so in the end either way is fine

Same here, slight preference for `coverage_is_valid` for the same reason: to somewhat group all coverage functionalities...

sgillies:
Certainly shapely 1.x namespaces were too nested, and exporting geometries classes from shapely instead of shapely.geometry, is good. Namespaces can be useful, though. Numpy uses them. `coverage_` as a name prefix might look like it groups functionality, but it does not do so in any sense of the Python language.

Ultimately, I think it comes down to the number of such functions. If we have ~4 such functions, a `coverage` module/namespace could be good. It wouldn't be worth it for just 2-3.

Without an actual, I think `coverage_is_valid` and `is_valid_coverage` are the same thing. Maybe it just comes down to which we want to appear first in a sorted list.

jorisvandenbossche:
Given that we are currently only talking about around 3 functions (simplify, is_valid, invalid_edges), I don't think a submodule is worth it. So then I went with the `coverage_is_valid()` naming.

theroggy (approved):

jorisvandenbossche:
GEOS has a `gapWidth` keyword (_"The maximum width of gaps to detect."_) for the validation method, which I still should expose here. But I am not entirely sure how this relates to the `tolerance` keyword of `coverage_simplify` (_"The degree of simplification roughly equal to the square root of the area of triangles that will be removed."_).

And the strange part is also that PostGIS calls this `tolerance` in  [ST_CoverageInvalidEdges](https://postgis.net/docs/ST_CoverageInvalidEdges.html) as well (this uses the same CoverageIsValid under the hood).

martinfleis:
`tolerance` is a different thing than `gapWidth`. It takes the role of `eps` from normal simplify to determine the degree of simplification. There's no way of plugging the concept of the `gapWidth` in and I am not sure how useful it is in the validation when what is important is how the coverage-based algos will treat the collection.

jorisvandenbossche:
Will already merge this, and do the invalid edges in a follow-up

jorisvandenbossche:
Start of a follow-up PR -> https://github.com/shapely/shapely/pull/2248

PR #2160, opened 2024-09-28
FIX: prevent crash on read nonlinear types from WKB/WKT

### LINKED ISSUE #2159: from_wkt and from_wkb crash on MULTISURFACE input
GEOS 3.13 introduced basic support for CURVE/SURFACE geometries, mainly to be able to parse them from WKB and WKT. However, `shapely.from_wkt` and `shapely.from_wkb` seem to crash on this type of input.

Noticed in a crashing test in pyogrio: https://github.com/geopandas/pyogrio/issues/478

Testing GEOS directly using `geosop` doesn't seem to trigger the problem (see issue above)?

I've had a quick look at the shapely code and don't immediately see where it can go wrong specifically on the curve types... but I'm not very familiar with the code, so probably missed something.

I also wonder what is going to happen if other shapely functions are going to be called on these geometries... but that's to be seen after this is solved :-)

Script to reproduce:

``` python
import shapely

wkt = "MULTISURFACE (CURVEPOLYGON ((-95.29698763468994 63.36578165250125, -94.5991423640996 53.06992070409323, -73.7073995758013 60.149472648492385, -95.29698763468994 63.36578165250125)))"
print(shapely.from_wkt(wkt))
print("here")
```

brendan-ward:
Resolves #2159

GEOS 3.13 added support for reading nonlinear types (CircularString, CompoundCurve, MultiCurve, CurvePolygon, MultiSurface) from WKB / WKT, but these types do not exist within the shapely geometry type registry, or are they likely handled correctly in various functions where we are using conditional statements on geometry type ID.  The crash is due to attempting to look up the type by index in the type registry and hitting an out-of-bounds error.

Until nonlinear types can be added more comprehensively, this prevents constructing nonlinear types after reading from WKB / WKT (when supported by GEOS) so that we don't run into crashes or unexpected behavior elsewhere in the codebase.

This adds WKT / WKB test cases from the GEOS test suite that are known to work with GEOS >= 3.13.

This should get backported to other supported versions of Shapely that would get compiled against GEOS 3.13.


jorisvandenbossche:
> This should get backported to other supported versions of Shapely that would get compiled against GEOS 3.13.

In practice that already happened for conda-forge (the latest shapely release was rebuilt for GEOS 3.13). I don't know if this is worth doing another 2.0.7 release, or otherwise we could also add this patch in the conda-forge recipe to fix those binaries that way (someone manually building shapely with latest GEOS of course still can run in that issue)

jorisvandenbossche (approved):
Thanks!

brendan-ward:
> also add this patch in the conda-forge recipe

I don't know what is involved there, but that is indeed probably the main place to fix.  So long as this goes in before wheels  get built with 3.13, that is probably sufficient outside of conda-forge.

mwtoews (approved):
Thanks, this is a good fix for now.

PR #2187, opened 2024-11-29
chore: upgrade GEOS_VERSION to 3.13.0 for wheels

Clarmy:
GEOS has been updated to version 3.13, and I would like to have Shapely built against this latest 3.13 version of GEOS. This is because we discovered that when the GEOS version is upgraded to 3.13, [the issue](https://github.com/SciTools/cartopy/pull/2463) we encountered with Cartopy (which depends on Shapely) disappears.


jorisvandenbossche (approved):
Thanks!

Issue #2195, opened 2024-12-19
Copy behaviour has changed but user manual has not


H0R5E:
The copying behaviour of objects like `Point` has changed between version 1 and 2. For instance, with version 1.8.5, the following is true:

```pycon
>>> from shapely.geometry import Point
>>> point = Point(0.0, 0.0)
>>> new_point = Point(point)
>>> new_point is point
False
```

while with version 2.0.6 the result is like so:

```pycon
>>> from shapely import Point
>>> point = Point(0.0, 0.0)
>>> new_point = Point(point)
>>> new_point is point
True
```

So, it seems like deep copies are not created from initializers any more, however the docs (e.g. [this line](https://github.com/shapely/shapely/blob/2ddeb0132a9cf3638b4c900292492f992a4a4862/docs/manual.rst?plain=1#L323)) do not indicate that the behaviour has changed:

> The `Point` constructor also accepts another `Point` instance, thereby making a copy.

I guess I'm not sure if this change of behaviour was intended? If so, I'm happy to offer an update to the docs, to make it clearer that copies are now shallow.

EDIT: Having read a little more, I assume that the immutability of the Geometry objects makes deep copying unnecessary, but I would still consider updating the docs to reflect this.

Thanks,

Mat

PR #2197, opened 2025-01-13
DOC: Be specific about copy operations in the user manual

H0R5E:
This PR improves the clarity of the user manual by updating sections where the word 'copy' is used, to be more precise. In version 1, all copies were deep, but with version 2 having immutable geometry objects some operations are now simply return references, but not all.

Fixes https://github.com/shapely/shapely/issues/2195

### DIFF
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -321,7 +321,7 @@ Coordinates may also be sliced. `New in version 1.2.14`.
   [(0.0, 0.0)]
 
 The `Point` constructor also accepts another `Point` instance, thereby making
-a copy.
+a shallow copy.
 
 .. code-block:: pycon
 
@@ -386,7 +386,7 @@ Coordinates may also be sliced. `New in version 1.2.14`.
   [(1.0, 1.0)]
 
 The constructor also accepts another `LineString` instance, thereby making a
-copy.
+shallow copy.
 
 .. code-block:: pycon
 
@@ -457,7 +457,7 @@ Defining coordinate values are accessed via the `coords` property.
   [(0.0, 0.0), (1.0, 1.0), (1.0, 0.0), (0.0, 0.0)]
 
 The `LinearRing` constructor also accepts another `LineString` or `LinearRing`
-instance, thereby making a copy.
+instance, thereby making a shallow copy.
 
 .. code-block:: pycon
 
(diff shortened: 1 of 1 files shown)


docs/manual.rst
```diff
@@ -321,7 +321,7 @@ Coordinates may also be sliced. `New in version 1.2.14`.
   [(0.0, 0.0)]
 
 The `Point` constructor also accepts another `Point` instance, thereby making
-a copy.
+a shallow copy.
```
sgillies, line 324: I don't think shallow copy quite captures the new behavior. How about something like this:

When given a `Point` instance, the `Point` constructor returns a reference to its argument. It does not make a copy.

sgillies (changes requested):
@H0R5E thank you for the contribution! I see what you're getting it, but I think it will be better to replace the mentions of copying with text that says no copy is made.

docs/manual.rst
```diff
@@ -566,10 +566,10 @@ To obtain a polygon with a known orientation, use
 
 .. function:: shapely.geometry.polygon.orient(polygon, sign=1.0)
 
-  Returns a properly oriented copy of the given polygon. The signed area of the
-  result will have the given sign. A sign of 1.0 means that the coordinates of
-  the product's exterior ring will be oriented counter-clockwise and the interior
-  rings (holes) will be oriented clockwise.
+  Returns a properly oriented deep copy of the given polygon. The signed area
```
sgillies, line 569: Yes.

docs/manual.rst
```diff
@@ -457,7 +457,7 @@ Defining coordinate values are accessed via the `coords` property.
   [(0.0, 0.0), (1.0, 1.0), (1.0, 0.0), (0.0, 0.0)]
 
 The `LinearRing` constructor also accepts another `LineString` or `LinearRing`
-instance, thereby making a copy.
+instance, making a deep copy or a shallow copy, respectively.
```
sgillies, line 460:
```suggestion
instance, returning a deep copy or a reference, respectively.
```

H0R5E:
@sgillies, thanks for the feedback. I was trying to leave a minimal footprint with my changes, but I'm happy to go further.

jorisvandenbossche (approved):
Thanks!

sgillies (approved):
Thank you very much @H0R5E !

PR #2199, opened 2025-01-13
FIX: remove unnecessary precision with GEOS 3.13 repr

mwtoews:
This fixes some confusing geometry repr outputs using GEOS 3.13.0:
```pycon
>>> import shapely
>>> shapely.geos_version_string
3.13.0
>>> line = shapely.LineString([(0.1, 0.1), (0.49, 0.51), (1.01, 0.89)])
>>> line.wkt
LINESTRING (0.1 0.1, 0.49 0.51, 1.01 0.89)
>>> line
<LINESTRING (0.1 0.1, 0.5 0.5, 1.01 0.9)>
```
the expected last line should be `<LINESTRING (0.1 0.1, 0.49 0.51, 1.01 0.89)>`.

This is was due to some logic in shapely for GEOS 3.13 that calls `GEOSWKTWriter_setRoundingPrecision` with a precision of 3 and trim enabled. This specific option combo does some weird things with some versions of GEOS (see https://github.com/libgeos/geos/pull/1199). A fix is to remove the call to setRoundingPrecision and keep GEOS' default.

Also, clean-up unused C code for `GeometryObject_repr`, since this is implemented in Python.

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -171,7 +171,7 @@ class BaseGeometry(shapely.Geometry):
     def __repr__(self):
         """Return a string representation of the geometry."""
         try:
-            wkt = super().__str__()
+            wkt = super().__str__()  # GeometryObject_ToWKT
         except (GEOSException, ValueError):
             # we never want a repr() to fail; that can be very confusing
             return f"<shapely.{self.__class__.__name__} Exception in WKT writer>"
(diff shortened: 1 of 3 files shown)


mwtoews:
On further inspection, for the geometry's repr, rounding precision was initially set to 3 for all GEOS versions. The primary issue is https://github.com/libgeos/geos/issues/1198 which impacts only GEOS 3.13.0. This should be resolved by GEOS 3.13.1.

PR #2200, opened 2025-01-15
DOC/TST: use scipy-doctest, update examples with complete imports

mwtoews:
This switches from using [`doctest`](https://docs.python.org/3/library/doctest.html) to [`scipy-doctest`](https://github.com/scipy/scipy_doctest) to run the doctesting. This is the same extension used by NumPy and SciPy.

The motivation for this switch is that it allows flexibility of minor floating-point differences in some outputs, and allows legacy NumPy repr formatting, e.g.:
```pycon
>>> import shapely
>>> from shapely import Point
>>> shapely.get_x(Point(1, 2))
1.0
```
instead of testing against `np.float64(1.0)` as with NumPy 2.0+. While this PR keeps the doctest CI job with NumPy 1.26.2, it runs the same locally with newer versions.

Most of the changes in this PR are to provide complete examples, and to call module-level ufuncs from the module. The previous version used (e.g.) `get_x(pt)` with an implicit `from shapely import get_x`. However, I think it is better to keep the module-level calls like `shapely.get_x(pt)`. An exception to this rule is the geometry classes (`Point`, etc.).

Other changes are to set conftest.py to the module root `shapely/`, and this is inspired directly from [NumPy's `numpy/conftest.py`](https://github.com/numpy/numpy/blob/v2.2.1/numpy/conftest.py). This allows some options  for scipy-doctest and fine-tuning filtering warnings in the doctests.


sgillies (commented):
@mwtoews obscuring the real type of values in the doctest might be a little misleading, though? Showing that a value is a Python int when it really is a numpy typed object could trip up somebody.

jorisvandenbossche:
It's certainly a trade-off, but the benefit of keeping the output as it is shown now in the docstrings (and so adopting scipy-doctest to be able to keep doing that also when updating numpy > 2) is that it gives less visual clutter in the example and focuses more on the message to show to the user. 

If we show an example like:

```
>>> shapely.has_z(Point(0, 0))
False
```

the main message is that it returns False for a XY point. If this shows `np.False_` (or `np.float64(1.0)` for 1.0), that feels like unnecessary clutter and potentially distracting the reader.

It's like putting a `print(..)` around each example, except for not doing that explicitly..

sgillies (approved):
Okay, I'm convinced about relaxing the type check.

mwtoews:
If/when NumPy changes their docstrings, then I'd be convinced to change it here too. For instance, see [numpy.sin](https://numpy.org/devdocs/reference/generated/numpy.sin.html) where currently `np.sin(np.pi/2.)` shows `1.0` rather than `np.float(1.0)` as with modern versions of NumPy.

Also, I should note that scipy-doctest has an option `dt_config.strict_check = True` that does dtype checks where applicable, which is related to https://github.com/numpy/numpy/pull/27272/commits/90c34cbbe8c87d49f17a7e657b4c4805797111a7. If we were to enable this feature, it would require changes like:
```pycon
>>> shapely.get_coordinate_dimension(Point(0, 0))
np.int32(2)
```
but this option still relaxes `np.float64(0.0)` stuff as `0.0`. But I don't think it's really important to most users that (e.g.) coordinate dimensions have a int32 type, so I didn't enable this feature. But perhaps it's more important for users to see (e.g.)
```pycon
>>> shapely.get_srid(shapely.set_srid(Point(0, 0), 1234567890))
np.int32(1234567890)
```

jorisvandenbossche:
> But perhaps it's more important for users to see (e.g.)

That also doesn't seem that important to know the SRID is returned as an int32 vs int64 (SRID numbers are typically smaller numbers than the int32 limits)

shapely/algorithms/polylabel.py
```diff
@@ -29,8 +29,9 @@ def polylabel(polygon, tolerance=1.0):
     shapely.errors.TopologicalError
         If the input polygon is not a valid geometry.
 
-    Example
-    -------
+    Examples
+    --------
+    >>> from shapely.algorithms.polylabel import polylabel
```
jorisvandenbossche, line 34:
```suggestion
    >>> from shapely.ops import polylabel
```

In the manual.rst we use this import

shapely/constructive.py
```diff
@@ -1131,53 +1158,54 @@ def snap(geometry, reference, tolerance, **kwargs):
...
-    >>> snap(point1, multipoint, 1)
+    >>> shapely.snap(point1, multipoint, 1)
     <POINT Z (0 0 1)>
 
     Snapping to an empty geometry has no effect:
 
-    >>> snap(line1, LineString([]), 0.25)
-    <LINESTRING (0.1 0.1, 0.49 0.51, 1.01 0.89)>
+    >>> shapely.snap(line1, LineString([]), 0.25).wkt  # TODO: why is '.wkt' needed?
+    'LINESTRING (0.1 0.1, 0.49 0.51, 1.01 0.89)'
```
jorisvandenbossche, line 1194: Should this todo get resolved first?

shapely/geometry/multilinestring.py
```diff
@@ -28,6 +28,7 @@ class MultiLineString(BaseMultipartGeometry):
     --------
     Construct a MultiLineString containing two LineStrings.
 
+    >>> from shapely.geometry import MultiLineString
```
jorisvandenbossche, line 31:
```suggestion
    >>> from shapely import MultiLineString
```

shapely/geometry/point.py
```diff
@@ -34,6 +34,7 @@ class Point(BaseGeometry):
     --------
     Constructing the Point using separate parameters for x and y:
 
+    >>> from shapely.geometry import Point
```
jorisvandenbossche, line 37:
```suggestion
    >>> from shapely import Point
```

shapely/geometry/point.py
```diff
@@ -151,11 +152,12 @@ def xy(self):
 
         Examples
         --------
-          >>> x, y = Point(0, 0).xy
-          >>> list(x)
-          [0.0]
-          >>> list(y)
-          [0.0]
+        >>> from shapely.geometry import Point
```
jorisvandenbossche, line 155:
```suggestion
        >>> from shapely import Point
```

shapely/geometry/polygon.py
```diff
@@ -44,6 +44,7 @@ class LinearRing(LineString):
     --------
     Construct a square ring.
 
+    >>> from shapely.geometry import LinearRing
```
jorisvandenbossche, line 47:
```suggestion
    >>> from shapely import LinearRing
```

shapely/geometry/polygon.py
```diff
@@ -212,6 +213,7 @@ class Polygon(BaseGeometry):
     --------
     Create a square polygon with no holes
 
+    >>> from shapely.geometry import Polygon
```
jorisvandenbossche, line 216:
```suggestion
    >>> from shapely import Polygon
```

shapely/ops.py
```diff
@@ -506,11 +506,13 @@ def split(geom, splitter):
         splitter : geometry
             The geometry that will split the input geom
 
-        Example
-        -------
+        Examples
+        --------
+        >>> import shapely.ops
+        >>> from shapely.geometry import Point, LineString
```
jorisvandenbossche, line 512:
```suggestion
        >>> from shapely import Point, LineString
```

jorisvandenbossche (approved):
Looks good!

Added some small comments for consistency

jorisvandenbossche:
I am going to merge those small suggestions, so I can merge this PR now to avoid another round of updating an fixing merge conflicts if I continue merging other PRs!

shapely/constructive.py
jorisvandenbossche, line 1194:
```suggestion
    >>> shapely.snap(line1, LineString([]), 0.25)
    <LINESTRING (0.1 0.1, 0.49 0.51, 1.01 0.89)>
```

jorisvandenbossche:
Thanks @mwtoews!

PR #2204, opened 2025-01-16
BLD: build linux aarch64 wheels on github actions

jorisvandenbossche:
Now there are linux arm64 runners available for public repos (https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/), we can try to move this configuration to github actions, potentially being able to do away with circle ci (and the manual wheel download/upload step).


.github/workflows/release.yml
```diff
@@ -53,21 +53,23 @@ jobs:
...
-          # - baseimage: manylinux2014_aarch64
-          #   cibw_build: "*manylinux_aarch64"
-          #   arch: aarch64
+          - os: ubuntu-24.04-arm
+            baseimage: manylinux2014_aarch64
+            cibw_build: "*manylinux_aarch64"
+            arch: aarch64
           # - baseimage: musllinux_1_2_aarch64
           #   cibw_build: "*musllinux_aarch64"
           #   arch: aarch64
```
mwtoews: It's possible to enable this too, since these numpy bdist wheels are available.

jorisvandenbossche: I had just done that and it seems to work!

PR #2211, opened 2025-01-30
Release 2.0.7

### LINKED ISSUE #2208: 2.0.7 release
@jorisvandenbossche support for gcc 14 hasn't been backported yet, right? What are the other most significant bug fixes we might want to backport?

sgillies:
Resolves #2208

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,7 +37,7 @@ jobs:
           python -m build --sdist
           twine check --strict dist/*
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
           path: ./dist/*.tar.gz
           retention-days: 30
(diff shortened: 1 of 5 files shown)


jorisvandenbossche (approved):

.github/workflows/release.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           python -m build --sdist
           twine check --strict dist/*
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
```
jorisvandenbossche, line 40: See the commit that changes this on main: https://github.com/shapely/shapely/commit/bc4632646009429952f344b8c7e11107c0fd513d

I think you will have to copy over the other changes as well to make this fully working (I assume github actions will also complain about the v3 in `actions/download-artifact@v3`)

jorisvandenbossche:
@sgillies I think it might be best that you remove the tag again from the commit in this PR. It has actually been useful that you did this to notice that the publish step is failing right now. But in general the release workflow assumes that the tag is done on the actual maint-2.0 branch

(if we want to be able to test doing a release before doing the actual release, we should set up the workflow with test-pypi I assume, so we can do that from a PR first. But for now I always just released from the maint branch as described in the wiki)

sgillies:
Yes, I see now that the Circle CI config has us building only on pushes to main and maint-2.0. I'll delete the tag, merge this and then force the tag on maint-2.0.

PR #2214, opened 2025-02-04
BUG: Gracefully handle calling orient() on empty polygon

sea-bass:
I recently had a user run into an error they couldn't figure out. Upon analyzing the issue, I pinpointed it to a very simple use case of trying to call `shapely.geometry.polygon.orient()` on an empty `Polygon` as follows:

```python
orient(Polygon())
```

This yields the following error:

```
  File "/home/sebastian/python-virtualenvs/pyrobosim/lib/python3.12/site-packages/shapely/geometry/polygon.py", line 346, in orient
    if signed_area(ring) / s >= 0.0:
       ^^^^^^^^^^^^^^^^^
  File "/home/sebastian/python-virtualenvs/pyrobosim/lib/python3.12/site-packages/shapely/algorithms/cga.py", line 11, in signed_area
    xs, ys = np.vstack([coords, coords[1]]).T
                                ~~~~~~^^^
IndexError: index 1 is out of bounds for axis 0 with size 0
```

Here is a simple check on polygon emptiness to prevent this in the future!

### DIFF
--- a/shapely/geometry/polygon.py
+++ b/shapely/geometry/polygon.py
@@ -321,6 +321,9 @@ shapely.lib.registry[3] = Polygon
 
 def orient(polygon, sign=1.0):
     """Return an oriented polygon."""
+    if polygon.is_empty:
+        return polygon
+
     s = float(sign)
     rings = []
     ring = polygon.exterior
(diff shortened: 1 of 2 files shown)


sgillies (approved):
Thanks @sea-bass !

PR #2216, opened 2025-02-10
STRtree: fix capitalization in the docs

adamjstewart:
These were the only misspellings I found.


sgillies (approved):
Thanks @adamjstewart !

PR #2227, opened 2025-02-16
MAINT: drop Python 3.9 support

jorisvandenbossche:
As discussed in https://github.com/shapely/shapely/issues/2119, similar like https://github.com/shapely/shapely/pull/2127 for Python 3.8.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,28 +17,28 @@ jobs:
         geos: [3.9.5, 3.10.6, 3.11.4, 3.12.2, 3.13.0, main]
         include:
           # 2020
-          - python: 3.9
+          - python: "3.10"
             geos: 3.9.5
-            numpy: 1.20.3
+            numpy: 1.21.6
           # 2021
           - python: "3.10"
             geos: 3.10.6
-            numpy: 1.21.3
+            numpy: 1.21.6
           # 2022
           - python: "3.11"
             geos: 3.11.4
-            numpy: 1.23.4
+            numpy: 1.23.5
           # 2023
           - python: "3.12"
             geos: 3.12.2
-            numpy: 1.26.2
+            numpy: 1.26.4
             matplotlib: true
             doctest: true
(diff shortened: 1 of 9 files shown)


mwtoews (approved):
Thanks! Looks like everything is covered, except for a minor detail here: https://github.com/shapely/shapely/blob/74eddd6af20566771ae4551a21e389e9ec188e66/docker/Dockerfile.valgrind#L7 (is this Dockerfile even used?)

jorisvandenbossche:
Good catch. Yes, we used that in the past to check for memory leaks in the C code. It's not something that is "used" automatically somewhere else in the repo or in CI though, it's for manual use.

martinfleis:
@sgillies I am not sure I understand the rationale here. If a user is running 3.9 while trying to install shapely without a pin, shapely 2.1 would not install on their machine but 2.0.x would. 

Also, note that numpy has also dropped 3.9 in  2.1 released back in August, so the latest shapely dependencies would also not install on 3.9.

mwtoews:
@sgillies this is somewhat discussed in #2119 where we'd essentially align closer to [SPEC 0](https://scientific-python.org/specs/spec-0000/) than the [supported Python versions](https://devguide.python.org/versions/). But not adhere strictly to any specific policy, as we like flexibility. Following NumPy's supported versions (which have a bit of flexibility) makes lots of practical sense, as I like to match the same cibuildwheel images to what they use.

jorisvandenbossche:
Also, by end of summer with python 3.14, we will already be supporting 5 Python versions if we just do >= 3.10. And I would like to avoid that we are releasing wheels for 6 or 7 python versions for the 2.1.x branch (like we now do for 2.0.x)

sgillies:
@martinfleis Right! I think I must be thinking of old times before PyPI metadata was reliable. As I just confirmed by trying `pip install numpy` with 3.9, the only harm is that pip uselessly fetches info about numpy 2.1.x releases. I've got no reservations then, let's do it.

PR #2228, opened 2025-02-18
DOC: clarify that polygon returned in oriented_envelope is not normalized

### LINKED ISSUE #2177: Corner points for : Oriented Envelope & minimum_rotated_rectangle

When using `oriented_envelope` or `minimum_rotated_rectangle` the corner points of the returned polygon are inconsistent. So far I've not found any underlying logic and it seems to randomly pick one of the corners. Below I show four different cases where the red dot is the `Shapely.Polygon.exterior.coords._coords[0]`.
(I've hidden the geometries I'm enveloping to simplify the visualization"
image

OpenCV's solution seems to be:
https://docs.opencv.org/3.4/db/dd6/classcv_1_1RotatedRect.html#a69d648b086f26dbce0029facae9bfb2d
{2A2B1DC2-454A-4135-A5FE-9C7E9CDBD1A0}
see also : 
https://stackoverflow.com/questions/69074165/order-of-corners-in-a-rotating-rectangle-in-opencv-python

Perhaps it would be nice to follow suit. Regardless of the strategy it would be nice to have additional documentation on this in the DocString.

## Operating system
Windows + WSL Ubuntu 20.04
(For example, Mac OS X 10.12.3)

`Shapely 2.0.6` installed from PyPI using pip

theroggy:
Closes #2177

### DIFF
--- a/shapely/constructive.py
+++ b/shapely/constructive.py
@@ -1260,8 +1260,8 @@ def oriented_envelope(geometry, **kwargs):
     coordinate axes. If the convex hull of the object is a degenerate (line
     or point) this degenerate is returned.
 
-    The starting point of the Polygon returned is not fixed. You can use
-    :func:`~shapely.normalize` to reorganize the Polygon to
+    The starting point of the rectangle is not fixed. You can use
+    :func:`~shapely.normalize` to reorganize the rectangle to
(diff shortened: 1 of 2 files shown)


jorisvandenbossche (approved):
Thanks!

PR #2231, opened 2025-02-18
DOC: Improve visibility of the Geometry creation functions

### LINKED ISSUE #1925: Include See also in docstrings to link between Polygon and polygons etc

### LINKED ISSUE #1954: Very slow MultiPoint initialization

theroggy:
Improve visibility of the Geometry creation functions:
- Add "See Also" links in the Constructor docstrings to the relevant bulk/array functions.
- Add references to them in the user guide.

Resolves #1925 
Resolves #1954

### DIFF
--- a/shapely/creation.py
+++ b/shapely/creation.py
@@ -149,9 +149,9 @@ def linestrings(
     out=None,
     **kwargs,
 ):
-    """Create an array of linestrings.
+    """Create an array of LineStrings.
 
-    This function will raise an exception if a linestring contains less than
+    This function will raise an exception if a LineString contains less than
     two points.
 
     Parameters
@@ -694,7 +694,7 @@ def multipolygons(geometries, indices=None, *, out=None, **kwargs):
 @deprecate_positional(["indices"], category=DeprecationWarning)
 @multithreading_enabled
 def geometrycollections(geometries, indices=None, out=None, **kwargs):
-    """Create geometrycollections from arrays of geometries.
+    """Create an array of GeometryCollections.
 
     Parameters
     ----------
--- a/shapely/creation.py
+++ b/shapely/creation.py
@@ -199,8 +199,8 @@ def linestrings(
 
     See Also
     --------
-    linearrings
-    multilinestrings
+    linearrings : Create an array of LinearRings.
+    multilinestrings : Create an array of MultiLineStrings.
 
     Examples
     --------
@@ -301,9 +301,9 @@ def linearrings(
 
     See Also
     --------
-    linestrings
-    polygons
-    multipolygons
+    linestrings : Create an array of LineStrings.
+    polygons : Create an array of Polygons.
+    multipolygons : Create an array of MultiPolygons.
 
     Examples
     --------
@@ -634,8 +634,8 @@ def multilinestrings(geometries, indices=None, *, out=None, **kwargs):
 
     See Also
     --------
-    linestrings
-    multipoints
+    linestrings : Create an array of LineStrings.
+    multipoints : Create an array of MultiPoints.
 
     """
     typ = GeometryType.MULTILINESTRING
--- a/shapely/creation.py
+++ b/shapely/creation.py
@@ -102,7 +102,7 @@ def points(
 
     See Also
     --------
-    multipoints
+    multipoints : Create an array of MultiPoints.
 
     Examples
     --------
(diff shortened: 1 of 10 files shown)


shapely/geometry/collection.py
```diff
@@ -18,6 +18,11 @@ class GeometryCollection(BaseMultipartGeometry):
     geoms : sequence
         A sequence of Shapely geometry instances
 
+    See Also
+    --------
+    geometrycollections : Create geometrycollections from arrays of geometries.
```
jorisvandenbossche, line 23:
```suggestion
    geometrycollections : Create an array of GeometryCollections.
```

?

shapely/geometry/linestring.py
```diff
@@ -23,6 +23,11 @@ class LineString(BaseGeometry):
         an array-like with shape (N, 2) or (N, 3).
         Also can be a sequence of Point objects, or combination of both.
 
+    See Also
+    --------
+    linestrings : Create an array of linestrings.
```
jorisvandenbossche, line 28:
```suggestion
    linestrings : Create an array of LineStrings.
```

I was going to say that I think we mostly use the camel case form (at least in this docstring), but definitely not consistently because the docstrings of the creation functions we are linking to indeed use the lower case version ..

theroggy: Yes, it is indeed not very consistent in general, and I indeed just reused the docstring of the creation function.

Maybe one or the other approach could be choosen and then make a PR to change it everywhere to that?

sgillies: @theroggy @jorisvandenbossche In my original manual, I tried to use "polygon" (for example) when talking abstractly about what a function does, like in the first paragraph of a docstring. But then I tried to always use "Polygon" when talking about the implementation. I didn't always succeed, but this was the intention.

Python's own docs are written this way. See https://docs.python.org/3/library/stdtypes.html#string-methods for example, and `str.split()` specifically:

> Return a list of the words in the string, using sep as the delimiter string.

It says "string" and not "str", because it is emphasizing the concepts, not the implementations. So, we can do the same thing. We can use "line string" when we want to highlight the concept and "LineString" when we want to emphasize the implementation.

jorisvandenbossche: That sounds as a good general policy to me! (although of course it will then still always be a bit ambiguous which of the two concepts you are talking about ;))

Here, we are creating the actual objects of class LineString. So would we then say `Create an array of LineStrings.`?

theroggy: I agree it is a good policy in theory, but indeed I'm afraid it will require some serious thinking and possibly discussing for each and every instance of these words...

shapely/creation.py
```diff
@@ -446,7 +461,7 @@ def polygons(geometries, holes=None, indices=None, *, out=None, **kwargs):
 @deprecate_positional(["ccw"], category=DeprecationWarning)
 @multithreading_enabled
 def box(xmin, ymin, xmax, ymax, ccw=True, **kwargs):
-    """Create box polygons.
+    """Create a box Polygon.
```
jorisvandenbossche, line 464:
```suggestion
    """Create a box Polygon (or an array of boxes).
```

All the others are "create an array of ..", and I understand here it is useful to emphasize it is in the first place also useful to create a single box Polygon (since there is no class instance constructor alternative for the scalar case), but at the same time good to still mention it can also create an array?

jorisvandenbossche (approved):
Apologies for the slow follow-up here! But this is looking good, just one small comment

jorisvandenbossche:
Thanks @theroggy!

PR #2232, opened 2025-02-18
DOC: improve doc regarding validness of polygons

### LINKED ISSUE #1292: Unexpected: a Polygon without coordinates is valid
I was quite surprised to find out that a Polygon without coordinates is valid. Not sure if that is a feature or a bug. 

Actual behavior: 

```python
>>> from shapely.geometry import Polygon
>>> Polygon([]).is_valid
True
```

Expected behavior: 

```python
>>> from shapely.geometry import Polygon
>>> Polygon([]).is_valid
False
```

See above

## Operating system

```shell
$ uname -srm
Linux 5.13.0-23-generic x86_64
$ cat /etc/os-release | grep PRETTY
PRETTY_NAME="Ubuntu 21.10"
```

```python
>>> import shapely
>>> shapely.__version__
'1.7.1'
```

Installed through pip.

theroggy:
Resolves #1292


mwtoews (approved):
Looks good to resolve the issue, thanks!

docs/manual.rst
```diff
@@ -499,8 +500,6 @@ Figure 4. On the left, a `Polygon` that is `invalid` because its exterior and
 interior rings touch along a line, and on the right, a `Polygon` that is
 `invalid` because its interior rings touch along a line.
 
-A `Polygon` has non-zero area and non-zero length.
```
jorisvandenbossche, line 502: This was removed on purpose? (I know it is not entirely correct because an empty polygon has a zero area .. but this sentence also served as intro to the next code block showing the area and length attribute)

theroggy: Yes, the issue being resolved was the comment that this sentence is misleading as it implies that an empty Polygon is invalid which is not the case...

Maybe this gives the best of both worlds?
`A `Polygon` has non-zero area and non-zero length if it isn't empty.`

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

PR #2233, opened 2025-02-20
MAINT: update ruff format/check to 0.9.6 with associated fixes

mwtoews:
Most of these changes were automated after updating Ruff from version v0.5.2 to v0.9.6 (latest). They are a combination of changes from `ruff format` and `ruff check --fix --unsafe-fixes`, which all look good to me.


theroggy (approved):

theroggy:
I think, based on the [documentation of ruff](https://docs.astral.sh/ruff/formatter/#format-suppression) the `# fmt: on` should be on a seperate line instead of after an `)`, but I don't quite understand why the formatting should be suppressed anyway... so I think removing it is fine.

theroggy (approved):

shapely/tests/geometry/test_decimal.py
```diff
@@ -55,31 +55,31 @@
...
     ],
     [
-        (Decimal(60.0), Decimal(80.0), Decimal(1)),
-        (Decimal(80.0), Decimal(80.0), Decimal(2)),
-        (Decimal(70.0), Decimal(60.0), Decimal(3)),
-        (Decimal(60.0), Decimal(80.0), Decimal(1)),
+        (Decimal("60.0"), Decimal("80.0"), Decimal(1)),
+        (Decimal("80.0"), Decimal("80.0"), Decimal(2)),
+        (Decimal("70.0"), Decimal("60.0"), Decimal(3)),
+        (Decimal("60.0"), Decimal("80.0"), Decimal(1)),
```
sgillies, line 82: Interesting. I like this choice.

sgillies (approved):
I love sorted lists of function and method names!

PR #2234, opened 2025-02-21
ENH: Add `include_m` to `get_coordinates`, warn with positional args

mwtoews:
This enhancement allows `shapely.get_coordinates()` to get M coordinate data with `include_m=True`. For shapely 2.1, this argument is after the keyword-only (`, *,`) marker. A DeprecationWarning is shown if the function is called with positional arguments for `include_z` or `return_index`, since a future version of shapely may require these as keyword-only. In the code I'm suggesting this be changed to a keyword-only signature with shapely 2.2, but this change schedule can be postponed later if needed.

From a low-level API, here are some notes:

- No changes to [`coordseq_to_buffer(..., int has_z, int has_m)`](https://github.com/shapely/shapely/blob/d62b3a9cfc35638789b4a31d2e83bad672a0753f/src/geos.h#L209-L210) which has a compatible signature with `has_m`.
- Change `GetCoords()` and `PyGetCoords()` (aka `shapely.lib.get_coordinates()`) to have signature order `..., include_z, include_m, return_index`.
- With C API in src/coords.c, change all `get_coordinates*()` functions to add `include_m` at the end.

Next steps:

- Update `set_coordinates` and check `coordseq_from_buffer` and other low- and high-level API functions
- Update `transform()` with `include_m: bool | None = False`
- Update `to_ragged_array` with `include_m=None`

### DIFF
--- a/shapely/decorators.py
+++ b/shapely/decorators.py
@@ -1,6 +1,8 @@
 """Decorators for Shapely functions."""
 
+import inspect
 import os
+import warnings
 from functools import wraps
 
 import numpy as np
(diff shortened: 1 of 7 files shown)


shapely/decorators.py
```diff
@@ -88,3 +90,44 @@ def wrapped(*args, **kwargs):
                 arr.flags.writeable = old_flag
 
     return wrapped
+
+
+def deprecate_positional(should_be_kwargs, category=DeprecationWarning):
```
jorisvandenbossche, line 95: Nice! 

Not for this PR, but I was thinking that we could also deprecate passing the first argument as a keyword argument, i.e. passing it _not_ positional.  
We are quite consistent using `geometry` for functions taking a single geometry/array (but I see we also have some use `geometries`), and here the keyword is quite descriptive (although still unnecessary I would say). But for example I don't think anyone should do `shapely.intersection(a=.., b=..)`. I don't know if it is worth deprecating, but if I would write those APIs now again, I think I would use positional-only keywords for those.

jorisvandenbossche (approved):

PR #2236, opened 2025-02-24
DOC: fix typo in LinearRing manual - can be constructed using a sequence of Points

nicohmje:
Pretty sure this isn't right. LinearRing instances can be constructed using a sequence of points, like LineStrings


jorisvandenbossche (approved):
Thanks for the catch and contribution to fix it! Constructing from Points indeed works and I think we indeed meant to say so

PR #2237, opened 2025-02-24
DEPR: deprecate positional args in some functions

mwtoews:
This adds a few deprecation warnings to a few functions:

- `shapely.transform()` - warns about positional `include_z`, also described in docs
- `shapely.points()` - warns about positional `indices`
- `shapely.linestrings()` - warns about positional `indices`
- `shapely.linearrings()` - warns about positional `indices`

The last three creation functions don't note the deprecation warning, as I'm guessing it's not commonly used? <s>The tests appeared to be missing use with indices, so these tests are added (both as positional and keyword).</s> (Found the creation tests with "indices" in a separate test file.)

Xref to #2234 that introduced the decorator function and raised warnings for `shapely.get_coordinates`.

Are there any other public functions that have positional arguments that should probably be shifted towards keyword-only over time? If so, mention them here, so the deprecation warnings can be introduced for shapely 2.1.

### DIFF
--- a/shapely/_geometry.py
+++ b/shapely/_geometry.py
@@ -5,7 +5,11 @@ import numpy as np
 
 from shapely import _geometry_helpers, geos_version, lib
 from shapely._enum import ParamEnum
-from shapely.decorators import multithreading_enabled, requires_geos
+from shapely.decorators import (
+    deprecate_positional,
+    multithreading_enabled,
+    requires_geos,
+)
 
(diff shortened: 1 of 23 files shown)


jorisvandenbossche (approved):
Looks good to me, thanks!

jorisvandenbossche (approved):

sgillies (approved):

Issue #2171, opened 2024-10-11
Shapely 2.1 + GEOS 3.12 now loads LineString M's properly, but .coords drops the M 


sentientnebula:
from shapely import wkt

# Works as expected with LINESTRING Z's
g = wkt.loads('LineString Z (1 2 3, 4 5 6)')
g
<LINESTRING Z (1 2 3, 4 5 6)>
list(g.coords)
[(1.0, 2.0, 3.0), (4.0, 5.0, 6.0)]

# Returns a LineString M now, but g.coords drops the M coordinate 
g = wkt.loads('LineString M (1 2 3, 4 5 6)')
g
<LINESTRING M (1 2 3, 4 5 6)>  <-- Now correct in 2.1
list(g.coords)
[(1.0, 2.0), (4.0, 5.0)] <-- Drops M coordinate

Expected Behavior:
list(g.coords)
[(1.0, 2.0, 3.0), (4.0, 5.0, 6.0)]

jmealo:
@sentientnebula: Where did you get Shapely 2.1? I only see 2.0.6 on GitHub and PyPi? 🤔

mwtoews:
Development "nightly" wheels can be installed [using these docs](https://shapely.readthedocs.io/en/latest/installation.html#installation-of-the-development-version-using-nightly-wheels)

jorisvandenbossche:
For a general discussion about how to add better support for M values across shapely's functionality, see https://github.com/shapely/shapely/issues/1648

PR #2238, opened 2025-02-25
FIX: allow .coords to include M coordinates if present

mwtoews:
This minor fix is an oversight related to #2234 where M coordinates should appear in the `.coords` property if present in the geometry.


jorisvandenbossche (approved):

jorisvandenbossche:
Good catch!

This also closes https://github.com/shapely/shapely/issues/2171 then, I think

PR #2239, opened 2025-02-26
ENH: Add m to points() / Point() functions

mwtoews:
This enhancement adds M coordinate support for point geometries, for example:
```pycon
>>> import shapely
>>> shapely.points(1, 2, 3, m=4)
<POINT ZM (1 2 3 4)>
>>> shapely.Point(1, 2, None, 4)
<POINT M (1 2 4)>
```

Currently these creation functions "work" with GEOS<3.12.0, but the M coordinates are ignored, creating XY or XYZ geometries instead of XYM and XYZM geometries. Should these attempts with older GEOS raise warnings?

The main `shapely.points()` function adds a few keyword-only arguments, `m=None` (array_like, optional) and `coords_has_m=None` (bool, optional). The last is used to evaluate `coords` with shape (N, 3) as either XYZ (default with `coords_has_m=None` or `False`) or XYM (`coords_has_m=True`).

It also changes some of the C API to allow other geometries to accept M coordinates, but this is a "minimal" PR to enable just point creations. A few "TODO" notes are added for future work.

Both the ufunc `points` and Cython `simple_geometries_1d()` functions replace `handle_nan` with an `options` int. This keeps the ufunc signature `(d),()->()` with a scalar to pass the options for the array. These options contain a simple bitmask for `geom_type`, `has_z`, `has_m` and `handle_nan`.

Here is a summary of other low-level API changes:

- `PyGEOS_CoordSeq_FromBuffer()`: replaces `unsigned int dims` with `char has_z, char has_m`
- `coordseq_from_buffer()`: replaces `unsigned int dims` with `char has_z, char has_m`
- Replace `PyGEOS_create3DEmptyPoint(GEOSContextHandle_t ctx)` with `PyGEOS_createEmptyPoint(GEOSContextHandle_t ctx, char has_z, char has_m)`
- `PyGEOS_createPoint`: add `double* m` parameter
- `create_point`: add `double* m` parameter

Next steps:

- [ ] Add M to `linestrings` / `LineString`
- [ ] Add M to collections like `multipoints` / `MultiPoint`

### DIFF
--- a/shapely/_geometry_helpers.pyx
+++ b/shapely/_geometry_helpers.pyx
@@ -67,7 +67,8 @@ cdef int _create_simple_geometry(
     const double[:, :] coord_view,
     Py_ssize_t idx,
     unsigned int n_coords,
-    unsigned int dims,
+    char has_z,
+    char has_m,
     int geometry_type,
     char is_ring,
     int handle_nan,
@@ -82,7 +83,7 @@ cdef int _create_simple_geometry(
     cdef unsigned int actual_n_coords = 0
 
     errstate = PyGEOS_CoordSeq_FromBuffer(
-        geos_handle, &coord_view[idx, 0], n_coords, dims,
+        geos_handle, &coord_view[idx, 0], n_coords, has_z, has_m,
         is_ring, handle_nan, &seq
     )
     if errstate != PGERR_SUCCESS:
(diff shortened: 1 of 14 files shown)


jorisvandenbossche:
(moved this to the 2.3 milestone)

PR #2241, opened 2025-02-27
coverage_simplify: raise error for non-polygonal elements instead of returning original input

### LINKED ISSUE #2240: Coverage simplify: raise error if there are non-polygonal geometries instead of silently returning input data?
Follow-up issue for https://github.com/shapely/shapely/pull/1969/, adding `shapely.coverage_simplify`.

I was demoing this new feature and trying to use it on the countries of Africa from the countries dataset at geoBoundaries (https://github.com/wmgeolab/geoBoundaries/raw/main/releaseData/CGAZ/geoBoundariesCGAZ_ADM0.gpkg). Running `coverage_simplify` on the original data gives a _"GEOSException: IllegalArgumentException: point array must contain 0 or >1 elements"_ error. Potentially something to report, but the data also turns out to contain quite some invalid geometries with small self-intersections, so in any case this is not guaranteed to work for such data. 
After calling `make_valid()` on the data and trying again, `coverage_simplify()` just silently did nothing. I was assuming this was because the data is not a good enough coverage (`coverage_is_valid()` also returns False).

But after some more looking at the data, this actually appeared to be because after `make_valid()`, some of the polygons turned into GeometryCollections with the polygon and a small linestring. With shapely 2.1 we can use the new keywords in `make_valid()` to get rid of those line elements. After doing that, `coverage_simplify` seems to work just fine (even though `coverage_is_valid` is still indicating it is not a proper coverage, although I have to expose a keyword to tweak the gap width to see if that helps).

So currently `shapely.coverage_simplify()` will return the input data if there are non-polygonal elements in the array (as documented). But I am wondering if it would be more user-friendly if we actually raise an error in that case? 
It would have been more informative in my case to figure out what was going on. But I don't know if there are good use cases for wanting to just get back the original result? cc @martinfleis

jorisvandenbossche:
Closes #2240


martinfleis:
Seeing the segfaults, it comes back to me that that might have been the reason! It was apparently fixed in GEOS as I reported that in https://github.com/libgeos/geos/issues/1031 but the reality is probably more complicated.

jorisvandenbossche:
Yeah, so it is specifically segfaulting on GEOMETRYCOLLECTION EMPTY. So in any case we will need to keep some validation on our side.

martinfleis:
Well, that is at least an improvement. GEOS 3.12.1 was segfaulting on anything else than polygons.

jorisvandenbossche:
It also seems to segfault on a GeometryCollection with polygon input. I was initially thinking that this might actually work because GEOS is doing

```
        for (const Geometry* g: m_input) {
            if (!g->isPolygonal())
                throw util::IllegalArgumentException("Argument is non-polygonal");
        }
```

for the input passed to the operation. And this `isPolygonal` will return True for a collection with a polygon, AFAIU. But then probably the rest of the code might not be expecting collections .. (something to report to GEOS).

But so will add back the validation in the C code to avoid segfaults. And then on the Python side we can still decide if we want to automatically apply it only on the Polygons/MultiPolygons or not.

PR #2243, opened 2025-02-28
DEPR: deprecate 'resolution' argument for BaseGeometry.buffer()

mwtoews:
Back in shapely-1.8, the `BaseGeometry.buffer()` function used `resolution=`
https://github.com/shapely/shapely/blob/9cf3cdfae7c9aa01b56c7151a51e6a517cceabdd/shapely/geometry/base.py#L541-L543
Since shapely-2.0 this has been renamed `quad_segs=`. This shows a FutureWarning if `resolution` is used.

### DIFF
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -561,7 +561,7 @@ class BaseGeometry(shapely.Geometry):
         if resolution is not None:
             warn(
                 "The 'resolution' argument is deprecated. Use 'quad_segs' instead",
-                FutureWarning,
+                DeprecationWarning,
                 stacklevel=2,
             )
             quad_segs = resolution
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -471,9 +471,6 @@ class BaseGeometry(shapely.Geometry):
         ----------
         distance : float
             The distance to buffer around the object.
-        resolution : int, optional
-            The resolution of the buffer around each vertex of the
-            object.
         quad_segs : int, optional
             Sets the number of line segments used to approximate an
             angle fillet.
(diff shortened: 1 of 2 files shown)


shapely/geometry/base.py
```diff
@@ -560,9 +557,13 @@ def buffer(
             )
             quad_segs = quadsegs
 
-        # TODO deprecate `resolution` keyword for shapely 2.1
         resolution = kwargs.pop("resolution", None)
         if resolution is not None:
+            warn(
+                "The 'resolution' argument is deprecated. Use 'quad_segs' instead",
+                FutureWarning,
```
jorisvandenbossche, line 564: Maybe also first a DeprecationWarning as you did for the positional/keyword arguments?

jorisvandenbossche (approved):

shapely/geometry/base.py
mwtoews: There is also a [`ShapelyDeprecationWarning`](https://github.com/shapely/shapely/blob/e74510756546656cc85869c73cbd7ad3ec9631e7/shapely/errors.py#L47-L48). This is a confusing topic for me, somewhat summarised in [this Q/A](https://stackoverflow.com/q/55377831/). My take on is is that DeprecationWarning is not visible to everyone (but shown for ipython), whereas FutureWarning is shown for everyone, thus is more visible. Why I didn't use FutureWarning or ShapelyDeprecationWarning for positional args? Not sure, but if we need to extend the deprecation for another release then it could be "upgraded" to a more visible FutureWarning or ShapelyDeprecationWarning. Opinions welcome!

sgillies: @mwtoews @martinfleis If I read https://peps.python.org/pep-0565/#additional-use-case-for-futurewarning correctly, we're right to not use FutureWarning. If GEOS, say, had a config file, and we were switching to a new major version with altered configuration, we might want to issue a FutureWarning for that.

mwtoews: I'm going to go with DeprecationWarning for "resolution", but keeping "quadsegs" unchanged with FutureWarning. They can both be removed at the same time "in a future release".

jorisvandenbossche (approved):

PR #2245, opened 2025-03-01
DEPR: deprecate 'project' and 'interpolate' linear referencing methods

mwtoews:
This proposed deprecation is for two linear referencing methods in the BaseGeometry class:

- Deprecate `project()`, use [`line_locate_point()`](https://shapely.readthedocs.io/en/latest/reference/shapely.LineString.html#shapely.LineString.line_locate_point) instead
- Deprecate `interpolate()`, use [`line_interpolate_point()`](https://shapely.readthedocs.io/en/latest/reference/shapely.LineString.html#shapely.LineString.line_interpolate_point) instead

The `project()` and `interpolate()` methods have been around since early shapely releases, whereas `line_locate_point()` and `line_interpolate_point()` were introduced with shapely 2.0 via PyGEOS.

While the new method names are longer, their names are more likely help users describe what they do (without needing to look up the docs). The meanings of "project" and "interpolate" vary in different geospatial contexts, thus they are less clear to users what exactly they do.

What do folks think about this proposal?

### DIFF
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -840,7 +840,7 @@ Shapely supports linear referencing based on length or distance, evaluating the
 distance along a geometric object to the projection of a given point, or the
 point at a given distance along the object.
 
-.. method:: object.interpolate(distance[, normalized=False])
+.. method:: object.line_interpolate_point(distance[, normalized=False])
 
   Return a point at the specified distance along a linear geometric object.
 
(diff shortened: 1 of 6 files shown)


sentientnebula:
I find them less clear, redundant, and overly verbose.

line.interpolate() is much cleaner than line.line_interpolate_point().

Why do I need to be reminded that my object is a line?  interpolate_point() without the line_ prefix would be OK if you want to be super explicit.  I'm not sure what else other than a point you'd be interpolating on a line, but line.interpolate_point is a lot better than line.line_interpolate_point.

I don't see the advantage of these depreciation.   As it stands now people who like the verbose versions can use them, and those who don't don't have to.

mwtoews:
Thanks all for speaking up, all opinions are welcome. As there is no rush to this decision, I'm postponing it for the next release milestone.

mwtoews:
Thanks all for the feedback! I'll drop this proposal, and `project()` / `interpolate()` will continue as maintained aliases.

PR #2248, opened 2025-03-03
ENH - coverage validation: add coverage_invalid_edges and expose gap_width

jorisvandenbossche:
Follow-up on https://github.com/shapely/shapely/pull/2156, adding the `gap_width` keyword to `coverage_is_valid()`, and also exposing the ability to return the invalid edges as a separate `coverage_invalid_edges()` (both use the same GEOS function under the hood, `GEOSCoverageIsValid_r`, which does both at the same time).

Still need to clean-up the implementation, add more tests, and add/update the docstring.


martinfleis:
I know that we discussed this and have no answer at the moment but the docstring of `coverage_is_valid` would really benefit from clearly spelled out relationship between its evaluation and `coverage_*` methods. If you get True, I suppose all should work. If you get False, `coverage_simplify` apparently _may_ work. In my experience, `coverage_union` is way more demanding in terms of proper topology, so shall I assume that it won't work? And than the main question - how does the `gap_width` relate to all of these?

jorisvandenbossche:
> And than the main question - how does the `gap_width` relate to all of these?

I added some small tests (but we should probably add some _visual_ examples to make this really clear to the reader) to see the effect (although one part I still didn't fully get, there is a question in the test code)>
So _"gaps are holes in the coverage which are narrower than a given width."_, and then you can set the maximum gap width, if narrow gaps are to be detected (because by default, gaps will be considered valid).

Quoting some more elaborate comments in the C++ GEOS code:

> A valid coverage may contain holes (regions of no coverage).
Sometimes it is desired to detect whether coverages contain narrow gaps between polygons (which can be a result of digitizing error or misaligned data).
This class can detect narrow gaps, by specifying a maximum gap width using `setGapWidth(double)`. Note that this will also identify narrow gaps separating disjoint coverage regions, and narrow gores. In some situations it may also produce false positives (i.e. linework identified as part of a gap which is wider than the given width).
To fully identify gaps it maybe necessary to use `CoverageUnion` and analyze the holes in the result to see if they are acceptable.

jorisvandenbossche:
To use a concrete example (from the tests), this is a valid coverage (difficult to see on the image, but so the vertices align on the shared edge):

coverage

```python
>>> shapely.coverage_is_valid([poly1, poly2_shift])
np.True_
>>> shapely.coverage_invalid_edges([poly1, poly2_shift])
array([<LINESTRING EMPTY>, <LINESTRING EMPTY>], dtype=object)
```

The gap has a width of 1, so if we now specify the maximum gap width to detect to a value of1 or higher, we can have it detect as a gap, i.e. invalid coverage:

```python
# gap is wider than 0.5, so considered as a valid hole and not invalid gap
>>> shapely.coverage_is_valid([poly1, poly2_shift], gap_width=0.5)
np.True_
# now it is detected as gap -> no valid coverage anymore
>>> shapely.coverage_is_valid([poly1, poly2_shift], gap_width=1.5)
np.False_
```

And then you can find those invalid edges:

```python
>>> shapely.coverage_invalid_edges([poly1, poly2_shift], gap_width=1.5)
array([<LINESTRING (10 7, 10 3)>, <LINESTRING (10 3, 11 5, 10 7)>],
      dtype=object)
```

coverage2


What I don't entirely understand is why, if we increase the gap width further, it stops finding the gap. This feels like a bug in GEOS? Or is there something I am missing?

```python
>>> shapely.coverage_is_valid([poly1, poly2_shift], gap_width=1.9)
False
>>> shapely.coverage_is_valid([poly1, poly2_shift], gap_width=2.0)
True
>>> shapely.coverage_is_valid([poly1, poly2_shift], gap_width=3.0)
True
```

jorisvandenbossche:
Opened an issue asking about this upstream: https://github.com/libgeos/geos/issues/1251

jorisvandenbossche:
Further feedback on the documentation is very welcome, and we can always improve the documentation later, but I am going to merge this to already have the function itself in main.

PR #2250, opened 2025-03-04
CI/RLS: update wheel builds to GEOS 3.13.1; other minor versions for tests

mwtoews:
GEOS 3.13.1, 3.12.3, 3.11.5, 3.10.7 and 3.9.6 were released ([download page](https://libgeos.org/usage/download/)).

This updates the release to use GEOS 3.13.1 with binary wheels. <s>Doctests are also now switched to GEOS 3.13.1</s> (note that 3.13.0 failed doctests). GEOS 3.13.1 is also used in two dockerfiles used for valgrind.

This PR also updates testing to use the other minor versions.


mwtoews:
Doctests are kept with Python 3.12 until a curious issue with Python 3.13's `sorted()` is sorted (#2251).

jorisvandenbossche (approved):

Issue #2005, opened 2024-03-05
TODO: M gets dropped from pickle serialization of LinearRing types


mwtoews:
On shapely `main` with #1808 to enable M and ZM in WKB outputs, one "TODO" item needs to be resolved for `test_pickle_m`. To demonstrate the issue:
```python
import pickle
import shapely

assert shapely.geos_version[0:2] >= (3, 12), shapely.geos_version_string

geom = shapely.from_wkt("LINEARRING M (0 0 1, 1 0 2, 1 1 3, 0 1 2, 0 0 1)")
pickled = pickle.dumps(geom)
actual = pickle.loads(pickled)

print(f"{geom!r} != {actual!r}")
# <LINEARRING M (0 0 1, 1 0 2, 1 1 3, 0 1 2, 0 0 1)> != <LINEARRING (0 0, 1 0, 1 1, 0 1, 0 0)>
```

I recall there is special handling for LinearRing types in shapely's `src/*` C code, as there is no WKB equivalent for LinearRing. It will need to be updated to handle M and ZM coordinate types.

jorisvandenbossche:
The special handling is currently not in the C code but here:

https://github.com/shapely/shapely/blob/e15799520838f9ba696cc107bc874f07644a474d/shapely/geometry/polygon.py#L16-L22

https://github.com/shapely/shapely/blob/e15799520838f9ba696cc107bc874f07644a474d/shapely/geometry/polygon.py#L113-L116

We should probably add a helper function in C that converts a LineString to a LinearRing by just getting the CoordSequence from the line and put that in a LinearRing, so that it preserves the full coord seq and its dimensionality.

PR #2253, opened 2025-03-07
BUG: preserve M values for pickle roundtrip of LinearRing

jorisvandenbossche:
Closes https://github.com/shapely/shapely/issues/2005


mwtoews (approved):
Great fix!

PR #2254, opened 2025-03-07
DOC/RLS: update changelog notes for 2.1.0 release

jorisvandenbossche:
Starting to update the release notes for 2.1.0. For now adding some missing things and aligning CHANGES.txt and 
docs/release/2.x.rst (some things were only added in one of both). 

If we follow the same approach as for 2.0.0, we could further expand the doc page a bit more (e.g. add an explicit section about M support or about coverage support), and at the same time reduce the content in CHANGES.txt. But also happy to keep both in sync (although it is not great to have to keep both)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -51,8 +51,8 @@ New functions:
 - Add function ``constrained_delaunay_triangles`` (#1685)
 - Add function ``coverage_simplify`` to allow topological simplification of polygonal
   coverages (#1969)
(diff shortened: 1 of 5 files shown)


mwtoews (approved):
Thanks, this looks good enough for a rc0! I don't see any important omissions.

jorisvandenbossche:
OK, merging then. More feedback to polish things before the final release is definitely welcome!

Issue #2027, opened 2024-03-25
Questionable behaviour of symmetric_difference_all


martinfleis:
My understanding of `symmetric_difference_all` is that it returns all parts of geometries that are unique to a single one, i.e. those parts that do not intersect any other. But that doesn't seem to  be the case as shown in the example below.

```py
import numpy as np
import shapely

geoms = np.array(
    [box(0, 0, 2, 2), box(1, 1, 3, 3), box(0, 0, 1.5, 1.5)]
)
shapely.symmetric_difference_all(geoms)
```

Visually:

You can see that the small square in the middle is shared by all three polygons.
![Screenshot 2024-03-25 at 14 30 21](https://github.com/shapely/shapely/assets/36797143/125f64fe-034f-49f4-9f11-b7257ef4c701)
Yet, it is present in the result.
![Screenshot 2024-03-25 at 14 31 21](https://github.com/shapely/shapely/assets/36797143/ec17f31f-4626-4053-98be-54b3e43274e6)

I would expect the result to not contain that small square. Is this my misunderstanding or a bug?

theroggy:
I also think this is not the expected result, so I'd think it is a bug.

I had a (very) quick look on how it is implemented, and the same `Y_Y_reduce_func` is used as is used as for `intersection_all`. I didn't look into details how the c function works, but I suppose this implements a loop through all inputs where each input geometry is, for `symmetric difference_all`, `symmetric differenced` from the result of the previous two... which leads to this result.

jorisvandenbossche:
Indeed, you essentially can't do this iteratively like we do now, in each step you need to consider all other geometries. 

The very dumb way I can currently think of is that for each geometry, you calculate the sym diff with the union (or just collection) of all other geometries. That sounds quite expensive, but not sure if there is a way to optimize this? (one could use a spatial index to do the sym diff with only the relevant (intersecting) other geometries)

@martinfleis did you have an actual use case for this, or just looking at it to consider adding the equivalent to geopandas? 
I am wondering if we shouldn't rather deprecate/remove this from shapely given it is not actually correct.

jorisvandenbossche:
cc @caspervdw

martinfleis:
> @martinfleis did you have an actual use case for this, or just looking at it to consider adding the equivalent to geopandas?
I am wondering if we shouldn't rather deprecate/remove this from shapely given it is not actually correct.

No use case. I hit this while trying to expose it on geopandas side. Deprecation is fine with me.

[...]

PR #2257, opened 2025-03-14
DEPR: deprecate symmetric_difference_all

jorisvandenbossche:
See https://github.com/shapely/shapely/issues/2027 for more context: the `symmetric_difference_all` has wrong behaviour which is not easy to fix, so rather deprecate and remove this function for now.

### DIFF
--- a/shapely/set_operations.py
+++ b/shapely/set_operations.py
@@ -1,5 +1,7 @@
 """Set-theoretic operations on geometry objects."""
 
+import warnings
+
 import numpy as np
 
 from shapely import Geometry, GeometryType, lib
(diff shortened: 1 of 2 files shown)


shapely/set_operations.py
```diff
@@ -317,6 +319,12 @@ def symmetric_difference_all(geometries, axis=None, **kwargs):
     If all elements of the given axis are None an empty GeometryCollection is
     returned.
 
+    .. deprecated:: 2.1.0
+
+        This function behaves incorrectlly and will be removed in a future
```
theroggy, line 324:
```suggestion
        This function behaves incorrectly and will be removed in a future
```

shapely/set_operations.py
```diff
@@ -357,6 +365,13 @@ def symmetric_difference_all(geometries, axis=None, **kwargs):
     <GEOMETRYCOLLECTION EMPTY>
 
     """
+    warnings.warn(
+        "The symmetric_difference_all function behaves incorrectlly and will be "
```
theroggy, line 369:
```suggestion
        "The symmetric_difference_all function behaves incorrectly and will be "
```

PR #2258, opened 2025-03-14
DEPR: bump some DeprecationWarnings from shapely 2.0 to more visible FutureWarning

jorisvandenbossche:
Those are deprecations that were already introduced in shapely 2.0. I think we can move those to a more visible FutureWarning now before actually removing them later.

### DIFF
--- a/shapely/errors.py
+++ b/shapely/errors.py
@@ -71,7 +71,7 @@ def __getattr__(name):
             f"{name} is deprecated and will be removed in a future version. "
             "Use ShapelyError instead (functions previously raising {name} "
             "will now raise a ShapelyError instead).",
-            DeprecationWarning,
+            FutureWarning,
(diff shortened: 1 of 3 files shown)


mwtoews:
Thanks, I was considering something exactly like this. One additional warnings to consider is in [`shapely/geos.py`](https://github.com/shapely/shapely/blob/474b2d7aa9d88c91984666809ed937f7c253f980/shapely/geos.py#L7-L13).

jorisvandenbossche:
> One additional warnings to consider is in [shapely/geos.py](https://github.com/shapely/shapely/blob/474b2d7aa9d88c91984666809ed937f7c253f980/shapely/geos.py#L7-L13).

That warning was only added in 2.1, so can stay at DeprecationWarning for now

PR #2259, opened 2025-03-15
TST: rename tests that use `handle_nan` option

mwtoews:
This is a relatively minor renaming of tests for the `handle_nan` option, which makes them more consistent, something along the lines of:

    def test_<geomtype>_handle_nan_<option>[_<behaviour]([args]):

There is also a minor fix for a geometry type in `test_linearrings_invalid_ndim` which I've stumbled on while developing M support for linestrings (for 2.2).


jorisvandenbossche (approved):
Thanks!

PR #2262, opened 2025-03-17
concave hull by length

jmsnn:
I'm in need for this functionality in some project and saw that it was on the roadmap of functions to be implemented, seemed like a good fit for a first contribution ! 

because the processing in the concave_hull_func() is the same whether using ratio or length, I just added a flag to know which one to perform. But the function is enclosed by some #if GEOS_SINCE_3_11_0 and the concave hull by length is only available in 3.12... not sure if it's better to create 2 separate functions each enclosed with correct version control ?

### DIFF
--- a/shapely/constructive.py
+++ b/shapely/constructive.py
@@ -427,7 +427,7 @@ def clip_by_rect(geometry, xmin, ymin, xmax, ymax, **kwargs):
 
 @requires_geos("3.11.0")
 @multithreading_enabled
-def concave_hull(geometry, ratio=0.0, allow_holes=False, **kwargs):
+def concave_hull(geometry, ratio=0.0, min_length=None, allow_holes=False, **kwargs):
     """Compute a concave geometry that encloses an input geometry.
 
     Parameters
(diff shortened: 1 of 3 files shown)


caspervdw:
@jmsnn  Thanks for you contribution (and sorry for the very late reply - we all have limited time for this project).

Looking at your own comments, I would prefer to put this in a different functions. Currently the concave_hull is categorized as a "special" function, not lumped together with other functions of the same call signature. However, with your addition, we now have 4 functions of the same callsignature `NPY_OBJECT, NPY_DOUBLE, NPY_BOOL, NPY_OBJECT`, namely:

- concave_hull
- concave_hull_by_length
- delaunay_triangles
- coverage_simplify

I think this warrants a generalization, something like `Ydb_Y`

I am currently working on an overhaul of `ufuncs.c` file. When doing that for the `Ydb_Y` functions, it is trivial to add the `by_length` version. So I can add that to my list, or if you feel up to it, you could try to do it yourself. An example:  https://github.com/shapely/shapely/pull/2353

PR #2268, opened 2025-03-24
DOC: add coverage_union to coverage API page

martinfleis:
Exposing these on coverage-dedicated API reference page to keep all relevant functions visible from a single place. Keeping it also in set operations page as it belongs to both and will have better discoverabilty that way.


jorisvandenbossche (approved):
Thanks!

Issue #2282, opened 2025-04-23
deprecate_positional is a performance bottleneck (300%-1000% slowdown) in Shapely 2.1


Zaczero:
I have the following code:

```py
@overload
def compressible_geometry(geometry: _GeomT, /) -> _GeomT: ...
@overload
def compressible_geometry(geometry: NDArray[np.float64], /) -> NDArray[np.float64]: ...
def compressible_geometry(
    geometry: _GeomT | NDArray[np.float64], /
) -> _GeomT | NDArray[np.float64]:
    """
    Make geometry easily compressible by reducing mentissa noise.
    It is then necessary to round the coordinates back.
    Inspired by http://www.danbaston.com/posts/2018/02/15/optimizing-postgis-geometries.html
    """
    if isinstance(geometry, BaseGeometry):
        return transform(geometry, compressible_geometry)
    view = geometry.view(np.uint64)
    view &= _MASK
    return geometry

```

And performance analysis indicate:

![Image](https://github.com/user-attachments/assets/f238ecdd-76a7-4555-a26f-d2dc8f8a8c81)

That only 17 seconds from 66 seconds total, is the implementation of `transform`. The remaining time is taken by the `deprecate_positional` decorator.

Zaczero:
Another example:

![Image](https://github.com/user-attachments/assets/f93dccb4-589e-45ea-a637-40e50f4bae82)

8.7s -> 82.8s increase !!

sgillies:
@Zaczero thank you for the report and fix !

Issue #2280, opened 2025-04-22
2.1 Polygon creation is much slower than 2.0.7


dbestPacksol:
It seems to be that creating Polygons in 2.1 is much slower (roughly 5-10x) slower than 2.0.7.

The following script takes roughly 0.1 seconds with Shapely 2.1 and 0.015 with Shapely 2.0.7. This is on Python 3.12

import time

import shapely

if __name__ == "__main__":

    start_time = time.time()
    for _ in range(1000):
        coords = ((0., 0.), (0., 1.), (1., 1.), (1., 0.), (0., 0.))
        polygon = shapely.Polygon(coords)

    print(time.time() - start_time)

jorisvandenbossche:
Thanks for the report. This slowdown seems to be due to the overhead of the decorator we added in https://github.com/shapely/shapely/pull/2237 to deprecate positional arguments. That decorator does inspect the signature, which in this case of a function call that itself is very fast, gives a relatively large overhead.

I know your example is a simplified code snippet to demonstrate the issue, _but_ if you are creating many polygons (which I assume is the case, otherwise the overhead of calling `Polygon()` only once should not matter), you can also do that by using the vectorized `shapely.polygons()` function instead of the `shapely.Polygon()` class constructor in a loop.

manuel-koch:
I see a similar performance degration in my code after upgrading from 2.0.6 to 2.1.0.

I.e. running profiler over my code shows that considerable amount of my execution time is now spent in this shapely decorator code to do inspect stuff ( and I don't even get one of those warnings ).
The cumtime numbers below seem to indicate that ~50% is spent in inspect stuff for my use case of shapely.

Would be nice if one could opt-out for this inspect stuff.
I noticed an even greater performance degration when running under a debugger.

```shell
         5328400 function calls (5328375 primitive calls) in 1.621 seconds

   Ordered by: cumulative time
   List reduced from 605 to 100 due to restriction <100>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    12/11    0.007    0.001    1.448    0.132 /.venv/lib/python3.12/site-packages/python_geometry/fit_polygon/fit_polygon.py:89(calc_fit_point_for_content_within_boundary)
  310/308    0.017    0.000    1.438    0.005 /.venv/lib/python3.12/site-packages/shapely/geometry/multipoint.py:44(__new__)
[...]

PR #2283, opened 2025-04-23
ENH: improve performance of deprecate_positional

Zaczero:
Closes https://github.com/shapely/shapely/issues/2282
Closes https://github.com/shapely/shapely/issues/2280

It's my first time contributing here.

- I'm aware that shapely is generally untyped.
- Types used here are very similar to shapely/strtree.py, and because of that I decided to include them.
- **Allow edits by maintainers** is checked so feel free to push commits for minor issues

Potentially useful info for reviewers:

`__code__` docs reference (from Python 3.9): https://docs.python.org/3.9/library/inspect.html
	
- co_argcount: number of arguments (not including keyword only arguments, * or ** args)
- co_varnames: tuple of names of arguments and local variables

This lets us skip the heavy inspect logic.

Another big optimization is the `n > warn_from` check and LRU cache of the messages, so people who don't resolve the deprecation warning aren't affected too much performance-wise.

### DIFF
--- a/shapely/decorators.py
+++ b/shapely/decorators.py
@@ -142,7 +142,7 @@ def deprecate_positional(
             return func
 
         # earliest position where a warning could occur
-        warn_from = min(idx for idx, _ in deprecate_positions)
+        warn_from = min(deprecate_positions)[0]
 
         @lru_cache(10)
         def make_msg(n_args: int):
--- a/shapely/decorators.py
+++ b/shapely/decorators.py
@@ -145,10 +145,8 @@ def deprecate_positional(
         warn_from = min(idx for idx, _ in deprecate_positions)
 
         @lru_cache(10)
-        def make_msg(n_args: int) -> str | None:
+        def make_msg(n_args: int):
             used = [name for idx, name in deprecate_positions if idx < n_args]
-            if not used:
-                return None
 
             if len(used) == 1:
                 args_txt = f"`{used[0]}`"
@@ -174,9 +172,7 @@ def deprecate_positional(
 
             n = len(args)
             if n > warn_from:
-                msg = make_msg(n)
-                if msg is not None:
-                    warnings.warn(msg, category=category, stacklevel=2)
+                warnings.warn(make_msg(n), category=category, stacklevel=2)
 
             return result
 
--- a/shapely/decorators.py
+++ b/shapely/decorators.py
@@ -4,6 +4,7 @@ import os
 import warnings
 from collections.abc import Callable, Iterable
 from functools import lru_cache, wraps
+from inspect import unwrap
 
 import numpy as np
 
@@ -124,7 +125,7 @@ def deprecate_positional(
     """
 
     def decorator(func: Callable):
-        code = func.__code__
+        code = unwrap(func).__code__
 
         # positional parameters are the first co_argcount names
         pos_names = code.co_varnames[: code.co_argcount]
--- a/shapely/decorators.py
+++ b/shapely/decorators.py
@@ -1,9 +1,9 @@
 """Decorators for Shapely functions."""
 
-import inspect
 import os
 import warnings
-from functools import wraps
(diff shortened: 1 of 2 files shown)


shapely/decorators.py
```diff
@@ -92,41 +93,90 @@ def wrapped(*args, **kwargs):
...
+        # early exit if there are no deprecated positional args
+        if not deprecate_positions:
+            return func
+
+        # earliest position where a warning could occur
+        warn_from = min(deprecate_positions)[0]
+
+        @lru_cache(10)
+        def make_msg(n_args: int):
+            used = [name for idx, name in deprecate_positions if idx < n_args]
```
mwtoews, line 151: There are some instances where `lru_cache` can cause memory leaks, if it cannot be garbage cleaned (e.g. [described here](https://docs.astral.sh/ruff/rules/cached-instance-method/). I'm not sure if this is a scenario where this may occur. If so, then perhaps this sub-method could be renamed and moved outside the method, adding `func_name` and `deprecate_positions` (as an immutable tuple) as arguments? E.g. maybe something like `_make_msg(func_name: str, n_args: int, dep_pos: tuple)`?

mwtoews (commented):
Thanks for your first, it looks great! Have you verified the timing to see if it improves the bottlenecks?

Shapely is mostly untyped for no good reason, perhaps just a low priority. But for new(ish) methods, adding typing hints is welcome from my perspective.

I'll find some more time in the next few days to take deeper look at this PR, but it looks pretty close to ready.

shapely/decorators.py
Zaczero: In this case, I don't believe there is a realistic memory leak scenario. The most common memory leak issue (which your link shows) occurs when a method accepts a volatile instance object (usually meaning it has a `self` argument). This makes the cache keep a reference to that object and prevents it from being freed up. The only remaining case to consider is what happens when the decorated method is cleaned up with `del` or replaced. I believe currently such an operation will properly clean the whole lru_cache object, as there are no other references to it. However, if we were to make this method global, it would cause the entire cache to live throughout the application lifetime, and if the original method were to be deleted or replaced, it wouldn't be able to free the related cache memory (small but still). So in that small edge case, it would be more *leaky* to have one global cache instead of individual function-specific caches. There are other **very minor** arguments against such edit: more arguments in cache key - slower access, global vs local scope - slower access, worse code locality.

Zaczero:
> Have you verified the timing to see if it improves the bottlenecks?

No. I expect there to be no more measurable performance overhead (except for the typical decorator wrap overhead) as there is now only one `len(...)` call, which is very lightweight, compared to the previous full `signature` machinery. I'll run a benchmark if I find time (or if you really insist). I would be surprised if there was anything more to it.

jorisvandenbossche:
@Zaczero thanks a lot for the PR! 

I had checked the timing and profile when checking https://github.com/shapely/shapely/issues/2280, so re-running that confirms your assumption there is now basically no overhead left:

```python
In [1]: import shapely

In [2]: %%timeit
   ...: for _ in range(1000):
   ...:     coords = ((0., 0.), (0., 1.), (1., 1.), (1., 0.), (0., 0.))
   ...:     polygon = shapely.Polygon(coords)
   ...: 
21.7 ms ± 1.16 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)   # shapely 2.0.6
107 ms ± 8.73 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)     # shapely main
22.5 ms ± 1.52 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)    # this branch
```

jorisvandenbossche (approved):

mwtoews (approved):
Finally got around to checking code, and indeed it restores the performance. Addition of missing tests is also very welcome. Thanks for the time assembling this PR, we should see this in a patch release anticipated soon!

PR #2291, opened 2025-05-02
BUG: avoid memory segfault in from_ragged_array

jorisvandenbossche:
Fixes the issue reported in https://github.com/shapely/shapely/discussions/2284

We were allocating a too small temporary array for storing the intermediate rings.

### DIFF
--- a/shapely/tests/test_ragged_array.py
+++ b/shapely/tests/test_ragged_array.py
@@ -586,3 +586,20 @@ def test_from_ragged_wrong_offsets():
             np.array([[0, 0], [0, 1]]),
             offsets=(np.array([0, 1]),),
         )
+
+
+def test_from_ragged_crash_2284():
+    # caused segfault in shapely 2.1.0
+    # https://github.com/shapely/shapely/discussions/2284
+
+    # one of the geometries has more rings than the total number of geometries
+    coords = np.random.randn(60, 2)
+    offsets1 = np.array([0, 10, 20, 30, 40, 50, 60], dtype=np.uint32)
+    offsets2 = np.array([0, 1, 5, 6], dtype=np.uint32)
+
+    for _ in range(10):
+        polygons = shapely.from_ragged_array(
+            shapely.GeometryType.POLYGON, coords, (offsets1, offsets2)
+        )
+        # just ensure it didn't crash
+        assert len(polygons) == 3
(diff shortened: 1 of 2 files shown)


shapely/tests/test_ragged_array.py
```diff
@@ -586,3 +586,28 @@ def test_from_ragged_wrong_offsets():
...
             offsets=(np.array([0, 1]),),
         )
+
+
+def test_from_ragged_crash_2284():
+    # caused segfault in shapely 2.1.0
+    # https://github.com/shapely/shapely/discussions/2284
+
+    # one of the geometries has more rings than the total number of geometries
+    coords = np.random.randn(60, 2)
```
mwtoews, line 596: New code should use the modern RNG methods ([docs](https://numpy.org/doc/2.2/reference/random/generated/numpy.random.randn.html)), also make the random coords stay the same-ish by specifying a seed.
```suggestion
    rng = np.random.default_rng(2284)
    coords = rng.standard_normal(120).reshape((60, 2))
```

mwtoews (approved):
I can reproduce the segfault and verify the fix. Looks like there was a mix-up with `n_parts` and `n_geoms`, which is now resolved. Only a small suggestion to apply for new NumPy code. Thanks for nicely spotted fix!

shapely/tests/test_ragged_array.py
jorisvandenbossche: > also make the random coords stay the same-ish by specifying a seed.

For testing we don't necessarily want to use a fixed seed? It seems useful to me to use varying values. Of course if you then run into an issue with certain values, it is not reproducible, so it is not necessarily very helpful, but personally I don't see the point of wanting to fix the random values for this use case

mwtoews: It probably doesn't make any difference on the randomness, as it's the data size that matters here. A non-seeded RNG is simply `rng = np.random.default_rng()`, if that's preferred.

jorisvandenbossche: Updated. 
(now, to be honest, I personally think this recommendation from numpy is too strong, and for testing purposes here where we don't care about the pseudo-random number generator theory (we are not doing statistical modelling), using `np.random.default_rng().standard_normal(120).reshape((60, 2))` instead of `np.random.randn(60, 2)`  is just unnecessary boilerplate and noise in the test code .. ;))

PR #2294, opened 2025-05-06
TST: update maximum_inscribed_circle tests for latest GEOS changes

jorisvandenbossche:
Some tests are failing on the GEOS main builds (eg https://github.com/shapely/shapely/actions/runs/14814815121/job/41593981112), related to some recent commits in GEOS (https://github.com/libgeos/geos/compare/4ae3f0536735923dc93d6ea8a1e0dadafc791af1...6f4a7396481ab28dd240cc7b1511949f1892ac2a)

### DIFF
--- a/shapely/tests/legacy/test_polylabel.py
+++ b/shapely/tests/legacy/test_polylabel.py
@@ -48,6 +48,9 @@ class PolylabelTestCase(unittest.TestCase):
             ]
         )
         label = polylabel(polygon)
+        if shapely.geos_version >= (3, 14, 0):
+            # TODO verify if this is a bug in GEOS?
+            assert label.coords[:] == [(32.722025, -117.195155)]
         if shapely.geos_version >= (3, 12, 0):
             # recent GEOS corrects for this
             assert label.coords[:] == [(32.722025, -117.201875)]
--- a/shapely/tests/legacy/test_polylabel.py
+++ b/shapely/tests/legacy/test_polylabel.py
@@ -49,7 +49,7 @@ class PolylabelTestCase(unittest.TestCase):
         )
         label = polylabel(polygon)
         if shapely.geos_version >= (3, 14, 0):
-            # TODO verify if this is a bug in GEOS?
+            # https://github.com/libgeos/geos/issues/1265
             assert label.coords[:] == [(32.722025, -117.195155)]
         elif shapely.geos_version >= (3, 12, 0):
             # recent GEOS corrects for this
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
The `test_rectangle_special_case` failure seems to be a regression in behaviour on the GEOS side.

```python
import matplotlib.pyplot as plt
import shapely
from shapely.plotting import plot_polygon, plot_line

polygon = shapely.Polygon(
    [
        (32.71997, -117.19310),
        (32.71997, -117.21065),
        (32.72408, -117.21065),
        (32.72408, -117.19310),
    ]
)
result = shapely.maximum_inscribed_circle(polygon)

fig, ax = plt.subplots()
plot_polygon(polygon, ax=ax)
plot_line(shapely.maximum_inscribed_circle(polygon), ax=ax, color="C1")
```

With GEOS 3.13.0:
test-geos-313

With GEOS 3.14.dev main:
test-geos-dev

jorisvandenbossche:
Actually, the images are misleading because I was not using equal aspect ratio ..

Plotting this again, you can see that both results are strictly correct (they give the same size of circle). It's only that for the purpose of "polylabel", the more centered circle might give a better result.

<details>
<summary>Updated code</summary>

```
import matplotlib.pyplot as plt
import shapely
from shapely.plotting import plot_polygon, plot_line

polygon = shapely.Polygon(
    [
        (32.71997, -117.19310),
        (32.71997, -117.21065),
        (32.72408, -117.21065),
        (32.72408, -117.19310),
    ]
)
result = shapely.maximum_inscribed_circle(polygon)

fig, ax = plt.subplots()
plot_polygon(polygon, ax=ax)
plot_line(result, ax=ax, color="C1")
ax.set_aspect("equal")
plot_line(shapely.get_point(result, 0).buffer(result.length).boundary, ax=ax, color="C1", add_points=False)
```

</details>

With GEOS 3.13.0:

test-geos-313


With GEOS 3.14.dev main:

test-geos-dev

mwtoews:
I agree with the upstream assessment that this is a "won't fix" since the input has a non-deterministic result. The input should be changed to yield a deterministic result.
```bash
$ pixi exec -s geos=3.10.1 geosop -f wkt -a "POLYGON ((100 0, 200 200, 300 200, 300 100, 100 0))" maxInscribedCircle 0.0001
LINESTRING (238.19656372070312 138.19656372070312, 182.91793823242188 165.83587646484375)
$ pixi exec -s geos=3.13.1 geosop -f wkt -a "POLYGON ((100 0, 200 200, 300 200, 300 100, 100 0))" maxInscribedCircle 0.0001
LINESTRING (238.19656372070312 138.19656372070312, 182.91793823242188 165.83587646484375)
```

jorisvandenbossche:
> The input should be changed to yield a deterministic result.

Then we can also just remove the test, since it was added specifically for this rectangle corner case (https://github.com/shapely/shapely/pull/644)

jorisvandenbossche:
Going to merge this already to have green CI on main, but happy to follow-up with either updating or removing the test

Issue #1319, opened 2022-02-16
mapping and rfc7946 ring ordering


rbuffat:
[RFC7946](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6) specifies an ordering for rings:

> A linear ring MUST follow the right-hand rule with respect to the
>       area it bounds, i.e., exterior rings are counterclockwise, and
>       holes are clockwise.
> 
>    Note: the [[GJ2008](https://datatracker.ietf.org/doc/html/rfc7946#ref-GJ2008)] specification did not discuss linear ring winding
>    order.  For backwards compatibility, parsers SHOULD NOT reject
>    Polygons that do not follow the right-hand rule.

Currently, mapping does not change / correct the ordering of rings. Should there be an option in mapping to generate GeoJSON Polygons according to RFC7946?

E.g. by orienting the rings first using something like 

```
    if isinstance(geometry, Polygon):
        return orient(geometry, sign=1.0) # type: ignore
    elif isinstance(geometry, MultiPolygon): # type: ignore
        return MultiPolygon([orient(geom, sign=1.0) for geom in geometry.geoms]) # type: ignore
```

jorisvandenbossche:
That sounds as a useful option, yes. 

I was looking at GEOS if they actually also support this, but it's not exposed in the C API. It is a bit unfortunate that the `normalize()` method (`GEOSNormalize`) seems to do the exact opposite than what we want here (outer ring clockwise, inner rings counter clockwise)

rkaravia:
With shapely 2.1.0 and [`shapely.orient_polygons()`](https://shapely.readthedocs.io/en/2.1.0/reference/shapely.orient_polygons.html) this looks relatively straightforward, see prototype in #2297, which adds a `force_ccw` flag to `mapping()`.

I would be happy to put in some more work to get this to a mergeable state (such as adding more tests/docs), but I'm not sure if the whole idea really makes sense.

From the docs about [mapping](https://shapely.readthedocs.io/en/2.1.0/manual.html#shapely.geometry.mapping), I saw that `mapping` should also work for "any object which implements `__geo_interface__`". This would require conversion to a shapely geometry before the call to `orient_polygons()`. I think it would be a bit weird to convert to a shapely geometry conditionally based on the `force_ccw` flag. What do you think?

rkaravia:
Thank you for the feedback! 

Yes, as I wrote above, I also have some doubts about this.

[...]

PR #2297, opened 2025-05-13
Prototype: Add force_ccw flag to shapely.geometry.mapping

rkaravia:


shapely/geometry/geo.py
```diff
@@ -118,7 +119,7 @@ def shape(context):
         raise GeometryTypeError(f"Unknown geometry type: {geom_type!r}")
 
 
-def mapping(ob):
+def mapping(ob, force_ccw=False):
```
sgillies, line 122: @rkaravia I'd like to see `ensure_ccw` or `orient_ccw` as the name of this keyword argument. Force, to me, means to push past warnings and ignore security or safety checks. We're not doing that here. We're only calling another method, which is named `orient_polygons()`.

shapely/tests/geometry/test_collection.py
```diff
@@ -82,6 +82,32 @@ def test_geointerface(geometrycollection_geojson):
...
+    assert mapping(geom, force_ccw=False) == {
+        "type": "GeometryCollection",
+        "geometries": [{
+            "type": "Polygon",
+            "coordinates": (
+                ((0.0, 0.0), (0.0, 1.0), (1.0, 1.0), (1.0, 0.0), (0.0, 0.0)),
+            )
+        }]
+    }
+    # force_ccw=True
```
sgillies, line 99: This should be a separate test function, so that failure in the first case doesn't cover up failure or success of the second case.

sgillies (commented):
@rkaravia I'm not sure I'm in favor of this overall. I left a couple of technical suggestions inline.

rkaravia:
Closing this because adding an extra flag is not the right solution, see https://github.com/shapely/shapely/issues/1319#issuecomment-2877782770

PR #2299, opened 2025-05-20
adding multi-core possibility to the validation.py

suriyahgit:
Added multi-core functionality to the validation.py

### DIFF
--- a/shapely/validation.py
+++ b/shapely/validation.py
@@ -1,55 +1,71 @@
 """Validate geometries and make them valid."""
-# TODO: allow for implementations using other than GEOS
+from concurrent.futures import ThreadPoolExecutor
+import numpy as np
+from functools import partial
 
 import shapely
 
-__all__ = ["explain_validity", "make_valid"]
+__all__ = ["explain_validity", "make_valid", "make_valid_parallel"]
 
 
 def explain_validity(ob):
-    """Explain the validity of the input geometry, if it is invalid.
-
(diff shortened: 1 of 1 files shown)


sgillies:
@suriyahgit I am not in favor of this. Shapely has many operations that one might want to do in parallel. It's not optimal to add parallel versions of these functions to the library because that will multiply the documentation, testing, and maintenance needs.

Instead, there should be a single function that can run *any* operation in parallel. That would increase the burden on the library by a small amount. In fact, that function already exists: it is `ThreadPoolExecutor.map()`.

Instead of using your `make_valid_parallel()`, we should simply use the body of that function in our own code. It's only four lines.

Now, for a GeoPandas user, it might be nice to have a parallel option (if that doesn't already exist). But that functionality belongs in GeoPandas, not Shapely.

caspervdw:
> Although it is not shapely-specific (and so not entirely sure it belongs in our docs), but we could add an example to the documentation about shapely not providing any parallelization itself but how you can easily do that with ThreadPoolExecutor.map()

I would lean towards the side of not documenting it (because it is really a generic thing). There are quite some flavours of parallelization and even more methods to implement them. It feels random to only show threads.

Not sure what is our policy on this, but as we have 2 downvotes of maintainers (actually 3 now), I think this PR should be closed.

Thanks for your time spending on contributing to shapely @suriyahgit .

PR #2300, opened 2025-05-23
MNT: rename first arg for __new__ staticmethods

mwtoews:
This PR started with `pre-commit autoupdate` which updated the ruff version and check rules. Thus the primary change here is to fix [bad-staticmethod-argument (PLW0211)](https://docs.astral.sh/ruff/rules/bad-staticmethod-argument/#bad-staticmethod-argument-plw0211), where I've renamed the first argument of the implicitly static methods `__new__(self...` to `__new__(cls...`.

Other changes:

- Rename `TCH` to `TC` for type-checking imports, as it seems that this rule has been renamed at some version.
- Remove a few unused ignores related to PLW/PLC; move one as an inline ignore.

Also, should this change be part of the yet-to-be created `maint-2.1` or just for `main` branch development for the next version? My instinct is for the later choice, thus I'd create a maintenance branch before merging.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
       - id: end-of-file-fixer
       - id: trailing-whitespace
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.9.6
+    rev: v0.11.11
     hooks:
       # Run the linter
       - id: ruff
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -115,7 +115,7 @@ select = [
     # implicit string concatenation
     "ISC",
     # type-checking imports
-    "TCH",
+    "TC",
     # comprehensions
     "C4",
     # Ruff-specific rules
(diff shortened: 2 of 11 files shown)


jorisvandenbossche (approved):

jorisvandenbossche:
> Also, should this change be part of the yet-to-be created `maint-2.1` or just for `main` branch development for the next version? My instinct is for the later choice, thus I'd create a maintenance branch before merging.

Strictly speaking this is a change that does not need to (or should not) be in a maintenance branch, but on the other hand if this would give small formatting differences, it might also avoid conflicts if this is included in the maint branch (although I don't expect there to be much formatting differences between the ruff versions, as it is quite stable, which can also be seen from this PR where there are essentially no changes to existing code). So I don't really have a strong opinion either way ;)

(in any case, also if this PR is already merged, we can always branch off from an older commit)

mwtoews:
Lets branch them now. This and subsequent PRs will go to `main`, while maintenance and backports to [`maint-2.1`](https://github.com/shapely/shapely/tree/maint-2.1)

PR #2301, opened 2025-05-25
MAINT: drop Python 3.10 support

mwtoews:
This drops Python 3.10 support for the next phase of development for shapely 2.2.

A few suggestions from ruff check are evaluated and applied, based on python 3.11+ support. Most of this is based on [zip-without-explicit-strict (B905)](https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/#zip-without-explicit-strict-b905).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,11 +17,11 @@ jobs:
         geos: [3.9.6, 3.10.7, 3.11.5, 3.12.3, 3.13.1, main]
         include:
           # 2020
-          - python: "3.10"
+          - python: "3.11"
             geos: 3.9.6
             numpy: 1.21.6
           # 2021
-          - python: "3.10"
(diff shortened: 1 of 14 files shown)


.github/workflows/tests.yml
```diff
@@ -55,19 +55,19 @@ jobs:
...
             python: "3.11"
             geos: 3.12.3
             numpy: 1.24.4
           # pypy (use explicit ubuntu version to not overwrite existing ubuntu-latest + geos 3.11.0 build)
           - os: ubuntu-22.04
-            python: "pypy3.10"
+            python: "pypy3.11"
             geos: 3.12.3
-            numpy: 1.24.4
+            numpy: 2.3.0rc1
```
mwtoews, line 70: I'll hold this PR until NumPy 2.3.0 is released.

PR #2306, opened 2025-06-17
exploding singular, linear objects into constituent segments

### LINKED ISSUE #2188: [ENH] exploding linestrings into constituent pairwise segments
Every so often the question arises[^1][^2] on the best way to explode singular linestrings into constituent pairwise segments. It doesn't seem like this solution is already baked into Shapely itself. In another ongoing project we have come up with this little solution that gets between ≈3x - ≈20x runtime reduction over a `map(Linsetring, zip(coords))` solution:

```python
def line_segments(line: shapely.LineString) -> np.ndarray:
    """Explode a linestring into constituent pairwise coordinates."""
    xys = shapely.get_coordinates(line)
    return shapely.linestrings(
        np.column_stack((xys[:-1], xys[1:])).reshape(xys.shape[0] - 1, 2, 2)
    )
```

See some timing experimentation in [this gist](https://gist.github.com/jGaboardi/754e878ee4cac986132295ed43e74512).

Is there any interest in getting it into Shapely? If so, I can put in a PR.

cc @martinfleis @anastassiavybornova

[^1]: [Exploding LineString into segments using Shapely](https://gis.stackexchange.com/questions/427932/exploding-linestring-into-segments-using-shapely)
[^2]: [ENH: explode lines into segments](https://github.com/geopandas/geopandas/issues/2476)

jGaboardi:
This resolves #2188


caspervdw:
Thanks for the contribution @jGaboardi  !

> Unless the function can optimize for an array of linestrings (which I don't think it can, this is an O(n) problem), I don't think there is a need to handle anything other than a single linestring. Collections of linestrings can be managed by the user, I think.

@sgillies Not sure if I am understanding you here; in my view shapely 2.0 optimized many "array of geometry" operations, while they are all O(n). The thing is that allowing array inputs keeps the python for-loop out of the equation. I believe the same situation applies here.

So I'd go with @martinfleis suggestion to have a similar API as get_parts. I think it can be done by using `get_coordinates(..., return_index=True)` and some smart numpy trick that @jGaboardi  already has in place for the coordinates.

shapely/ops.py
```diff
@@ -718,3 +723,50 @@ def orient(geom, sign=1.0):
 
     """
     return shapely.orient_polygons(geom, exterior_cw=sign < 0)
+
+
+def get_segments(geometry):
```
caspervdw: Not completely sure about this, but looking at the tests being in the `legacy` folder, I think the `ops.py` file is there for historic reasons.  More recently we are trying to put the algorithms into categories. In this case I think the function should best be placed in the same file as `get_coordinates`. The tests may be moved as well.

shapely/ops.py
```diff
@@ -718,3 +723,50 @@ def orient(geom, sign=1.0):
...
+    """Return the constituent segments a line.
```
caspervdw:
```suggestion
    """Get segments of each linear geometry object.
```

Looking at consistency of docstrings of `get_parts` and `get_coordinates`

shapely/ops.py
```diff
@@ -718,3 +723,50 @@ def orient(geom, sign=1.0):
...
+    geometry : Geometry
+        A single linear object.
+
+    Returns
+    -------
+    ndarray of constituent pairwise segments.
+
+    See Also
+    --------
+    get_parts
```
caspervdw:
```suggestion
    get_parts, get_coordinates
```

shapely/ops.py
```diff
@@ -718,3 +723,50 @@ def orient(geom, sign=1.0):
 
     """
     return shapely.orient_polygons(geom, exterior_cw=sign < 0)
+
+
+def get_segments(geometry):
+    """Return the constituent segments a line.
+
+    Here 'segments' is defined as the individual pairwise coordinates
+    comprising a (non-multi) linear feature.
```
caspervdw:
```suggestion
    comprising a LineString or LinearRing. Multi* geometry objects are not supported.
```

shapely/tests/legacy/test_get_segments.py
```diff
@@ -0,0 +1,40 @@
+import numpy
```
caspervdw: This is a convention that you'll find throughout the repo:

```suggestion
import numpy as np
```

caspervdw (commented):
Thanks again for the contribution, there is one more high-level remark I left in a separate comment. Best to wait for that discussion to be resolved before iterating on this, I would say.

caspervdw:
> Thanks again for the contribution, there is one more high-level remark I left in a separate comment. Best to wait for that discussion to be resolved before iterating on this, I would say.

@jGaboardi  It's been more than a month, and both me and @martinfleis  are on board with this change. So I think we can be positive on going forward with this. Whenever you have time, please have a look at the comments. I'll try to provide feedback soon after that.

jGaboardi:
@caspervdw -- sounds good

jGaboardi:
@caspervdw 

> @sgillies Not sure if I am understanding you here; in my view shapely 2.0 optimized many "array of geometry" operations, while they are all O(n). The thing is that allowing array inputs keeps the python for-loop out of the equation. I believe the same situation applies here.

So for your comment above, are we thinking to allow for the passing in of:
1. ***only*** single linear objects (current implementation); or
2. both single linear objects ***and*** array-like of  linear objects?

martinfleis:
As said above, ensure that the API matches `get_parts`. That also means that this should allow arrays or LineStrings.

jGaboardi:
> As said above, ensure that the API matches `get_parts`. That also means that this should allow arrays or LineStrings.

I misread the above (blame lack of caffeine or too much turkey).

I'll push up my current iteration (that may or may not be close to what we'd like in the signature, but surely not the final product). For now it still resides in `ops.py` since putting it in either `coordinates.py` or `_geometry.py` leads to a circular import error due to being reliant on `linestrings()`.

jGaboardi:
1187505 moves to addressing @caspervdw's comments for an API in line with `get_parts()` while also supporting some of the arguments from `get_coordinates()` and the ability to pass keywords in to `linestrings()`.

The logic is much more complex when we allow for passing arrays of geometries in, since we then need to provide the option to return an index -- along with the option for handling `Z` values -- among other caveats that crop up.

I put some testing in there, but didn't get too deep in the weeds yet so I could get yalls feedback on this iteration.

shapely/ops.py
caspervdw: In shapely, many functions are thin wrappers around C-implemented ones. Take for instance the `get_coordinates`: its implementation is just a  direct call into `lib.get_coordinates`:

```
lib.get_coordinates(
        np.asarray(geometry, dtype=np.object_), include_z, include_m, return_index
    )
```

What this wrapper adds is 1) type coercion and 2) docstrings. You don't need those, as long as you are doing the type coercion correctly if necessary.

So I suggest not depending on `creation.py` and `coordinates.py`, but directly calling into `lib.get_coordinates` and `lib.linestrings`. Most modules already depend on `lib` so that will be fine. Then you'll have free choice of where to put the function. 

My first choice would be `_geometry.py` next to `get_parts`.

shapely/ops.py
```diff
@@ -684,3 +686,155 @@ def orient(geom, sign=1.0):
...
+
+    def _vectorized_isolate_segments(coords, idx, include_z, **kwargs):
+        """Isolate segments from all input linear features."""
+        n_xys = coords.shape[0]
+        idx_splitter = np.unique(idx.reshape(n_xys, 1), return_index=True)[1][1:]
+        xys_by_input_feature = np.split(xys, idx_splitter)
+        signature = _determine_signature(xys_by_input_feature)
+
+        return np.concatenate(
+            np.vectorize(_isolate_segments, signature=signature, otypes=[np.ndarray])(
```
caspervdw: It will probably be hard, but try to do this without the `np.vectorize`. We can discuss about it if you can't find a solution.

caspervdw (commented):
Great work, @jGaboardi , thanks for resuming it.

I went through the code and put some comments.

shapely/tests/legacy/test_get_segments.py
```diff
@@ -0,0 +1,72 @@
...
+    "geoms,expected",
+    [
+        [line_string, np.array(out_elbow)],
+        [[line_string], np.array(out_elbow)],
+        [[[line_string]], np.array(out_elbow)],
+        [[[[line_string]]], np.array(out_elbow)],
+        [[line_string, line_string], np.array(out_elbow + out_elbow)],
+        [[[[line_string, line_string]]], np.array(out_elbow + out_elbow)],
+        [[[line_string], [line_string]], np.array(out_elbow + out_elbow)],
+        [line_string_z, np.array(out_elbow)],
```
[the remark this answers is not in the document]
jGaboardi: > I would expect out values with a z coordinate here

Even when `include_z=False`, as it is here? (I have not added tests for anything more than default keyword arguments at this point – but I absolutely intend to).

sgillies:
@caspervdw @jGaboardi sorry for the late reply! I read through the thread above and everything looks like it's going in a good direction.

jGaboardi:
Just checking in here, been slammed w/ work + general "time-of-the-year" activities. But still trying to carve out time for a next iteration on this.

jGaboardi:
7ae53e9 provides timing comparison of:

* loop
* list comprehension
* map

These 3 method seem to be highly uniform.

```
In [1]: from shapely import get_segments
   ...: from shapely.tests.common import line_string, linear_ring
   ...: 
   ...: geoms = [line_string, linear_ring]
   ...: print(f"\ninput lines: {len(geoms)}")
   ...: %timeit get_segments(geoms, create_style="loop")
   ...: %timeit get_segments(geoms, create_style="list-comprehension")
   ...: %timeit get_segments(geoms, create_style="map")
   ...: 
   ...: geoms = [line_string, linear_ring] * 10
   ...: print(f"\ninput lines: {len(geoms)}")
   ...: %timeit get_segments(geoms, create_style="loop")
   ...: %timeit get_segments(geoms, create_style="list-comprehension")
   ...: %timeit get_segments(geoms, create_style="map")
   ...: 
   ...: geoms = [line_string, linear_ring] * 100
   ...: print(f"\ninput lines: {len(geoms)}")
   ...: %timeit get_segments(geoms, create_style="loop")
   ...: %timeit get_segments(geoms, create_style="list-comprehension")
   ...: %timeit get_segments(geoms, create_style="map")
   ...: 
   ...: geoms = [line_string, linear_ring] * 1000
   ...: print(f"\ninput lines: {len(geoms)}")
   ...: %timeit get_segments(geoms, create_style="loop")
   ...: %timeit get_segments(geoms, create_style="list-comprehension")
   ...: %timeit get_segments(geoms, create_style="map")
   ...: 
   ...: geoms = [line_string, linear_ring] * 10_000
   ...: print(f"\ninput lines: {len(geoms)}")
   ...: %timeit get_segments(geoms, create_style="loop")
   ...: %timeit get_segments(geoms, create_style="list-comprehension")
   ...: %timeit get_segments(geoms, create_style="map")

input lines: 2
27.3 μs ± 123 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
27.4 μs ± 105 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
28.6 μs ± 1.12 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

input lines: 20
90.7 μs ± 1.92 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
91.3 μs ± 3.43 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
87.2 μs ± 585 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

input lines: 200
681 μs ± 4.94 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
671 μs ± 4.52 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
692 μs ± 17.9 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

input lines: 2000
6.79 ms ± 126 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
6.64 ms ± 52 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
6.72 ms ± 63.5 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

input lines: 20000
67.5 ms ± 452 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
67.5 ms ± 429 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
67.5 ms ± 334 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

-----------------------

Tests are still in `legacy/test_get_segments.py` until functionality is decided upon, at which point I'll need to add more tests for `include_z`

jGaboardi:
@caspervdw @sgillies 

My next iteration in is https://github.com/shapely/shapely/commit/7ae53e93b20e680e6ce9191f34207ff75ed31b90.

See comments for feedback in:

* [#2306-3679959923](https://github.com/shapely/shapely/pull/2306#issuecomment-3679959923)
* [#2306-3602276657](https://github.com/shapely/shapely/pull/2306#pullrequestreview-3602276657)

Once I have yalls feedback on that I'll implement more testing.

caspervdw:
> [7ae53e9](https://github.com/shapely/shapely/commit/7ae53e93b20e680e6ce9191f34207ff75ed31b90) provides timing comparison of:
> 
>     * loop
> 
>     * list comprehension
> 
>     * map
> 

Thanks for the detailed comparison. There is little difference between the implemenation because in the end you are always iterating over the segments in Python. I added a variant that is entirely based on numpy ufuncs (which has the iteration done in compiled code):


```
In [7]: geoms = [line_string, linear_ring]
   ...: print(f"\ninput lines: {len(geoms)}")
   ...: %timeit get_segments(geoms, create_style="loop")
   ...: %timeit get_segments(geoms, create_style="indices")
   ...: 
   ...: geoms = [line_string, linear_ring] * 100
   ...: print(f"\ninput lines: {len(geoms)}")
   ...: %timeit get_segments(geoms, create_style="loop")
   ...: %timeit get_segments(geoms, create_style="indices")
   ...: 
   ...: geoms = [line_string, linear_ring] * 10_000
   ...: print(f"\ninput lines: {len(geoms)}")
   ...: %timeit get_segments(geoms, create_style="loop")
   ...: %timeit get_segments(geoms, create_style="indices")
   ...: 

input lines: 2
91.6 μs ± 163 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
94.9 μs ± 1.73 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

input lines: 200
1.59 ms ± 1.79 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
571 μs ± 13.4 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

input lines: 20000
159 ms ± 333 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
48.6 ms ± 2.04 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

jGaboardi:
@caspervdw very cool implementation. With further fleshed out comparison I am seeing 3x performance boost with your `indices` variant in all but the most trivial case.

```python
from shapely import get_segments
from shapely.tests.common import line_string, linear_ring

geoms = [line_string, linear_ring]
multipliers = [1, 10, 100, 1_000, 10_000, 100_000, 1_000_000]
create_styles = ["loop", "list-comprehension", "map", "indices"]

for multiplier in multipliers:
    _geoms = geoms * multiplier
    print(f"\n================================\ninput lines: {len(_geoms)}")
    for create_style in create_styles:
        print(f"-- {create_style} --")
        %timeit -o get_segments(_geoms, create_style=create_style)
```

jGaboardi:
@caspervdw  so your `indices` variant is by far the most performant and as equally robust as the other implementations. Shall I scrap the other variants in favor of that one now or should we get others' opinions?

martinfleis:
I would say go ahead with the numpy route. I don't immediately see any better solution.

jGaboardi:
#### still needs attention

* [x] [how to best handle NaN](https://github.com/shapely/shapely/pull/2306#pullrequestreview-3602276657)
  * Only allow as in the PR currently?
  * Expose [`HandleNaN`](https://github.com/shapely/shapely/blob/a26048bb2c0b357579965f5ab1b540cadb82a48f/shapely/creation.py#L27) to use in `_geometry.py`?
* [x] [add explicit `z` tests](https://github.com/shapely/shapely/pull/2306#discussion_r2592254536)
* [x] add explicit `return_index` tests
* [x] [add explicit empty linestrings and `None` values](https://github.com/shapely/shapely/pull/2306#discussion_r2592250638)
   * should not error, but return no segments 
* [x] move tests into `test_geometry.py`

caspervdw (commented):
This is getting real good! I think you covered all on your checklist, just some feedback on it:

- Handling NaN; I'd say, remove it from the PR altogether; NaNs in, NaNs out.
- Handling Z; You should indeed go through the 4 possible combinations of 2D/3D geometries with include_z True/False in your unittests
- `return_index` should be tested indeed
- test coverage should be expanded to empty linestrings and None input (expect no errors, just no output)
- Test should be in `test_geometry.py`, aligning with where you put the function.

jGaboardi:
Now that I have (hopefully) adequate testing cases/coverage, it seems the tests involving `line_string_nan` are causing some hiccups in several CI environments. Thoughts on how to handle this?

caspervdw:
> Now that I have (hopefully) adequate testing cases/coverage, it seems the tests involving `line_string_nan` are causing some hiccups in several CI environments. Thoughts on how to handle this?

Great! We have seen "invalid value encountered" more often, and it is due to C++ internals that set a warning flag on NaN values in some conditions and on some platforms. There is an "ignore_nan" decorator in other unittests in use (specifically: here https://github.com/shapely/shapely/blob/main/shapely/tests/test_geometry.py#L250). You can use the same approach here.

shapely/tests/test_geometry.py
```diff
@@ -693,3 +695,143 @@ def test_force_3d(geom, expected):
...
+        [linear_ring, np.array(out_ring)],
+        [[line_string, linear_ring], np.array(out_elbow + out_ring)],
+        [[None, empty_line_string, line_string], np.array(out_elbow)],
+        [line_string_nan, line_string_nan],
+        [empty_line_string, np.array([])],
+    ],
+)
+def test_get_segments_defaults(geoms, expected):
+    actual = shapely.get_segments(geoms)
+    assert_geometries_equal(actual, expected)
```
caspervdw: The z and m coordinatse should be dropped here, please test that

caspervdw (approved):
Some minor points, can be merge once they are addressed.

Thanks for the contribution @jGaboardi !

jGaboardi:
@caspervdw – should I adapt the failing `z` tests here from GEOS 3.{10.11} or simply skip if GEOS<3.12?

caspervdw:
> @caspervdw – should I adapt the failing `z` tests here from GEOS 3.{10.11} or simply skip if GEOS<3.12?

Thats right, the has_z acts differently on NaN coordinates for older GEOS. Nothing we can do about it. A version-dependent skip is allright.

jGaboardi:
@caspervdw fully green CI now

martinfleis:
This will need a note in the changelog.

jGaboardi:
> This will need a note in the changelog.

Something like [this](https://github.com/shapely/shapely/blob/9334a3bac2af533cdf9f1acd28d752920314a5c0/CHANGES.txt#L145-L151) in the [`v2.2.0` section](https://github.com/shapely/shapely/blob/9334a3bac2af533cdf9f1acd28d752920314a5c0/CHANGES.txt#L4)?

caspervdw:
Just merged, thanks for the work @jGaboardi

jGaboardi:
xref https://github.com/shapely/shapely/issues/2413

PR #2313, opened 2025-07-11
ENH: Add minimum_width function (minimum diameter)

### LINKED ISSUE #2312: Add support for GEOS Minimum Diameter/Width functionality
# Add support for GEOS Minimum Diameter/Width functionality

## Summary

This issue proposes adding support for the GEOS `MinimumDiameter` functionality to Shapely, providing access to compute the minimum width (minimum diameter) of geometries.

## Background

GEOS provides a `MinimumDiameter` class that computes "the width of the smallest band that contains the geometry, where a band is a strip of the plane defined by two parallel lines." This is also known as the minimum diameter.

This functionality is complementary to the existing `minimum_rotated_rectangle` (oriented envelope) function, but serves a different purpose:
- `minimum_rotated_rectangle`: Returns the minimum area rectangle that encloses the geometry
- `minimum_width`: Returns a LineString representing the shortest distance across the geometry

## Proposed API

Add a new function `minimum_width()` to `shapely.constructive`:

```python
def minimum_width(geometry, **kwargs):
    """Compute the minimum width (minimum diameter) of a geometry.

    The minimum width is defined as the width of the smallest band that 
    contains the geometry, where a band is a strip of the plane defined 
    by two parallel lines. This is also known as the minimum diameter.

    For most geometry types, this function returns a LineString representing
    the minimum width line segment. For degenerate cases (points), it may
    return the point itself.

    Parameters
    ----------
        Geometry or geometries for which to compute the minimum width.
    **kwargs
        See NumPy ufunc docs for other keyword arguments.

    Returns
    -------
    LineString or Geometry
        A LineString representing the minimum width, or the geometry itself
        for degenerate cases.

    Examples
    --------
    >>> import shapely
    >>> from shapely import Point, LineString, Polygon
    >>> 
    >>> # Square has minimum width equal to side length
    >>> square = Polygon([(0, 0), (1, 0), (1, 1), (0, 1), (0, 0)])
    >>> width_line = shapely.minimum_width(square)
    >>> width_line.length
    1.0
    >>> 
    >>> # Rectangle's minimum width is the shorter side
    >>> rect = Polygon([(0, 0), (3, 0), (3, 1), (0, 1), (0, 0)])
    >>> width_line = shapely.minimum_width(rect)
    >>> width_line.length
    1.0
    """
    return lib.minimum_width(geometry, **kwargs)
```

## Example Usage

```python
import shapely
from shapely import Polygon, LineString

# Find minimum width of various shapes
square = Polygon([(0, 0), (1, 0), (1, 1), (0, 1), (0, 0)])
rectangle = Polygon([(0, 0), (3, 0), (3, 1), (0, 1), (0, 0)])
triangle = Polygon([(0, 0), (2, 0), (1, 2), (0, 0)])

print("Square minimum width:", shapely.minimum_width(square).length)       # 1.0
print("Rectangle minimum width:", shapely.minimum_width(rectangle).length) # 1.0  
print("Triangle minimum width:", shapely.minimum_width(triangle).length)   # ~1.79

# Compare with minimum rotated rectangle
min_width_line = shapely.minimum_width(rectangle)
min_rect = shapely.minimum_rotated_rectangle(rectangle)

print("Minimum width line:", min_width_line)  # LINESTRING (0 1, 0 0)
print("Minimum rectangle:", min_rect)         # POLYGON ((0 0, 0 1, 3 1, 3 0, 0 0))
```

## Implementation Details

The implementation would:

1. Add `GEOSMinimumWidth_r` to the C extension ufuncs (similar to existing `GEOSMinimumRotatedRectangle_r`)
2. Follow the same patterns as `minimum_bounding_circle` and `oriented_envelope`
3. Include comprehensive tests covering various geometry types
4. Support vectorized operations via NumPy ufuncs
5. Maintain compatibility with existing Shapely conventions

## Relationship to Existing Functions

- **`minimum_rotated_rectangle`/`oriented_envelope`**: Returns minimum area rectangle
- **`minimum_bounding_circle`**: Returns minimum enclosing circle  
- **`minimum_width`** (proposed): Returns minimum width line segment

These functions complement each other for different geometric analysis needs.

## GEOS Compatibility

This feature requires GEOS with `MinimumDiameter` support. The function is available in GEOS C API as `GEOSMinimumWidth_r`.

---

Would this be a valuable addition to Shapely? I'm happy to work on the implementation if there's interest from the maintainers.

adriancaruana:
Add support for GEOS Minimum Diameter/Width functionality

Closes #2312

### DIFF
--- a/shapely/tests/test_constructive.py
+++ b/shapely/tests/test_constructive.py
@@ -80,7 +80,7 @@ def test_no_args_array(geometry, func):
 def test_float_arg_array(geometry, func):
     if (
         func is shapely.offset_curve
-        and shapely.get_type_id(geometry) not in [1, 2]
+        and shapely.get_type_id(geometry) not in {1, 2}
         and shapely.geos_version < (3, 11, 0)
     ):
         with pytest.raises(GEOSException, match="only accept linestrings"):
@@ -1153,13 +1153,13 @@ def test_voronoi_polygons_ordered_raise():
 
 @pytest.mark.parametrize("geometry", all_types)
 def test_maximum_inscribed_circle_all_types(geometry):
-    if shapely.get_type_id(geometry) not in [3, 6]:
+    if shapely.get_type_id(geometry) not in {3, 6}:
         # Maximum Inscribed Circle is only supported for (Multi)Polygon input
         with pytest.raises(
             GEOSException,
             match=(
-                "Argument must be Polygonal or LinearRing|"  # GEOS < 3.10.4
-                "must be a Polygon or MultiPolygon|"
+                r"Argument must be Polygonal or LinearRing|"  # GEOS < 3.10.4
+                r"must be a Polygon or MultiPolygon|"
                 "Operation not supported by GeometryCollection"
             ),
         ):
@@ -1168,7 +1168,7 @@ def test_maximum_inscribed_circle_all_types(geometry):
 
     if geometry.is_empty:
         with pytest.raises(
-            GEOSException, match="Empty input(?: geometry)? is not supported"
+            GEOSException, match=r"Empty input(?: geometry)? is not supported"
         ):
             shapely.maximum_inscribed_circle(geometry)
         return
@@ -1201,7 +1201,7 @@ def test_maximum_inscribed_circle_empty():
     with pytest.raises(
         GEOSException,
         match=(
-            "Argument must be Polygonal or LinearRing|"  # GEOS < 3.10.4
+            r"Argument must be Polygonal or LinearRing|"  # GEOS < 3.10.4
             "must be a Polygon or MultiPolygon"
         ),
     ):
@@ -1209,7 +1209,7 @@ def test_maximum_inscribed_circle_empty():
 
     geometry = shapely.from_wkt("POLYGON EMPTY")
     with pytest.raises(
-        GEOSException, match="Empty input(?: geometry)? is not supported"
+        GEOSException, match=r"Empty input(?: geometry)? is not supported"
     ):
         shapely.maximum_inscribed_circle(geometry)
 
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -8,7 +8,7 @@ jobs:
     steps:
       - uses: actions/checkout@v5
 
-      - uses: actions/setup-python@v5
+      - uses: actions/setup-python@v6
         with:
           python-version: 3.x
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,7 +33,7 @@ jobs:
           fetch-depth: 0
 
       - name: Set up Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           python-version: "3.x"
 
@@ -44,7 +44,7 @@ jobs:
           python -m build --sdist
           twine check --strict dist/*
 
-      - uses: actions/upload-artifact@v4
+      - uses: actions/upload-artifact@v5
         with:
           name: release-sdist
           path: ./dist/*.tar.gz
@@ -113,7 +113,7 @@ jobs:
         shell: bash
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v3.1.3
+        uses: pypa/cibuildwheel@v3.2.1
         env:
           CIBW_ARCHS: ${{ matrix.arch }}
           # TEMP don't use automated/isolated build environment, but manually
@@ -131,7 +131,7 @@ jobs:
           CIBW_MUSLLINUX_AARCH64_IMAGE: ${{ matrix.baseimage }}_geos:${{ env.GEOS_VERSION }}
 
       - name: Upload artifacts
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
           name: release-linux-${{ matrix.baseimage }}
           path: ./wheelhouse/*.whl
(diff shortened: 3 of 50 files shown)


adriancaruana:
@sgillies Can you please approve the workflows for my most recent changes?

adriancaruana:
@sgillies Thoughts on merging this?

shapely/constructive.py
```diff
@@ -1429,6 +1430,44 @@ def minimum_bounding_circle(geometry, **kwargs):
...
+@multithreading_enabled
+def minimum_width(geometry, **kwargs):
+    """Compute the minimum width (minimum diameter) of a geometry.
+
+    The minimum width is defined as the width of the smallest band that
+    contains the geometry, where a band is a strip of the plane defined
+    by two parallel lines. This is also known as the minimum diameter.
+
+    For most geometry types, this function returns a LineString representing
+    the minimum width line segment. For degenerate cases (points), it may
```
jorisvandenbossche, line 1442: Straight lines are also such an example of a degenerate case?

shapely/tests/test_constructive.py
```diff
@@ -972,6 +972,34 @@ def test_minimum_bounding_circle(geometry, expected):
     assert_geometries_equal(actual, expected)
 
 
+@pytest.mark.parametrize("geometry", all_types)
+def test_minimum_width_all_types(geometry):
+    actual = shapely.minimum_width([geometry, geometry])
```
jorisvandenbossche, line 977: You could also add `shapely.minimum_width` to `CONSTRUCTIVE_NO_ARGS` defined above and then this test should be covered by an existing one

shapely/tests/test_constructive.py
```diff
@@ -972,6 +972,34 @@ def test_minimum_bounding_circle(geometry, expected):
...
+        (LineString([(0, 0), (1, 1)]), LineString),
+        (Polygon([(0, 0), (1, 0), (1, 1), (0, 1), (0, 0)]), LineString),
+        (empty_point, LineString),
+        (empty_line_string, LineString),
+        (empty_polygon, LineString),
+    ],
+)
+def test_minimum_width(geometry, expected_type):
+    actual = shapely.minimum_width(geometry)
+    assert isinstance(actual, expected_type)
```
jorisvandenbossche, line 998: In the docstring you mention the degenerate case of a Point returning a Point instead of a LineString, but the above test conflicts with that?

jorisvandenbossche (commented):
@adriancaruana apologies for the slow follow-up, and thanks a lot for the PR!

This is a nice addition, just a few minor comments on the tests

shapely/constructive.py
```diff
@@ -1429,6 +1430,44 @@ def minimum_bounding_circle(geometry, **kwargs):
...
+        Geometry or geometries for which to compute the minimum width.
+    **kwargs
+        See :ref:`NumPy ufunc docs <ufuncs.kwargs>` for other keyword arguments.
+
+    Examples
+    --------
+    >>> import shapely
+    >>> from shapely import Point, LineString, Polygon
+    >>> polygon = Polygon([(0, 0), (1, 0), (1, 1), (0, 1), (0, 0)])
+    >>> shapely.minimum_width(polygon)  # doctest: +SKIP
```
jorisvandenbossche, line 1457: This skips should not be needed?

```suggestion
    >>> shapely.minimum_width(polygon)
```

shapely/constructive.py
jorisvandenbossche: So quickly testing with your PR, it seems that even in the case of points, a 0-length Linestring gets returned?

adriancaruana: Good catch. I've reconsidered this, and IMHO, the behaviour of this function should match GEOS,
which always returns a LineString. I'm happy to discuss if you disagree thouguh.

adriancaruana:
Just bumping this again @jorisvandenbossche. 😄

shapely/tests/test_constructive.py
```diff
@@ -972,6 +973,24 @@ def test_minimum_bounding_circle(geometry, expected):
...
+    [
+        (Point(0, 0), LineString),
+        (LineString([(0, 0), (1, 1)]), LineString),
+        (Polygon([(0, 0), (1, 0), (1, 1), (0, 1), (0, 0)]), LineString),
+        (empty_point, LineString),
+        (empty_line_string, LineString),
+        (empty_polygon, LineString),
+    ],
+)
+def test_minimum_width(geometry, expected_type):
```
jorisvandenbossche, line 987:
```suggestion
@pytest.mark.parametrize("geometry", all_types)
def test_minimum_width(geometry):
```

Given that we are not testing exact results but just that it returns a LineString, I think we can also use the `all_types` list to avoid redefining some specific geometries

shapely/tests/test_constructive.py
```diff
@@ -972,6 +973,24 @@ def test_minimum_bounding_circle(geometry, expected):
...
+    actual = shapely.minimum_width(geometry)
+    assert isinstance(actual, expected_type)
```
jorisvandenbossche, line 989:
```suggestion
    assert isinstance(actual, LineString)
```

Since it is always the same, can hardcode this instead of having it part of the parametrization

jorisvandenbossche (approved):
Thanks for the update! Two small remaining suggestions on the test, then good to go in

adriancaruana:
Thanks @jorisvandenbossche, should be ready to merge now once the tests go green.

PR #2317, opened 2025-08-05
CI: update windows-2019 to windows-2022

jorisvandenbossche:
The windows-2019 image was deprecated and now the builds were no longer running


jorisvandenbossche:
Failing tests on main are being addressed in https://github.com/shapely/shapely/pull/2311

PR #2318, opened 2025-08-06
TST: update test_coverage_union_overlapping_inputs for upstream GEOS change

jorisvandenbossche:
Upstream GEOS recently changed so this now raises an error instead of silently returning the invalid input unchanged.

xref https://github.com/libgeos/geos/pull/1279#discussion_r2253380355

### DIFF
--- a/shapely/set_operations.py
+++ b/shapely/set_operations.py
@@ -562,6 +562,9 @@ def coverage_union(a, b, **kwargs):
 
     This is an optimized version of union which assumes the polygons to be
     non-overlapping.
+    If this assumption is not met, the exact result is not guaranteed
+    (depending on the GEOS version, it may return the input unchanges or raise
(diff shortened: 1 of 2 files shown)


shapely/set_operations.py
```diff
@@ -562,6 +562,9 @@ def coverage_union(a, b, **kwargs):
 
     This is an optimized version of union which assumes the polygons to be
     non-overlapping.
+    If this assumption is not met, the exact result is not guaranteed
+    (depending on the GEOS version, it may return the input unchanges or raise
```
dbaston, line 566:
```suggestion
    (depending on the GEOS version, it may return the input unchanged or raise
```

jorisvandenbossche: Thanks for the catch!

PR #2319, opened 2025-08-12
Request Add Feature : maximum_inscribed_rectangle similar to maximum_inscribed_circle

shiv111202:
# Feature Request: `maximum_inscribed_rectangle` similar to `maximum_inscribed_circle`

## Description
Currently, Shapely provides `maximum_inscribed_circle` but no equivalent for rectangles.  
I propose adding a `maximum_inscribed_rectangle` function to compute the largest **axis-aligned** rectangle fully contained within a polygon.

---

## Proposed Functionality
```python
from shapely.algorithms import maximum_inscribed_rectangle

rect = maximum_inscribed_rectangle(polygon, resolution=20)
```

**Parameters:**
- `polygon` – Shapely `Polygon`
- `resolution` – number of grid points to sample in each axis

**Returns:**
- Shapely `Polygon` representing the largest inscribed rectangle

---

## Algorithm Outline
1. Generate an evenly spaced grid over the polygon’s bounding box.
2. Use a binary mask to mark grid cells inside the polygon.
3. For each grid row, use a histogram-based largest rectangle search.
4. Adjust rectangle edges to fully fit inside the polygon.
5. Return the rectangle with the largest area found.

---

## Performance Benchmarks (local tests)
- **Resolution = 100** for a 1,000-vertex polygon: ~10–15 ms  
- Scales linearly with resolution.

---

## Example Output
_Input polygon (concave):_  
example_input

_Largest inscribed rectangle:_  
example_output

---

## Edge Cases Handled
- Concave polygons
- Empty geometries (returns empty polygon instead of `None`)

---

## Why This is Useful
- Architectural layout optimizations
- Agricultural field planning
- Packing and cutting problems


shiv111202:
Working on better version

PR #2343, opened 2025-10-31
perf: Use a threadlocal GEOS context

### LINKED ISSUE #1283: Raising GEOS "notice" messages as warnings not working?
Moved from https://github.com/pygeos/pygeos/issues/214, and related to the logging discussion at #998

Currently, we have the intention to use warnings for the "notice" messages of GEOS (and exceptions for "error" messages):

https://github.com/shapely/shapely/blob/48cf81e5ba3e48b779e81be543ab42858385c37d/src/geos.h#L130

https://github.com/shapely/shapely/blob/48cf81e5ba3e48b779e81be543ab42858385c37d/src/geos.h#L64-L67

Looking into the GEOS source code to see what these "notices" are used for, I *think* it is actually only used for the `is_valid` operation (see https://github.com/libgeos/geos/blob/27516e0bdb7526699c00fef328c61561ab7a4b47/capi/geos_ts_c.cpp#L688). All other errors are sent to the "error" handler. 

From my reading of the GEOS code, I expected that this would raise a warning:

```python
import warnings
warnings.simplefilter("always")

import shapely
import shapely.geometry
coords = [(0, 0), (0, 2), (1, 1), (2, 2), (2, 0), (1, 1), (0, 0)]
bowtie = shapely.geometry.Polygon(coords)
assert bowtie.is_valid is False
assert shapely.is_valid_reason(bowtie) == "Ring Self-intersection[1 1]"
```

However, running the above code does not actually raise a warning on the main branch.

As comparison, the equivalent with Shapely 1.8 and setting up the logger to print to stdout, _does_ show the message from GEOS:

```python
import sys
import logging

logger = logging.getLogger('shapely.geos')
logging_handler_out = logging.StreamHandler(sys.stdout)
logger.addHandler(logging_handler_out)
logger.setLevel(logging.INFO)

coords = [(0, 0), (0, 2), (1, 1), (2, 2), (2, 0), (1, 1), (0, 0)]
bowtie = Polygon(coords)

In [9]: bowtie.is_valid
Ring Self-intersection at or near point 1 1
Out[9]: False
```

### LINKED ISSUE #1523: PERF: Reintroduce a global (or threadlocal) GEOS context

caspervdw:
**Description**

This PR changes the GEOS context creation and finalization **for every GEOS call**. Instead of creating / destroying the GEOS context at every call, it is reused from a threadlocal.

It does so by changing the `GEOS_INIT[_THREADS]` macros (and the finish macros accordingly).

It should give a performance increase of about 75 ns per call, which is especially apparent for scalar calls. See #2035 for benchmarks

Closes #1523 

Also closes #1283 , because the Notice/warning handlers are completely removed. They weren't used anyway.

### DIFF
--- a/src/geos.h
+++ b/src/geos.h
@@ -46,8 +46,9 @@
  * - Errors are captured in the 'errstate' and 'last_error' variables, that are
  *   initialized in GEOS_INIT.
  * - Errors are converted to appropriate Python exceptions in the GEOS_FINISH macro.
- *   This means that PyErr_SetString is called; it is up to the caller to return NULL
- *   in that case (after GEOS_FINISH, by checking `errstate == PGERR_SUCCESS`).
+ *   PyErr_SetString will be called if `errstate != PGERR_SUCCESS`. When writing Python
+ *   C extension functions, the caller should check `errstate` and return `NULL` in that
+ *   case. For NumPy ufuncs, this is not applicable because they return `NULL` anyway.
  * - In case of a GEOS error, the 'last_error' buffer is used as a temporary storage
  *   for the error message. This buffer is reused (as threadlocal).
  *
@@ -63,11 +64,18 @@
  *     goto finish;
  * }
  *
- * // More GEOS operations...
+ * // More operations (that may adjust `errstate`)...
  *
  * finish:
  *   GEOS_FINISH;  // Clean up context and convert errors to Python exceptions
  *                 // Use GEOS_FINISH_THREADS if GEOS_INIT_THREADS was used
+ *
+ * // Python C extension function should check errstate and return NULL on error
+ * if (errstate != PGERR_SUCCESS) {
+ *     return NULL;
+ * }
+ * return Py_BuildValue(...);  // Return appropriate Python object
+ *
  */
 
 // Define the error states


shapely/_geos.pyx
```diff
@@ -20,26 +13,11 @@ cdef class get_geos_handle:
...
+        self.handle = PyGEOS_InitGEOSContext()
+        self.last_error = PyGEOS_InitGEOSErrorBuffer()
         return self.handle
 
     def __exit__(self, type, value, traceback):
-        try:
-            if self.last_error[0] != 0:
-                raise GEOSException(self.last_error)
-            if self.last_warning[0] != 0:
-                warnings.warn(self.last_warning)
```
jorisvandenbossche: You removed the warning part because that doesn't get used in practice?

caspervdw:
> Looking good!

Thanks for the review!

> So this uses a thread-local GEOS context handle for everything, right? I.e. not just for the scalars calls under the GIL, but also for the ufuncs (which right now still initialized a context per ufunc call)?

Yes it does change the initialization for everything. It really needs some battle-testing (especially using multithreading).
I rewrote the docstring in `geos.h` and also put some description on this PR.

src/geos.h
```diff
@@ -21,33 +21,54 @@
...
+ *   thread and sets this in the local `ctx` variable.
+ * - The GEOS_INIT_THREADS does the same, but also releases the GIL allowing other Python
+ *   threads to run concurrently. GEOS_FINISH_THREADS must be used to reacquire the GIL.
+ *
+ * Finalization and Error Handling:
+ * - Errors are captured in the 'errstate' and 'last_error' variables, that are
+ *   initialized in GEOS_INIT.
+ * - Errors are converted to appropriate Python exceptions in the GEOS_FINISH macro.
+ *   This means that PyErr_SetString is called; it is up to the caller to return NULL
+ *   in that case (after GEOS_FINISH, by checking `errstate == PGERR_SUCCESS`).
```
jorisvandenbossche: This is not directly clear to me (without looking at our C code to check for examples, but also there we typically don't do anything _after_ GEOS_FINISH, except for freeing some intermediate data?)

Could this be reflected in the example below?

PR #2345, opened 2025-11-12
perf: Improve performance of scalar Y -> b functions

caspervdw:
This extends the work started in #2339 for the boolean functions.

Two additional remarks:

- I removed some obsolete comments from `geos_funcs_Y_d.c`
- The `is_geometry` was defined twice in the `ufuncs.c`. Once in the `DEFINE_Y_b` and once in the `DEFINE_O_b`. It should have been only `DEFINE_O_b`. I guess it was no problem because the second definition was the correct one and it probably overwrites the first. In this PR I did not move this mistake to the new `geos_funcs_Y_b.c` file.

### DIFF
--- a/src/ufuncs.c
+++ b/src/ufuncs.c
@@ -3573,11 +3573,6 @@ finish:
 }
 static PyUFuncGenericFunction to_geojson_funcs[1] = {&to_geojson_func};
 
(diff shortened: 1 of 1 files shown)


caspervdw:
Test failures are unrelated and fixed in https://github.com/shapely/shapely/pull/2346

PR #2346, opened 2025-11-12
maint: Replace unittest.SkipTest with pytest.skip

caspervdw:
This fixes test failures with pytest 9 on main.

### DIFF
--- a/shapely/tests/legacy/test_locale.py
+++ b/shapely/tests/legacy/test_locale.py
@@ -4,6 +4,8 @@ import locale
 import sys
 import unittest
 
+import pytest
+
 from shapely.wkt import dumps, loads
 
 # Set locale to one that uses a comma as decimal separator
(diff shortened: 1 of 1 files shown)


jorisvandenbossche (approved):

caspervdw:
@jorisvandenbossche  this probably needs a backport, how do you do this normally? Through a PR or just pushing to the correct branch?

jorisvandenbossche:
I don't know if we are planning another release on 2.1.x (we did one in september for Python 3.14 support, for now I don't think there any other issues for which we would do another bugfix release)

For keeping the maint branch CI passing, feel free to push a commit once in a while (if it is a backport that applies cleanly, it the logic is customized for the maint branch, I think it is better for visibility to go through a PR). 
When actually working towards a release, and at that point backporting more things (including bug fixes), I typically create a PR with a bunch of the commits in one go (which makes it easier to check CI etc). See eg https://github.com/shapely/shapely/pull/2329 for the last release.

caspervdw:
Backported into `maint-2.1` (be4cdf4b1cfe0b62604e3ccad3462416c6c5d5e8)

PR #2355, opened 2025-11-30
perf: Improve performance of scalar Y, i -> Y functions

caspervdw:
This extends the work started in https://github.com/shapely/shapely/pull/2339 for the integer functions.

**Notes**

My first attempt used `METH_VARARGS` for the scalar case. I got 319 ns for a typical `get_geometry_scalar` call. Then I changed it to using `METH_FASTCALL`, which is new since Python 3.10. This improved performance to 232 ns per call. Consequence of this is that we have to cast to `int` on the python side, else it will error on numpy int types.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   test:
-    name: ${{ matrix.os }}-${{ matrix.architecture }} Py${{ matrix.python }} GEOS ${{ matrix.geos }}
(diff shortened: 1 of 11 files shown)


jorisvandenbossche:
> Consequence of this is that we have to cast to int on the python side, else it will error on numpy int types.

We have some tests that cover passing a numpy scalar to the scalar function?

jorisvandenbossche (approved):

caspervdw:
Test failures are unrelated and handled by #2367

PR #2357, opened 2025-12-03
BLD: Use meson-python for build-backend

### LINKED ISSUE #2136: Build backend for next development cycle
Now's an OK time to discuss ideas for Shapely's build backend (this caries on a discussion from #2132).

Like many mature Python projects, Shapely's build backend has always been based on setuptools with much of the control in `setup.py`. But since [PEP 517](https://peps.python.org/pep-0517/), alternative build backends have popped up.

Some options:

1. No change; keep with [setuptools](https://setuptools.pypa.io/)
2. Migrate to [meson-python](https://github.com/mesonbuild/meson-python) which is based on the [Meson Build System](https://mesonbuild.com/)
3. Migrate to [scikit-build-core](https://github.com/scikit-build/scikit-build-core) which is based on [CMake](https://cmake.org/)

----

FWIW, I've been exploring meson-python in [one of my branches](https://github.com/mwtoews/shapely/tree/meson-python). I might explore a scikit-build-core option too (after all, I'm more familiar with CMake than Meson). But the easy option 1 could ultimately be fine too!

mwtoews:
This changes the project build-backend from setuptools to meson-python (aka mesonpy).

For now, it disables versioneer from creating fancy versions from git tags. This can be re-enabled and/or refactored, similar to scipy and pandas (which also use same build backend and fancy version generation). Depending on how this PR goes, I'll work on that next step.

Closes #2136


sgillies:
@mwtoews as I wrote in https://github.com/shapely/shapely/issues/2136, I'm not enthusiastic about this.

meson.build
```diff
@@ -0,0 +1,51 @@
...
+cython_args = ['-3', '--fast-fail', '--output-file', '@OUTPUT@', '--include-dir', '@BUILD_ROOT@', '@INPUT@']
+cython_gen = generator(cython,
+  arguments : cython_args,
+  output : '@BASENAME@.c',
+)
+
+numpy_dep = dependency('numpy')
+incdir_numpy = run_command(
+    py,
+    [
```
eli-schwartz: I don't think you need both of these -- a number of projects have it due to predating https://mesonbuild.com/Dependencies.html#numpy and numpy 2.

@rgommers can correct me if I'm wrong, but I'm not aware of any post-1.4.0 users who still need this? But if you do need to support numpy < 2, you may want to limit to only the incdir approach?

shapely/meson.build
```diff
@@ -0,0 +1,58 @@
+py.extension_module(
+  '_geometry_helpers',
+  cython_gen.process('_geometry_helpers.pyx'),
+  include_directories: ['../src', inc_np],
+  dependencies : [py_dep, numpy_dep, geos_dep],
```
eli-schwartz: Note since meson 0.63.0, `py_dep` is implicit: https://mesonbuild.com/Python-module.html#extension_module

meson.build
rgommers: To be fair I've been using both in other projects, to combine the better logging of `dependency('numpy')` (it shows you the numpy version number) with the numpy 1.x support of `np.get_include()`. 

One other case where it marginally helps is cross-compiling using `crossenv` (which, e.g., conda-forge still uses), where `dependency('numpy')` only works when the recipe is changed to use `numpy.pc` while the manual thing works out of the box (because crossenv contains the hacks needed to make a native package act like a host one, so no QEMU needed).

tl;dr I think it's fine to leave as is, and move to `dependency('numpy')`-only only after scipy et al do.

pyproject.toml
```diff
@@ -21,12 +21,11 @@ maintainers = [
 description = "Manipulation and analysis of geometric objects"
 readme = "README.rst"
 keywords = ["geometry", "topology", "gis"]
-license = {text = "BSD 3-Clause"}
+license = "BSD-3-Clause"
```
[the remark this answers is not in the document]
mwtoews: Thanks for the suggestion. Since this draft PR, I've already merged #2369. I was considering also adding `license-files` too, but I probably held back since it would only add a slightly outdated [`CREDITS.txt`](https://github.com/shapely/shapely/blob/main/CREDITS.txt) file.

mwtoews:
This PR is nearly ready for review. Everything builds fine! Docs might need more work. However, currently stuck with this somewhat common runtime issue only for Windows editable installs:
```
    import shapely; print(f'GEOS version: {shapely.geos_version_string}')
    ^^^^^^^^^^^^^^
  File "D:\a\shapely\shapely\shapely\__init__.py", line 3, in <module>
    from shapely.lib import GEOSException
ImportError: DLL load failed while importing lib: The specified module could not be found.
Error: Process completed with exit code 1.
```
There is no obvious solution to this one. Here are some candidate ideas:

- Copy GEOS dependency DLLs (geos.dll and geos_c.dll) to a "magic path", possibly one from on `sys.path`. A similar approach was used for setuptools, where DLLs were copied into the local `shapely/` package directory, but no longer works. Also copying these to `build/cp3xx/shapely/` (with `*.pyd` files) does not work.
- Add some hook to `shapely/__init__.py` to call (e.g.) `os.add_dll_directory(libs_dir)` similar to what delvewheel does, except this seems invasive to suite the needs of one install type for Windows.
- Somehow append some logic to meson python's `site-packages/_shapely_editable_loader.py` (based on [`mesonpy/_editable.py`](https://github.com/mesonbuild/meson-python/blob/main/mesonpy/_editable.py)) to call `os.add_dll_directory(libs_dir)`

Ideas welcome!

eli-schwartz (changes requested):

rgommers (commented):
Overall this looks quite good, just a few more comments from me.

meson.build
```diff
@@ -0,0 +1,92 @@
...
+python_version = py.language_version()
+if python_version.version_compare(f'<@min_python_version@')
+  error(f'Minimum supported Python version is @min_python_version@, found @python_version@')
+endif
+
+# silence warnings
+add_project_arguments('-DNPY_NO_DEPRECATED_API=0', language : 'c')
+
+# see also min_numpy_version
+add_project_arguments('-DNPY_TARGET_VERSION=NPY_1_23_API_VERSION', language : 'c')
```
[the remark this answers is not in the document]
mwtoews: This was translated from setup.py, and is likely a left-over, now removed.

.github/workflows/release.yml
```diff
@@ -35,10 +33,13 @@ jobs:
...
         uses: actions/setup-python@v6
         with:
-          python-version: "3.x"
+          python-version-file: "pyproject.toml"
 
       - name: Build a source tarball
         run: |
+          sudo -E apt-get update
+          sudo -E apt-get autopurge -y needrestart
+          sudo -E apt-get -yq --no-install-suggests --no-install-recommends install libgeos-dev
```
jorisvandenbossche, line 43: Building the sdist now requires GEOS to be available? (but I assume it still does not actually use it, it is just the meson setup that looks for it unconditionally?)

.github/workflows/release.yml
```diff
@@ -188,35 +181,26 @@ jobs:
           arch: ${{ matrix.msvc_arch }}
         if: ${{ matrix.msvc_arch }}
 
+      - name: Remove /usr/bin/link (MSVC workaround)
+        shell: bash
+        run: rm -rfv /usr/bin/link
+        if: ${{ matrix.msvc_arch }}
```
jorisvandenbossche: Do you remember what this was needed for? If it is still needed, might be good to expand with some more detailed comment

.github/workflows/release.yml
```diff
@@ -188,35 +181,26 @@ jobs:
...
+
       - name: Build wheels
         uses: pypa/cibuildwheel@v3.3.1
         env:
           CIBW_ARCHS: ${{ matrix.arch }}
           CIBW_ENVIRONMENT_MACOS:
             GEOS_INSTALL=${{ runner.temp }}/geos-${{ env.GEOS_VERSION }}
-            GEOS_CONFIG=${{ runner.temp }}/geos-${{ env.GEOS_VERSION }}/bin/geos-config
```
jorisvandenbossche, line 191: Is the `GEOS_INSTALL` one still used? 
(since you specify `setup-args="-Dcmake_prefix_path=${{ runner.temp }}/geos-${{ env.GEOS_VERSION }}"` below that does not use the GEOS_INSTALL env variable but just defines it itself, I would expect not?)

docs/environment.yml
```diff
@@ -11,3 +11,5 @@ dependencies:
   - numpydoc=1.1
   - matplotlib
   - pip
+  - cmake
+  - meson
```
jorisvandenbossche, line 15: meson-python as well?

docs/installation.rst
```diff
@@ -55,49 +55,69 @@ On macOS::
...
-enable you to build Shapely yourself:
-
-- Get a C compiler applicable to your Python version (https://wiki.python.org/moin/WindowsCompilers)
-- Download and install a GEOS binary (https://trac.osgeo.org/osgeo4w/)
-- Set GEOS_INCLUDE_PATH and GEOS_LIBRARY_PATH environment variables (see below
-  for notes on GEOS discovery)
-- Run ``pip install shapely --no-binary``
-- Make sure the GEOS .dll files are available on the PATH
+    $ GEOS_INSTALL=/path/to/geos-3.14.1
+    $ pip install shapely --no-binary shapely -Csetup-args=-Dcmake_prefix_path="$GEOS_INSTALL"
```
jorisvandenbossche, line 63: Is this example specific to linux? Or should passing `-Dcmake_prefix_path` work on every platform (assuming the path is valid for that platform, of course)

docs/installation.rst
```diff
@@ -55,49 +55,69 @@ On macOS::
...
     $ git clone git@github.com:shapely/shapely.git
     $ cd shapely/
 
-Install it in development mode using ``pip``::
+Local development with ``pip`` is enabled with an "editable" mode, where
+edits to the source code become effective without the need of a new
+installation step. As of Shapely 2.2, the build backend was changed to
+meson-python, which requires `a few other instructions
+<https://mesonbuild.com/meson-python/how-to-guides/editable-installs.html>`__,
+such as ``--no-build-isolation``.
```
jorisvandenbossche, line 85: (not super important, but I don't know if this is actually true. At least, I have been using `--no-build-isolation` already for a long time for all packages with extension modules)

docs/installation.rst
```diff
@@ -55,49 +55,69 @@ On macOS::
...
-    (env) $ pip install -e .[test]
+    $ python3 -m venv .venv
+    $ source .venv/bin/activate
+    $ pip install -r requirements-dev.txt
+
+Install shapely in editable mode using ``pip``:
+
+.. code-block:: console
+
+    $ pip install --no-build-isolation -e .[test]
```
jorisvandenbossche: As recommendation for local development, I would also include `-Ceditable-verbose=true`

docs/installation.rst
```diff
@@ -55,49 +55,69 @@ On macOS::
...
+
+Conda environments
+^^^^^^^^^^^^^^^^^^
 
-Or using ``conda``:
+Development within conda environments requires additional packages installed
+with the ``conda`` command, including pre-built distributions of GEOS.
```
jorisvandenbossche, line 119: (not too important, but it feels a bit strange to call out that you need to install GEOS in the conda case, and not in the venv case)

docs/installation.rst
jorisvandenbossche: (also, once you update with latest main, be sure to get the `--group dev` change instead of the extra)

docs/installation.rst
```diff
@@ -111,31 +131,40 @@ Shapely can be tested using ``pytest``::
...
 
-On Windows, there is no ``geos-config`` and the include and lib folders need to
-be specified manually in any case::
+.. code-block:: console
+
+    $ pip install --no-build-isolation -e . -Csetup-args="-Dcmake_prefix_path=$HOME/opt/geos"
+
+For Linux, if the GEOS library is not on the dynamic linker run-time path (i.e.
+it was installed to a custom path), then the ``LD_LIBRARY_PATH`` environment
+variable needs to be set before running, for example:
```
jorisvandenbossche, line 170: This is for discovery at runtime?

shapely/algorithms/meson.build
jorisvandenbossche, line 1: For pure-python-file directories, is it needed to list all the files explicitly? (like done below as well for `shapely/geometry` and the various test directories)

Looking at the pandas meson.build, it seems to do something with `install_subdir` without an explicit list. 
And it seems numpy is also doing something similar: https://github.com/numpy/numpy/blob/140ddeb611d8e30f7e2753558e618b8235b844ae/numpy/meson.build#L365-L371

meson.build
```diff
@@ -0,0 +1,98 @@
...
+
+cy = meson.get_compiler('cython')
+
+# generator() doesn't accept compilers, only found programs - cast it.
+cython = find_program(cy.cmd_array()[0])
+cython_args = ['-3', '--fast-fail', '--output-file', '@OUTPUT@', '--include-dir', '@BUILD_ROOT@', '@INPUT@']
+if cy.version().version_compare('>=3.1.0')
+  cython_args += ['-Xfreethreading_compatible=True']
+endif
+cython_gen = generator(cython, arguments : cython_args, output : '@BASENAME@.c')
```
jorisvandenbossche: Why is this `cython_gen` is needed? Or, why does `py.extension_module(..)` not work without specifying a custom cython generator? (the custom `cython_args` can also be passed to `py.extension_module()`)

pyproject.toml
```diff
@@ -59,22 +59,19 @@ docs = [
...
 skip = ["pp*", "*_i686", "*_ppc64le", "*_s390x"]
 build-verbosity = 1
+before-build = "pip install -r requirements-dev.txt"
+build-frontend = { name = "pip", args = ["--no-build-isolation"] }
 test-requires = "pytest"
 test-command = "pytest --pyargs shapely.tests"
 enable = ["cpython-freethreading"]
 
+[tool.cibuildwheel.windows]
+before-build = "pip install -r requirements-dev.txt delvewheel"
```
jorisvandenbossche: With the latest cibuildwheel, delvewheel is now used by default, and so we don't have to install it explicitly anymore (so windows can also use the `before-build` defined above)

pyproject.toml
```diff
@@ -59,22 +59,19 @@ docs = [
...
+config-settings = { setup-args = ["--vsenv"] }
```
jorisvandenbossche: Is this needed if we have this `Activate MSVC` step in the release workflow?

jorisvandenbossche (commented):
@mwtoews finally gave this a try! 

Seems to work fine locally in my development setup (only needed a small workaround for properly finding cython, but that is a known upstream meson issue)

And did a review pass over the changes, added some small comments and questions.

.github/workflows/release.yml
eli-schwartz: meson-python utilized `meson dist` under the hood and can run any registered `meson.add_dist_script()`, which means it has to be able to successfully complete `meson setup`.

In theory there could be infrastructure to setup for building an sdist with -D running-in-sdist-only=true upon which dependencies aren't searched for. That being said, do people actually build sdists that they then don't attempt to compile and install?

docs/installation.rst
eli-schwartz: Meson-python will attempt to rerun `ninja` to ensure the editable install is up to date.

This is a tradeoff, because if you edit a:
- *.py file, then the setuptools process of "well idk, just assume edits will be picked up automatically on import", does work and you don't need to have build dependencies available, but meson-python requires it
- *.c(pp)? file, then you were always kind of screwed with setuptools, as edits would NOT be picked up unless you reran `pip install -e` but you'd have no idea that that was necessary


The former case only worked because it was tantamount to editing `site-packages/shapely/*.py` in *non* editable mode, so it "made sense" that you don't need to disable build isolation. :D

docs/installation.rst
eli-schwartz: editable-verbose isn't strictly necessary. There is an in-progress PR so that the next version of meson-python will print the ninja output including compiler errors, in the traceback when editable rebuilding fails. This is the main reason why verbose mode is tempting, I think, so maybe you don't actually want to see full compile logs every single time it compiles fine anyway.

shapely/algorithms/meson.build
eli-schwartz: It's a style choice. For setuptools, one could list all packages, or use `packages = find:` (that is the `find_packages()` function but for setup.cfg).

The former gives you control to avoid strange cases like accidentally installing stale or experimental files and means that the contents of the package are always defined by the meson.build / setup.cfg tracked by git, and the latter saves you some typing.

.github/workflows/release.yml
jorisvandenbossche: I don't think we mind (and the added complexity is probably not worth it for just being able to build an sdist as-is), just wondering out loud about the difference

docs/installation.rst
jorisvandenbossche: Personally I have find it helpful in practice to actually see the output when it rebuilds (instead of potentially a few seconds of seemingly "stalling" before doing anything), that's the reason that I would put it here in the documented development workflow. 

But good to hear this is going to improve also without enabling it!

mwtoews:
Thanks for thorough review @jorisvandenbossche and feedback by @eli-schwartz, I'll try find time over the weekend to address the comments while there is momentum.

meson.build
```diff
@@ -0,0 +1,92 @@
...
+    'These previously-supported environment variable(s) are not supported:',
+    ignored_env_vars,
+    '; use `cmake_prefix_path` approach instead')
+endif
+
+# A custom GEOS installation can be provided with configuration option:
+# -Dcmake_prefix_path=/path/to/geos-install
+
+geos_dep = dependency('geos', version : f'>=@min_geos_version@',
+  method : 'cmake',  modules : ['GEOS::geos_c'])
```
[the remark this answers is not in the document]
jorisvandenbossche: From _my_ user perspective (i.e. my specific case of using various conda environments to test different dependencies setups), I would expect meson to always (and automatically, without me having to do something specific) first find dependencies in the conda env, and only then falls back to system ones.

For something like Python, that works that way (i.e. it finds the Python from my conda env, not my system python).
(disregarding the details of why that works for Python for a moment, just describing what I observe as a user)

But for GEOS that does not happen if I remove the `method : 'cmake'` on the above line

meson.build
```diff
@@ -0,0 +1,98 @@
+project(
+  'shapely',
+  'c', 'cython',
+  version: '2.2.0.dev0',
+  license: 'BSD-3',
+  meson_version: '>=1.2.1',
+  default_options: [
+    'buildtype=release',
+    'c_std=c11',
+    'warning_level=2',
```
jorisvandenbossche: A warning level of 2 includes `-Wall -Wextra`, and this cause a flood of "unused parameter" warnings, making the logs very long. Maybe keep it at 1 for now? (which only does `-Wall`, and `unused-parameter` is only included in the extras) Or pass `-Wno-unused-parameter` ?

meson.build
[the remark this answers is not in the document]
eli-schwartz: But this is obviously an inaccurate analysis.

> If method is left to it's default behaviour and specify `-Dcmake_prefix_path=${GEOS_PREFIX}` to a custom installation

> I would expect meson to always (and automatically, without me having to do something specific) first find dependencies in the conda env, and only then falls back to system ones.
> [...]
> But for GEOS that does not happen if I remove the `method : 'cmake'` on the above line

You cannot logically claim that the case of finding GEOS in the conda env constitutes "without me having to do something specific", if you had to do something specific by setting CMAKE_PREFIX_PATH to point to conda.

The fact that pkg-config and cmake behave differently is then solely down to the fact that you did something specific for the latter, but did not do something specific for the former. If you find manually passing -Dcmake_prefix_path to be better for your workflow than passing -Dpkg_config_path, then fair enough, but my point here is simply that they are both specific manual things you've done.

...

Anyways I'm just surprised that conda doesn't automatically initialize a conda shell with the necessary environment variables, I guess.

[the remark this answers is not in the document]
jorisvandenbossche: Hmm, I was testing it with a conda env that had `cmake` installed (because it was the env to test with GEOS main), so that is the reason it found it with the current version forcing cmake. 
If you have a fully "clean" conda env (that has just installed GEOS and shapely's other build dependencies), you end up with an environment that has neither cmake or pkg-config, and it actually also does not find GEOS.

So in practice, to work properly with meson in a conda env, you essentially have to ensure `pkg-config` is always installed in the conda env? (which we should document in that case)

The upside of forcing "cmake" is that we _could_ add cmake to the `[build-system].requires` in pyproject.toml since cmake is available on PyPI, ensuring that a source install would work out of the box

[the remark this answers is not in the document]
jorisvandenbossche: > You cannot logically claim that the case of finding GEOS in the conda env constitutes "without me having to do something specific", if you had to do something specific by setting CMAKE_PREFIX_PATH to point to conda.

To be clear, what you are quoting are the probably separate use cases of Mike and myself (sorry for that confusion by adding a different use case in the same thread). 
Mike wants to be able to build with a custom path to the GEOS installation (as we do in CI, and there of course you have to provide _some_ indication of that path), and I was testing locally with a conda env (where I do expect it to find it since it is located in the standard place in conda's prefix)

docs/installation.rst
mwtoews: Yes, this is platform agnostic, so either `path/to/prefix` or `D:\path\to\prefix`. It is then passed to CMake's [find_package()](https://cmake.org/cmake/help/latest/command/find_package.html) command.

.github/workflows/release.yml
mwtoews: As I've completely forgotten, I've re-run it and see an error:
```
  ..\meson.build:1:0: ERROR: Found GNU link.exe instead of MSVC link.exe in C:\Program Files\Git\usr\bin\link.EXE.
  This link.exe is not a linker.
  You may need to reorder entries to your %PATH% variable to resolve this.
```
so I'll add a comment to help guide what this is for.

.github/workflows/release.yml
```diff
@@ -10,9 +10,7 @@ on:
     paths:
       - ".github/workflows/release.yml"
       - "ci/*"
-      - "MANIFEST.in"
       - "pyproject.toml"
-      - "setup.py"
```
jorisvandenbossche, line 15: Can you add the `meson.build` files here?

jorisvandenbossche:
> However, currently stuck with this somewhat common runtime issue only for Windows editable installs:

@mwtoews about the Windows issue not finding GEOS you described at https://github.com/shapely/shapely/pull/2357#issuecomment-3787756377, and where you listed some possible options. In the meantime, I think you added the `geos_libdir` meson option to address this, which if specified adds a file to the source, and which is then inspected in the `__init__` to call `os.add_dll_directory` if needed.

I think I am fine with that solution, but just wanted to highlight that in rasterio/fiona/pyogrio, something similar is done, but instead of letting the user specify a specific path during the build, we essentially just check if the relevant library (GDAL in their case) is available on the PATH. 
Rasterio (and fiona has the same): https://github.com/rasterio/rasterio/blob/5044f0c8cd08655c63c8b2e452618254c15d8f29/rasterio/__init__.py#L11-L23
Pyogrio: https://github.com/geopandas/pyogrio/blob/88c25164a3c46482a1f4f41651f5a0dc613b4d3e/pyogrio/_env.py#L30-L59 (slight variation on the same pattern to only add the dir temporarily while importing the library)

I don't have a strong preference either way, though (the PATH is of course more broad, and does not only kick-in when you built (and set `geos_libdir`) manually, but I am not really sure if in practice it covers other uses cases than local source builds)

.github/workflows/release.yml
jorisvandenbossche: Hmm, I would have expected the `ilammy/msvc-dev-cmd` action to take care of that .. :shrug:

meson.build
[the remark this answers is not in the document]
eli-schwartz: > So in practice, to work properly with meson in a conda env, you essentially have to ensure `pkg-config` is always installed in the conda env? (which we should document in that case)
> 
> The upside of forcing "cmake" is that we _could_ add cmake to the `[build-system].requires` in pyproject.toml since cmake is available on PyPI, ensuring that a source install would work out of the box

It is recommended to have pkgconfig installed, yeah. If you do have it installed from conda, it will pick up the conda version of GEOS. If pkgconfig is installed from the parent system, it's probably not configured to look in the conda env and the -Dpkg_config_path workaround is needed.

Forcing cmake is a bit awkward, since cmake is both very large and not necessarily needed (and if you build in a venv on a regular Linux system rather than inside conda, you probably already have cmake and cmake is configured to search the same place anyway, but it doesn't come with *.dist-info metadata for the python package manager, so you'll be forced to install a new copy from PyPI and it doesn't work on all CPU architectures). In general it's frustrating that the PyPI ecosystem doesn't have a good way to define "recommended build requirements for a smooth default experience" versus "actual required requirements".

.github/workflows/release.yml
mwtoews: The `BEFORE_ALL_*` steps call `./ci/install_geos.sh` / `ci\install_geos.cmd` which require `GEOS_INSTALL` and `GEOS_VERSION`, so this is needed to build the prerequisites lib.

docs/installation.rst
mwtoews: Yes run-time only; not needed at build-time.

.github/workflows/release.yml
mwtoews: I was looking for ways to skip the dependency check for sdist, but couldn't find any easy wins. I was hoping to find some meson variable to indicate that `meson dist` was running to skip these, but gave up.

pyproject.toml
mwtoews: Thanks for reminding me to check this, and it looks OK without, so is removed.

mwtoews:
A note on my experiment yesterday. I attempted simplifying the NumPy build version to >= 2.0.0 while also allowing older NumPy 1.23 runtime. This attempt first installed NumPy 2 for editable build, then re-installed and older NumPy 1 for runtime. But the "editable" nature did not appreciate that change and doing a simple `import shapely` broke and threw lots of errors. We can keep NumPy>=1.23 for this upcoming release (note that pandas still allows NumPy>=1.26.0), but after I'd like to require NumPy>=2 to simplify the build stuff.

jorisvandenbossche:
But if you do your editable install with `--no-build-isolation`, that should not happen? 
(I don't think we should care about the use case of editable without that flag)

meson.build
mwtoews: I've also passed a few `-Wno-*` options in `meson.build` with a TODO note that it would be nice to eventually resolve these.

.github/workflows/release.yml
mwtoews: Note that this issue / workaround is [specifically noted](https://github.com/ilammy/msvc-dev-cmd#name-conflicts-with-shell-bash) with ilammy/msvc-dev-cmd. It is not mentioned with TheMrMilchmann/setup-msvc-dev, so I'll re-try disabling this workaround to see after #2486 is merged.

jorisvandenbossche: No luck, still the same error about "Found GNU link.exe instead of MSVC link.exe". But passing `--vsenv` to meson on Windows might also address that?

jorisvandenbossche:
> A note on my experiment yesterday. I attempted simplifying the NumPy build version to >= 2.0.0 while also allowing older NumPy 1.23 runtime. This attempt first installed NumPy 2 for editable build, then re-installed and older NumPy 1 for runtime. But the "editable" nature did not appreciate that change and doing a simple `import shapely` broke and threw lots of errors.

To get back to this once more: I think this is not related to building/running with numpy >= 2.0 and <2. I quickly tried and I get exactly the same error when having an environment with eg numpy 2.2 and not doing build isolation (in which case it will build with numpy 2.5 and run with numpy 2.2). 
I think that just boils down to what you noted in the docs: when doing an editable install with meson, you now always need to specify `--no-build-isolation` (to have it work with the auto-rebuilding; in theory with setuptools you don't need the build isolation if you were only touching pure python files since it wouldn't rebuild the extensions)

pyproject.toml
```diff
@@ -1,13 +1,12 @@
...
 requires = [
     "Cython",
-    # Starting with NumPy 1.25, NumPy is (by default) as far back compatible
-    # as oldest-support-numpy was (customizable with a NPY_TARGET_VERSION
-    # define).
-    "numpy>=1.25,<3",
-    "setuptools>=77.0.3",
+    "meson-python>=0.15.0",
+    # Exclude 2.5.0: wheels built against it segfault on older numpy
+    "numpy>=1.23,!=2.5.0",
```
jorisvandenbossche: I would be inclined to leave the numpy specification as it was .. (although I suppose in practice it will also not matter that much).

For a normal install (not editable and with build isolation), building with numpy 1.25 or 2.0 should be fine when running on 1.23. While _in theory_ with `>=1.23` you could get a situation of building with numpy 1.24 and running with 1.23, which will break (again, in practice I can't directly think of an environment where this would happen, since we also don't support such old Python versions where only those numpy versions would be available).
And if you for whatever reason can only install numpy 1.23 (also for the build), then you still have the `--no-build-isolation` escape.

The last time I updated the numpy requirement to `numpy>=1.25,<3` (https://github.com/shapely/shapely/pull/2129), I included the `>1.25` lower bound in the idea that if you can't install numpy 2.0 for some reason, then that will also be the case for the runtime, and then building with 1.25 is still fine. But nowadays the numpy docs indicate `>=2.0` for compatibility with numpy 1.x and 2.x (https://numpy.org/devdocs/dev/depending_on_numpy.html#numpy-2-0-abi-handling). 
(I suppose now we dropped the older Python versions that were not yet supported by numpy 2.x)

.github/workflows/release.yml
mwtoews: It was just an error for `tests.yml` and not for `release.yml` which probably has a different shell environment, i.e. in cibuildwheel. I've posted this issue to TheMrMilchmann/setup-msvc-dev.

shapely/algorithms/meson.build
mwtoews: I implemented the [`install_subdir()`](https://mesonbuild.com/Reference-manual_functions_install_subdir.html) approach with 00337d04acc971151eb26f7fbf20fbd02eafbb8b which also cleans up several `build.meson` files. I think this looks better and will be easier to manage.

meson.build
mwtoews: Thanks for raising this point. It was probably from copying other projects, but looking at [meson's Cython docs](https://mesonbuild.com/Cython.html), it works without needing to make a generator.

docs/installation.rst
mwtoews: I've added some words above here to say that the non-conda approach relies on a local or system installation of GEOS, since it is not available as a package on the PyPI ecosystem.

meson.build
eli-schwartz: IIRC @rgommers used a generator initially in SciPy due to some bugs in very old versions of Meson, but I forget the exact details? It has mostly spread from there.

rgommers: It's still useful in SciPy to use a generator I believe, because we have a fair bit of templated Cython code, and using a generator avoids duplicating the `depends : [_cython_tree, cython_shared_module]` for multiple extension modules. 

If you don't use Tempita or another templating engine that causes `.pyx` files to land in the build directory, then yes I think a generator is only extra boilerplate.

shapely/__init__.py
```diff
@@ -1,5 +1,30 @@
...
+    pth = importlib.resources.files("shapely").joinpath("geos_libdir.txt")
+    try:
+        geos_libdir = pth.read_text().strip()
+    except FileNotFoundError:
+        return
+    os.add_dll_directory(geos_libdir)
+
+
+_set_geos_libdir()
+del _set_geos_libdir
```
sgillies: @mwtoews I don't love this kind of file opening and reading during module loading.

What's the advantage over a good old-fashioned environment variable?

mwtoews: This is my least favourite part of this PR too. I'll explore some potential lightweight options. And to re-emphasize, this is required only for Windows.

rgommers: This is an annoying wart to deal with indeed. And the docs at https://mesonbuild.com/meson-python/how-to-guides/shared-libraries.html#external-shared-libraries don't quite cover editable installs.

I see in CI you use a `GEOS_INSTALL` env var; `-Csetup-args="-Dgeos_libdir` is similar and cleaner - but both could perhaps be used to dynamically and this `os.add_dll_directory` snippet, so it does the right thing but isn't part of regular builds/wheels, only editable ones?

We had a fair amount of discussion at https://github.com/mesonbuild/meson-python/issues/256, but only one real use case to know at configure time whether a build is editable or not. It looks like this is a second one.

The reason we don't have a problem with this in numpy/scipy when linking a BLAS library on Windows is that we use [spin](https://github.com/scientific-python/spin) rather than a bare editable install from `pip`/`uv`. It takes care of this, writing out a shim dynamically at build time: https://github.com/numpy/numpy/blob/82d8dc995535bbfc34b4f091b18490900f099777/.spin/cmds.py#L622-L633

jorisvandenbossche: FWIW, I commented about this a few weeks ago in the main thread: https://github.com/shapely/shapely/pull/2357#issuecomment-5142840411. And so what we currently do in similar other projects (rasterio, fiona, pyogrio), is checking for the library on the PATH.

That avoids opening and reading a file during module load, but not sure if that is necessarily a cleaner solution (it still globs several directories to see if a file is present). Unless we would not use PATH, but require the user to set a specific env variable.

BTW, I am not sure if this is limited to editable installs? If you build locally in a non-editable way with a pre-installed GEOS somewhere on your Windows system, out extension modules still need to find that library?

rgommers: Yes you're right - the difference is local build vs. distributable wheels (or conda packages, or ...) where you run `delvewheel` or its distro equivalent.

mwtoews: I found some time on a Windows computer to hopefully work out something much simpler. The brave few Windows developers that are working with editable builds will need to do (e.g.) `set GEOS_LIBDIR=C:\OSGeo4W\bin` or wherever their GEOS DLLs are located at runtime.
The binary wheels, used by the majority of Windows users, don't need to do anything, because delvewheel inserts a patch to `__init__.py` to load the DLLs locally from `shapely.libs`.

mwtoews:
This is _finally_ ready for review. Thanks to the meson[-python] devs for helping several aspects of this PR!

Some follow-up work would be to (probably) replace versioneer with setuptools-scm or other modern git-based versioning tool. And perhaps improving the install docs to be more coherent.

pyproject.toml
```diff
@@ -98,6 +88,10 @@ ignore = ["D000", "D004"]
 allow-long-titles = true
 max-line-length = 79
 
+[tool.meson-python.args]
+# setup = ["--vsenv"]  # ERROR: Tried to modify read only option "vsenv"
+# setup = ["--clearcache"]  # uncomment to clear build cache with local builds
```
sgillies, line 90: @mwtoews both of these options are opaque to me. The first looks like a feature that doesn't work? The second could maybe be deleted too? When would one want to use it?

I'm pretty surprised that meson-python isn't driven more by standard options in this section, and instead has a lot of Python code in various meson.build modules.

sgillies (commented):
@mwtoews I checked out your branch, tried the install instructions, and everything worked as advertised. Thank you very much for the detailed instructions.

I trust you and @jorisvandenbossche on the potential benefits of using meson-python, but I feel let down by it in some ways.

1. When running `pip install -e .` in shapely, with no GEOS specified, the new build backend didn't provide a significantly better experience than what we had before. Our new backend didn't fetch and build GEOS, it just failed with a "GEOS not found" message. Is resolving the GEOS dependency something that meson can do? Or is it out of scope?

2. I expected meson-python to be more mature, with more declarative options covering our use cases. Instead, we have a traded a single `setup.py` module for multiple `meson.build` modules that include custom Python logic, again, just written differently.

Can either of these be addressed, or will we have to work around them?

eli-schwartz:
> I trust you and @jorisvandenbossche on the potential benefits of using meson-python, but I feel let down by it in some ways.
> 
>  1. When running `pip install -e .` in shapely, with no GEOS specified, the new build backend didn't provide a significantly better experience than what we had before. Our new backend didn't fetch and build GEOS, it just failed with a "GEOS not found" message. Is resolving the GEOS dependency something that meson can do? Or is it out of scope?

Meson can definitely do this! See the documentation on subprojects: https://mesonbuild.com/Subprojects.html

tl;dr for upstream projects using meson or cmake, you can essentially plop a single INI file in subprojects/*.wrap, containing the tarball download url, checksum, and dependency lookup names. (CMake is less well supported than meson due to the nature of cross buildsystem tooling.)

GEOS upstream uses cmake. It should probably work okay but double check.

Subprojects are automatically downloaded and used, but only if the system dependency isn't found. No actual modifications to the build files are needed to enable a subproject, it is automatic as long as `dependency()` is used instead of `cc.find_library()`.

> 2. I expected meson-python to be more mature, with more declarative options covering our use cases. Instead, we have a traded a single `setup.py` module for multiple `meson.build` modules

Meson doesn't require using multiple `meson.build` files. A somewhat popular third-party setuptools extension enables using multiple `setup.py` files, and meson simply supports this built-in. Same way cmake supports multiple CMakeLists.txt and autotools supports multiple Makefile.am or extending configure.ac via *.m4.

People often choose to use multiple files regardless of build system if they personally feel it aids in clarity, but if you wish to, you could use a single root meson.build file.

>  that include a custom Python logic, written differently.

The custom python logic exists for two reasons:
- the PR author wishes to provide a helpful warning to users that certain environment variables from setup.py are no longer supported.
- Old versions of numpy (less than 2.x) don't support the new upstream numpy dependency detection tooling that numpy added specially for modern build systems, therefore detection requires using custom python code as originally designed for numpy.distutils (which is even more custom python code).

Both of these can be dropped, if desirable. That's up to you all, I guess. :)

(The support for old numpy is surely going to be eventually dropped regardless.)

pyproject.toml
eli-schwartz: --vsenv is a `meson setup` option that forces meson to search for Microsoft `cl.exe` even if there is already a mingw gcc.exe or clang.exe in `%PATH%`. It is essentially choosing to force a toolchain decision. No comment on whether or not it is a good idea.

The read-only option thing is a meson bug that I'm pretty sure was fixed a while ago?

The clearcache thing looks strange to me. This is an option you can pass to `meson setup` if you're doing incremental builds but you think you've accidentally wedged your build by detecting the wrong installation of your dependencies, and you'd like meson to stop caching that lookup in your incremental build and slowly go search for it all over again (which may require running a bunch of `pkg-config` subprocesses or writing out a CMakeLists.txt file and running cmake to see if cmake find_package() can detect it, and depending on the specific third-party dependencies you have, possibly also running qmake or `ncurses-config` or various other things; even geos-config could be supported, but that will require updating Meson's source code to make `dependency('geos')` automatically try it for a new version of meson. If shapely would very much like this, I'm open to adding it to meson upstream!).

***You don't need clearcache*** unless you have an old build and you want to delete incremental build caches! It feels very odd to set this in pyproject.toml and it's not obvious that leaving it commented out for users to uncomment it will be any better.

rgommers:
Let me confirm that the NumPy handling will reduce to a one-liner:
```meson
numpy_dep = dependency('numpy')
```
or, if you want to enforce a version check at build time (as is done in this PR now):
```meson
numpy_dep = dependency('numpy', version : f'>=@min_numpy_build_version@')
```

This required NumPy 2.0 shipping an entrypoint for a `numpy-config` utility, which is a nice way of doing it - we just didn't ever think of that in the 1.x era because downstream packages were also using distutils/setuptools and (un)happily executing Python code.

***

Re GEOS: +1 to what @eli-schwartz said. In addition, I'll confirm that use of external shared libraries is a first-class citizen for `meson-python`: we now have [good docs on it](https://mesonbuild.com/meson-python/how-to-guides/shared-libraries.html) and I'd like to add more support to make this easier in the future (there are still usability issues around RPATH handling) - this thread gave me another nudge to work on that.

pyproject.toml
mwtoews: The commented line with `--vsenv` can be removed, as I'd prefer to have the environment activated with the correct toolset to reliably compile GEOS beforehand (as done in CI).

The second commented line with `--clearcache` I wouldn't mind keeping for a bit longer, as I've used this in a few scenarios. Or it can be moved to a less obvious comment to use `pip install ... -Csetup-args="--clearcache"`. I'll follow-up with this as a possible issue upstream when time permits.

meson.build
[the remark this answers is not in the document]
mwtoews: I've implemented another strategy for handling "method" in `dependency('geos')`:

1. If a user specifies `-Dcmake_prefix_path=` then this method would try "cmake".
2. And if the user specifies `-Dpkg_config_path=` then the method would try "pkg-config".
3. And if neither options are specified (or dependency not found with the previous methods), then use whatever default method that meson uses.

It takes a few lines of logic, but it should honour the method if the setup built-in option is specified, which is important when I have multiple installations of GEOS from the system or in a custom prefix.

meson.build
```diff
@@ -0,0 +1,95 @@
...
+ignored_env_vars = run_command(py, ['-c', '''
+from os import environ
+for name in ['GEOS_INCLUDE_PATH', 'GEOS_LIBRARY_PATH', 'GEOS_CONFIG']:
+    if name in environ:
+        print(name + "=" + environ[name])
+'''], check : true).stdout().strip().split()
+if ignored_env_vars != []
+  warning(
+    'These previously-supported environment variable(s) are not supported:',
+    ignored_env_vars, '; use `cmake_prefix_path` approach instead')
```
jorisvandenbossche:
```suggestion
    ignored_env_vars, '; use `cmake_prefix_path` or 'pkg_config_path' approach instead')
```

meson.build
```diff
@@ -0,0 +1,95 @@
...
+endif
+
+# GEOS dependency method priority if built-in option is specified
+min_geos_version = '3.10'
+geos_dep = dependency('', required: false)
```
jorisvandenbossche: Why this `dependency('', ..)`? What does it do exactly? Wouldn't this just always fail, since you are not actually specifying a dependency?

meson.build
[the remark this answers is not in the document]
jorisvandenbossche: Thanks for that update! This ensures it works out of the box when specifying one of those variables, but my understanding is that if you are testing with conda/pixi envs (in which case you should not need to pass any of those flags), and you have a system install of GEOS, it will typically still pick up that system install instead of the conda one, _unless_ you have `pkg-config` installed in the conda env (assuming pkg-config is installed on the system, which is the case for me, and in that case `meson` will give priority to pkg-config, even if cmake would find the local conda installation). 

Maybe we should add that to the documentation to mention it is recommended to include `pkg-config` in your conda env?

meson.build
jorisvandenbossche: Ah, I assume is is just a dummy to create the `geos_dep` variable with the correct type for the repeated `if not geos_dep.found()` checks. 

Maybe add something like this?
```suggestion
geos_dep = dependency('', required: false)  # dummy to have dependency variable
```

docs/environment.yml
```diff
@@ -11,3 +11,6 @@ dependencies:
   - numpydoc=1.1
   - matplotlib
   - pip
+  - cmake
```
jorisvandenbossche:
```suggestion
  - pkg-config
```

Shouldn't actually matter, but since that is what meson will try first before cmake, could also use that

docs/installation.rst
```diff
@@ -55,87 +55,131 @@ On macOS::
     $ pip install shapely --no-binary shapely
 
 If you've installed GEOS to a standard location on Linux or macOS, the
-installation will automatically find it using ``geos-config``. See the notes
-below on GEOS discovery at compile time to configure this.
+installation will automatically find it using CMake. If GEOS is installed
```
jorisvandenbossche:
```suggestion
installation will automatically find it using pkg-config or CMake. If GEOS is installed
```

docs/installation.rst
```diff
@@ -55,87 +55,131 @@ On macOS::
...
 
-Or using ``conda``:
+Install shapely in editable mode using ``pip``:
 
 .. code-block:: console
 
-    $ conda create -n env python=3 geos numpy cython pytest
-    $ conda activate env
-    (env) $ pip install -e .
+    (env) $ pip install --no-build-isolation -e .
```
jorisvandenbossche, line 106: I think I commented about that before, but I would personally show people to use the verbose mode:


```suggestion
    (env) $ pip install --no-build-isolation -e . -Ceditable-verbose=true
```

Or at least mention it in the text. If you are not already using meson, this is otherwise not obvious

docs/installation.rst
```diff
@@ -55,87 +55,131 @@ On macOS::
...
+Conda environments
+^^^^^^^^^^^^^^^^^^
+
+Development within conda environments requires additional packages installed
+with the ``conda`` command, including pre-built distributions of GEOS.
+For example:
+
+.. code-block:: console
+
+    $ conda create -n shapely-dev cmake cython meson meson-python ninja numpy geos pip pytest
```
jorisvandenbossche:
```suggestion
    $ conda create -n shapely-dev cython meson meson-python pkg-config ninja numpy geos pip pytest
```

For the reason mentioned in one of the previous comments, I think it is a bit safer to recommend installing pkg-config, since whether we like it or not, meson will use that first to search for GEOS (and thus if you have system installs of pkg-config and geos, and pkg-config is not in the conda env, AFAIK it will pick up the wrong one)

docs/installation.rst
```diff
@@ -55,87 +55,131 @@ On macOS::
...
+
+Activate the environment and install shapely in editable mode using ``pip``:
+.. code-block:: console
+
+    $ conda activate shapely-dev
+    (shapely-dev) $ pip install --no-build-isolation -e .
```
jorisvandenbossche:
```suggestion
    (shapely-dev) $ pip install --no-build-isolation -e . -Ceditable-verbose=true
```

shapely/__init__.py
```diff
@@ -1,5 +1,22 @@
 """Manipulation and analysis of geometric objects in the Cartesian plane."""
 
+import os
+import platform
+
+if platform.system() == "Windows":
+    try:
+        from shapely.lib import GEOSException
+    except ImportError:  # DLL load failed while importing lib...
+        # This is possibly an editable build, which requires an
```
jorisvandenbossche:
```suggestion
        # This is possibly a local build, which requires an
```

It's not just for editable builds, any local build from source that uses a custom GEOS install will need this

pyproject.toml
jorisvandenbossche: (this comment still stands, although I assume that in practice almost no-one will run into this, so maybe I should not care ;))

jorisvandenbossche (commented):
Thanks for the updates @mwtoews, did another round of review (mostly just small nitpicks).

jorisvandenbossche:
> I trust you and @jorisvandenbossche on the potential benefits of using meson-python, but I feel let down by it in some ways.

I can assert that it will be give a Quality-of-Life improvement for developing locally ;) 
(but for sure, every build system has it quirks, so if you are used to one, it is always a bit adapting (your habits/understanding) to another; it also took we some time to get accustomed to meson)

Let's leave automated fetching of GEOS as a subproject for a potential follow-up discussion. While it can be nice that installing shapely from source "just works", that probably comes with its own set of caveats and considerations that we should think about.

docs/installation.rst
mwtoews: Sure, mentioned it below in text. I find that pip's `-v` is handy enough to see the initial compilation.

pyproject.toml
mwtoews: This is mostly to please the CI, which has two "numpy: 1.23.5" items. I'll adjust the CI and restore most of the previous comment.

mwtoews:
Apologies @sgillies for delayed reply...

> 1. When running `pip install -e .` in shapely, with no GEOS specified, the new build backend didn't provide a significantly better experience than what we had before. Our new backend didn't fetch and build GEOS, it just failed with a "GEOS not found" message. Is resolving the GEOS dependency something that meson can do? Or is it out of scope?

I've done some changes to allow `pkg-config` to help with the dependency search. It all depends on how/where GEOS was installed. If in a predicable system path, just do:
```
pip install --no-build-isolation -e .
```

If this isn't working, toss any cached stuff with `rm -rf build` and start again. If GEOS is in a custom place, I've put a few examples how to discover this with cmake or pkg-config methods.

> 2. I expected meson-python to be more mature, with more declarative options covering our use cases. Instead, we have a traded a single `setup.py` module for multiple `meson.build` modules that include custom Python logic, again, just written differently.

I could wrap all of the `meson.build` logic into one file, but I'd rather have three, as it's usually common to have one of these in directories with code that gets compiled. This seems to be the norm with other projects.

mwtoews:
> Thanks for the updates @mwtoews, did another round of review (mostly just small nitpicks).

Thanks @jorisvandenbossche for looking at the details, minor or otherwise. I've restyled the meson.build files one last time with a few more comments. I don't think there is too much more to address at this stage.

shapely/__init__.py
```diff
@@ -1,5 +1,22 @@
 """Manipulation and analysis of geometric objects in the Cartesian plane."""
 
+import os
+import platform
+
+if platform.system() == "Windows":
+    try:
+        from shapely.lib import GEOSException
+    except ImportError:  # DLL load failed while importing lib...
+        # This is possibly a local` build, which requires an
```
jorisvandenbossche:
```suggestion
        # This is possibly a local build, which requires an
```

docs/installation.rst
jorisvandenbossche: > I find that pip's `-v` is handy enough to see the initial compilation.

Yep, but as you say, that is only the initial compilation. I have found it nice to also see the compilation happening when importing shapely (or running tests) after an automatic rebuild, instead of having this seemingly hang for a while with nothing happening.

shapely/__init__.py
jorisvandenbossche:
```suggestion
        # This is possibly a local build, which requires an
```

docs/installation.rst
```diff
@@ -55,87 +55,139 @@ On macOS::
...
-    $ conda activate env
-    (env) $ pip install -e .
+    (env) $ pip install --no-build-isolation -e .
+
+This approach relies on a local or system installation of GEOS, since it is
+not available as a package on the PyPI ecosystem (unlike conda).
+
+To see compilation details, add ``-v`` (or ``--verbose``) to ``pip install``.
+Further inspection of compilation details can be enabled by also adding
+``-Ceditable-verbose=true`` to the command.
```
jorisvandenbossche:
```suggestion
To also see compilation details when an automatic rebuild of the extension modules happens in editable mode, add
``-Ceditable-verbose=true`` to the command.
```

(or something like that. I think the current sentence is not super clear what this "further inspection" is about if you are not familiar with meson-pythons automatic rebuilding, since this is new compared to setuptools)

jorisvandenbossche (approved):

docs/installation.rst
mwtoews: Yes, this is better wording for what this particular feature does, thanks.

mwtoews:
Thanks everyone in this loooong thread for feedback and comments! Any new issues related to this change can be opened in a new issue, and we'll help you out with the transition.

sgillies:
Thanks for leading this @mwtoews !

I don't like the embedded Python code in the build files at all, but it all has the potential to go away over time when we delete the environment sniffing, stop supporting numpy 1, and can rely on the build front end to enforce the Python version requirement. I can live with it for now 👍

PR #2359, opened 2025-12-08
perf: Improve performance of scalar usage of O -> b functions

caspervdw:
This extends the work started in https://github.com/shapely/shapely/pull/2339 for the functions that take any object and return a bool.

**Notes**

Although these functions are not used in their scalar versions, I refactored them now because they can be used for the scalar calls of (for example) the `YY_Y` functions.

Because when looking at a method like `Geometry.union`, it depends if the argument is a scalar on an array whether we should call into the numpy ufunc. Something like this should be done:

```
def union(self, other):
    if isinstance(other, Geometry) or other is None):
        return lib.union_scalar(self, other)
    else:
        return lib.union(self other)
```

Now after this PR, we can use the `lib.is_valid_input_scalar` to do this check. This is slightly faster (50 ns instead of 80 ns).

I did some benchmarks to compare the current situation (1) with pre-emptive checking (2) or handling typeerrors (3). Looking at the timings, I think option (2) is best to implement in the `Geometry` methods that accept an argument that may be scalar or array (such as `.union`). It will speed up scalar calls by about 800 ns, at the expense of adding 50 of latency to array-based calls.

It is up to debate to put this check in the function `union` or in the `Geometry` method `union`. Putting it in the function will impact scalar usage of that function, however, the API will change as well because return values will be actual Python scalars instead of 0-dimensinonal numpy arrays. I left that out of this PR.

NB: I did the benchmarks with `is_prepared` methods, to reduce variability of the GEOS part, but the measured overhead is also valid for all other functions.

```
import shapely

# current implementation
def is_prepared_1(geometry, **kwargs):
    return shapely.lib.is_prepared(geometry, **kwargs)

# pre-emptive check
def is_prepared_2(geometry, **kwargs):
    if not kwargs and shapely.lib.is_valid_input_scalar(geometry):
        return shapely.lib.is_prepared_scalar(geometry)
    return shapely.lib.is_prepared(geometry, **kwargs)

# handling TypeError
def is_prepared_3(geometry, **kwargs):
    if not kwargs:
        try:
            return shapely.lib.is_prepared_scalar(geometry)
        except TypeError:
            pass
    return shapely.lib.is_prepared(geometry, **kwargs)

In [5]: %timeit is_prepared_1(None)
940 ns ± 3.03 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [6]: %timeit is_prepared_2(None)
131 ns ± 0.272 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [7]: %timeit is_prepared_3(None)
106 ns ± 0.226 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [12]: x = np.array([None])

In [13]: %timeit is_prepared_1(x)
675 ns ± 2.36 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [14]: %timeit is_prepared_2(x)
719 ns ± 2.79 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [15]: %timeit is_prepared_3(x)
980 ns ± 7.48 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

### DIFF
--- a/shapely/creation.py
+++ b/shapely/creation.py
@@ -748,6 +748,8 @@ def prepare(geometry, **kwargs):
     prepared geometries.
 
     This function returns True where geometries were prepared, and False otherwise.
+    When it returns False, this does not mean that the function failed; it merely
+    indicates that the geometry was already prepared.
 
     Parameters
     ----------


jorisvandenbossche:
> It is up to debate to put this check in the function `union` or in the `Geometry` method `union`. Putting it in the function will impact scalar usage of that function, however, the API will change as well because return values will be actual Python scalars instead of 0-dimensinonal numpy arrays. I left that out of this PR.

Yeah, for that reason my preference would be to keep that out of the functions, and only do it in the class methods/attributes. I think it is clearer and more predictable that the function is as close as possible to a ufunc (although we already do wrap the actual ufunc), and so that it doesn't matter if you call it on a scalar or array.  
(of course that does mean that when you are working with scalars, it would actually be advisable to use the class methods and not the functions) 

But in any case, such a change is indeed best to discuss in a separate issue/PR, since it is not just a refactor or perf boost.

src/geos_funcs_O_b.c
```diff
@@ -0,0 +1,288 @@
...
+ * ========================================================================
+ *
+ * Function signature for operations that take a Python object and return a bool: O->b.
+ * These functions do not raise errors on non-geometry objects (unlike Y_b functions).
+ *
+ * Parameters:
+ *   obj: Input Python object (may be a geometry or any other type)
+ *
+ * Returns:
+ *   0 for false, 1 for true
```
jorisvandenbossche: Not entirely correct? It returns an error code, and return false/true to the return pointer (the GEOS functions it wraps do that, though)

jorisvandenbossche: I suppose the difference here is that there is no "core operation logic" section that typically used this pattern?

src/geos_funcs_O_b.c
```diff
@@ -0,0 +1,288 @@
...
+ *   data: User data passed from ufunc creation (contains function pointer)
+ */
+static void O_b_func(char** args, const npy_intp* dimensions, const npy_intp* steps, void* data) {
+  // Extract the specific function from the user data
+  FuncO_b* func = (FuncO_b*)data;
+
+  // Initialize GEOS context with thread support (releases Python GIL)
+  GEOS_INIT_THREADS;
+
+  (void)ctx;   // dims warning;
```
jorisvandenbossche: What is this exactly?

caspervdw:
Thanks for the review @jorisvandenbossche . I iterated on your remarks and refined the code a bit myself too.

src/geos_funcs_O_b.c
```diff
@@ -0,0 +1,288 @@
...
+    *result = 0;  // Valid input (None), but not prepared
+  } else if (((GeometryObject*)obj)->ptr_prepared == NULL) {
+    *result = 0;
+  } else {
+    *result = 1;
+  }
+  return PGERR_SUCCESS;
+}
+
+static char PrepareGeometry(GEOSContextHandle_t context, PyObject* obj, char* result) {
```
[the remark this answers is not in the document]
caspervdw: Yes this gives some extra information to the user, while it doesn't cost us anything (it is very cheap to return a bool; I benchmarked it and saw no difference). At the same time it removes some code duplication.

I added it to the docstring, tests and changelog.

src/geos_funcs_O_b.c
```diff
@@ -0,0 +1,288 @@
...
+ *
+ * Parameters:
+ *   obj: Input Python object (may be a geometry or any other type)
+ *
+ * Returns:
+ *   0 for false, 1 for true
+ */
+typedef char FuncO_b(GEOSContextHandle_t context, PyObject* obj, char* result);
+
+static char IsMissing(GEOSContextHandle_t context, PyObject* obj, char* result) {
```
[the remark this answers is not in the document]
caspervdw: No they don't, I just wanted to reuse the code. I documented this a bit more clearly now.

shapely/creation.py
```diff
@@ -747,7 +747,7 @@ def prepare(geometry, **kwargs):
     it is efficient to call this function on an array that partially contains
     prepared geometries.
 
-    This function does not return any values; geometries are modified in place.
+    This function returns True where geometries were prepared, and False otherwise.
```
jorisvandenbossche: Should we stress that the function always succeeds regardless of the return value (unless it would raise an error)?

PR #2362, opened 2025-12-09
Generalize svg style mechanism

hillwithsmallfields:
I needed more control of the SVG attributes (colors, stroke thicknesses etc) than shapely provides, but it was perfect for the project apart from that... so I've made a branch in which any attributes supported by SVG may be given as keyword arguments to shapely objects' `svg()` methods.  The attributes are passed through, with underscores in the Python keyword argument names being converted to dashes.

I think I've maintained compatibility with the old names, and the old default values.

(I thought I had squashed it into a single commit, but the version that's got through to here has got it as several separate stages of work.)


sgillies:
@hillwithsmallfields we'll need a cleaner PR. That said, I think expanding the style options will be widely welcomed. I've often wished there was a way to set a global style in a notebook, but never had the time to do anything about that.

hillwithsmallfields:
Thanks --- I'll tidy it up sometime in the next few days, and look at how to incorporate a global style.

PR #2363, opened 2025-12-11
BLD: use manylinux_2_28 for x86_64 and aarch64

mwtoews:
From the [NumPy 2.3.0 Release Notes](https://numpy.org/doc/stable/release/2.3.0-notes.html):
> Note that we have also upgraded from manylinux2014 to manylinux_2_28

This change should be also reflected in this repo. Otherwise it seems that numpy is built from source by [release.yml](https://github.com/shapely/shapely/blob/main/.github/workflows/release.yml) which adds a few extra minutes for the CI.

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -3,13 +3,18 @@ ARG GEOS_VERSION=3.13.1
 
 FROM $BASEIMAGE
 ARG GEOS_VERSION
+# Workaround -- https://github.com/pypa/manylinux/issues/1886
+RUN if [ "${AUDITWHEEL_POLICY}" == "manylinux_2_28" ] && \
+            [ -f "/etc/ld.so.conf.d/00-manylinux.conf" ] && \
+            [ -d "/usr/local/lib64" ]; then \
+        echo "/usr/local/lib64" >> /etc/ld.so.conf.d/00-manylinux.conf && \
+        ldconfig; \
+    fi
 ENV GEOS_VERSION=${GEOS_VERSION}
 RUN curl -LOSs --retry 5 https://download.osgeo.org/geos/geos-$GEOS_VERSION.tar.bz2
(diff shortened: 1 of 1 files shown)


mwtoews:
This change of images shares [this unresolved issue with auditwheel](https://github.com/pypa/auditwheel/issues/531). My workaround is to conditionally install GEOS twice. Once to the standard /usr/local base (which is used for headers, but is not on LD_LIBRARY_PATH) and another to DEVTOOLSET_ROOTPATH (if present) since it is on the default LD_LIBRARY_PATH, so it can be picked-up by auditwheel.

jorisvandenbossche:
> Otherwise it seems that numpy is built from source by [release.yml](https://github.com/shapely/shapely/blob/main/.github/workflows/release.yml) which adds a few extra minutes for the CI.

FWIW, that in itself could be solved by installing a slightly older version of numpy (we don't have to build with the latest numpy version, in the past we always built with an older version), eg automatically by requiring binary-only for numpy

Now, if numpy moves to only release such wheels with newer manylinux base, I think we can certainly consider doing that as well

mwtoews:
Rather than manipulating the version, you can also provide [`--prefer-binary`](https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-prefer-binary) for pip. But this only works for some of the Python versions, i.e. image quay.io/pypa/manylinux2014_x86_64:2025.08.02-1 with cp314 will always build from source (with numpy-2.3.5). Whereas with the same image it is happy to install numpy-2.2.6-cp313-cp313-manylinux_2_17_x86_64.manylinux2014_x86_64.whl. I've also looked at `PIP_PREFER_BINARY`, but this doesn't seem to always work. So a better solution is to follow the images that NumPy are using.

Re: the auditwheel issue with LD_LIBRARY_PATH, I've posted this issue upstream with pypa/manylinux. I'd like to see if there is resolution there before merging here.

PR #2364, opened 2025-12-11
CI: replace macos-13 with macos-latest; use macos-15-intel for x86_64

mwtoews:
Recently [`macos-13` images were shut down](https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/), so this update prefers `macos-latest`, which are currently an alias for `macos-15` which is arm64-based. I've added `macos-15-intel` for x86_64 testing.

I've also changed `windows-2022` to `windows-latest`, which is currently an alias for `windows-2025`.

See also https://github.com/actions/runner-images

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -56,7 +56,10 @@ jobs:
             python: "pypy3.11"
             geos: 3.12.3  # keep distinct from free threaded geos version
             numpy: 2.3.2
-
+        exclude:
+          # skip floating-point exceptions from GEOS that are now resolved
+          - os: macos-latest
+            geos: 3.13.1
     env:
       GEOS_VERSION: ${{ matrix.geos }}
       GEOS_VERSION_SPEC: ${{ matrix.geos }}


.github/workflows/tests.yml
```diff
@@ -38,31 +37,20 @@ jobs:
           # 2025
           - python: "3.14"
             geos: 3.14.1
-            numpy: 2.3.3
+            numpy: 2.3.5
           # free threaded Python (no numpy version to indicate installing nightly cython and numpy)
-          - os: ubuntu-22.04
+          - os: ubuntu-latest
```
jorisvandenbossche: I think we used a different os entry here than the default in the matrix, because otherwise that overrides the other one? (for the pypy entry below there is an explicit comment to that effect)

mwtoews:
<s>The change to add macos-15-intel</s> Modern macos-latest (arm64) raises floating-point exceptions:

- RuntimeWarning: divide by zero encountered in oriented_envelope
- RuntimeWarning: invalid value encountered in minimum_width

I'm pretty sure these are tracked upstream in GEOS, but are difficult to resolve.

Update: looking at results more closely, it's just macos-latest with GEOS 3.13.1 that raises these issues. That's a release where those FPE were raised. I'll seek to exclude this entry...

mwtoews:
Mysteriously, macos-latest only has FPE-related failures with GEOS main and the now-excluded GEOS 3.13.1, but not with GEOS 3.14.1 (which has a few FPE warnings). This can be followed-up later.

The three failures with the main branch of GEOS are expected due to upstream changes, and can be resolved with a follow-up. They are all:
```
_____________ test_remove_repeated_points_invalid_result[geom0-2] ______________

geom = <POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0))>, tolerance = 2

    @pytest.mark.skipif(shapely.geos_version < (3, 12, 0), reason="GEOS < 3.12")
    @pytest.mark.parametrize(
        "geom, tolerance", [[Polygon([(0, 0), (1, 0), (1, 1), (0, 1), (0, 0)]), 2]]
    )
    def test_remove_repeated_points_invalid_result(geom, tolerance):
        # Requiring GEOS 3.12 instead of 3.11
        # (GEOS 3.11 had a bug causing this to intermittently not fail)
>       with pytest.raises(shapely.GEOSException, match="Invalid number of points"):
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       Failed: DID NOT RAISE <class 'shapely.errors.GEOSException'>

shapely/tests/test_constructive.py:453: Failed
```

Issue #2358, opened 2025-12-04
Error calling unary_union with very niche inputs


MattTheCuber:
## Problem

Running the minimal example code below throws this `GEOSException`. I have been working with @mikedh on an issue in https://github.com/mikedh/trimesh/issues/2488 and it has led us here. Although, we were able to resolve the problem with a workaround, I thought it would be beneficial to share the bug report here.

`GEOSException: TopologyException: Ring edge missing at -1.5732562941774 7.904434391775184`

<details><summary>Stack trace</summary>

```
---------------------------------------------------------------------------
GEOSException                             Traceback (most recent call last)
...

File .venv/lib/python3.13/site-packages/shapely/ops.py:117, in CollectionOperator.unary_union(self, geoms)
    111 def unary_union(self, geoms):
    112     """Return the union of a sequence of geometries.
    113 
    114     Usually used to convert a collection into the smallest set of polygons
    115     that cover the same area.
    116     """
--> 117     return shapely.union_all(geoms, axis=None)

File .venv/lib/python3.13/site-packages/shapely/decorators.py:173, in deprecate_positional.<locals>.decorator.<locals>.wrapper(*args, **kwargs)
    171 @wraps(func)
    172 def wrapper(*args, **kwargs):
--> 173     result = func(*args, **kwargs)
    175     n = len(args)
    176     if n > warn_from:

File .venv/lib/python3.13/site-packages/shapely/decorators.py:88, in multithreading_enabled.<locals>.wrapped(*args, **kwargs)
     86     for arr in array_args:
     87         arr.flags.writeable = False
---> 88     return func(*args, **kwargs)
     89 finally:
     90     for arr, old_flag in zip(array_args, old_flags):

File .venv/lib/python3.13/site-packages/shapely/set_operations.py:553, in union_all(geometries, grid_size, axis, **kwargs)
    549         raise ValueError("grid_size parameter only accepts scalar values")
    551     return lib.unary_union_prec(collections, grid_size, **kwargs)
--> 553 return lib.unary_union(collections, **kwargs)

GEOSException: TopologyException: Ring edge missing at -1.5732562941774 7.904434391775184
```

</details>

```python
import numpy as np
from shapely import ops
from shapely.geometry import Polygon

polygons = np.array(
    [
        [
            [-9.107924729871753, -49.578770022602505],
            [-11.477396284004609, -55.48211427005445],
            [-8.779906989660152, -68.10055885244654],
            [-9.107924729871753, -49.578770022602505],
[...]

PR #2365, opened 2025-12-11
TST: add test on failing union_all

theroggy:
(Temporary) PR to run a failing union_all on windows, linux and macos. It confirms that it only fails on linux.

Reference:
- #2358
- https://github.com/libgeos/geos/issues/1338

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -56,6 +56,13 @@ jobs:
             python: "pypy3.11"
             geos: 3.12.3  # keep distinct from free threaded geos version
             numpy: 2.3.2
+          - os: ubuntu-24.04-arm  # linux-arm
+            python: "3.13"
+            geos: 3.14.1
+          - os: windows-11-arm  # windows-arm
+            python: "3.13"
+            geos: 3.14.1
+
         exclude:
           # skip floating-point exceptions from GEOS that are now resolved
           - os: macos-latest


pramsey:
Is the failure associated with architecture (intel vs arm) or OS?

theroggy:
> Is the failure associated with architecture (intel vs arm) or OS?

@pramsey It seems to be the OS.

For information an overview of runners used in the tests with their architecture + os ([reference](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories)):

| runner                    | architecture | os          | result         |
| ---------------         | ------------- | ----        | ------------ |
| ubuntu-latest         | intel X64      | linux       | fails          |
| ubuntu-24.04-arm | arm              | linux       | fails          |
| windows-latest      | intel X64      | windows | OK           |
| windows-11-arm   | arm              | windows | build fails |
| macos-15-intel      | intel X64      | macos     | OK           |
| macos-latest         | arm              | macos      | OK           |

The windows and macos runner with GEOS main fail as well, but that's on other tests, so unrelated.

jorisvandenbossche:
If you trigger the wheel builds, it will also test on other linux architectures (eg make a dummy change in the config)

theroggy:
> If you trigger the wheel builds, it will also test on other linux architectures (eg make a dummy change in the config)

I added a linux-arm runner in the tests...

PR #2369, opened 2025-12-15
MNT: implement PEP 639 for improved license metadata

mwtoews:
This implements [PEP 639](https://peps.python.org/pep-0639/) licence metadata changes and silences package build warnings with recent setuptools versions.


jorisvandenbossche (approved):
Thanks!

PR #2370, opened 2025-12-15
BUG: fix (rare) error in split of a line with a (multi)point

### LINKED ISSUE #2087: Crash when splitting a multilinestring with a multipoint

Here is what seems to be a minimal example:

```python
pts = shapely.from_wkt("MULTIPOINT (-8e-16 0, -9e-16 0)")
line = shapely.from_wkt("LINESTRING (-1.5 0, 1.5 0)")
shapely.ops.split(line, pts)
```

Produces a crash with `shapely.errors.GEOSException: IllegalArgumentException: point array must contain 0 or >1 elements`

Those points are very close to each other but not quite identical. Changing the `9` to an `8` fixes the problem, changing the exponent from `-16` to `-15` fixes the problem, and making the multipoint coordinates positive fixes the problem, so I imagine this is tickling some floating-point issue.

## Operating system

Mac OS X 14.4.1
Python 3.11.5

2.0.4 via `poetry`. I also reproduced it on a google colab with `shapely` version 2.0.4.

theroggy:
In some rare cases, when the split-point is very close to a vertex of the linestring to split and when floating points precision issues lead to this point being treated as the same distance from the previous vertex, `split` gives the following error: `shapely.errors.GEOSException: IllegalArgumentException: point array must contain 0 or >1 elements`.

PR fixes this by not splitting (for that split-point) in this case.

Resolves #2087


jorisvandenbossche:
As mentioned on the issue, the GEOSSplit function seems to handle this "fine" (to some extent) and is splitting it. We still need our own implementation for a while, so it would certainly be good to still have a better behaviour than the current strange error. But wouldn't the correct fix be to split it?

shapely/tests/legacy/test_split.py
```diff
@@ -133,6 +133,15 @@ def test_split_line_with_point(self):
...
         self.helper(self.ls, splitter, 1)
 
+    def test_split_line_with_point_fp_precision_issue(self):
+        """Test split of a line with a point triggering fp precision issues."""
+        line = LineString([(-1.5, 0), (-8e-16, 0)])
+
+        # point within floating point precision to 2nd vertex but still on interior
+        # of line --> return equal
+        splitter = Point(-9e-16, 0)
+        self.helper(line, splitter, 1)
```
jorisvandenbossche, line 143: For this one, GEOS also returns the input:

```
>>> geometry = LineString([(-1.5, 0), (-8e-16, 0)])
>>> splitter = Point(-9e-16, 0)
>>> shapely.split(geometry, splitter).wkt
'GEOMETRYCOLLECTION (LINESTRING (-1.5 0, -8e-16 0))'
```

shapely/tests/legacy/test_split.py
```diff
@@ -146,6 +155,17 @@ def test_split_line_with_multipoint(self):
         splitter = MultiPoint([(1, 1), (1.5, 1.5), (1, 1)])
         self.helper(self.ls, splitter, 3)
 
+    def test_split_line_with_multipoint_fp_precision_issue(self):
+        """Test split of a line with a multipoint triggering fp precision issues.
+
+        This tests the handling of floating point precision issues when the
+        splitter point is extremely close to a vertex of the line.
```
jorisvandenbossche, line 162: The splitter point is not actually close to any of the vertices?  Only close to (or on) the line segment, but it is still quite in the middle of the segment?

```
>>> shapely.distance(shapely.get_point(line, 0), splitter.geoms[0])
np.float64(1.4999999999999991)
>>> shapely.distance(shapely.get_point(line, 1), splitter.geoms[0])
np.float64(1.5000000000000009)
>>> shapely.distance(line, splitter.geoms[0])
np.float64(0.0)
```

So here it seems more logical that the expected result is actually splitting the line?

jorisvandenbossche: Ah, I misinterpreted your (still wrong) comment I think, because looking at the code, you actually do expect one split now. 

So I assume the issue is that after splitting with the first point of the MultiPoint, then splitting the remainder with the second point of the MultiPoint, then you have a case of a vertex being very close

PR #2376, opened 2025-12-16
BUG: fix to_ragged_array to work with read-only ndarray input

jorisvandenbossche:
xref https://github.com/geopandas/geopandas/issues/3697

Similar as https://github.com/shapely/shapely/pull/1744,  but then for the geometry input

### DIFF
--- a/shapely/_geometry_helpers.pyx
+++ b/shapely/_geometry_helpers.pyx
@@ -214,7 +214,7 @@ cdef const GEOSGeometry* GetRingN(GEOSContextHandle_t handle, GEOSGeometry* poly
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
-def get_parts(object[:] array, bint extract_rings=0):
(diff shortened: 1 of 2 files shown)


shapely/_geometry_helpers.pyx
```diff
@@ -214,7 +214,7 @@ cdef const GEOSGeometry* GetRingN(GEOSContextHandle_t handle, GEOSGeometry* poly
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
-def get_parts(object[:] array, bint extract_rings=0):
+def get_parts(np.ndarray[object] array, bint extract_rings=0):
```
jorisvandenbossche: typically when working with memoryviews, we would make this `const object[:]`, like was done in https://github.com/shapely/shapely/pull/1744, but that doesn't work for object dtype. Using the numpy ndarray syntax allows read-only by default

caspervdw (approved):

PR #2381, opened 2025-12-19
BUG: fix split if point is on start or end coordinate while line self-intersects on this point

### LINKED ISSUE #1569: Should I expect ops.split to split Linestrings at its ending points when that is self-intersecting?
Example script:

```
import shapely

foo = shapely.geometry.LineString([(0,0), (0,1), (1,1), (1,0), (0,0), (-1,0)])
bar = shapely.geometry.LineString([(0,-1), (0,0), (0,1), (1,1), (1,0), (0,0), (-1,0)])
baz = shapely.geometry.LineString([(0,0), (0,1), (1,1), (1,0), (-1,0)])
qux = shapely.geometry.LineString([(0,-1), (0,1), (1,1), (1,0), (-1,0)])
splitter = shapely.geometry.Point(0,0)

print(shapely.__version__)
for geom in [foo, bar, baz, qux]:
    print(shapely.ops.split(geom, splitter).wkt)
```
That returns:
```
1.8.4
GEOMETRYCOLLECTION (LINESTRING (0 0, 0 1, 1 1, 1 0, 0 0, -1 0))
GEOMETRYCOLLECTION (LINESTRING (0 -1, 0 0), LINESTRING (0 0, 0 1, 1 1, 1 0, 0 0, -1 0))
GEOMETRYCOLLECTION (LINESTRING (0 0, 0 1, 1 1, 1 0, -1 0))
GEOMETRYCOLLECTION (LINESTRING (0 -1, 0 0), LINESTRING (0 0, 0 1, 1 1, 1 0, -1 0))
```
I am expecting the geometries always break at (0, 0) in the case of `foo` and `baz`, no matter if the `splitter` is an ending point of the geometry or not.

Did I misunderstood the manual or is this potentially a bug?

The build is:
geos                      3.11.0               h27087fc_0    conda-forge
python                    3.8.13          h582c2e5_0_cpython    conda-forge
shapely                   1.8.4            py38h3b45516_0    conda-forge

Many thanks.

theroggy:
If a line self-intersects on the start or end point, and the split-point happens to be that point, it seems logical that the line is split on that location, but this is not the case.

Reason: there was a check that if `relate(line, split_point)` resulted in the `split_point` not being on the interior, that no actual split was needed. However, `relate` does not make a distinction if the line is self-intersecting or not, which leads to this issue.

This PR fixes this.

resolves #1569

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -20,6 +20,10 @@ in a more performant way than `shapely.ops.transform` does. `shapely.ops.transfo
 Added return values to `prepare` and `destroy_prepared`. They now return whether something was done;
 calling `prepare` twice on a geometry will first return `True`, and then `False`.
 
+Bug fixes:
+
+- Fix `split` does not split self intersecting lines if point on start or end
+  coordinate (#2381).
 
 2.1.2 (2025-09-24)
 ------------------
--- a/shapely/ops.py
+++ b/shapely/ops.py
@@ -373,16 +373,35 @@ class SplitOp:
         if not isinstance(splitter, Point):
             raise GeometryTypeError("Second argument must be a Point")
 
-        # check if point is in the interior of the line
-        if not line.relate_pattern(splitter, "0********"):
-            # point not on line interior --> return collection with single identity line
-            # (REASONING: Returning a list with the input line reference and creating a
-            # GeometryCollection at the general split function prevents unnecessary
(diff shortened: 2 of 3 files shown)


jorisvandenbossche:
It seems to make sense to fix this, but before I review, want to be sure that the upstream version in GEOS will also do this (https://github.com/libgeos/geos/issues/1494)

PR #2385, opened 2025-12-21
PERF: use prepared versions of relate and relate_pattern if possible

theroggy:
Prepared variants available since GEOS 3.13.

reference #1570, #2382


jorisvandenbossche:
Thanks @theroggy!

PR #2392, opened 2025-12-30
DOC: Fix + improve doc of within

### LINKED ISSUE #1081: Documentation error: description of method "within"
The current description of method "within" is:

>   Returns ``True`` if the object's `boundary` and `interior` intersect only
>   with the `interior` of the other (not its `boundary` or `exterior`).

However, in fact the object's boundary can intersect the boundary of the other. For example:

> aa=Polygon([(0,0),(0,2),(2,2),(2,0)])
> bb=Polygon([(0,0),(0,1),(1,0)])
> bb.within(aa)

> True

If the current definition were correct, "within" would not be the inverse of "contains".

theroggy:
Resolves #1081

### DIFF
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -1260,10 +1260,11 @@ example, the following lines touch at ``(1, 1)``, but do not overlap.
 
 .. method:: object.within(other)
 
-  Return ``True`` if the object is completely inside the other. For more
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Thanks!

PR #2414, opened 2026-01-30
doc: add 'See Also' sections to geometry class docstrings

### LINKED ISSUE #1925: Include See also in docstrings to link between Polygon and polygons etc

AmmarYasser455:
## Summary

This PR adds "See Also" sections to geometry class docstrings, linking each geometry class to its corresponding vectorized function.

Closes #1925

## Changes

| Class | Links to |
|-------|----------|
| `Point` | `shapely.points` |
| `LineString` | `shapely.linestrings` |
| `Polygon` | `shapely.polygons`, `shapely.box` |
| `LinearRing` | `shapely.linearrings` |
| `MultiPoint` | `shapely.multipoints`, `Point` |
| `MultiLineString` | `shapely.multilinestrings`, `LineString` |
| `MultiPolygon` | `shapely.multipolygons`, `Polygon` |

## Motivation

As mentioned in #1925, users often miss the existence of vectorized functions like `shapely.polygons` because the `Polygon` class documentation doesn't reference them. This leads to inefficient code where users write for-loops instead of using the faster vectorized alternatives.

By adding "See Also" sections following NumPy docstring conventions, we help users discover these efficient functions directly from the class documentation.

### DIFF
--- a/shapely/geometry/linestring.py
+++ b/shapely/geometry/linestring.py
@@ -24,6 +24,11 @@ class LineString(BaseGeometry):
         an array-like with shape (N, 2) or (N, 3).
         Also can be a sequence of Point objects, or combination of both.
 
+    See Also
+    --------
+    shapely.linestrings :
+        Create multiple line strings efficiently from coordinate arrays.
+
     Examples
     --------
     Create a LineString with two segments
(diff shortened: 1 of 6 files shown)


theroggy:
Oops, sorry, I thought this was a PR I created :-(...

theroggy (commented):
A detail, but in the numpy style guide the "See Also" section is before the "Examples" section. Hence, maybe them before "Examples"?

AmmarYasser455:
Thanks for the review feedback @mwtoews and @theroggy! I've addressed both comments:

1. Fixed the long line in `linestring.py` to pass pre-commit ✅
2. Moved all 'See Also' sections before 'Examples' sections to follow NumPy docstring conventions ✅

Ready for re-review when you get a chance!

PR #2415, opened 2026-01-30
 perf: Improve performance of scalar YY->d functions

caspervdw:
This extends the work started in https://github.com/shapely/shapely/pull/2339 for the  (Y,Y -> d) functions (like `distance`)

In this PR I also removed a protection that was necessary for GEOS < 3.8.2, which we don't support anymore.  See relevant issue https://trac.osgeo.org/geos/ticket/1058 and the fix I can find in GEOS 3.10 (https://github.com/libgeos/geos/blob/b6f4a63116a16fa698f41df31004b4a4ad35c0b6/capi/geos_ts_c.cpp#L3461)

Also I added a pre-emptive "isEmpty" check on the input geometries. This because tests were giving sporadic failures:

- `distance(geometry, empty)` returned `Inf` instead of `NaN` with GEOS main: https://github.com/shapely/shapely/actions/runs/21527444949/job/62525317983
- invalid value encountered with empty geometries: https://github.com/shapely/shapely/actions/runs/21684108311/job/62526302122?pr=2415

In any case, the pre-emptive check was already there for `frechet_distance`, `line_locate_point` and `line_locate_point_normalized` so I effectively only added it for `distance` and `haussdorfdistance`

### DIFF
--- a/src/geos_funcs_YY_d.c
+++ b/src/geos_funcs_YY_d.c
@@ -96,7 +96,7 @@ static char core_YY_d_operation(GEOSContextHandle_t context, FuncGEOS_YY_d* func
   }
 
   // GEOS behaves as follows for empty inputs:
-  //  - Project[Normalized] segfault
+  //  - Project[Normalized] segfaults (for geom2) or sets a GEOS error (for geom1)
   //  - FrechetDistance sets a GEOS error
   //  - Distance gives 0.0 or Infinity as output
   //  - HausdorffDistance (sometimes?) sets the "invalid value encountered" flag
@@ -248,8 +248,8 @@ DEFINE_YY_d(GEOSProjectNormalizedWrapped_r);
  * 2. A scalar function (e.g., "distance_scalar") for single geometry pair operations
  *
  * Parameters:
+ *   func_name: GEOS function to call
  *   py_name: Python function name (e.g., distance)
- *   func: GEOS function to call
  *
  */
 


caspervdw:
I am seeing a big improvement in the scalar case, but a slight (4%) degrade in the vectorized case, probably because of the pre-emptive isEmpty check.

GEOS behaves as follows for empty inputs:
 - Project[Normalized] segfault
 - FrechetDistance sets a GEOS error
 - Distance gives 0.0 or Infinity  as output
 - HausdorffDistance: returns NaN (what we want)

```
In [1]: import shapely
In [2]: import numpy as np
In [3]: p = shapely.Point(0, 0)
In [4]: x = shapely.points(np.arange(1000), np.arange(1000))
In [5]: %timeit p.distance(x[0])
482 ns ± 2.66 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)  # this branch
4.13 μs ± 75.2 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each) # main
In [6]: %timeit p.distance(x)
276 μs ± 850 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each) # this branch
265 μs ± 1.13 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each) # main
```

src/geos_funcs_YY_d.c
```diff
@@ -0,0 +1,292 @@
...
+/*
+ * We use a macro to register both ufunc and scalar versions of a function with Python.
+ *
+ * This creates two Python-callable functions:
+ * 1. A NumPy ufunc (e.g., "distance") for array operations
+ * 2. A scalar function (e.g., "distance_scalar") for single geometry pair operations
+ *
+ * Parameters:
+ *   py_name: Python function name (e.g., distance)
+ *   func: GEOS function to call
```
jorisvandenbossche:
```suggestion
 *   func_name: GEOS function to call
 *   py_name: Python function name (e.g., distance)
```

src/geos_funcs_YY_d.c
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  // Handle missing geometry (None) - return NaN
+  if ((geom1 == NULL) || (geom2 == NULL)) {
+    *result = NPY_NAN;
+    return PGERR_SUCCESS;
+  }
+
+  // GEOS behaves as follows for empty inputs:
+  //  - Project[Normalized] segfault
```
jorisvandenbossche:
```suggestion
  //  - Project[Normalized] segfaults (for geom2) or sets a GEOS error (for geom1)  
```

Not too important, but just to preserve the information we had previously in the comments. Unless you actually noticed this is no longer true?

jorisvandenbossche:
> * `distance(geometry, empty)` returned `Inf` instead of `NaN` with GEOS main:

That was actually a conscious change on the GEOS side, see https://github.com/libgeos/geos/pull/1345, which _in theory_ we could follow. However, in GEOS that change fixed distance() returning 0 for empty geometries, while in shapely we have always already ensured to return NaN in that case. So I assume we probably want to continue doing that.

Now, I actually don't understand why that wasn't failing in CI for GEOS main at the moment, because we currently have that empty check only if the result is 0. 
@caspervdw you mention in the top post this was giving "sporadic failures" with this PR before adding the pre-emptive check. So there was no specific test that was consistently failing?

> I am seeing a big improvement in the scalar case, but a slight (4%) degrade in the vectorized case, probably because of the pre-emptive isEmpty check.

I think that is fine. I assume this degrade will also only be the case for `distance` and `hausdorff_distance`, as for the other methods there we already had custom checks in their wrapped versions anyway. 
In theory we could keep those empty-checks only in the wrapped versions of the function that need it (and keep the post-check of the result), but I think the simplified code is probably worth it.

caspervdw:
> Now, I actually don't understand why that wasn't failing in CI for GEOS main at the moment, because we currently have that empty check only if the result is 0. @caspervdw you mention in the top post this was giving "sporadic failures" with this PR before adding the pre-emptive check. So there was no specific test that was consistently failing?

So it's been a while, but looking at https://github.com/shapely/shapely/actions/runs/21527444949/job/62525317983 I see that without our pre-emptive IsEmpty, the `test_distance_empty` is failing on GEOS main only (it returns Inf instead of NaN. But as you say, you'd expect main failing as well. I went through the old code and indeed the Inf should theoretically not be caught 
```
      /* in case the outcome is 0.0, check the inputs for emptyness */
      if (*op1 == 0.0) {
        if (GEOSisEmpty_r(ctx, in1) || GEOSisEmpty_r(ctx, in2)) {
          *(double*)op1 = NPY_NAN;
        }
      }
```
> I think that is fine. I assume this degrade will also only be the case for distance and hausdorff_distance, as for the 
other methods there we already had custom checks in their wrapped versions anyway.

Exactly. Pre-checking for Project/FrechetDistance and post-checking for Distance/HausdorffDistance complicates matters, let's accept this performance degrade for the sake of code clarity.

PR #2432, opened 2026-04-07
CI: address warnings from cibuildwheel

jorisvandenbossche:
Small follow-up on https://github.com/shapely/shapely/pull/2431, addressing part of the warnings that can be seen at https://github.com/shapely/shapely/actions/runs/24053890587.  
There is another node warning coming from `ilammy/msvc-dev-cmd@v1` (but that first needs a new release in that action), and about the macosx deployment target (but cibuildwheel is bumping that correctly anyway, and to different values depending on the build)


mwtoews (approved):
Looks good, thanks for this fix!

PR #2436, opened 2026-04-18
perf: Improve performance of scalar Ydb->Y functions

caspervdw:
This extends the work started in https://github.com/shapely/shapely/pull/2339 for the (Y,d,b -> Y) functions

### DIFF
--- a/src/geos_funcs_Ydb_Y.c
+++ b/src/geos_funcs_Ydb_Y.c
@@ -1,0 +1,345 @@
+#define PY_SSIZE_T_CLEAN
+
+#include <Python.h>
+#include <math.h>
+
+#define NO_IMPORT_ARRAY
+#define NO_IMPORT_UFUNC
+#define PY_ARRAY_UNIQUE_SYMBOL shapely_ARRAY_API
+#define PY_UFUNC_UNIQUE_SYMBOL shapely_UFUNC_API
+#include <numpy/ndarraytypes.h>
+#include <numpy/npy_3kcompat.h>
+#include <numpy/ufuncobject.h>
+
+#include "fast_loop_macros.h"
+#include "geom_arr.h"
+#include "geos.h"
+#include "pygeos.h"
+#include "pygeom.h"
+#include "signal_checks.h"
+
+/* ========================================================================
+ * GEOS WRAPPER FUNCTIONS
+ * ========================================================================
+ *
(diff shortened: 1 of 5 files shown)


jorisvandenbossche:
@caspervdw this gets us into the "custom" ufuncs. While for this PR, there are still three that happen to have the same signature (and there are probably still a few other categories to distinguish, like O_Y), did you already think about what you want to do in general for those custom ones? 
(or leave those as they are currently in the ufuncs.c file? Also not all of the custom ones probably needs a scalar equivalent)

caspervdw:
> @caspervdw this gets us into the "custom" ufuncs. While for this PR, there are still three that happen to have the same signature (and there are probably still a few other categories to distinguish, like O_Y), did you already think about what you want to do in general for those custom ones? (or leave those as they are currently in the ufuncs.c file? Also not all of the custom ones probably needs a scalar equivalent)

I don't have a particular strategy in mind yet. The goals are clearer though: to give the ufuncs a more performant scalar version, and at the same time make the code more maintainable.

Pulling functions out with the same call signature makes less sense from a code reuse perspective, but still, scalar versions might be useful. Also, having the call signatures fully split out makes extensibility to other ufuncs better as long as the call signature is already present. That a function happens to have a unique call signature should not influence the scalar version being present or not.

So I am tending towards just keeping up the same patterns and making files with just single functions (like `dddd_Y` for box).

src/geos_funcs_Ydb_Y.c
```diff
@@ -0,0 +1,345 @@
...
+  return GEOSConcaveHull_r(ctx, g, ratio, (unsigned int)allow_holes);
+}
+#endif  // GEOS_SINCE_3_11_0
+
+#if GEOS_SINCE_3_12_0
+static GEOSGeometry* ShapelyCoverageSimplifyVWValidated_r(GEOSContextHandle_t ctx,
+                                                           const GEOSGeometry* g,
+                                                           double tolerance,
+                                                           int simplify_boundary) {
+  int num_geoms = GEOSGetNumGeometries_r(ctx, g);
```
jorisvandenbossche, line 57:
```suggestion
  // Validate the geometries in the collection
  int num_geoms = GEOSGetNumGeometries_r(ctx, g);
```

jorisvandenbossche (approved):
Looks good!

PR #2438, opened 2026-04-25
Refactor shortest_line into the other YY_Y functions

caspervdw:
This merges the implementation of the "custom ufunc" `shortest_line` into the other ufuncs with the same call signature (YY->Y, for example `intersection`).

I added three extra conditionals in the main path of the other YY-Y function. Theoretically, these should be very fast (order nanoseconds), but I did benchmark the difference. I didn't measure anything larger than the benchmark noise. So I think we're good here.


jorisvandenbossche (approved):
Looks good, thanks!

Issue #2338, opened 2025-10-26
Remove `@multithreading_enabled` decorator usage


caspervdw:
This proposal suggests removing the `@multithreading_enabled` decorator from all Shapely functions to improve performance. The decorator adds significant overhead (~2x slower for scalar operations) while providing limited thread safety benefits that should be handled at the application level.

## Performance

Performance analysis ib #2035 shows the decorator adds 1 microsecond overhead per function call.

- `shapely.get_x(point)`: ~2,250ns
- `shapely.lib.get_x(point)` (bypassing the decorator): ~1,120ns  

This affects most functions across Shapely's modules.

## NumPy's approach

NumPy handles thread safety by putting responsibility on users. From the [NumPy documentation](https://numpy.org/doc/stable/reference/thread_safety.html):

> "It is possible to share NumPy arrays between threads, but extreme care must be taken to avoid creating thread safety issues when mutating arrays that are shared between multiple threads. If two threads simultaneously read from and write to the same array, they will at best produce inconsistent, racey results that are not reproducible, let alone correct. It is also possible to crash the Python interpreter by, for example, resizing an array while another thread is reading from it to compute a ufunc operation."

Contrary to Numpy, Shapely does release the GIL when iterating on an `object`-dtyped array. This adds an additional concern; it is possible that one thread overwrites an array position, thereby reducing the refcount of the Geometry to 0 and deleting it, while another thread is operating on the same Geometry. 

## The decorator is no lock

The `@multithreading_enabled` decorator only makes arrays read-only temporarily. It can still lead to interpreter crashes in the following scenario:

- Four threads start operations on the same array. Thread A sets it to read-only, threads B, C, D see it's already read-only and proceed.
- Thread A terminates quickly (maybe it had many None values or simple geometries), setting the array back to writable.
- Now anything can happen between the remaining threads - thread B might overwrite a geometry that thread C is still operating on, causing thread C to crash the interpreter.

The decorator creates a false coordination mechanism where early-finishing threads can "unlock" arrays that other threads are still using.

# Proposed Changes

1. **Remove `@multithreading_enabled`** from all Shapely functions
2. **Keep the decorator itself** but add a deprecation warning when used  
3. **Update documentation** with proper thread safety guidance (similar to NumPy)

jorisvandenbossche:
I think I am fine with removing the decorator. 

[...]

PR #2445, opened 2026-06-12
Fix ValueError in multithreading_enabled decorator when restoring writeable flags for array views and original arrays

### LINKED ISSUE #2410: BUG: passing view and original array to binary method releasing the GIL gives ValueError in multithreading_enabled decorator

```python
>>> import shapely

>>> shapely.equals(arr, arr)
array([ True])
array([ True])

Cell In[69], line 1

     89 finally:

```

Federicorao:
Fixes #2410.

Fix ValueError in multithreading_enabled decorator when restoring writeable flags for array views and original arrays. The multithreading_enabled decorator collects all object-dtype ndarray arguments, sets their writeable flag to False, calls the function, and then restores each flag. When a view and its original array are both passed, restoring the view's flag can raise ValueError because the original array's flag is still False at that point (it is restored later). The fix wraps the flag restoration in a try-except to ignore ValueError, which matches the suggestion in the issue and ensures the decorator does not crash in this case.

I checked that the changed Python files still parse correctly.

### DIFF
--- a/shapely/decorators.py
+++ b/shapely/decorators.py
@@ -88,7 +88,10 @@ def multithreading_enabled(func):
             return func(*args, **kwargs)
         finally:
             for arr, old_flag in zip(array_args, old_flags, strict=False):
-                arr.flags.writeable = old_flag
+                try:
+                    arr.flags.writeable = old_flag
+                except ValueError:
+                    pass
 
     return wrapped
 


caspervdw:
This decorator is pending removal (see #2338), so I don't think this fix is worth it.

If you have the bandwith to implement the proposal I did in #2338 that would be much appreciated though :)

PR #2447, opened 2026-06-13
perf: Refactor ufuncs with CUSTOM call signature

caspervdw:
This extends the work started in https://github.com/shapely/shapely/pull/2339 for a couple of functions that have unique call signatures.

It's a lot of code, and of course I had help of an LLM. I did carefully review and compare to the old code and similar implementations. Also I checked the test coverage of the touched geometry methods and added a path that was untested.

Some remarks:

- For the `is_valid_reason`, the `relate` and `relate_pattern` I chose not to work with a function pointer so that we can extend later with other GEOS functions. The reason for this are the special call signatures including a string.
- For the `set_precision` I moved the check for valid values of the `flags` parameter into the python code. 
- The `clip_by_rect` not requires the minx/maxx/miny/maxy to be scalars anymore. The function just loops over all arrays.
- In `equals_exact` the outcome of the ufunc was checked for errors like this: `((ret != 0) && (ret != 1))`. I changed that to `(ret == 2)` like all others.


shapely/geometry/base.py
```diff
@@ -900,6 +902,12 @@ def dwithin(self, other, distance):
 
         Refer to `shapely.dwithin` for full documentation.
         """
+        try:
+            distance = float(distance)
+        except (TypeError, ValueError):
+            pass
+        if isinstance(distance, float) and shapely.lib.is_valid_input_scalar(other):
+            return shapely.lib.dwithin_scalar(self, other, float(distance))
```
jorisvandenbossche, line 910:
```suggestion
            return shapely.lib.dwithin_scalar(self, other, distance)
```

The isinstance check already ensures it is a float

src/geos_funcs_Y_O.c
```diff
@@ -0,0 +1,151 @@
...
+    PyErr_Format(PyExc_TypeError, "expected 1 argument, got %zd", nargs);
+    return NULL;
+  }
+
+  GEOS_INIT;
+
+  errstate = core_is_valid_reason(ctx, args[0], &result);
+
+  GEOS_FINISH;
+
```
jorisvandenbossche, line 83: Adding a 
```
 if (errstate != PGERR_SUCCESS) {
    return NULL;  // Python exception was set by GEOS_FINISH
  }
```

is not needed because `result` will be null in that case?

jorisvandenbossche (approved):
Looks good!

src/shapely/lib/geos_funcs_Y_O.c
```diff
@@ -0,0 +1,151 @@
...
+    PyErr_Format(PyExc_TypeError, "expected 1 argument, got %zd", nargs);
+    return NULL;
+  }
+
+  GEOS_INIT;
+
+  errstate = core_is_valid_reason(ctx, args[0], &result);
+
+  GEOS_FINISH;
+
```
caspervdw: Yes I think it is better to return `result` in all cases because if it is not-NULL, it is a Python object and then it needs to be returned or else you get a memory leak.

caspervdw:
@jorisvandenbossche  I fixed the merge conflicts and addressed your comments, but because of the release work I leave the actual merge to you.

jorisvandenbossche:
Thanks! I think we can certainly still merge this for 2.2.0

jorisvandenbossche:
@caspervdw are you planning to do a similar refactor for the creation ufuncs? (like `points`)

PR #2454, opened 2026-06-19
feat: Add coverage_clean function

JuhoErvasti:
Implement `GEOSCoverageCleanWithParams_r` from the GEOS >=3.14.0 C API as `coverage_clean`.

### DIFF
--- a/shapely/_coverage.py
+++ b/shapely/_coverage.py
@@ -210,7 +210,7 @@ def coverage_clean(
     """Return a cleaned version of an input geometry.
 
     Assumes that the geometry forms a polygonal coverage. Ensures that no polygons
-    overlap and removes small gaps.
+    overlap, small gaps are removed, and all shared edges are exactly identical.
 
     A collection of valid polygons is considered a coverage if the polygons are:
 
@@ -225,15 +225,22 @@ def coverage_clean(
     Parameters
     ----------
     geometry : Geometry or array_like
-    gap_width: float
-        Gaps smaller than this value will be removed.
-    snapping_distance: float
+    gap_width : float, default 0.0
+        Gaps smaller than this value are merged with the adjacent polygon with
+        longest shared shared border. Set to 0.0 to (default) for no removal
+        of gaps.
+    snapping_distance : float
         Determines the node snapping step when nearby vertices are snapped
-        together. Set to -1 by default, which automatically finds a snapping
-        distance based on the input. Set to 0.0 to disable snapping.
-    merge_strategy:
-        Determines how overlapping areas are handled by choosing which polygons
-        to merge them into.
+        together. Set to -1 by default, which automatically finds a small snapping
+        distance based on the extend of the input. Set to 0.0 to disable snapping.
+    merge_strategy : {'longest_border', 'max_area', 'min_area', 'min_index'}, default 'longest_border'
+        Determines how overlapping areas are handled by choosing which polygons to merge
+        them into. CoverageCleanMergeStrategy.longest_border ('longest_border') chooses
+        polygon with longest common border. CoverageCleanMergeStrategy.max_area
+        ('max_area') chooses polygon with largest area.
+        CoverageCleanMergeStrategy.min_area ('min_area') chooses polygon with smallest
+        area. CoverageCleanMergeStartegy.min_index ('min_index') chooses the first
+        encountered polygon.
 
     Returns
     -------
@@ -245,13 +252,13 @@ def coverage_clean(
 
     Examples
     --------
-    >>> import shapely
-    >>> from shapely import box
-    >>> polygons = [box(0, 0, 1, 1), box(0.9, 0, 2, 1)]
-    >>> shapely.coverage_clean(polygons)
+    >>> import shapely  # doctest: +SKIP
+    >>> from shapely import box  # doctest: +SKIP
+    >>> polygons = [box(0, 0, 1, 1), box(0.9, 0, 2, 1)]  # doctest: +SKIP
+    >>> shapely.coverage_clean(polygons)  # doctest: +SKIP
     array([<POLYGON ((0 0, 0 1, 0.9 1, 1 1, 1 0, 0.9 0, 0 0))>,
        <POLYGON ((1 1, 2 1, 2 0, 1 0, 1 1))>], dtype=object)
-    """
+    """  # noqa: E501
     if isinstance(merge_strategy, str):
         merge_strategy = CoverageCleanMergeStrategy.get_value(merge_strategy)
 
--- a/shapely/_coverage.py
+++ b/shapely/_coverage.py
@@ -259,8 +259,16 @@ def coverage_clean(
     array([<POLYGON ((0 0, 0 1, 0.9 1, 1 1, 1 0, 0.9 0, 0 0))>,
        <POLYGON ((1 1, 2 1, 2 0, 1 0, 1 1))>], dtype=object)
     """  # noqa: E501
+    if not np.isscalar(merge_strategy):
+        raise TypeError("merge_strategy only accepts scalar values")
     if isinstance(merge_strategy, str):
         merge_strategy = CoverageCleanMergeStrategy.get_value(merge_strategy)
+    if isinstance(merge_strategy, int) and not CoverageCleanMergeStrategy.has_value(
+        merge_strategy
+    ):
+        raise ValueError(f"not a valid merge strategy value: {merge_strategy}.")
+    if isinstance(merge_strategy, CoverageCleanMergeStrategy):
+        merge_strategy = merge_strategy.value
 
     geometries = np.asarray(geometry)
     # we always consider the full array as a single coverage -> ravel the input
--- a/shapely/_coverage.py
+++ b/shapely/_coverage.py
@@ -1,10 +1,40 @@
 import numpy as np
 
 from shapely import Geometry, GeometryType, lib
+from shapely._enum import ParamEnum
 from shapely._geometry import get_parts
 from shapely.decorators import multithreading_enabled, requires_geos
 
-__all__ = ["coverage_invalid_edges", "coverage_is_valid", "coverage_simplify"]
+__all__ = [
+    "coverage_clean",
+    "coverage_invalid_edges",
+    "coverage_is_valid",
+    "coverage_simplify",
+]
+
+
+class CoverageCleanMergeStrategy(ParamEnum):
+    """Enumeration of coverage clean merge strategies.
+
+    Determines which neighboring polygon to merge overlapping areas into.
+
+    Attributes
+    ----------
+    longest_border : int
+        Polygon with longest common border is chosen.
+    max_area : int
+        Polygon with largest area is chosen.
+    min_area : int
+        Polygon with smallest area is chosen.
+    min_index : int
+        Polygon with smallest input index is chosen.
+
+    """
+
+    longest_border = 0
+    max_area = 1
+    min_area = 2
+    min_index = 3
 
 
 @requires_geos("3.12.0")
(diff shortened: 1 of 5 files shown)


JuhoErvasti:
AFAICT the test failures are because

a) in GEOS 3.15-dev the position of the first vertex in the results are different. I'm not sure how to fix this, should I normalize the results?

b) the doctests fails because it's run with GEOS 3.13. I assume there's a way to skip certain tests since `coverage_is_valid` etc. don't fail, but I don't really see how.

shapely/_coverage.py
```diff
@@ -166,3 +196,71 @@ def coverage_simplify(geometry, tolerance, *, simplify_boundary=True):
...
+    geometry,
+    *,
+    gap_width=0.0,
+    snapping_distance=-1,
+    merge_strategy="longest_border",
+):
+    """Return a cleaned version of an input geometry.
+
+    Assumes that the geometry forms a polygonal coverage. Ensures that no polygons
+    overlap and removes small gaps.
```
jorisvandenbossche:
```suggestion
    overlap, small gaps are removed, and all shared edges are exactly identical.
```

shapely/_coverage.py
```diff
@@ -166,3 +196,71 @@ def coverage_simplify(geometry, tolerance, *, simplify_boundary=True):
...
+
+    If there are other geometry types than Polygons or MultiPolygons present,
+    the function will raise an error.
+
+    .. versionadded:: 2.2.0
+
+    Parameters
+    ----------
+    geometry : Geometry or array_like
+    gap_width: float
```
jorisvandenbossche:
```suggestion
    gap_width : float, default 0.0
```

shapely/_coverage.py
```diff
@@ -166,3 +196,71 @@ def coverage_simplify(geometry, tolerance, *, simplify_boundary=True):
...
+        Gaps smaller than this value will be removed.
```
jorisvandenbossche: Maybe add about "Gaps are merged with the adjacent polygon with longest shared border" and that "set to 0.0 (default) for no removal of gaps" ?

shapely/_coverage.py
```diff
@@ -166,3 +196,71 @@ def coverage_simplify(geometry, tolerance, *, simplify_boundary=True):
...
+    snapping_distance: float
```
jorisvandenbossche:
```suggestion
    snapping_distance : float
```

shapely/_coverage.py
```diff
@@ -166,3 +196,71 @@ def coverage_simplify(geometry, tolerance, *, simplify_boundary=True):
...
+        Determines the node snapping step when nearby vertices are snapped
+        together. Set to -1 by default, which automatically finds a snapping
+        distance based on the input. Set to 0.0 to disable snapping.
```
jorisvandenbossche:
```suggestion
        together. Set to -1 by default, which automatically finds a small snapping
        distance based on the extent of the input. Set to 0.0 to disable snapping.
```

shapely/_coverage.py
```diff
@@ -166,3 +196,71 @@ def coverage_simplify(geometry, tolerance, *, simplify_boundary=True):
...
+    merge_strategy:
```
jorisvandenbossche:
```suggestion
    merge_strategy : {'longest_border', 'max_area', 'min_area', 'min_index'}, default 'longest_border'
```

shapely/_coverage.py
```diff
@@ -166,3 +196,71 @@ def coverage_simplify(geometry, tolerance, *, simplify_boundary=True):
...
+        Determines how overlapping areas are handled by choosing which polygons
+        to merge them into.
```
jorisvandenbossche: Can you briefly list the options here? (essentially copy the explanation of the attributes in the docstring of `CoverageCleanMergeStrategy` in here as a list or so)

src/ufuncs.c
```diff
@@ -1168,6 +1168,120 @@ static PyUFuncGenericFunction coverage_invalid_edges_funcs[1] = {&coverage_inval
...
+      n_geoms++;
+    }
+    collection =
+        GEOSGeom_createCollection_r(ctx, GEOS_GEOMETRYCOLLECTION, geoms, n_geoms);
+    if (collection == NULL) {
+      errstate = PGERR_GEOS_EXCEPTION;
+      goto finish;
+    }
+
+    result_collection = GEOSCoverageCleanWithParams_r(ctx, collection, params);
```
jorisvandenbossche, line 1261: Do we have to set an errstate if the return value is NULL here?

shapely/tests/test_coverage.py
```diff
@@ -247,3 +247,203 @@ def test_coverage_unsupported_geos():
...
+    # polygons with a gap
+    poly1 = shapely.from_wkt("POLYGON ((0 11, 10 11, 10 -1, 0 -1, 0 11))")
+    poly2 = shapely.from_wkt(
+        "POLYGON ((10 10, 10 13, 20 13, 20 -2, 10 -2, 10 0, 10.5 1, 10.5 8.5, 10 10))",
+    )
+
+    # gap width 0.0
+    expected_1 = shapely.from_wkt(
+        [
+            "POLYGON ((0 -1, 0 11, 10 11, 10 10, 10 0, 10 -1, 0 -1))",
```
jorisvandenbossche, line 292: This is exactly the same as poly1, poly2 ? (except from ordering of coordinates?) 
To make that easier to follow, you could also put those poly1, poly2 in an array and normalize them (if those are actually normalized?)

jorisvandenbossche (commented):
@JuhoErvasti thanks a lot for this PR!

Looking good, I added a bunch of comments but mostly minor documentation suggestions

Can you also add a test for ensuring we get an error when passing non-polygonal input geometries? (like `test_coverage_simplify_geom_types`)

jorisvandenbossche:
> a) in GEOS 3.15-dev the position of the first vertex in the results are different. I'm not sure how to fix this, should I normalize the results?

If this is not consistent across GEOS versions, normalizing both the result and expected values is indeed the best way to deal with this

> b) the doctests fails because it's run with GEOS 3.13. I assume there's a way to skip certain tests since coverage_is_valid etc. don't fail, but I don't really see how.

I suppose this is the first function that requires GEOS 3.14, and so for the other `coverage_..` functions, the docstrings will work. 
For now, easiest might be to skip the testing of the docstring examples in the PR here by adding `  # doctest: +SKIP` comments to each input line in the example (and will see to update the docstring running to use GEOS 3.14 in a separate PR)

shapely/tests/test_coverage.py
JuhoErvasti: Yes, the first case should not do anything, but (depending on version) GEOS changes the coordinate order. As per your later comment, I've changed these so that the test results are normalized, and I also changed these inputs to be normalized so hopefully it is clearer now.

JuhoErvasti:
@jorisvandenbossche Thank you for the review, everything should now be addressed.

shapely/_coverage.py
```diff
@@ -166,3 +196,78 @@ def coverage_simplify(geometry, tolerance, *, simplify_boundary=True):
...
+    >>> import shapely  # doctest: +SKIP
+    >>> from shapely import box  # doctest: +SKIP
+    >>> polygons = [box(0, 0, 1, 1), box(0.9, 0, 2, 1)]  # doctest: +SKIP
+    >>> shapely.coverage_clean(polygons)  # doctest: +SKIP
+    array([<POLYGON ((0 0, 0 1, 0.9 1, 1 1, 1 0, 0.9 0, 0 0))>,
+       <POLYGON ((1 1, 2 1, 2 0, 1 0, 1 1))>], dtype=object)
+    """  # noqa: E501
+    if isinstance(merge_strategy, str):
+        merge_strategy = CoverageCleanMergeStrategy.get_value(merge_strategy)
+
```
theroggy, line 270: A detail, and on other places the check is often very similar, but when an integer is passed in with an invalid value (e.g. 5), there won't be a clean error. Maybe add a `has_value` method to the `ParamEnum` class to check in the else if the value passed in is fine?

shapely/tests/test_coverage.py
```diff
@@ -247,3 +247,229 @@ def test_coverage_unsupported_geos():
...
+
+
+@pytest.mark.skipif(shapely.geos_version < (3, 14, 0), reason="requires >= 3.14")
+def test_coverage_clean_invalid_merge_strategy():
+    with pytest.raises(ValueError, match="'invalid' is not a valid option"):
+        shapely.coverage_clean([], merge_strategy="invalid")
+
+
+@pytest.mark.skipif(shapely.geos_version < (3, 14, 0), reason="requires >= 3.14")
+def test_coverage_clean_scalar_parameters():
```
theroggy, line 422: Another detail, but the `merge_strategy` parameter should be a scalar as well...

theroggy (commented):
Nice! Two minor details...

JuhoErvasti:
@theroggy Thanks for the review!

I added the additional checks, `merge_strategy` being a scalar is checked in the ufunc but that won't be reached via the Python function because the code will error out at `np.intc()`, so I added an additional scalar check on the Python side along with the `has_value` check.

theroggy (approved):
Thanks, look fine!

JuhoErvasti:
Do I need to change something still? As best I can tell the failing check is unrelated.

jorisvandenbossche:
Not necessarily! Mostly waiting on me getting back from holidays, will do a final look now. The CI failure is indeed unrelated (and fixed by https://github.com/shapely/shapely/pull/2462)

src/ufuncs.c
```diff
@@ -1223,6 +1223,7 @@ static void coverage_clean_func(char** args, const npy_intp* dimensions, const n
     CHECK_SIGNALS(i);
 
     if (errstate == PGERR_PYSIGNAL) {
+      Py_BLOCK_THREADS;
```
jorisvandenbossche, line 1226: Is there something (eg failing test) that triggered you to add this? 
It actually looks correct to do this, but then I missed doing that in `coverage_invalid_edges` as well ..

shapely/_coverage.py
```diff
@@ -166,3 +196,86 @@ def coverage_simplify(geometry, tolerance, *, simplify_boundary=True):
...
+    if not np.isscalar(merge_strategy):
+        raise TypeError("merge_strategy only accepts scalar values")
+    if isinstance(merge_strategy, str):
+        merge_strategy = CoverageCleanMergeStrategy.get_value(merge_strategy)
+    if isinstance(merge_strategy, int) and not CoverageCleanMergeStrategy.has_value(
+        merge_strategy
+    ):
+        raise ValueError(f"not a valid merge strategy value: {merge_strategy}.")
+    if isinstance(merge_strategy, CoverageCleanMergeStrategy):
+        merge_strategy = merge_strategy.value
```
jorisvandenbossche: This one is actually not needed, since we made the enums to inherit from IntEnum, they essentially are already acting as its integer value when passing to `np.intc(..)` below

jorisvandenbossche (approved):
Thanks for the updates!

src/ufuncs.c
JuhoErvasti: Not this line specifically but `test_coverage_clean_geom_types` segfaulted on non-polygonal geometries hence I added the GIL release to each error state.

jorisvandenbossche (approved):

src/ufuncs.c
jorisvandenbossche: Ah, I see. For `coverage_invalid_edges` we don't have such a test since it allows (ignores) non-polygonal geometries, so we didn't catch this.

(but will add those in a follow-up)

jorisvandenbossche:
@JuhoErvasti Thanks again for the PR! The failures are unrelated (there are a few other PRs open at the moment to address those)

JuhoErvasti:
No problem! Thanks for the review!

PR #2455, opened 2026-06-21
Fix writeable flag restore for array views

### LINKED ISSUE #2410: BUG: passing view and original array to binary method releasing the GIL gives ValueError in multithreading_enabled decorator

gaoflow:
## Summary
- Retry restoring object-array writeable flags when NumPy rejects making a view writeable before its base array is restored.
- Add regression coverage for passing a base array and a view to a `multithreading_enabled` function in either order.

Fixes #2410

## Tests
- `.venv/bin/python -m pytest shapely/tests/test_misc.py -q`
- `.venv/bin/python -m pytest shapely/tests/test_misc.py::test_multithreading_enabled_preserves_view_and_base_flags -q`
- Manual reproduction script for `shapely.equals(arr.view(), arr)`
- `uvx ruff check shapely/decorators.py shapely/tests/test_misc.py`
- `git diff --check`

### DIFF
--- a/shapely/decorators.py
+++ b/shapely/decorators.py
@@ -87,8 +87,21 @@ def multithreading_enabled(func):
                 arr.flags.writeable = False
             return func(*args, **kwargs)
         finally:
-            for arr, old_flag in zip(array_args, old_flags, strict=False):
-                arr.flags.writeable = old_flag
+            pending = list(zip(array_args, old_flags, strict=False))
+            while pending:
+                remaining = []
+                last_error = None
+                for arr, old_flag in pending:
+                    try:
+                        arr.flags.writeable = old_flag
+                    except ValueError as err:
+                        if not old_flag:
+                            raise
+                        remaining.append((arr, old_flag))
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
This decorator is pending removal (see https://github.com/shapely/shapely/issues/2338), so this might not be worth fixing (until we might do another 2.1.x release). 
The test might be worth adding though, in the future.

In addition, it seems you are producing a huge number of PRs across a wide set of projects, so I am assuming you are automating this and not actually contributing this PR with a genuine interest in shapely. This is not welcome, or at least then I am not planning to provide much feedback here form my side.

binggao1230:
Thanks for the feedback. I understand this is not worth pursuing while the decorator is pending removal, and I also hear your concern about high-volume drive-by PRs. I am closing this so it does not take more maintainer time.

PR #2456, opened 2026-06-22
Use PEP-735 for dependency groups

mwtoews:
This implements [PEP 735](https://peps.python.org/pep-0735/) to use dependency groups for "test" and "docs", as well as a new "dev" group, which includes "test" and "pre-commit".

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -80,7 +80,7 @@ installing Shapely with pip from PyPI, you clone the package from Github::
 
 Install it in development mode using ``pip``::
 
-    $ pip install -e .[test]
+    $ pip install -e . --group dev
 
 For development, use of a virtual environment is strongly recommended. For
 example using ``venv``:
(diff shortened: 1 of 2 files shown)


jorisvandenbossche (approved):
Looks good to me!

jorisvandenbossche:
I was wondering if we could actually use it ourselves (instead of only mentioning it in the docs), but the install step in `tests.yml` is a bit too complex (depending on the exact matrix params) it seems

mwtoews:
> I was wondering if we could actually use it ourselves (instead of only mentioning it in the docs), but the install step in `tests.yml` is a bit too complex (depending on the exact matrix params) it seems

I was looking leverage this feature in tests.yml too, but it's a bit gnarly at the moment. Worth considering on the next CI workflow refresh.

Nevertheless, this PR helps my local development workflow which generally relies on `uv` which installs "dev" by default.

PR #2462, opened 2026-06-30
TST: use Python 3.14t for free-threaded tests

mwtoews:
This changes testing free-threaded Python from 3.13t (which was [regarded as experimental](https://cibuildwheel.pypa.io/en/stable/changelog/#v400)) to 3.14t. This fixes an issue where the latest numpy does not have the experimental 3.13t wheels anymore; cibuildwheel v4.0 builds cp314t onwards.

Two other minor changes are put here for convenience (of not creating more PRs):

- Ignore `*.lock` files, e.g., uv, pixi, Pipfile
- Rename dependency group "test" to "tests", to align with the "tests" directory name (in "shapely")

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,9 +44,9 @@ dependencies = [
 [dependency-groups]
 dev = [
     "pre-commit",
-    {include-group = "test"},
+    {include-group = "tests"},
 ]
-test = [
+tests = [
     "pytest",
     "pytest-cov",
     "scipy-doctest",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -44,9 +44,9 @@ dependencies = [
 [dependency-groups]
 dev = [
     "pre-commit",
-    {include-group = "test"},
+    {include-group = "tests"},
 ]
-test = [
+tests = [
```
caspervdw, line 49: This is sort-of a breaking change. Not sure how many people are actually installing with `[test]`, but its in the README, so people might actually do this.

What's the reason for changing it?

mwtoews: It's not exactly breaking, as I removed the "test" extra last week and re-labelled it as a [PEP 735](https://peps.python.org/pep-0735/) dependency group e10c6da077744005290f77eaa725777d1c631f66 . It's just a renaming of the group name. There isn't anything in the README, but docs/installation.rst now suggests using `--group dev` to get a few other development dependencies.

PR #2463, opened 2026-06-30
Fix multithreading flag restoration for array views

### LINKED ISSUE #2410: BUG: passing view and original array to binary method releasing the GIL gives ValueError in multithreading_enabled decorator

Whning0513:
## Summary
- restore ndarray writeable flags in a view-safe order
- add a regression test for passing a view before its base array

Fixes #2410

## Testing
- LD_LIBRARY_PATH=/data/whn/.hermes/hermes-agent/venv/lib/python3.11/site-packages/shapely.libs:${LD_LIBRARY_PATH} python -m pytest -q shapely/tests/test_misc.py -k multithreading_enabled
- python -m ruff check shapely/decorators.py shapely/tests/test_misc.py

### DIFF
--- a/shapely/decorators.py
+++ b/shapely/decorators.py
@@ -12,6 +12,15 @@ from shapely import lib
 from shapely.errors import UnsupportedGEOSVersionError
 
 
+def _array_view_depth(arr: np.ndarray) -> int:
+    depth = 0
+    base = arr.base
+    while isinstance(base, np.ndarray):
+        depth += 1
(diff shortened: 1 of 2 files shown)


caspervdw:
Closing this with the same reason as the other solutions of #2410 .

PR #2464, opened 2026-06-30
PERF: remove `@multithreading_enabled` decorator usage

### LINKED ISSUE #2338: Remove `@multithreading_enabled` decorator usage

caspervdw:
Closes #2338 

See issue for an extensive description.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -14,6 +14,10 @@ Improved performance by changing the GEOS context handling. Instead of initializ
 on every call, GEOS contexts are reused through a threadlocal. It should give a performance
 increase of about 75 nanoseconds per call.
 
+Improved performance by removing the multithreading protection, which was not complete protection
+anyway. Now it is up to the user to avoid race conditions when mutating arrays of geometries
+from multiple threads (#2338).
+
 Added `transform_coordseq`, which applies a transformation function per coordinate sequence
 in a more performant way than `shapely.ops.transform` does. `shapely.ops.transform` is now deprecated.
 
--- a/README.rst
+++ b/README.rst
@@ -68,7 +68,13 @@ Shapely functions generally support multithreading by releasing the Global
 Interpreter Lock (GIL) during execution. Normally in Python, the GIL prevents
 multiple threads from computing at the same time. Shapely functions
 internally release this constraint so that the heavy lifting done by GEOS can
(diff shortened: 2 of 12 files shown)


jorisvandenbossche (approved):
This looks good to me!

PR #2465, opened 2026-07-02
Preserve precision grid size when copying geometries

### LINKED ISSUE #2073: deepcopy() looses the precision property if set

I would expect that a deepcopy of any geometry returns a geometry with the precision preserved, but this is not the case.

```
from copy import deepcopy
from shapely import Point, set_precision, get_precision

p = set_precision(Point([0,0]), 0.1)
print(f"original precision: {get_precision(p)}")
print(f"precision after deepcopy: {get_precision(deepcopy(p))}")
```
Output
```
original precision: 0.1
precision after deepcopy: 0.0
```
I tried with LineString and Polygon and I get the same result.

I think it's discussed in this issue #1821

## Operating system
Ubuntu 22.04

Python 3.10.12
shapely==2.0.4 installed from PyPI using pip

apoorva-01:
`deepcopy` and pickle lose the precision grid size from `set_precision` (`get_precision` comes back `0.0`). The reduce methods only store WKB, and WKB doesn't carry the grid size.

So I stash `get_precision` next to the WKB and re-apply it on unpickle. Regression test in `test_misc.py`, plus a CHANGES entry. Suite's green.

One call I'd like your take on: @jorisvandenbossche noted set_precision might change the geometry, and it does for polygons. The default mode renormalizes the ring start vertex (geometry's still equal, just a different starting point) but keeps Z and M. `mode="pointwise"` avoids that but drops M, so I went with the default to not lose data. Fine as is, or would you rather have a lower-level grid-size setter?

Fixes #2073

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -26,6 +26,11 @@ coordinates in each input LineString or LinearRing (#2306).
 `Polygon.from_bounds` now builds the ring in counterclockwise order, consistent
 with `shapely.box`, so the result satisfies the GeoJSON right-hand rule (#2396).
 
+Bug fixes:
+
+- `copy.deepcopy` and pickling now preserve the precision grid size set with
+  `set_precision` (#2073).
+
 
 2.1.2 (2025-09-24)
 ------------------
--- a/shapely/geometry/base.py
+++ b/shapely/geometry/base.py
@@ -79,6 +79,14 @@ def _maybe_unpack(result):
         return result
 
 
+def _unpickle(wkb, precision=0.0):
+    geometry = shapely.from_wkb(wkb)
+    if precision:
+        # WKB does not carry the precision grid size, so restore it.
+        geometry = shapely.set_precision(geometry, precision)
+    return geometry
+
+
 class CAP_STYLE:
     """Buffer cap styles."""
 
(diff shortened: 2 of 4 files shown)


jorisvandenbossche:
> set_precision might change the geometry, and it does for polygons.

If we don't have a way to prevent that, my feeling is that we should rather accept that pickle/copy does not preserve the precision (because my feeling is that changing the coordinates, whether it is changing the order or dropping some dimension, is worse), and document that as such

fallenmi (changes requested):
The new unpickle path can erase geometries created with a non-default precision mode.

At exact base `330e9d3` and head `9a2c152`, the base/head `__reduce__` implementations were executed with:

```python
g = shapely.set_precision(
    LineString([(0.1, 0.1), (0.2, 0.2)]),
    1.0,
    mode="pointwise",
)
```

`g` is the intentionally preserved, non-empty `LINESTRING (0 0, 0 0)`. The base copy remains that same non-empty geometry (while losing the precision metadata); the head copy becomes `LINESTRING EMPTY` with precision `1.0`.

This happens because `_unpickle` re-applies the grid with the default `valid_output` mode, which removes collapsed pointwise coordinates. That turns the existing metadata-loss limitation into geometry data loss during `deepcopy`/pickle. Please preserve the WKB geometry exactly while restoring the precision model, or retain/document the old behavior if no non-mutating restoration path exists, and add regression coverage for `pointwise` and `keep_collapsed` geometries. Ruff check/format, compile, and diff checks passed on the changed files.

Reviewed with OpenAI Codex assistance.

PR #2470, opened 2026-07-24
ENH: vectorize affinity transformations

Kkkakania:
Closes #1923.

## Summary

- let `affine_transform` and `translate` transform geometry arrays in one coordinate pass
- support shaped arrays in `rotate`, `scale`, and `skew`, interpreting keyword origins independently for each geometry
- preserve array shape, missing values, and mixed 2D/3D dimensions
- document the array behavior and add focused regressions for every affinity operation

## Validation

- `PATH="/opt/homebrew/opt/geos/bin:$PATH" uv run pytest shapely/tests -q` (`6574 passed, 90 skipped, 3 xfailed`)
- `uvx pre-commit run --files shapely/affinity.py shapely/tests/legacy/test_affinity.py`
- `git diff --check`

I used an AI coding assistant while developing this contribution. I reviewed the implementation and tests, and ran the full Shapely test suite locally.

PR #2471, opened 2026-07-24
DOC: clarify transform API choices

### LINKED ISSUE #1897: DOC: clarify use of shapely.transform vs shapely.ops.transform
Documentation of [`shapely.ops.transform`](https://github.com/shapely/shapely/blob/main/docs/manual.rst?plain=1#L2038) does not reference `shapely.transform` (https://github.com/shapely/shapely/blob/main/shapely/coordinates.py#L17), and vice-versa.  Despite the clarification that `shapely.transform` is a newer addition that was ported in from `pygeos.apply`, I'm still confused as to which `transform` I should use when, where and why.

The documentation can be improved by:

- Referencing the alternate transform function from within the documentation of at least one option
- Explaining which one is preferable for some/all use cases

### Discussed in https://github.com/shapely/shapely/discussions/1895

<div type='discussions-op-text'>

<sup>Originally posted by **ijessen-mitll** September 26, 2023</sup>
The two `transform` functions seem to significantly overlap each other:

> shapely.ops.transform(func, geom)
> Applies func to all coordinates of geom and returns a new geometry of the same type from the transformed coordinates.

> shapely.transform(geometry, transformation, include_z=False)
> Returns a copy of a geometry array with a function applied to its coordinates.

There are very slight differences in their parameters, and some logic within `ops.transform` seems to try to suss out the interface of the provided transform function.  Besides that, can anyone speak to why both exist?  What are the independent use cases for each?  Can the documentation be improved to direct the use of one over the other?</div>

Kkkakania:
## Summary

Update the manual's transformation section to reflect the current Shapely 2.2 API:

- recommend `shapely.transform` for array-capable, coordinate-count-preserving transformations
- explain when `shapely.transform_coordseq` is needed
- identify `shapely.ops.transform` as the deprecated Shapely 1.x compatibility API
- replace the legacy pyproj example with the current `interleaved=False` call pattern
- remove the obsolete pyproj `<2.1` example

Closes #1897.

## Validation

- Executed both new examples locally, including the pyproj reprojection
- `uvx pre-commit run --files docs/manual.rst`
- Full Sphinx HTML build completed successfully (15 pre-existing documentation warnings)

---

*AI disclosure: this documentation update was AI-assisted. I checked the wording against the current implementations and deprecation notice, executed the examples, and built the documentation locally.*


Kkkakania:
I reviewed the failed test matrix. The GEOS-main jobs fail the same pre-existing split and make-valid expectations seen on #2470 after upstream GEOS behavior changes; the Windows fixed-GEOS failures are the existing `test_binary_op_grid_size[union]` access violation, and the macOS fixed-GEOS failures are unrelated warnings promoted to errors. This PR changes only API documentation, and its pre-commit and documentation checks pass, so there is no appropriate code change to make here.

PR #2450, opened 2026-06-16
TMP: debug windows crash with GEOS main

jorisvandenbossche:
The tests on Windows with GEOS main are crashing since a few days:

```
shapely\tests\geometry\test_geometry_base.py ........................... [  3%]
Windows fatal exception: access violation

Current thread 0x000006d4 (most recent call first):
  File "D:\a\shapely\shapely\shapely\geometry\base.py", line 748 in union
  File "D:\a\shapely\shapely\shapely\decorators.py", line 173 in wrapper
  File "D:\a\shapely\shapely\shapely\tests\geometry\test_geometry_base.py", line 105 in test_binary_op_grid_size
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\_pytest\python.py", line 167 in pytest_pyfunc_call
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\pluggy\_callers.py", line 121 in _multicall
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\pluggy\_manager.py", line 120 in _hookexec
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\pluggy\_hooks.py", line 512 in __call__
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\_pytest\python.py", line 1707 in runtest
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\_pytest\runner.py", line 184 in pytest_runtest_call
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\pluggy\_callers.py", line 121 in _multicall
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\pluggy\_manager.py", line 120 in _hookexec
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\pluggy\_hooks.py", line 512 in __call__
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\_pytest\runner.py", line 250 in <lambda>
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\_pytest\runner.py", line 361 in from_call
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\_pytest\runner.py", line 249 in call_and_report
[...]

PR #2473, opened 2026-07-27
TST: update tests for upstream changes in GEOS 3.15

jorisvandenbossche:
This:

- updates the expected value for the `make_valid` test when running with GEOS 3.15
- for now just xfails the failing split tests

Those split failures will have to be fixed "properly" by using the split functionality included in GEOS 3.15, since our custom logic has been broken by GEOS now merging linear results from an overlay operation (we currently rely on `line.difference(splitter)` resulting in a MultiLineString with the two splitted parts, but GEOS now returns this as a single LineString (https://github.com/libgeos/geos/pull/1459)). But will do that in a separate PR.

Closes https://github.com/shapely/shapely/issues/2467

PR #2474, opened 2026-07-27
CI: fix Windows segfaults for GEOS<3.15 by setting older MSVC toolchain

jorisvandenbossche:
Initially Windows was only segfaulting with GEOS main (https://github.com/shapely/shapely/pull/2450, and this has been fixed upstream in the meantime: https://github.com/libgeos/geos/issues/1449), and so the Windows build with GEOS main is now green. 
However, the issue also occurs with all other GEOS versions, and so now those are crashing as well. Awaiting bug-fix releases from the GEOS side, trying to "fix" that here by forcing an older MSVC toolset version, avoiding the buggy optimization.


jorisvandenbossche:
The other failures with GEOS main are covered by https://github.com/shapely/shapely/pull/2473

mwtoews (commented):
This is a simple workaround which is fine to apply. But a few considerations:

- I suspect `release.yml` would fail, if it weren't using cached builds, so a similar fix would be appropriate there.
- [ilammy/msvc-dev-cmd](https://github.com/ilammy/msvc-dev-cmd) appears to be unmaintained (e.g. raises "Node 20 is being deprecated" warnings that have been unaddressed upstream). A maintained fork is [TheMrMilchmann/setup-msvc-dev](https://github.com/TheMrMilchmann/setup-msvc-dev) which can be used as a mostly drop-in replacement except that it requires `arch: x64` to be specified.

jorisvandenbossche:
> I suspect release.yml would fail, if it weren't using cached builds, so a similar fix would be appropriate there.

Yeah, I know. But for now I just wanted to get green CI where it is currently failing. And for the release I hope that GEOS will actually do bug-fix releases that includes this (although adding it in the release workflow for now as well is easy to do, will push that)

jorisvandenbossche:
Going to merge this to have green CI. And thanks for the link to `TheMrMilchmann/setup-msvc-dev`, will do that in a follow-up

PR #2477, opened 2026-07-28
ENH: add top-level split() function

jorisvandenbossche:
This PR exposes the new `GEOSSplit` C API function (new in upcoming GEOS 3.15) as a top-level `shapely.split()` ufunc.

In addition, it updates our existing custom `shapely.ops.split()` function to use this GEOS version if GEOS>=3.15.  
The reason for this, apart from that it is now supported by GEOS directly, is also because another change in GEOS 3.15 actually breaks certain cases in our custom implementation (GEOS now merges linear results from an overlay operation, while we currently rely on `line.difference(splitter)` resulting in a MultiLineString with the two splitted parts, but so GEOS now returns this as a single LineString (https://github.com/libgeos/geos/pull/1459)).

There are a few behaviour differences with the new GEOS split function vs our own implementation. The ones I noticed from the tests:

- ~It is now supported to split with a (Multi)Polygon with a Multi(Polygon (taking the boundary of the splitter polygon, just as we also already do ourselves in case of splitting LineString input with a polygon)~ -> aligned that in [#2505](https://github.com/shapely/shapely/pull/2505) to also allow this in our own implementation.
- It is now allowed to have overlap between the input geometry and splitter (our custom implementation raises an error for that). The overlapping segment is split as a separate part in the result.

Some questions:

- The top-level `shapely.split()` will be a new function in shapely 2.2, but will still only work if you have shapely with GEOS 3.15. Since we have a custom implementation already, I could also add a simple fallback in `shapely.split()` to still use that for older GEOS versions, such that `shapely.split()` is available always regardless of the GEOS version you are running with.
  - We have a similar fallback for eg `shapely.minimum_rotated_rectangle` and `shapely.orient_polygons`
- [minor] I currently put it in the `constructive.py` file, but I could also put it in `set_operations.py` (just a matter of where it gets documented in the reference) (`geosop` lists it as one of the overlay functions)

Note: we have a bunch of [open issues](https://github.com/shapely/shapely/issues?q=is%3Aissue%20state%3Aopen%20label%3Asplitting) related to `split`. I am planning to go through them and check which are already fixed by the new GEOS version (or potentially report upstream), and then add tests to close those (maybe as a follow-up PR)


shapely/tests/test_constructive.py
```diff
@@ -1321,6 +1323,452 @@ def test_orient_polygons_array_like():
     assert_geometries_equal(np.asarray(actual), expected)
 
 
+@pytest.mark.skipif(shapely.geos_version < (3, 15, 0), reason="GEOS < 3.15")
```
jorisvandenbossche, line 1329: All the tests that I added here, is essentially a translation of the existing `legacy/test_split.py` tests. Once we don't have the custom implementation anymore for older GEOS versions, the legacy tests can just be removed.

(will add a note about that in the legacy file)

shapely/tests/test_constructive.py
```diff
@@ -1321,6 +1323,452 @@ def test_orient_polygons_array_like():
...
+def test_split_unsupported_geometry_type():
+    msg = "Splitting a Polygon with a point is not supported"
+    with pytest.raises(GEOSException, match=msg):
+        shapely.split(polygon, point)
+
+    with pytest.raises(GEOSException, match=msg):
+        shapely.split(polygon, multi_point)
+
+    # Point/MultiPoint seems supported, but this is essentially a no-op
+    # but also returning Point instead of GeometryCollection
```
jorisvandenbossche: This seems wrong, reported upstream -> https://github.com/libgeos/geos/issues/1490

jorisvandenbossche:
In the last commit, I added support for `shapely.split()` also for older GEOS versions, by using our own implementation (which is used for `shapely.ops.split()` on those older versions). That ensures the new top-level function can be used regardless of the GEOS version, and this addresses the first question from the top post:

> * The top-level `shapely.split()` will be a new function in shapely 2.2, but will still only work if you have shapely with GEOS 3.15. Since we have a custom implementation already, I could also add a simple fallback in `shapely.split()` to still use that for older GEOS versions, such that `shapely.split()` is available always regardless of the GEOS version you are running with.
> 
>     * We have a similar fallback for eg `shapely.minimum_rotated_rectangle` and `shapely.orient_polygons`

But I can easily revert that last commit if we want to keep it more minimal.

If we keep it, I can add a brief note about the differences in behaviour depending on the GEOS version in the docstring (i.e. with newer GEOS, it supports splitting overlapping segments and splitting Polygons with Polygons (by taking the rings))

sgillies:
@jorisvandenbossche Hyrum's Law says that somebody out there depends on the different behavior of `shapely.ops.split()`. We eventually want them to switch to the new function with the GEOS behavior and semantics. I think it would be okay to deprecate `shapely.ops.split()` and mark its bugs as "won't fix", or effectively fixed by switching to `shapely.split()`.

jorisvandenbossche:
@sgillies I understand what you say in general, but not what you are trying to suggest with that concretely for which action to take in this PR.

First, for people relying on `shapely.ops.split()`, there will be cases that work fine now (with GEOS < 3.15) that will be broken when using the same shapely code with GEOS 3.15 (because of a behaviour change in GEOS, a behaviour which we relied upon in the logic inside `shapely.ops.split()`. See also the tests that are currently skipped with https://github.com/shapely/shapely/pull/2473).  
(and while we can control including GEOS 3.14 in the wheels for the upcoming release, people can have custom installs, and also on conda-forge shapely will likely be updated to build against GEOS 3.15 in the near future (while in theory we could block such update, that is annoying since it would block migration of the ecosystem in general to the latest GEOS))

So there are indeed differences in behaviour between our `shapely.ops.split()` and GEOS-based `split()`, and so I understand that for that reason we might want to keep them separate. But I think those differences in behaviour are rather minor (it's mostly improvements, i.e. cases we explicitly don't support (with clear error messages) in our python version, which GEOS does support), and certainly more minor that the buggy behaviour `shapely.ops.split()` will get by _not_ using GEOS' split under the hood for GEOS 3.15.

jorisvandenbossche:
Just to illustrate the bug that existing users of `shapely.ops.split()` will run into, the following basic "split line by another intersecting line" stops working silently:

```python
>>> import shapely
>>> shapely.ops.split(shapely.LineString([(0, 0), (1, 1)]), shapely.LineString([(0, 1), (1, 0)]))
# return value with GEOS 3.14 or below -> correctly split into two parts
<GEOMETRYCOLLECTION (LINESTRING (0 0, 0.5 0.5), LINESTRING (0.5 0.5, 1 1))>
# return value with GEOS 3.15 -> no split (the intersection point is only added to the line)
<GEOMETRYCOLLECTION (LINESTRING (0 0, 0.5 0.5, 1 1))>
```

So it is not just about marking existing bugs as won't-fix. Users would silently get wrong behaviour for something that worked fine before (if they are using GEOS 3.15 and we don't update `shapely.ops.split()` to use GEOSSplit for GEOS>=3.15)

sgillies:
@jorisvandenbossche I overlooked that shapely's old splitter was using GEOSDifference. Your fallback seems best in this case.

I'll add a few more comments inline.

src/shapely/tests/legacy/test_split.py
```diff
@@ -106,12 +106,19 @@ def test_split_poly_with_multiline(self):
         self.helper(self.poly_hole, splitter, 2)
 
     def test_split_poly_with_other(self):
-        with pytest.raises(GeometryTypeError):
+        error = GeometryTypeError if geos_version < (3, 15, 0) else GEOSException
```
sgillies, line 109: I'd prefer that `split()` continue to raise GeometryTypeError in these cases. A "raw" GEOS exception is a downgrade.

The complication is that the GEOS C API throws away exception categories and the C++ info about `geos::util::IllegalArgumentException` is lost, so we'd need to parse GEOSException messages. The good news is that they are fairly normalized.

src/shapely/tests/legacy/test_split.py
```diff
@@ -106,12 +106,19 @@ def test_split_poly_with_multiline(self):
...
+
+        with pytest.raises(error):
             split(self.poly_simple, Point(1, 1))
-        with pytest.raises(GeometryTypeError):
+        with pytest.raises(error):
             split(self.poly_simple, MultiPoint([(1, 1), (3, 4)]))
-        with pytest.raises(GeometryTypeError):
-            split(self.poly_simple, self.poly_hole)
+
+        if geos_version < (3, 15, 0):
```
sgillies, line 116: Can we lift the version check out of the test? By using a pair of params?

src/shapely/tests/test_constructive.py
```diff
@@ -1323,6 +1325,504 @@ def test_orient_polygons_array_like():
...
+            poly_hole,
+            Polygon([(0.2, 3), (0.2, -3), (1.7, -3), (1.7, 3)]),
+            3,
+            marks=pytest.mark.xfail(
+                shapely.geos_version < (3, 15, 0),
+                reason="GEOS 3.15 required for splitting overlapping input",
+            ),
+        ),
+    ],
+)
```
sgillies, line 1755: How did you come up with these parameters? Is any of this generated code? Or are they from the GEOS test suite?

jorisvandenbossche: Nope, all manual work ;) They are coming from our own test suite. See my comment at https://github.com/shapely/shapely/pull/2477#discussion_r3665639580 that I added as clarification (but because of my latest commit, I suppose that doesn't show up anymore when looking at the diff by being outdated ..).

I essentially converted our existing legacy tests to tests for the new `split()` function (but with some more guarantees, i.e. for some simple test cases I am now checking the actual result, not just that it roundtrips)

(it might be debatable whether it is more readable as the old tests (although I like it), but it at least matches better with the rest of the file here)

src/shapely/tests/legacy/test_split.py
jorisvandenbossche: Yeah, the error messages are at least all good in those cases here. 
(in the new `test_split_unsupported_geometry_type` test I added in the other file, I included error messages in the `pytest.raises`, so we check that GEOS keeps giving a decent error mesage)

For `shapely.ops.split()`, I could move some of the checks more up front so it is before we dispatch to the GEOS split (for GEOS>=3.15), but for the new top-level `shapely.split()`, I would prefer to just go with what GEOS gives us (doing those checks here is also harder given that the inputs do not need to be just scalar geometries, and thus if we would add checks, those would ideally happen in the C code)

sgillies: Yeah, maybe my itch is better scratched in the GEOS C API. I started asking at https://github.com/libgeos/geos/issues/1529.

Still, Shapely is (has always aspired to be) more than only the most obvious wrapper around GEOS, and that applies to exceptions. We should do better than the GEOS C API when we can.

jorisvandenbossche: One problem here is that the GEOS exception type (which is indeed included in the error message of what we now raise as `GEOSException`) is still quite crude. 
For example, here a `IllegalArgumentException` is raised for the case of passing unsupported geometry types, e.g. splitting a polygon with a point gives "IllegalArgumentException: Splitting a Polygon with a point is not supported.".

But it is not that we can just translate an "IllegalArgumentException" `GEOSException` to a `GeometryTypeError`, because GEOS also uses this for non-geometry input errors. For example:

```
>>> shapely.segmentize(shapely.LineString([(0, 0), (1, 1)]), max_segment_length=-1)
...
GEOSException: IllegalArgumentException: Tolerance must be positive
```

(of course, in this specific example, we could also rather easily check this up front and give a better error message, but this is just one example, and the question is if we should start repeating all input validation that GEOS also does)

src/shapely/tests/legacy/test_split.py
jorisvandenbossche: Given that the tested code is a bit different (calling `split()` vs calling `self.helper`) that seems a bit complicated to do as parametrization. I could split the test in one for pre-geos 3.15 and one for geos-3.15+, each with the appropriate skips, if you would prefer.

jorisvandenbossche:
Regarding one of the (explicit) behaviour differences between our version and GEOSSplit, i.e. allowing to split a polygon by a polygon, I think we could easily make that difference go away by also allowing that on our end. Made a small PR to do that: https://github.com/shapely/shapely/pull/2505

src/shapely/tests/legacy/test_split.py
sgillies: Thats true about the complication. In rasterio I'm planning to eliminate GDAL version checks within test functions to make the version cases selectable using `pytest -k` https://github.com/rasterio/rasterio/issues/3627. But I don't object to your approach for this PR. Changes could be made later.

src/shapely/tests/legacy/test_split.py
sgillies: No, I don't want to repeat the validation.

Giving up the specific `GeometryTypeError`, but translating `GEOSException("IllegalArgumentException: Tolerance must be positive")` to a general `ValueError("Tolerance must be positive")` could be a good compromise and still very useful. It's nice to get `ValueError` when the function can't be executed only because of improper input values, not because of some small topological incorrectness or algorithmic issue.

I'm starting a new job this week, but will try to find the time to look into our low-level exception handling to see if there's a not terrible solution (in another PR).

sgillies (approved):

PR #2478, opened 2026-07-28
TST: run doctests with latest GEOS 3.14

jorisvandenbossche:
Moving the doctests (and "error on warnings") to the build with the latest releases, i.e. using GEOS 3.14


mwtoews (approved):
Changes look good, thanks!

PR #2479, opened 2026-07-28
TST: ignore 'invalid value' warnings for multipoint/coverage/shared_path on MacOS

jorisvandenbossche:
After https://github.com/shapely/shapely/pull/2473 and https://github.com/shapely/shapely/pull/2474, this should be the last PR to get CI green again.

This suppresses a few "RuntimeWarning: invalid value encountered in .." warnings, that have been showing up only on MacOS.
They happen since end of June for GEOS main, and since this week with released GEOS >= 3.12, so probably not related to a change in GEOS, but "uncovered" by a toolchain update (e.g. I am seeing GEOS being built with AppleClang 21.0.0.21000101 on the recent builds, and with AppleClang 17.0.0.17000013 a month ago, i.e. macos-26-arm64 vs macos-15-arm64 image).

I can't reproduce locally, so testing with CI which ones need to be ignored.


jorisvandenbossche:
Reported upstream: https://github.com/libgeos/geos/issues/1492

mwtoews (approved):
Appreciate your efforts to finding these fixes with macOS via CI! I wonder if it would be worthwhile creating a macOS-specific `ignore_invalid_on_darwin` method to import.

jorisvandenbossche:
> I wonder if it would be worthwhile creating a macOS-specific `ignore_invalid_on_darwin` method to import.

We also have a bunch that do `with ignore_invalid(shapely.geos_version < (3, 12, 0))`. Let's see if more of those turn up, we can always add a shortcut for it later.

PR #2480, opened 2026-07-28
BUG: forward **kwargs in line_interpolate_point and line_locate_point

uttam12331:
### Problem

`line_interpolate_point` and `line_locate_point` both accept `**kwargs` and document them —

```
**kwargs
    See :ref:`NumPy ufunc docs <ufuncs.kwargs>` for other keyword arguments.
```

— but the bodies never forward `**kwargs` to the underlying ufunc, so ufunc keyword arguments such as `out=` and `where=` are silently ignored:

```python
import numpy as np, shapely
from shapely import LineString

line = LineString([(0, 2), (0, 10)])
out = np.empty(2, dtype=object)
r = shapely.line_interpolate_point([line, line], [2.0, 4.0], out=out)

out          # [None, None]   -- expected [<POINT (0 4)>, <POINT (0 6)>]
out is r      # False          -- expected True
```

For comparison, sibling functions honor `out=` correctly (e.g. `shapely.distance([...], [...], out=out)` fills and returns `out`). A bogus keyword is also swallowed silently instead of raising `TypeError`, confirming the kwargs reach the function and are then dropped.

### Cause

```python
# line_interpolate_point
if normalized:
    return lib.line_interpolate_point_normalized(line, distance)   # **kwargs dropped
else:
    return lib.line_interpolate_point(line, distance)              # **kwargs dropped
```

Every other ufunc wrapper in `linear.py` (`line_merge`, `shared_paths`, `shortest_line`) and the two-branch functions in `measurement.py` (`hausdorff_distance`, `frechet_distance`) already forward `**kwargs`; only these two linear-referencing functions drop them. The `multithreading_enabled` decorator also special-cases `out`/`where` in kwargs, so the framework expects them to flow through.

### Fix

Forward `**kwargs` in both branches of both functions:

```python
if normalized:
    return lib.line_interpolate_point_normalized(line, distance, **kwargs)
else:
    return lib.line_interpolate_point(line, distance, **kwargs)
```

When `kwargs` is empty (the default), `f(a, b, **{})` is identical to `f(a, b)`, so existing calls are unaffected.

### Tests

Added `test_line_interpolate_point_out` and `test_line_locate_point_out` (both `normalized` values), modeled on the existing `test_unary_with_kwargs` pattern — they assert `func(..., out=out) is out`. Both fail on `main` and pass with the fix; the existing `test_linear.py` suite still passes.


jorisvandenbossche:
@uttam12331 thanks for the PR. Can you merge latest main? (that should resolve the CI failures)

PR #2481, opened 2026-07-29
Add actionlint to catch GitHub Actions bugs early

chernojagne:
Adds actionlint static analysis for the three GitHub Actions workflow files in this repo.

**What this adds:**
- `.github/actionlint.yaml` — configuration: no self-hosted runners, ShellCheck enabled with four common false-positive rules suppressed (`SC2086`, `SC2129`, `SC2155`, `SC1091` — each noisy due to `${{ }}` interpolation in workflow run blocks)
- `.github/workflows/actionlint.yml` — dedicated workflow triggered on changes to `.github/workflows/**`, runs actionlint via the upstream download script
- `.pre-commit-config.yaml` — adds the `actionlint` pre-commit hook (pinned `v1.7.7`) for local feedback before push

**Why:**
Actionlint catches typos in step names, invalid expressions, missing permissions, and ShellCheck issues that otherwise only surface as broken CI runs. The repository's 204-line multi-OS/multi-Python/multi-GEOS matrix (`tests.yml`) and multi-arch release workflow (`release.yml`) make early detection particularly valuable.

This change was generated by [Autar](https://autar.ai) and reviewed by a human before submission.

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -1,0 +1,34 @@
+name: Actionlint
+
+on:
+  push:
+    paths:
+      - '.github/workflows/**'
+      - '.github/actionlint.yaml'
+      - '.github/actionlint.yml'
+  pull_request:
+    paths:
+      - '.github/workflows/**'
+      - '.github/actionlint.yaml'
+      - '.github/actionlint.yml'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: actionlint-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  actionlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+      - name: Download actionlint
+        id: download
+        run: bash <(curl -sSf https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash)
+        shell: bash
+      - name: Run actionlint
+        run: |
+          ./actionlint -color
+        shell: bash
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -23,12 +23,5 @@ jobs:
   actionlint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
-      - name: Download actionlint
-        id: download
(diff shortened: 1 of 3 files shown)


chernojagne:
The fork CI (which runs immediately for first-time contributors) shows failing Windows test jobs (`windows-latest Py3.11 GEOS 3.10.7`, etc.), but these failures appear on upstream's `main` branch itself ([e.g. run 30429428782](https://github.com/shapely/shapely/actions/runs/30429428782)) and predate this PR by several commits.

Our change only adds `.github/actionlint.yaml`, `.github/workflows/actionlint.yml`, and a pre-commit hook entry — none of which touch Python source, tests, or GEOS builds. Happy to investigate further if useful, but this looks like a pre-existing Windows/GEOS issue unrelated to this contribution.

.github/workflows/actionlint.yml
```diff
@@ -0,0 +1,34 @@
...
+  cancel-in-progress: true
+
+jobs:
+  actionlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+      - name: Download actionlint
+        id: download
+        run: bash <(curl -sSf https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash)
```
martinfleis: This is probably the most dangerous line of GHA code I've seen in a while.

chernojagne: You're right — thank you for catching it. `bash <(curl .../main/...)` fetches and executes unversioned remote code, which is a supply-chain risk.

I've replaced the two steps with the official `rhysd/actionlint` action pinned to `v1.7.12`, which installs the binary safely and auto-discovers `.github/actionlint.yaml`. Also updated `actions/checkout` to `@v7` to match the rest of the repo.

Fix: 16fe272

chernojagne:
Thanks again for flagging the install step earlier — you were right to call it out. I've since switched it to the pinned official action, but rather than keep this PR lingering I'd rather close it and fix the tooling that generated it at the source (pinned actions only, no dynamic installers). I'll come back with something clean if it's ever useful. Apologies for the noise, and thanks for taking the time to review.

PR #2484, opened 2026-07-30
CI: cleanup build setup workarounds for freethreading

jorisvandenbossche:
Clean up some workarounds that were added in https://github.com/shapely/shapely/pull/2138 to support free-threading, when we still needed to install a pre-release cython version (which is no longer the case for some time)


mwtoews (approved):
Changes look good, and skimming a few CI logs for "Build wheels" I don't see any errors or related warnings.

PR #2485, opened 2026-08-02
DOC: cross-link geometry classes and their bulk constructors (#1925)

### LINKED ISSUE #1925: Include See also in docstrings to link between Polygon and polygons etc

JayeshSuryavanshi:
Closes #1925.

The geometry classes and their array-based (bulk) constructors were not cross-linked in the docs. As reported in #1925, a user missed `shapely.polygons` and wrote a Python loop over `Polygon` instead, because the `Polygon` docstring never mentioned `polygons`. @jorisvandenbossche agreed the docstrings could be much better linked.

This adds a numpydoc `See Also` section to every class and its bulk constructor:

| class | constructor |
| --- | --- |
| `Point` | `points` |
| `LineString` | `linestrings` |
| `LinearRing` | `linearrings` |
| `Polygon` | `polygons` |
| `MultiPoint` | `multipoints` |
| `MultiLineString` | `multilinestrings` |
| `MultiPolygon` | `multipolygons` |
| `GeometryCollection` | `geometrycollections` |

Each class now points to its bulk constructor, and each constructor points back to its class (extending the existing `See Also` block where one was already present). Documentation only, no behaviour change, and `ruff` is clean.


jorisvandenbossche:
There are already open PRs for the issue you link, so therefore closing this one

PR #2486, opened 2026-08-03
CI: replace ilammy/msvc-dev-cmd action with TheMrMilchmann/setup-msvc-dev

jorisvandenbossche:
xref https://github.com/shapely/shapely/pull/2474#pullrequestreview-4839815436 from @mwtoews:

> [ilammy/msvc-dev-cmd](https://github.com/ilammy/msvc-dev-cmd) appears to be unmaintained (e.g. raises "Node 20 is being deprecated" warnings that have been unaddressed upstream). A maintained fork is [TheMrMilchmann/setup-msvc-dev](https://github.com/TheMrMilchmann/setup-msvc-dev) which can be used as a mostly drop-in replacement except that it requires `arch: x64` to be specified.


mwtoews (approved):
Just need to add `arch: x64` for `test.yml` (already specified for `release.yml`)

jorisvandenbossche:
(I thought to first test it without as the docs seemed to indicate it is optional, but so actually it explicitly says it is not when specifying toolset ..)

PR #2487, opened 2026-08-04
BLD/CI: add 'build' dependency group, modernize ci/install_geos.* scripts

mwtoews:
This adds a `build` dependency group to install all the tools required to build both GEOS and Shapely. This is used for both CI tests and with cibuildwheel. It also will allow `pip install --no-build-isolation` to be used, only done for cibuildwheel in this PR.

Also simplify the ci/install_geos.* scripts to use modern CMake -S source and -B build directory options. Always disable build testing, which requires GEOS 3.9 or later.

Hopefully addresses https://github.com/shapely/shapely/pull/2456#issuecomment-4776793243


jorisvandenbossche:
> It also will allow `pip install --no-build-isolation` to be used, only done for cibuildwheel in this PR.

What is the reason for enabling no-build-isolation when building wheels? While I use it constantly for development, my feeling is that for building distributions it is good practice to use the standard mechanism

(the rest of the PR looks good!)

mwtoews:
Ok, I see this cibuildwheel arg option is not necessary, and I see the value of the standard mechanism for building wheels. Nevertheless, the build group seems to work with `--no-build-isolation` in tests.yml.

jorisvandenbossche (approved):

jorisvandenbossche:
> Nevertheless, the build group seems to work with `--no-build-isolation` in tests.yml.

Yes, that is a nice simplification!

PR #2488, opened 2026-08-05
CI/RLS: publish workflow to use trusted publishing + update github release action

### LINKED ISSUE #2342: Switch to trusted publishing for package upload to PyPI in CI
Trusted publishing (with attestations) means I have high confidence that what I download from PyPI is the same artefact which was generated in GitHub CI, meaning that what I see in GitHub is the same as what is installed - handy for auditing.

See [the Python packaging documentation](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#configuring-trusted-publishing), [the PyPI documentation](https://docs.pypi.org/trusted-publishers/), and [the official pypi-publish GitHub action documentation](https://github.com/pypa/gh-action-pypi-publish?tab=readme-ov-file#trusted-publishing) on trusted publishing.

<details><summary>Implementation (click to expand)</summary>

* Configure (or use an existing) GitHub environment, and register with PyPI
* Add the [environment definition](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idenvironment) to the same `publish` job
* Add `id-token: write`, `contents: write` and `packages: write` [permissions](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idpermissions) to the same `publish` job
  * specifying any permission restricts the token, and creating a GitHub release requires `contents: write` and potentially `packages: write`
* Remove `user` and `password` arguments in the "Upload Release Assets to PyPI" step of the `publish` job of the [Build and publish](https://github.com/shapely/shapely/blob/main/.github/workflows/release.yml) CI workflow
* Optionally remove the `pypi_password` project secret

</details>

jorisvandenbossche:
Closes https://github.com/shapely/shapely/issues/2342 to use the PyPI trusted publishing setup. 

And also updates the action to create the github release (`actions/create-release` was archived for 5 years, `softprops/action-gh-release` is what we have been using for other projects as well)


.github/workflows/release.yml
```diff
@@ -238,37 +245,12 @@ jobs:
           path: dist
 
       - name: Create GitHub Release
-        id: create_release
-        uses: actions/create-release@v1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # This token is provided by Actions, you do not need to create your own token
+        uses: softprops/action-gh-release@v3
```
sgillies, line 248: @jorisvandenbossche who or what is softprops? Being new to our publishing stream, how is the software and its authors vetted?

jorisvandenbossche: I don't know the person (https://github.com/softprops), but the action (https://github.com/softprops/action-gh-release) is, to the extent this is somewhat an indirect vetting, quite widely used: 5.7k github stars (and the most popular action from the 4 alternatives that are listed in the deprecated https://github.com/actions/create-release), and I have personally used it (or seen used) in various other projects for several years (GEOS, GDAL, geopandas, pyogrio)

Nowadays it might be that this could be rather easily replaced with some `gh` cli call, though. I could take a look at that if we would prefer that.

jorisvandenbossche: Based on a quick Google AI question (without check), this would be something like:

```
- name: Create Release
  env:
    GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
  run: |
    gh release create ${{ github.ref_name }} dist/*.tar.gz
```

sgillies: I'm happy to pick up the work of implementing our own action. Meanwhile, I approve this.

sgillies (approved):

.github/workflows/release.yml
jorisvandenbossche: -> tracked as https://github.com/shapely/shapely/issues/2489

PR #2492, opened 2026-08-11
CI/TST: test with Python 3.15

jorisvandenbossche:
Now that the upcoming Python 3.15 is in the RC phase, it's a good time to start testing it (https://github.com/shapely/shapely/pull/2491 also starts building wheels)

Like https://github.com/shapely/shapely/pull/2330 / https://github.com/shapely/shapely/pull/2331 for Python 3.14


shapely/tests/legacy/test_wkb.py
```diff
@@ -149,7 +149,7 @@ def test_dump_binary_load_hex(some_point, tmpdir):
 
     # TODO(shapely-2.0) on windows this doesn't seem to error with pygeos,
     # but you get back a point with garbage coordinates
-    if sys.platform == "win32":
+    if sys.platform == "win32" and sys.version_info < (3, 15):
```
jorisvandenbossche: I think with the changes in Python 3.15 about using UTF-8 by default (https://docs.python.org/3.15/whatsnew/3.15.html#whatsnew315-utf8-default), this is now consistently failing also on Windows. Which makes me wonder if we can simplify this by forcing the encoding always.

mwtoews: Agree that this Python 3.15 change is important to note here, and removing L150-L157 (0ce19f5cc1de909083b3ec2bf273e306a9d57906) is the correct change.

mwtoews (approved):

sgillies (approved):

PR #2493, opened 2026-08-19
MNT: Move pytest configuration from setup.cfg to pyproject.toml

mwtoews:
This moves the pytest configuration from `setup.cfg` to `pyproject.toml`, as supported since [version 6](https://docs.pytest.org/en/8.4.x/reference/customize.html#pyproject-toml) under `[tool.pytest.ini_options]`. No options are changed, just translated. The motivation is to clear out and remove `setup.cfg`.

An upgrade to require pytest>=9 with options slightly re-written under `[tool.pytest]` can happen at a later time.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,6 +98,10 @@ ignore = ["D000", "D004"]
 allow-long-titles = true
 max-line-length = 79
 
+[tool.pytest.ini_options]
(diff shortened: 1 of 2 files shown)


jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

PR #2496, opened 2026-09-01
BLD: harden Windows ARM64 wheel releases

yeelam-gordon:
## Summary

- authenticate and SHA-256 verify GEOS before extraction or cache reuse
- acquire matching native Windows ARM64 CPython runtimes and verify repaired wheels across `cp311` through `cp315t`
- retain producer/verifier provenance and gate publication on finalized evidence
- make PyPI staging and GitHub Release publication fail-closed and safely rerunnable

## Why

Shapely already has a `windows-11-arm` build lane, but a green build alone does not prove wheel tags, native PE architecture, repaired GEOS dependencies, licenses, installed-wheel tests, or publication provenance. This change makes those checks explicit and prevents publication when native evidence is incomplete.

## Scope

This PR contains one squashed commit and only nine release/CI implementation and test files. Fork-only hackathon documentation and demo assets are intentionally excluded.

## Local validation

- `python -m py_compile ci/arm64_python_runtime.py ci/wheel_evidence.py ci/release_transaction.py`
- `python -m pytest ci/tests/test_arm64_python_runtime.py ci/tests/test_install_geos.py ci/tests/test_release_transaction.py ci/tests/test_wheel_evidence.py` — 62 passed
- workflow YAML and release-invariant checks passed
- live `cp315` and `cp315t` discovery resolved the current official `3.15.0rc1` ARM64 manifest

The local host was x64. This draft should remain unmerged until the PR workflow completes its non-publishing native `windows-11-arm` producer, all seven verifier jobs, and evidence finalization. `nightly_upload`, `publish`, Anaconda upload, PyPI upload, and GitHub Release mutation must remain skipped on the pull-request event.

## Ownership boundary

This is the pre-step repository change. Merge, tag creation, package publication, and any public ARM64 support claim remain maintainer-owned post-steps. An official support claim should follow a package-owned release that publishes verified `win_arm64` wheels.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,6 +20,8 @@ on:
 
 env:
   GEOS_VERSION: "3.13.1"
+  GEOS_SHA256: "df2c50503295f325e7c8d7b783aca8ba4773919cde984193850cf9e361dfd28c"
+  GEOS_CHECKSUM_PROVENANCE_URL: "https://github.com/libgeos/geos/releases/tag/3.13.1"
 
 jobs:
   build_sdist:
@@ -162,7 +164,7 @@ jobs:
         uses: actions/cache@v6
         with:
           path: ${{ runner.temp }}/geos-${{ env.GEOS_VERSION }}
-          key: ${{ matrix.os }}-${{ matrix.arch }}-${{ env.GEOS_VERSION }}-${{ hashFiles('ci/*') }}
+          key: ${{ matrix.os }}-${{ matrix.arch }}-${{ env.GEOS_VERSION }}-${{ env.GEOS_SHA256 }}-${{ hashFiles('ci/*') }}
 
       - name: Add GEOS LICENSE
         run: cp ci/wheelbuilder/LICENSE_GEOS .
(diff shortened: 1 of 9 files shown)


jorisvandenbossche:
@yeelam-gordon you dump a 4000+ lines (!) change on us, that will require a bit more rationale of why you think this is needed, why the standard cibuildwheel and delve/auditwheel processes are not enough, and why this is specifically needed only for Windows ARM64.

Even if there are parts of this PR that are useful, it will have to come in much more bite-sized, targeted changes. At the moment I am not even going to "approve workflows to run" as I would have no idea what I am approving

sgillies:
I'm with @jorisvandenbossche . I don't think it's Shapely's responsibility to be on the forefront of making Windows packaging better and safer. I prefer that this project follow practices that are already established by, for example, the Python Packaging Authority or Numpy.

yeelam-gordon:
Closing this draft because the latest scope review showed it is not a minimal Arm64-enablement PR.

Current `shapely/main` already has a native `windows-11-arm` release lane, `CIBW_ARCHS=ARM64`, and cibuildwheel-installed wheel tests via `test-command = "pytest --pyargs shapely.tests"`. This PR added a large evidence/provenance/release-transaction layer around an already-supported native path, which is not the right upstream change for claiming Arm64 support.

If maintainers want a future contribution here, it should be a much smaller, separately justified change (for example Windows GEOS download hardening), not this bundled release-logic rewrite.

jorisvandenbossche:
> If maintainers want a future contribution here, it should be a much smaller, separately justified change (for example Windows GEOS download hardening), not this bundled release-logic rewrite.

It is good to hear that your LLM agent has understood our concern 🙃

PR #2499, opened 2026-09-08
MAINT: bump minimum numpy version to 1.26

jorisvandenbossche:
NumPy 1.26.0 is 3 years old (Sep 16, 2023), so thought to directly bump to that instead of 1.25 (only a few months older).

Since we are about to release, I would keep numpy 1.x support now (it doesn't really cost us anything, I also didn't find anything to clean up). But then after that, for shapely 2.3.x, I think we could drop support for numpy 1.x.


pyproject.toml
mwtoews: Here too?

jorisvandenbossche: Ah, yes, updated (I shouldn't actually matter in practice, since binaries built with 1.25 run fine on 1.26, but probably not a reason to not keep this minimum version consistent)

mwtoews (approved):
👍

Issue #2413, opened 2026-01-29
2.2.0 release


jorisvandenbossche:
Issue to track the 2.2.0 release (2.1.0 was released in May 2025).

Main changes for an upcoming release we already have or are ongoing:

- GEOS updated to 3.14 in the wheels
- Python support bumped to >=3.11 (dropped Python 3.10) and GEOS to >=3.10 (dropped 3.9)
- Improved performance, i.e. all work related to the refactor Casper is doing ([#2339](https://github.com/shapely/shapely/pull/2339) and linked PRs)
- New methods: `minimum_witdh`, `get_segments`, `transform_coordseq`
- Improved M support for creation functions?

Other issues/PRs on the milestone: https://github.com/shapely/shapely/milestone/59

See https://github.com/shapely/shapely/issues/1570 for some GEOS functions that can still be wrapped for new methods (GEOSCoverageCleanWithParams from GEOS 3.14 would be interesting)

cc @shapely/shapely-team

theroggy:
It would indeed be great if `GEOSCoverageCleanWithParams` would be exposed...

MugundanMCW:
Hi Team,

Could you please share the ETA for the next release?

Thanks

chourmo:
Having GEOSGeomGetEndPoint and GEOSGeomGetStartPoint would be super great for linestring work. Even better with a GetStartEndPoint !

jorisvandenbossche:
We should certainly do a release in the coming week(s), so we have a wheel for Python 3.15 and to ensure our released tests run with latest GEOS 3.15.

@shapely/shapely-team any preferences between:

- Do a quick 2.1.3 release with Python 3.15 support (and test fixes for GEOS 3.15), and take a bit more time for finalizing shapely 2.2.0
  - (I don't think it should be much work to get a 2.1.3 ready, and I could do that effectively today, or let's say this week)
- Release 2.2.0 essentially as is, e.g. also this week or next week
  - We have more than enough new stuff to do a release, I would only want to ensure to merge https://github.com/shapely/shapely/pull/2464 and https://github.com/shapely/shapely/pull/2477 (or at least the `shapely.ops.split` part of that PR, not necessarily the top-level function)
  - One question here is which GEOS version to include in the wheels. We already bumped that from 3.13 to 3.14, but in theory we could also bump it further to the just released 3.15 (I think my preference here is to stick with 3.14, but try to have a quicker shapely 2.3 (not another year) that then includes 3.15)

[...]

PR #2500, opened 2026-09-08
DOC/RLS: update changelog notes for 2.2.0 release

jorisvandenbossche:
Start release notes to work towards a 2.2.0 release (https://github.com/shapely/shapely/issues/2413)

Still have to clean-up some sections and still have to consolidate this with the content in CHANGES.txt (for now I just reused the content from there)

### DIFF
--- a/docs/release/2.x.rst
+++ b/docs/release/2.x.rst
@@ -1,6 +1,106 @@
 Version 2.x
 ===========
 
+
+.. _version-2-2-0:
+
+Version 2.2.0 (2026-09-xx)
+--------------------------
+
+New features
+^^^^^^^^^^^^
+
+Performance improvements for scalar function calls
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+While the shapely 2.0 release provided significant performance improvements for
+vectorized operations on arrays of geometries (through the new top-level
+universal functions), this came at the cost of the performance of scalar
+function calls (#2035, #2339).
+
+TODO add benchmarks example of creating point and calling some attribute
+
+Improved performance of the following geometry methods: `area`, `length`,
+`minimum_clearance`, `x`, `y`, `z`, `m`, `has_z`, `has_m`, `is_empty`,
+`is_ring`, `is_closed`, `is_simple`, `geom_type`, `__len__`, `_ndim`.
+Benchmarks showed a change of call duration from 2.6 microseconds to 0.2
+microseconds (on the `Point.x` attribute).
+
+Improved performance by changing the GEOS context handling. Instead of
+initializing/finalizing on every call, GEOS contexts are reused through a
+threadlocal variable. It should give a performance increase of about
+75 nanoseconds per call.
+
+New functions
+~~~~~~~~~~~~~
+
+Several new functions from GEOS are now exposed in Shapely as top-level
+vectorized functions:
+
+- :func:`.minimum_width` to compute the minimum width (minimum diameter)
+  of a geometry (#2313).
+- :func:`.coverage_clean`, which returns a cleaned version of geometries
+  forming a polygonal coverage (#2454)  (requires GEOS >= 3.10).
+
+In addition, two other functions have been added to the top-level namespace:
+
+- :func:`.transform_coordseq`, which applies a transformation function per
+  coordinate sequence in a more performant way than
+  :func:`shapely.ops.transform` does (which is now deprecated, see below)
+  (#1902).
+- :func:`.get_segments`, which returns individual linear features from all
+  pairwise coordinates in each input LineString or LinearRing (#2306).
+
+Other improvements
+~~~~~~~~~~~~~~~~~~
+
+- The :func:`.relate` and :func:`.relate_pattern` functions now support
+  prepared geometries, i.e. they will use a faster implementation if you ensure
+  the left input geometry has been prepared first using :func:`.prepare`, which
+  can be beneficial when testing that geometry against multiple other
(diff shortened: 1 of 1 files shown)


docs/release/2.x.rst
```diff
@@ -1,6 +1,106 @@
...
+Improved performance of the following geometry methods: `area`, `length`,
+`minimum_clearance`, `x`, `y`, `z`, `m`, `has_z`, `has_m`, `is_empty`,
+`is_ring`, `is_closed`, `is_simple`, `geom_type`, `__len__`, `_ndim`.
+Benchmarks showed a change of call duration from 2.6 microseconds to 0.2
+microseconds (on the `Point.x` attribute).
+
+Improved performance by changing the GEOS context handling. Instead of
+initializing/finalizing on every call, GEOS contexts are reused through a
+threadlocal variable. It should give a performance increase of about
+75 nanoseconds per call.
```
sgillies, line 32: These numbers are compelling!

jorisvandenbossche:
For the first section on scalar performance, I was running some benchmarks locally to get a good idea of what actually improved. Essentially testing creating a point, accessing a cheap attribute, and calling a method, equivalent of:

```python
import shapely
p = shapely.Point(1.0, 2.0)
p.x
p.buffer(1.0)
```

<details>

```bash
echo "Shapely version: $(python -c 'import shapely; print(shapely.__version__)')"
echo "GEOS version: $(python -c 'import shapely.geos; print(shapely.geos.geos_version_string)')"
echo "Python version: $(python -c 'import platform; print(platform.python_version())')"
echo "NumPy version: $(python -c 'import numpy; print(numpy.__version__)')"
echo ""
echo "Point creation: p = Point(1.0, 2.0)"
python -m pyperf timeit "Point(1.0, 2.0)" -s "from shapely.geometry import Point"
sleep 3
echo ""
echo "Coordinate (attribute) access: p.x"
python -m pyperf timeit "p.x" -s "from shapely.geometry import Point; p = Point(1.0, 2.0)"
sleep 3
echo ""
echo "Buffer creation: p.buffer(1.0)"
python -m pyperf timeit "p.buffer(1.0)" -s "from shapely.geometry import Point; p = Point(1.0, 2.0)" 
```

</details>

The results:

```
Shapely version: 1.8.5.post1
GEOS version: 3.10.3-CAPI-1.16.1
Python version: 3.11.16
NumPy version: 1.26.4

Point creation: p = Point(1.0, 2.0)
.....................
Mean +- std dev: 4.33 us +- 0.14 us

Coordinate (attribute) access: p.x
.....................
Mean +- std dev: 3.92 us +- 0.04 us

Buffer creation: p.buffer(1.0)
.....................
Mean +- std dev: 16.1 us +- 0.2 us
```

vs

```
Shapely version: 2.0.7
GEOS version: 3.11.4-CAPI-1.17.4
Python version: 3.11.16
NumPy version: 1.26.4

Point creation: p = Point(1.0, 2.0)
.....................
Mean +- std dev: 2.74 us +- 0.06 us

Coordinate (attribute) access: p.x
.....................
Mean +- std dev: 2.62 us +- 0.03 us

Buffer creation: p.buffer(1.0)
.....................
Mean +- std dev: 16.5 us +- 0.2 us
```

vs


```
Shapely version: 2.1.2
GEOS version: 3.13.1-CAPI-1.19.2
Python version: 3.11.16
NumPy version: 1.26.4

Point creation: p = Point(1.0, 2.0)
.....................
Mean +- std dev: 3.53 us +- 0.09 us

Coordinate (attribute) access: p.x
.....................
Mean +- std dev: 2.72 us +- 0.05 us

Buffer creation: p.buffer(1.0)
.....................
Mean +- std dev: 16.0 us +- 0.3 us
```

vs

```
Shapely version: 2.1.1+109.g4d4f5a77
GEOS version: 3.14.1-CAPI-1.20.5
Python version: 3.11.16
NumPy version: 1.26.4

Point creation: p = Point(1.0, 2.0)
.....................
Mean +- std dev: 2.39 us +- 0.04 us

Coordinate (attribute) access: p.x
.....................
Mean +- std dev: 70.7 ns +- 8.6 ns

Buffer creation: p.buffer(1.0)
.....................
Mean +- std dev: 13.4 us +- 0.1 us
```
(of course also GEOS versions change across shapely versions, so some variation will also be because of that)

So first, shapely 2.0 was not actually slower than 1.8 in practice (based on those limited timings), but so the main point is that _attribute access_ got a lot faster now (thanks to Casper's refactors), and for something like a cheap x coordinate look-up, is now only around 70ns instead of 2-3 microseconds. 
(of course it is still recommended to use the vectorized functions when having to get this for many points when possible)

docs/release/2.x.rst
```diff
@@ -1,6 +1,107 @@
...
+significantly. For example, considering a single ``Point`` geometry (``p =
+shapely.Point(1.0, 2.0)``), accessing a cheap attribute such as its
+x-coordinate (``p.x``) now takes around 50-100 nanoseconds instead of 2-3
+microseconds.
+
+Many of the other attributes and methods have also received performance
+improvements. This is achieved by a combination of specialized scalar versions
+(avoiding the overhead of the NumPy ufunc machinery) and improved GEOS context
+handling (instead of initializing/finalizing on every call, GEOS contexts are
+reused through a thread-local variable).
```
jorisvandenbossche, line 32: cc @caspervdw see the above attempt to summarize this topic (I removed the explicit list of methods that have been improved, since this was outdated, and would now also get too long to list all of them)

jorisvandenbossche:
Further feedback is certainly welcome, but going to merge already to have something rendered in the latest docs to point to.

PR #2502, opened 2026-09-08
BLD: move to a src layout

mwtoews:
This moves the source code from a flat layout to a src layout. Specifically these two directory changes:

- `src` →`src/shapely/lib`: C code for the `shapely.lib` Python module
- `shapely` →`src/shapely`: Python and Cython code for the `shapely.*` modules

See the [Python Packaging User Guide on the topic](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/), which describes some of the advantages of the src layout. Pytest generally recommends a src layout in their ["Good Integration Practices"](https://docs.pytest.org/en/8.0.x/explanation/goodpractices.html#src-layout) doc.


mwtoews:
Any thoughts whether moving this project layout is worth the effort or not?

sgillies:
@mwtoews I'm in favor of a src layout.

jorisvandenbossche:
Apologies, I just merged https://github.com/shapely/shapely/pull/2464, which touches a lot of files. Hopefully the conflicts are easy to solve

---

I am also very much in favor of a src layout (I never bothered to propose it, but happy you did). 
This also solves the somewhat (in my mind) strange layout that the c code is separate from the actual package, instead of part of it.

jorisvandenbossche (approved):

mwtoews:
Conflicts were easy to resolve, and merging existing PRs or any existing work-in-progress shouldn't be much of an issue. I find this layout is much more intuitive to see, with the C code in the "lib" subdir to match the `shapely.lib` module. I'll merge this sometime tomorrow, unless I see anything else to consider.

jorisvandenbossche:
All green, so let's merge before I create more merge conflicts?

PR #2504, opened 2026-09-10
BLD: Exclude files from sdist and bdist wheels created by meson-python

mwtoews:
One oversight from the meson-python transition (#2357) is to inspect the build artefacts. This PR addresses two aspects:

1. The source distribution supported by meson-python ([sdist docs](https://mesonbuild.com/meson-python/how-to-guides/sdist.html)) is based on Git's revision control system. This means that everything in the Git repo is included with the source build. Whereas the prior setuptools setup was controlled with a [`MANIFEST.in`](https://github.com/shapely/shapely/blob/4ea509980a4a06de0559dc9f32c87326a0bde741/MANIFEST.in) file. This adds a few `export-ignore` items to `.gitattributes` to ignore a few files and directories. The new sdist, however, keeps these:
   - `docs/` (in full -- former only copied `*.rst`)
   - <s>`benchmarks/` -- could be handy to anyone with a source distribution</s> decided to remain excluded
   - `CODE_OF_CONDUCT.md` and `CONTRIBUTING.md` docs
   - Note that `ISSUE_TEMPLATE.md` is removed in this PR since it is not used by GitHub (unsure when this was unsupported)
2. The binary distribution (bdist / wheel) is controlled by `meson.build` files wherever `install_subdir()`/`py.install_sources()` is used. Some special treatment of `install_subdir()` is needed for the `tests/legacy` subdir, which contains non-Python files. This includes:
   - `rungrind.dist` and `valgrind-python.supp` - I'm tempted to clean these up, and can do with this PR - update these are removed
   - `data/` which contains version <2 geometry pickle files -- these might be cleaned up someday when [<2 geometry pickle doomsday approaches](https://github.com/shapely/shapely/pull/2260), maybe after 2.2 is out

   A result of these exclusions is that the bdist contents are nearly identical (for Linux at least; Windows includes a few `*.lib` ar files)


.gitattributes
```diff
@@ -1 +1,18 @@
...
+
 shapely/_version.py export-subst
+
+# Exclude files from a sdist generated by meson-python
+######################################################
+
+.github/* export-ignore
+.git* export-ignore
+ci/* export-ignore
+*.json export-ignore
```
jorisvandenbossche, line 10: If we keep the benchmarks, we should probably also keep the asv.conf.json ?

.gitattributes
```diff
@@ -1 +1,18 @@
...
+*.yml export-ignore
+*.yaml export-ignore
+.dockerignore export-ignore
```
jorisvandenbossche: Alternatively, could we simplify this by ignoring everything that starts with a `.` ? That should cover many things listed here (eg all yml and yaml files currently)

jorisvandenbossche:
> The new sdist, however, keeps these:

+1 

> Note that `ISSUE_TEMPLATE.md` is removed in this PR since it is not used by GitHub (unsure when this was unsupported)

Maybe in the past it didn't need to be inside a `.github/ISSUE_TEMPLATE` directory? (anyway, fine to remove, if we want to use templates we can add a new one in a separate PR)

> [wheels] Some special treatment of `install_subdir()` is needed for the `tests/legacy` subdir, which contains non-Python files.

Why go through the effort to exclude the test data? 
I understand in general that is not needed (the test will skip if the data is not there), and in other projects we also do some effort to exclude large data files to minimize wheel size, but here this is about 1.4kB in total, so not really relevant for that, I think


---

One other thing that doesn't seem to work anymore with the new setup is including the licenses (in the nightly wheel from before the meson switch, there are licenses in `../shapely-2.1.1+102.gd9b82cc0.dist-info/licenses/`

jorisvandenbossche:
> One other thing that doesn't seem to work anymore with the new setup is including the licenses

In the release workflow, we do have:

https://github.com/shapely/shapely/blob/30a32b2629a96c3764c268492bf29981f3cd4db4/.github/workflows/release.yml#L169-L176

But that seems to no longer work (meson does not pick up those files automatically? Not sure how setuptools decided to include them)

NumPy seems to use a CIBW_BEFORE_BUILD hook to append their license with the ones included in binary wheels: https://github.com/numpy/numpy-release/blob/ff6643cbdf5bcd6f5ef5430682d2cb1b385a2e72/tools/wheels/cibw_before_build.sh#L7-L17

I suppose that only works for appending to an existing file that meson will include. If we want to keep adding files (but appending is maybe fine as well?), we might need a custom CIBW_REPAIR_WHEEL_COMMAND?

.gitattributes
mwtoews: Maybe it might be easier to not start including benchmarks.

shapely/meson.build
```diff
@@ -56,12 +56,19 @@ py.install_sources([
...
   'vectorized',
 ]
 foreach subdir : subdirs_list
   install_subdir(subdir, install_dir: py.get_install_dir() / 'shapely')
 endforeach
+
+# Filter-out a few files from "tests" for binary wheel distribution
+install_subdir('tests', install_dir: py.get_install_dir() / 'shapely',
+  exclude_directories: 'legacy/data',  # keep until <2 geometry pickle support is removed
+  exclude_files: ['legacy/rungrind.dist', 'legacy/valgrind-python.supp'],
```
mwtoews: Thinking about these files a bit more - I think they probably should not be here, so I might remove them.

.gitattributes
jorisvandenbossche: (I don't have a strong opinion either way, but there is nothing "difficult" about just removing the line I commented on ;))

mwtoews:
> One other thing that doesn't seem to work anymore with the new setup is including the licenses (in the nightly wheel from before the meson switch, there are licenses in `../shapely-2.1.1+102.gd9b82cc0.dist-info/licenses/`

Thanks for this reminder. I checked the core files, but didn't really check this directory. I'll find a solution to bring into this PR.

.gitattributes
mwtoews: Using `.*` seems to work and is much more simpler.

mwtoews:
I'm inspired by meson-python's [`test_dynamic_license_files`](https://github.com/mesonbuild/meson-python/blob/c4ad3bc0b7cae176eddb06409e6bb75b075cd0fa/tests/test_project.py#L149) to use PEP 639's `license-files`, made dynamic with a simple script to print which licence applies within an environment. This needs to bump some requirements to work. This appears to work as expected.

pyproject.toml
```diff
@@ -44,8 +44,8 @@ dependencies = [
 build = [
     # keep in sync with a superset of [build-system.requires]
     "Cython",
-    "meson>=1.2.1",
-    "meson-python>=0.15.0",
+    "meson>=1.6.0",
+    "meson-python>=0.21.0",
```
jorisvandenbossche:
```suggestion
    "meson-python>=0.18.0",
```

Support for PEP 639 (dynamic) license files sees to be added in 0.18? (0.21 is only three days old)

src/shapely/meson.build
```diff
@@ -57,13 +57,17 @@ py.install_sources([
...
-  'tests',
   'vectorized',
 ]
 foreach subdir : subdirs_list
   install_subdir(subdir, install_dir: py.get_install_dir() / 'shapely')
 endforeach
+
+# Filter-out a few files from "tests" for binary wheel distribution
+install_subdir('tests', install_dir: py.get_install_dir() / 'shapely',
+  exclude_directories: 'legacy/data',  # keep until <2 geometry pickle support is removed
```
jorisvandenbossche: See also my comment at https://github.com/shapely/shapely/pull/2504#issuecomment-5615324437, I feel this is not necessary to save 1.4kB

(essentially the default behaviour here seems just fine / an improvement over what setuptools did before?)

.github/workflows/release.yml
```diff
@@ -166,15 +166,6 @@ jobs:
           path: ${{ runner.temp }}/geos-${{ env.GEOS_VERSION }}
           key: ${{ matrix.os }}-${{ matrix.arch }}-${{ env.GEOS_VERSION }}-${{ hashFiles('ci/*') }}
 
-      - name: Add GEOS LICENSE
```
jorisvandenbossche, line 169: There is another identical section above (in the linux wheel build) that can also be removed

jorisvandenbossche:
> I'm inspired by meson-python's [`test_dynamic_license_files`](https://github.com/mesonbuild/meson-python/blob/c4ad3bc0b7cae176eddb06409e6bb75b075cd0fa/tests/test_project.py#L149) to use PEP 639's `license-files`, made dynamic with a simple script to print which licence applies within an environment.

Good, that seems like a nice solution!

pyproject.toml
mwtoews: Thanks for the catch! Also forgot to update `build-system.requires` too.

src/shapely/meson.build
mwtoews: Sure, we can include them.

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks for the updates!

One final question: do we care about the nesting of the licenses in dist-info? Because the additional licenses are now included as `../shapely-x.y.z.dist-info/licenses/ci/wheelbuilder/LICENSE_GEOS`, instead of the current `../shapely-x.y.z.dist-info/licenses/LICENSE_GEOS`

I think _if_ we want to solve that, we need to keep the copy step we had in the cibuildwheel workflow (i.e. which is copying the files to the root of the repo), so that then the `print_license_files.py` file you added can print them as if they are in the root of the repo (and since that is behind a cibuildwheel check, that should be fine locally).

mwtoews:
Aha, I finally figured out the licence stuff. Before we were relying on setuptools' `licence-files` default `LICEN[CS]E*` glob, whereas meson-python adheres strictly to the specifications. Specifically [PEP 639](https://peps.python.org/pep-0639/) shows an example with this same glob, but it doesn't specify that this is any default. So the solution is to keep the licence file copies done in CI for cibuildwheel and specify a glob pattern in `licence-files` to pick these up for the wheels. (A cleaner solution would be to do a before-all hook copy of these files into the build wheel, rather than relying on GHA, but this can be done later.)

I have one last mystery to solve before I'm happy with the transition to meson-python, related to the version. I'll hopefully follow up this soon with a PR.

jorisvandenbossche:
@mwtoews https://github.com/mesonbuild/meson-python/pull/845 seems to contains some latest practices for meson-python for dynamic versions (I can also start a PR if you want, since this is breaking geopandas' CI)

jorisvandenbossche:
I actually kind of like the idea of still having a static version pyproject.toml and only dynamically add a git hash, as decribed in the "Storing the git commit hash inside your package" section in https://github.com/mesonbuild/meson-python/pull/845 (the `dynamic-version-from-script` worked example). However it is a bit more custom setup.

For the versioneer / setuptools-scm way, https://github.com/tiran/pycxxfilt/pull/16 might be a good minimal example of what we probably want (that was linked from the doc PR mentioned above)

mwtoews:
I was looking at those two PRs too! There are nearly too many options to consider for dynamic (or mixture) versioning at the moment, and this could all very well change in another year from now. I think the easiest option is to refactor versioneer by mostly removing it, then consider another option when they mature.

PR #2505, opened 2026-09-14
ENH: allow splitting Polygon by another Polygon (taking boundary)

jorisvandenbossche:
Triggered by https://github.com/shapely/shapely/pull/2477, and can be done as a precursor. 

One of the behaviour differences between our `shapely.ops.split()` implementation and the new `GEOSSplit` is that we raise an error when splitting a polygon by another polygon, while GEOSSplit allows this by taking the boundary of the splitter polygon. 
But we also already do exactly that when splitting a line by a polygon, so it seems fine to do exactly the same when splitting a polygon by a polygon.
This makes the cases of splitting line or polygon more consistent, and makes us consistent with the upstream `GEOSSplit` (and so there is less behaviour difference between the `split()` based on our implementation vs GEOS, which will be used depending on which GEOS version you are using)


src/shapely/ops.py
```diff
@@ -314,8 +314,12 @@ def _split_polygon_with_line(poly, splitter):
         """Split a Polygon with a LineString."""
         if not isinstance(poly, Polygon):
             raise GeometryTypeError("First argument must be a Polygon")
+        if splitter.geom_type in {"Polygon", "MultiPolygon"}:
+            splitter = splitter.boundary
```
jorisvandenbossche, line 318: This is copied from the `_split_line_with_line` function

jorisvandenbossche:
I am going to merge this, since it simplifies https://github.com/shapely/shapely/pull/2477 a bit, which is ready. But if someone has thoughts about this, still happy to hear them

PR #2506, opened 2026-09-14
MAINT: refactor versioneer

mwtoews:
This refactors versioneer in a few ways. Firstly, note that the prior `shapely/_version.py` looked like this:
```python
# This file was generated by 'versioneer.py' (0.28) from
# revision-control system data, or from the parent directory name of an
# unpacked source archive. Distribution tarballs contain a pre-generated copy
# of this file.

import json

version_json = '''
{
 "date": "2025-09-24T15:22:48+0200",
 "dirty": false,
 "error": null,
 "full-revisionid": "5fb639d1056888d135fe56bfaf750c9648addeec",
 "version": "2.1.2"
}
'''  # END VERSION_JSON


def get_versions():
    return json.loads(version_json)
```
This refactor removes the vendored versioneer code, using the version from PyPI instead. And only uses a few features from it accessed from `generate_version.py`.  The git tag versioning pattern remains similar to before, since it still is versioneer under the hood.

When [vcs-versioning](https://github.com/pypa/setuptools-scm/tree/main/vcs-versioning) looks more mature, it may replace versioneer. This change would slightly alter the tag version pattern (e.g. `uvx vcs-versioning` gives `2.1.2.dev113+ga1d55e112`)


jorisvandenbossche:
This is essentially what pandas does as well? (I don't know if you based in on theirs, or if the pandas setup was also based on something else)

jorisvandenbossche:
You will need to remove the ` -Ceditable-verbose=true` from `tests.yml` to fix the doctests (it should not matter there anyway, since we only build once)

mwtoews:
Yes, inspiration from several sources, including numpy and pandas.

jorisvandenbossche:
Can you add `versioneer` to the conda environments (or example in the docs how to setup local dev env with conda)?

mwtoews:
As noted above, I was considering vcs-versioning. But as of [v2.4.0](https://github.com/pypa/setuptools-scm/tree/e78b379a387ade703acde8c06a8780c11061b95c/vcs-versioning) it doesn't yet look mature for several reasons. Firstly there are no docs -- just source code and tests. But there is still more work underway in decoupling it from setuptools-scm. It looks like a more modern and promising candidate with it looks "ready".

Also, we don't need any of these external packages. Just subprocess + git in generate_version.py could do too.

jorisvandenbossche:
Yeah, I might be inclined to go the numpy way of having a minimal script ourselves that just calls `git` in a subprocess and constructs the version from that (https://github.com/numpy/numpy/blob/main/numpy/_build_utils/gitversion.py), instead of using `vcs-versioning`. The latter might have the benefit of some more fallbacks if the version is not available through the standard method (eg when install from git archive)?



> When [vcs-versioning](https://github.com/pypa/setuptools-scm/tree/main/vcs-versioning) looks more mature, it may replace versioneer. This change would slightly alter the tag version pattern (e.g. `uvx vcs-versioning` gives `2.1.2.dev113+ga1d55e112`)

I assume that we should be able to configure it such that we can keep the same pattern, given the many options and knobs that it provides, if we want. In any case, I would prefer to keep the current scheme with the distance in the local part (after the `+`)

jorisvandenbossche (approved):

mwtoews:
I'll have a "go" later today at a more minimal git + subprocess approach like numpy. Fewer deps is a good thing.

jGaboardi:
As an unsolicited comment - we are using [`setuptools-scm`](https://pypi.org/project/setuptools-scm/) over in PySAL for some years now with nice success. (if that is something to consider here).

mwtoews:
Thanks for chiming in @jGaboardi I was considering that too. But the reason "why not" is that setuptools-scm does too many other things, like source code management (with a MANIFEST.in) and other things already done by meson-python. The setuptools-scm team have been splitting the dynamic version work into vcs-versioning (thus better integration with non-setuptools projects). But this "splitting" appears to still be in progress, as I've noted above.

mwtoews:
Looking further at git + subprocess here, it turns out there are several git commands and string manipulations done by versioneer / vcs-versioning, so not worth re-writing into `generate_version.py`. I have since found [a few more docs for vcs-versioning](https://github.com/pypa/setuptools-scm/blob/main/docs/integrators.md), but read that some of these are still experimental. Let stick with versioneer one last round, then re-evaluate other options when they are ready.

mwtoews:
A working proof-of-concept with vcs-versioning is done as the previous commit. This only uses a small subset of the features from the (current) public API. The main difference that I can see is here:

 - versioneer: 2.1.1+117.g9fa10afd
 - vcs-versioning: 2.1.2.dev118+gda423e287

Already, I prefer the vcs-versioning style because it has "dev" to clearly mark it as a development version, and it has "2.1.2" which is the same as the current release. I don't know why versioneer uses the previous "2.1.1" version.

Any preference or opinions?

---

Note that this PR fixes the version in the "latest" docs. And I've also tested tagging a dummy "2.2.0" version locally, and both sdist and wheel artefacts are correctly built with that version, so I don't anticipate any surprises on release.

jorisvandenbossche:
I would propose to stick with the previous commit using versioneer for now, to merge this so we fix the version of the latest nightly. 
And we can then further discuss using `vcs-versioning` (or something else custom) in a follow-up?

---

Some feedback: if using `vcs-versioning`, I would expect the full `generate_version.py` writing a version file would go away? (since that is the point that `vcs-versioning` does this for us? Also otherwise it does not feel much like an improvement)

On the versioning scheme:

> I don't know why versioneer uses the previous "2.1.1" version.

versioneer simply does no magic guessing of the next version, it uses the current tag and adds a `+` part to it. Personally I like this simplicity (setuptools-scm/vcs-versioning has a whole set of varying rules how to determine or guess the version (and with or without release branches etc) to choose from, one of them also being to just use the tag as is). 
And so the reason it is `2.1.1` now, is because that is the latest tag on main (2.1.2 was tagged in the release branch, and I forgot to tag main as `2.2.0.dev0` after branching (which I had done for 2.1, see https://github.com/shapely/shapely/tags?after=2.0.2)).
(and FWIW, the guessed version of `2.1.2.dev118` by vcs-versioning is also kind of "wrong" since this is the main branch with lots of changes not meant for a bug fix release. I suppose we would have to test "semver-pep440" from https://setuptools-scm.readthedocs.io/latest/extending/#available-implementations)

Personally I don't mind just tagging ourselves how we want the dev version to look like (and having the simplicity of not having to understand the details of the guessing scheme), but if one of the schemes works how we want, I can also live with trying that ;) 

But as I mentioned above, I would prefer using something like `2.1.2.dev0+118` instead of `2.1.2dev118`, so it is only the local identifier part (the `+` part) that changes with every commit

mwtoews:
Prior versioneer version is now restored, so this can be merged shortly.
Thanks for the details on why versioneer goes from `2.1.1` -- makes sense. I'm still navigating the `vcs-versioning` codebase to "see" how it works, and it does write files, but not clear when/how. Given that it's actively being developed, it might need to wait a few months to see a good example.

mwtoews:
Oh, before merge. I could just restore the vendored `versioneer.py` and configuration at `setup.cfg` (and remove the new requirements).

jorisvandenbossche:
It would indeed be a smaller change, but personally I don't mind adding versioneer as build dependency for now (and does the vendored version only work with setup.cfg and not with the configuration in pyproject.toml? or is that because the vendored version is old?)

mwtoews:
It's not too old, we are vendoring 0.28 and the latest is 0.29 (https://github.com/python-versioneer/python-versioneer/compare/0.28...0.29). The configuration in `setup.cfg` is readable by all Python versions, whereas from `pyproject.toml` is needs to be either Python 3.11 or with `tomli` as a dependency.

Let's consider the smaller diff with no new deps...

PR #2510, opened 2026-09-17
Translate GEOS exception "InvalidArgumentException: .." into a ValueError

### LINKED ISSUE #2509: Convert GEOSException("IllegalArgumentException: ...") to ValueError("...")
We have several special cases of raising ValueError in GEOS_FINISH. I propose to convert *all* GEOS IllegalArgumentException errors to Python ValueError. ValueError is what a Python programmer is expects and so this makes Shapely fail in a more normal way. Some background here: https://github.com/shapely/shapely/pull/2477#discussion_r3980761320.

I did a search of geopandas and did not find instances of `except shapely.errors.GEOSException`, so I'm optimistic that geopandas won't be adversely affected. On the other side of the coin, I don't see where new ValueErrors from shapely functions would trip up geopandas.

A global search of GitHub shows code that catches `shapely.errors.GEOSException` without also catching `ValueError`. That doesn't mean that such code always break, but it could.

Note that the wkb reader/writer are the only components in shapely that advertise that they raise `GEOSException`. The exceptions coming from other shapely functions are almost entirely unspecified in code or docs. Adding a new one doesn't break a written contract except in those wkb cases.

sgillies:
Resolves #2509.

I'm expecting test failures, will update those accordingly.


sgillies:
More changes needed for GEOS versions < 3.13.

src/shapely/lib/geos.h
```diff
@@ -110,7 +110,12 @@ enum ShapelyErrorCode {
                       "Geometry objects.");                                              \
       break;                                                                             \
     case PGERR_GEOS_EXCEPTION:                                                           \
-      PyErr_SetString(geos_exception[0], last_error);                                    \
+      if (last_error != NULL && strlen(last_error) > 26 && strncmp("IllegalArgumentException: ", last_error, 26) == 0) { \
```
sgillies, line 113: Do some safety checks and then check that `last_error` starts with "IllegalArgumentException: ".

src/shapely/lib/geos.h
```diff
@@ -110,7 +110,12 @@ enum ShapelyErrorCode {
...
+        PyErr_SetString(PyExc_ValueError, last_error + 26);                              \
```
sgillies, line 114: Strip off the "IllegalArgumentException: " prefix, it's superfluous.

src/shapely/tests/test_constructive.py
```diff
@@ -1216,21 +1216,22 @@ def test_maximum_inscribed_circle(geometry, expected):
...
-    geometry = shapely.from_wkt("POLYGON EMPTY")
-    with pytest.raises(
-        GEOSException, match=r"Empty input(?: geometry)? is not supported"
-    ):
+        ("POLYGON EMPTY", r"Empty input(?: geometry)? is not supported"),
+    ],
+)
+def test_maximum_inscribed_circle_empty(wkt, pattern):
+    geometry = shapely.from_wkt(wkt)
+    with pytest.raises(ValueError, match=pattern):
```
sgillies, line 1234: Took the opportunity to parametrize this test.

src/shapely/tests/test_io.py
```diff
@@ -752,15 +753,18 @@ def test_to_wkb_none():
...
-    with pytest.raises(shapely.GEOSException):
-        shapely.to_wkb(point, output_dimension=5)
-
-    with pytest.raises(ValueError):
-        shapely.to_wkb(point, flavor="other")
+@pytest.mark.parametrize(
+    "arg, kwargs, exception",
+    [
+        # Should this be ValueError, too?
+        (1, {}, TypeError),
```
sgillies, line 760: Making this a ValueError seems helpful. See the wkt test too.

src/shapely/tests/test_linear.py
```diff
@@ -129,13 +129,19 @@ def test_line_locate_point_empty(normalized):
     )
 
 
+@pytest.mark.parametrize(
+    "args, exception",
+    [
+        # C API throws its own GEOSException here, not IllegalArgumentException.
+        # See https://github.com/libgeos/geos/issues/1534.
+        ((line_string, line_string), shapely.errors.GEOSException),
```
sgillies, line 137: Found a GEOS bug.

src/shapely/tests/test_io.py
jorisvandenbossche: I think the TypeError is fine here, although the distinction is often not a strict line. 

Eg based on https://docs.python.org/3/builtins/exceptions.html#TypeError, since the first arguments should be geometry objects, passing an int is the wrong type, not just a wrong value.

While for the other two cases below, the tested values are of the correct type, but not a correct value.

jorisvandenbossche:
Code changes look good! I merged the new `split` function, so in its tests the `test_split_unsupported_geometry_type` will have to be updated as well now (and `test_split_poly_with_other` in `src/shapely/tests/legacy/test_split.py`)

src/shapely/tests/test_io.py
mwtoews: TypeError is the correct error to raise in here and with the wkt test.

mwtoews (approved):
Thanks @sgillies the changes look good, and I see that you've cleaned up almost all instances of `IllegalArgumentException` (which originates from libgeos).

sgillies:
@jorisvandenbossche @mwtoews thanks for the feedback on type/value errors. GEOS doesn't distinguish between the two, but I agree that we should, if not perfectly.

I'll update the new tests.

Thinking more a documentation of our exceptions... it's impractical to spell them out for every function, but fortunately all of the core geometry operations go through the same GEOS_FINISH macro, which allows us to have a general documentation topic on exceptions in the API reference. What do you think about me adding something there in a new PR?

sgillies:
@jorisvandenbossche I'm all done here. Will you merge before the release candidate?

jorisvandenbossche:
I was planning to, but see my comment on the issue: https://github.com/shapely/shapely/issues/2509#issuecomment-5759345832. So on second thought, I think we have to consider a bit more the impact and whether we are fine with doing it as a breaking change now, as a less-breaking change (the ShapelyValueError idea that inherits from both GEOSException and ValueError), or as a breaking change in the future (i.e. leave such a change for a hypothetical shapely 3.0)

jorisvandenbossche:
@sgillies maybe a short-term question: would you be OK with keeping this for 2.3? (and take some more time to discuss / implement / test)

sgillies:
It's hard to imagine a 3.0 in our timeline. Not raising ValueError in these cases is a 2.x defect, but yes, this could wait for 2.3. What do you think about a warning in 2.2.0? That would be a trivial change to GEOS_FINISH.

jorisvandenbossche:
What kind of warning would you imagine? 
Adding a warning every time we currently raise such an InvalidArgumentException GEOSException would get annoyingly noisy, I think (libraries/code that currently catches such errors, would also have to start catching the warning to avoid users to see it)

(unfortunately, there is AFAIK not a good way to deprecate a raised exception class. Like for `instance(.., class)` you can do some metaclass magic to warn for that kind of usage, but I don't think anything similar is possible for `except class`)

sgillies:
@jorisvandenbossche Deprecation warnings are only reported in `__main__` or in dev mode, so users of published packages will rarely see them if at all. Packaged code that catches `GEOSException` doesn't have to change anything until 2.3. Developers of research code and scripts can set `once::DeprecationWarning`.

sgillies:
So as not to hold up the 2.2 release candidate (it'll take me a couple hours to implement the warning and catching in tests), how about this @jorisvandenbossche?

* Document upcoming change to ValueError for 2.4 (no PendingDeprecationWarning)
* Add ShapelyValueError in 2.3
* DeprecationWarning in 2.4

I also just realized that GeoPandas users in IPython or notebooks are exposed to Python deprecation warnings, of course. The documentation is more important for them than a pending deprecation warning would be.

PR #2511, opened 2026-09-20
MAINT: move versioneer config to pyproject.toml; remove setup.cfg

mwtoews:
This moves the versioneer config from setup.cfg to pyproject.toml, since this project has a minimum Python 3.11 where tomllib is part of the standard library (rather than requiring a build dep). This cleans up the last setup file.

Two semi-related things are added here for convenience:

1. `packaging` is not used in this project, so remove it from the isort configuration (historic note: it used to be vendored up until the [shapely-1.8.* series](https://github.com/shapely/shapely/tree/maint-1.8/_vendor/packaging))
2. Add `test_git_version` now that `__git_version__` is added.


pyproject.toml
```diff
@@ -197,3 +196,11 @@ allow-dunder-method-names = [
     "_repr_svg_",
 ]
 max-positional-args = 8  # TODO: revise with deprecate_positional changes
+
+[tool.versioneer]
+VCS = "git"
+style = "pep440-branch"
```
mwtoews: Oh right, forgot that I had tested changing this. Not sure if this is welcome, but it always add "dev0" to the version name, e.g. [in docs](https://shapely--2511.org.readthedocs.build/en/2511/) and in build artefacts. I can restore this to `pep440` again if desired.

jorisvandenbossche: I would maybe keep it as is for now, so we can discuss changing the style together with our tagging practices and potentially switching from versioneer to another tool, etc 

(eg the `"pep440-branch"` option also seems to be based on the assumption of a "master" branch, while we have "main", and so it adds this `dev0` also to the version of the main branch, which then does not seem to be its intention? And that would also make the current nightly version sort as older compared to the current latest release on PyPI)

jorisvandenbossche:
```suggestion
style = "pep440"
```

mwtoews: Aggree, restored

PR #2512, opened 2026-09-20
MAINT: Update pre-commit and apply changes for ruff check

mwtoews:
This first does `pre-commit autoupdate` to update ruff to the latest release, and applies fixes to the implicit-string-concatenation-in-collection-literal rule (ISC004).

A few "preview" rules are checked and fixed as well, where they make sense. These preview rules may eventually get enforced with future releases. These were:

- <s>E266: multiple-leading-hashes-for-block-comment</s>
- PLR6201: literal-membership


src/shapely/tests/test_constructive.py
```diff
@@ -1328,7 +1328,7 @@ def test_orient_polygons_array_like():
 @pytest.mark.parametrize(
     "geometry,splitter, expected",
     [
-        ## LineString with Point
+        # LineString with Point
```
jorisvandenbossche: FWIW, I know this gets start nitpicking and I would like to avoid discussing style semantics, but those "double hashes" were added very much on purpose to provide some more structure in the comments. I would personally prefer to keep this, and just ignore that rule (or not yet opt in right now, and ignore it if it would ever be enforced by default)

mwtoews: I have no issue with style semantic nitpicking, so I'll revert these. If this gets out of "preview" I'll ignore it then (mostly because the preview rules are also considering changing all "A1" codes with "a-different-style-like-this", and I don't want to make too many changes based on preview speculation.)

jorisvandenbossche (approved):

